PR #21, opened 2016-03-11
Set up explicit target for Eclipse Mars

elharo:
@akerekes @briandealwis mostly copied from Brian's work on the legacy plugin

### DIFF
--- a/eclipse/mars/gcp-eclipse-mars.target
+++ b/eclipse/mars/gcp-eclipse-mars.target
@@ -1,0 +1,27 @@
+<?xml version="1.0"?>
+<?pde?>
+<target name="Google Cloud Platform for Eclipse Mars" sequenceNumber="1456419225">
+  <locations>
+    <location includeMode="planner" includeAllPlatforms="false" includeSource="true" 
+              includeConfigurePhase="false" type="InstallableUnit">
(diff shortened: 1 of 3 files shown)


akerekes:
LGTM

briandealwis:
We should pull in the .tpd file: it provides a much saner way to set up .target files than using PDE's target editor directly.

elharo:
Is there a way we can do this without also checking in the .target file? I
prefer not to check in generated code. Can Tycho generate .target from .tpd?

On Mon, Mar 14, 2016 at 3:59 PM, Brian de Alwis <notifications@github.com>
wrote:

> We should pull in the .tpd file: it provides a much saner way to set up
> .target files than using PDE's target editor directly.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/21#issuecomment-196497783>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

briandealwis:
The .target file provides specific versions of the artifacts, and is the equivalent of version numbers in Maven's <dependency> elements and are the key for reproducible builds.  We only regenerate them when something external has changed (e.g., Mars.2 released, or Neon released).

If we don't want target files, then we need to resolve artifacts directly against the various repositories as we were doing before.  But then our builds are slower.

PR #22, opened 2016-03-15
correct eclipse version

elharo:
@akerekes @briandealwis @rdayal Maven builds are annoyingly not hermetic so this issue was missed.


akerekes:
LGTM

PR #36, opened 2016-03-29
Fixes #16 - Add rule to ensure we don't link to commons collections 3.2.1

### LINKED ISSUE #16: Add rule to ensure we don't link to commons collections 3.2.1
http://www.geekinasuit.com/2016/03/apache-commons-collections-321.html

akerekes:


briandealwis:
I posted on a similar issue on GPE:

> I realized earlier today that this fix only prevents Maven dependencies. I don't think it will do anything for dependencies brought in via OSGi/Tycho, and I don't think it will do anything for dependencies that ship commons-collections as an included jar. I think we can do something for the former with a `p2.inf` containing negative requirement on `org.apache.commons.collections`.

I think we can do this by adding a `target-platform-configuration` filter like the following to the main `pom.xml` (note: untested):
```
<plugin>
  <artifactId>target-platform-configuration</artifactId>
  <configuration>
    <filters>
      <filter>
        <type>eclipse-plugin</type>
        <id>org.apache.commons.collections</id>
        <restrictTo>
          <versionRange>[3.2.2,3)</versionRange>
        </restrictTo>
      </filter>
    </filters>
  </configuration>
</plugin>
```

elharo:
lgtm

akerekes:
@briandealwis with a minor correction your proposal works, so I incorporated into the PR

briandealwis:
lgtm!  I wonder if that jetty version restriction is necessary; I don't know why it was there in GPE.

PR #40, opened 2016-03-31
Fixes #29 - One Google Cloud Platform for Eclipse Mars target

### LINKED ISSUE #29: One Google Cloud Platform for Eclipse Mars target
Whehn we setup a build we get two "Google Cloud Platform for Eclipse Mars" targets, though they're actually the same. 

Set the checkbox next to one of the `Google Cloud Platform for Eclipse Mars`
     target platforms (they are pointing to the same file), and click `Apply`.

See if we can figure out how to have only one.

akerekes:
Added `features`, `plugins`, `eclipse` to resource filter on parent project

### DIFF
--- a/.project
+++ b/.project
@@ -14,4 +14,42 @@
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
+		<filter>
+			<id>1459454994333</id>
+			<name></name>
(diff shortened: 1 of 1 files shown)


briandealwis:
Need to add `gcp-repo` to the list too: look for `category.xml`

briandealwis:
But otherwise lgtm

PR #42, opened 2016-03-31
Adds classpath entries to the server definition

akerekes:
It turns out these entries are required for Eclipse. If they are missing, an error is shown in the Server Runtime Environment wizard dialog.
However running the server does not need them.


elharo:
lgtm

PR #44, opened 2016-04-01
Add Travis profile, and allow Maven toolchains can be overridden

briandealwis:
Allow Travis to opt-out from the toolchains enforcement as its build is configured to always with with `jdk7`.  Uses a Travis profile in case we need to add more options later.


elharo:
lgtm

PR #51, opened 2016-04-06
Ensure Travis doesn't pull stale artifacts from the cached repository

briandealwis:
Configure Tycho to ignore copies of build artifacts found in the local repository.  This should prevent Tycho from resolving missed references to renamed/deleted artifacts from Travis' cached repository.


elharo:
lgtm 👍

PR #53, opened 2016-04-06
Add license to feature.xml

elharo:
@briandealwis 
fix #50

PR #57, opened 2016-04-07
Start filling out the whitelist plugin

elharo:
@briandealwis @akerekes


features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
```diff
@@ -1,20 +1,30 @@
...
-      Copyright 2016
-   </copyright>
+  <copyright>
+    Copyright 2016
+  </copyright>
 
-   <license>
-      Todo
+  <license>
+    Licensed under the Apache License, Version 2.0 (the "License");
```
akerekes: Check out this commit: https://github.com/GoogleCloudPlatform/google-plugin-for-eclipse/pull/118/commits/cdc02aad2c3d2e7ed95c55a1f990a7e12222ed38 I think we should use the license feature provided by Eclipse instead of adding the license contents explicitly here.

akerekes:
After resolving my license feature comment, LGTM

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/plugin.xml
```diff
@@ -1,4 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
+  <extension
+      id="appEngineWhitelist"
+      point="org.eclipse.core.resource.builders">
+    <builder hasNature="true">
```
briandealwis: I think `hasNature="true"` may mean we can't add the builder to a project via WTP facets?

That said, I think we can have a WTP facet add/remove a nature from the project.

plugins/com.google.cloud.tools.eclipse.appengine.whitelist.test/src/com/google/cloud/tools/eclipse/appengine/whitelist/XmlTest.java
```diff
@@ -10,23 +11,36 @@
...
+  public void testPluginXml() {
+    Assert.assertEquals("plugin", document.getDocumentElement().getNodeName());
+  }
+  
+  @Test
+  public void testLoadExtensionClass()  {
+    Element runElement = (Element) document.getDocumentElement().getElementsByTagName("run").item(0);
+    String className = runElement.getAttribute("class");
+    try {
+      Class.forName(className);
```
briandealwis: I like this idea and this works for now, but `Class#forName()` isn't a 1-1 match with `IConfigurationElement#createExecutableExtension()`.  For example, it doesn't test that the class is instantiable with a 0-arg constructor.  And `createExecutableExtension()` allows specifying additional data with the class name like `com.google.cloud.Foo:additional data` that is provided via the `IExecutableExtension` interface.

features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
briandealwis: I agree, but I think we can do that separately.

briandealwis:
Since you merged in the `license` branch here, close #53?

plugins/com.google.cloud.tools.eclipse.appengine.whitelist.test/src/com/google/cloud/tools/eclipse/appengine/whitelist/XmlTest.java
elharo: On Thu, Apr 7, 2016 at 1:45 PM, Brian de Alwis <notifications@github.com>
wrote:

> I like this idea and this works for now, but Class#forName() isn't a 1-1
> match with IConfigurationElement#createExecutableExtension(). For
> example, it doesn't test that the class is instantiable with a 0-arg
> constructor. And createExecutableExtension() allows specifying additional
> data with the class name like com.google.cloud.Foo:additional data that
> is provided via the IExecutableExtension interface.
>

This test is sort of ad hoc. Is there a way to load up and test the
plugin.xml or an individual extension element file to see that everything
matches up and works?

-- 
Elliotte Rusty Harold
elharo@macfaq.com

briandealwis: On 8-Apr-2016, at 7:59 AM, Elliotte Rusty Harold <notifications@github.com> wrote:
> > + @Test
> > +  public void testLoadExtensionClass()  {
> > +    Element runElement = (Element) document.getDocumentElement().getElementsByTagName("run").item(0);
> > +    String className = runElement.getAttribute("class");
> > +    try {
> > +      Class.forName(className);
> On Thu, Apr 7, 2016 at 1:45 PM, Brian de Alwis <notifications@github.com> wrote:
>> I like this idea and this works for now, but Class#forName() isn't a 1-1 match with IConfigurationElement#createExecutableExtension(). For example, it doesn't test that the class is instantiable with a 0-arg constructor. And createExecutableExtension() allows specifying additional data with the class name like com.google.cloud.Foo:additional data that is provided via the IExecutableExtension interface.
> 
> This test is sort of ad hoc. Is there a way to load up and test the plugin.xml or an individual extension element file to see that everything matches up and works?

Since we’re executing within a plugin context, we can use the org.eclipse.core.runtime.IExtensionRegistry directly.  It doesn’t have an Path-Like mechanism, which is why I’m guessing GPE created some com.google.gdt.eclipse.core.extensions.ExtensionQuery* helper classes.

Keeping the Document parser is still a good idea as the extension registry silently ignores broken plugin.xml files.

Brian.

PR #60, opened 2016-04-08
whitelist all non-JRE classes

elharo:
@briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
@@ -1910,6 +1910,8 @@ public class AppEngineJreWhitelist {
         || className.startsWith("sun.util.") 
         || className.startsWith("org.xml.sax.") 
         || className.startsWith("org.w3c.dom.") 
+        || className.startsWith("org.omg.") 
+        || className.startsWith("org.ietf.jgss") 
         || className.startsWith("com.sun.")) {
       return WHITELIST.contains(className);
     } else { // not a JRE class
--- a/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
@@ -1905,7 +1905,17 @@ public class AppEngineJreWhitelist {
    * @return true if this class is allowed in Java 7 on App Engine, false otherwise
    */
   public static boolean contains(String className) {
-    return WHITELIST.contains(className);
+    if (className.startsWith("java.") 
+        || className.startsWith("javax.") 
+        || className.startsWith("sun.util.") 
+        || className.startsWith("org.xml.sax.") 
+        || className.startsWith("org.w3c.dom.") 
+        || className.startsWith("com.sun.")) {
+      return WHITELIST.contains(className);
+    } else { // not a JRE class
+      return true;
+    }
+    
   }
   
 }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
@@ -1901,7 +1901,7 @@ public class AppEngineJreWhitelist {
               ));
   
   /**
-   * @param className fully package qualified className
+   * @param className fully package qualified class name
    * @return true if this class is allowed in Java 7 on App Engine, false otherwise
    */
   public static boolean contains(String className) {
--- a/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
@@ -1905,8 +1905,13 @@ public class AppEngineJreWhitelist {
    * @return true if this class is allowed in Java 7 on App Engine, false otherwise
    */
   public static boolean contains(String className) {
-    if (className.startsWith("java.") 
-        || className.startsWith("javax.") 
+    if (className.startsWith("javax.") ){
+      if (isBundledInJre(className)) {
+        return WHITELIST.contains(className);
+      }
+      return true;
+    }
+    else if (className.startsWith("java.") 
         || className.startsWith("sun.util.") 
         || className.startsWith("org.xml.sax.") 
         || className.startsWith("org.w3c.dom.") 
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
```diff
@@ -1901,11 +1911,55 @@
...
+      }
+      return true;
+    }
+    else if (className.startsWith("java.") 
+        || className.startsWith("sun.util.") 
+        || className.startsWith("org.xml.sax.") 
+        || className.startsWith("org.w3c.dom.") 
+        || className.startsWith("org.omg.") 
+        || className.startsWith("org.ietf.jgss") 
+        || className.startsWith("com.sun.")) {
```
briandealwis: This com.sun seems too coarse as it will catch any other Sun-produced libraries

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
```diff
@@ -1901,11 +1911,55 @@
...
+      return WHITELIST.contains(className);
+    } else { // not a JRE class
+      return true;
+    }
+    
+  }
+
+  private static boolean isBundledInJre(String className) {
+    if (className.startsWith("javax.accessibility")
+        || className.startsWith("javax.activation")
```
briandealwis: Some of these have trailing dots and others don't. They should probably all have trailing dots?

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
```diff
@@ -1900,12 +1910,73 @@
...
+  
   /**
-   * @param className fully package qualified className
-   * @return true if this class is allowed in Java 7 on App Engine, false otherwise
+   * @param className fully package qualified class name
+   * @return true if this class is allowed in Java 7 on App Engine Standard, false otherwise
    */
   public static boolean contains(String className) {
-    return WHITELIST.contains(className);
+    if (className.startsWith("javax.") ){
```
briandealwis: Huh, good point: there are other javax.* implementations that can be carried along as a library.

I'm finding this logic a bit tortuous though.  Could we not merge the else-if logic (lines 1924-1931) into the isBundledInJre() and 
simplify it as:
```
return !isBundledInJre(className) || WHITELIST.contains(className));
```

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
```diff
@@ -1900,12 +1910,73 @@
...
+        || className.startsWith("com.sun.org.apache.")) {
+      return WHITELIST.contains(className);
+    } else { // not a JRE class
+      return true;
+    }
+  }
+
+  // javax packages are tricky. Some are in the JRE. Some aren't. 
+  private static boolean isBundledInJre(String className) {
+    if (className.startsWith("javax.accessibility.")
```
briandealwis: This list is becoming big enough that we may need a different data structure for fast lookups.

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
elharo: good idea. done.

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
elharo: Let's see if this becomes a user noticeable issue. I'm half-tempted to remove the packages no one should be using anyway. Or we could make it another HashMap.

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
briandealwis: > good idea. done.

I don't see it in the latest commit c9ef72f?

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
```diff
@@ -1910,24 +1910,21 @@
...
-      }
-      return true;
+      return !isBundledInJre(className) || WHITELIST.contains(className);
     } else if (className.startsWith("java.") 
         || className.startsWith("sun.util.") 
         || className.startsWith("org.xml.sax.") 
         || className.startsWith("org.w3c.dom.") 
         || className.startsWith("org.omg.") 
         || className.startsWith("org.ietf.jgss.") 
+        // com.sun and com.oracle packages are tricky. Some are in the JRE. Some aren't.
```
briandealwis: It seems funny to include these here as there are no com.sun.* or com.oracle classes in the whitelist.

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/src/com/google/cloud/tools/eclipse/appengine/whitelist/AppEngineJreWhitelist.java
```diff
@@ -1910,24 +1910,21 @@
...
    * @return true if this class is allowed in Java 7 on App Engine Standard, false otherwise
    */
   public static boolean contains(String className) {
     if (className.startsWith("javax.") ){
-      if (isBundledInJre(className)) {
-        return WHITELIST.contains(className);
-      }
-      return true;
+      return !isBundledInJre(className) || WHITELIST.contains(className);
     } else if (className.startsWith("java.")
```
briandealwis: I had meant that all of these packages should be moved into isBundledInJre() too.

briandealwis:
I took a step back and looked at the files in entirety, and here are my thoughts:
- I find it hard to reason about why a class/package will be allowed or not.  For example, the `com.oracle.nio` package is referenced, but there are no classes in the whitelist.
- The current organization means that whitelisting a class means that there are multiple places that might need to be modified — I not only have to add it to WHITELIST, I may also need to ad the package name to an `if()` statement too.
- I see that other classes have been added to `WHITELIST` than are listed in the [JRE Class Whitelist](https://cloud.google.com/appengine/docs/java/jrewhitelist).  I think these additional classes should be separated out to another list both to record their provenance and to simplify maintenance.  When the JRE  Class Whitelist is updated (e.g., AE supports Java8), we can just replace the entirety of the WHITELIST field. It also means we could use them for warnings too ("Warning: {0} is not an officially whitelisted class and may not be available in the future").
- It looks like we have a set of blacklisted packages too, and those should be explicit with a reference to the JRE that they were pulled from?

Maybe this would be easier as having an explicit blacklist.
```
public static boolean contains(String className) {
  if(WHITELIST.contains(className)) {
    return true;
  }
  // some bundled classes are grey listed
  if(isGreylisted(className)) {
    return true;
  }
  // known classes that we will never support
  if(isBlacklisted(className)) {
    return false;
  }
  // anything that's not blacklisted and not bundled in the JRE is ok
  return !isBundledInJre(className);
}

/** Known internal classes */
private static boolean isBlacklisted(String className) {
  // disallow any internal classes to any JRE, incl JRockit, J9, …
  return className.startsWith("sun.") || …;
}

/**
 * Well-Known classes that aren't officially part of the JRE but commonly
 * expected to be available
 */
private static boolean isGreylisted(String classname) {
  return className.startsWith("com.sun.rmi.") || …;
}
```

elharo:
The TLDR version is:

Anything in the Java 7 JRE that is not whitelisted is blacklisted.
The list on https://cloud.google.com/appengine/docs/java/jrewhitelist is not complete. In particular it does not include JRE internal classes that usually shouldn't be used directly.
Some packages (com.sun, com.oracle, and especially javax) have members both inside and outside the JRE. All classes outside the JRE are allowed.

briandealwis:
In the absence of a better source of information, LGTM.

PR #64, opened 2016-04-15
initial boilerplate for new project plugin

elharo:
@briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.google.cloud.tools.eclipse</groupId>
+    <artifactId>trunk</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
+    <relativePath>../../</relativePath>
+  </parent>
+  <groupId>com.google.cloud.tools.eclipse</groupId>
```
briandealwis: Can drop the <groupId> since it's inherited.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+  <artifactId>com.google.cloud.tools.eclipse.appengine.newproject.test</artifactId>
+  <packaging>eclipse-test-plugin</packaging>
+
+  <build>
+    <sourceDirectory>src</sourceDirectory>
```
briandealwis: I don't think `<sourceDirectory>` is necessary?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/XmlTest.java
```diff
@@ -0,0 +1,53 @@
...
+public class XmlTest {
+    
+  private Document document;
+	
+  @Before
+  public void setUp() throws ParserConfigurationException, IOException, SAXException {
+    DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+    factory.setNamespaceAware(true);
+    DocumentBuilder builder = factory.newDocumentBuilder();
+    document = builder.parse(
```
briandealwis: It would be nice to avoid embedding relative pathnames, especially if we decide to move the tests out to a different root in the repository (e.g., `plugins/`, `features/`, and `tests/`).  But that can wait until we refactor these with a utility class.

But if you want to, you can resolve this file using OSGi (which is necessary to use the extension registry as per below) by fetch the file from the bundle.  Normally finding a bundle would require iterating through the loaded bundles looking at them by name, but since these tests are in a fragment then we share the same bundle.

```
Bundle bundleUnderTest = FrameworkUtil.getBundle(this.getClass());
URL pluginXml = bundleUnderTest.getEntry("plugin.xml");
try(InputStream input = pluginXml.openStream()) {
  document = builder.parse(input);
}
```

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/XmlTest.java
```diff
@@ -0,0 +1,53 @@
...
+        new File("../com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml"));
+  }
+    
+  @Test
+  public void testPluginXml() {
+    Assert.assertEquals("plugin", document.getDocumentElement().getNodeName());
+  }
+  
+  @Test
+  public void testLoadExtensionPoints() {
```
briandealwis: This method might be better named `testValidExtensionPoints` or `Actual`

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,9 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Cloud Tools for Eclipse App Engine New Project Generation
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.newproject;singleton:=true
+Bundle-Version: 0.1.0.qualifier
+Bundle-Vendor: Google, Inc.
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Require-Bundle: javax.servlet;bundle-version="3.1.0";visibility:=reexport,
```
briandealwis: I thought AppEngine Standard only supported 2.5?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
```
briandealwis: Assuming something will be added here. Eclipse warns about empty `plugin.xml`s.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/pom.xml
```diff
@@ -0,0 +1,18 @@
...
+    <groupId>com.google.cloud.tools.eclipse</groupId>
+    <artifactId>trunk</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
+    <relativePath>../../</relativePath>
+  </parent>
+  <artifactId>com.google.cloud.tools.eclipse.appengine.newproject</artifactId>
+  <version>0.1.0-SNAPSHOT</version>
+  <packaging>eclipse-plugin</packaging>
+  
+  <dependencies>
```
briandealwis: Remove this.  m2e complains about empty `<dependencies>`

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <plugins>
+      <plugin>
```
briandealwis: We can skip explicitly including `tycho-surefire-plugin` as it's bound in as part of the `eclipse-test-plugin`'s lifecycle.  It's only needed if we're tweaking the plugin's configuration.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
elharo: Are we likely to need to do that though? I'd like to have the places we need to add things to in place, rather than trying to figure them out down the line when we do need them. Depending on hidden defaults is often confusing.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/XmlTest.java
elharo: "The import org.osgi cannot be resolved"

I'm not too worried about tests that break when code changes. We have tests precisely so we'll notice this.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/XmlTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
elharo: added

plugins/com.google.cloud.tools.eclipse.appengine.newproject/pom.xml
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/pom.xml
briandealwis: > Depending on hidden defaults is often confusing.

Welcome to Maven ;-)

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.properties
```diff
@@ -0,0 +1,5 @@
+google = Google
+Bundle-Vendor = Google Inc.
+Bundle-Name = Wizards for Google Cloud Platform Apps
```
briandealwis: This Bundle-Name is at odds with the much more constrained bundle symbolic name (`…appengine…`).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
```diff
@@ -1,5 +1,28 @@
...
 <plugin>
-  
+   <extension point="org.eclipse.ui.newWizards">
+    <wizard
+            id="com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandard"
+            name="App Engine Standard Java Project"
+            class="com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardFactory"
+            icon="icons/app_engine_standard_16x16.png"
+            project="true"
+            finalPerspective="org.eclipse.jdt.ui.JavaPerspective"
```
briandealwis: I leave out the `finalPerspective` as many developers will want to use the _Java EE_ perspective instead.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
```diff
@@ -1,5 +1,28 @@
...
+         <description>%wizard.description</description>
+         <selection class="org.eclipse.core.resources.IResource"/>
+      </wizard>
+      <wizard
+            id="com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandardDotDotDot"
+            name="App Engine Standard Java Project"
+            class="com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardFactory"
+            icon="icons/app_engine_standard_16x16.png"
+            project="true"
+            finalPerspective="org.eclipse.jdt.ui.JavaPerspective"
```
briandealwis: Ditto

elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.properties
elharo: Yes: http://spring.io/blog/2008/02/18/creating-osgi-bundles/
I thought the Bundle name was more human readable for display to people. If that's not true, what's the difference between Bundle-Symbolic_Name and Bundle-Name? Or what would you suggest here for the value of Bundle-Name?

elharo:
Currently dying at

org.apache.maven.InternalErrorException: Internal error: java.lang.RuntimeException: org.osgi.framework.BundleException: Bundle com.google.cloud.tools.eclipse.appengine.newproject cannot be resolved
Resolution errors:
   Bundle com.google.cloud.tools.eclipse.appengine.newproject - Missing Constraint: Import-Package: org.eclipse.ui; version="0.0.0"

something about MANIFEST.MF may be off

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
```diff
@@ -1,5 +1,28 @@
...
+            category="org.eclipse.ui.Basic">
```
briandealwis: Doesn't seem like the right category

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.properties
briandealwis: "Wizards for Google Cloud Platform's App Engine"?

An aside, but the difference between bundle-name, symbolic-name, and id:
- the Bundle-Symbolic-Name (BSN) is a partial identifier
- each bundle is uniquely identified by its BSN _and_ its version number
   - given two bundles with the same BSN and version, the OSGi framework assumes they are identical and won't load the second
- each bundle when loaded into the OSGi Framework is assigned a unique numeric id
- the Bundle-Name is a human meaningful term and ignored by the framework; the Bundle Name doesn't show up in much of the UI

You'll often see the bundle id in logs like the following: `Could not resolve module: org.eclipse.pde.api.tools.tests [670]`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
elharo: This I copied from the old plugin I think. For the moment I'll take the second wizard out. We'll likely define an App Engine specific category in the future.

briandealwis: Or a category for GCP?  Data flow projects, GCE, GKE, etc.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,15 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Cloud Tools for Eclipse App Engine New Project Generation
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.newproject;singleton:=true
+Bundle-Version: 0.1.0.qualifier
+Bundle-Vendor: Google, Inc.
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Require-Bundle: org.eclipse.jface.viewers,
```
briandealwis: These aren't right: `org.eclipse.jface.viewers` and `org.eclipse.jface.wizard` are packages, not bundles.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,15 @@
...
+  org.eclipse.jface.wizard,
+  org.eclipse.ui,
+  javax.servlet;bundle-version="2.5.0";visibility:=reexport,
```
briandealwis: And please don't re-export `javax.servlet` and `javax.servlet.jsp`: it's an anti-pattern.  Eclipse has several bundles that does this for historical reasons; we'd like to remove them but it causes API breakage.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
elharo: so this sis what I keep running into. Where is any of this documented? How do I find out what bundles exist and what's contained in them and vice versa?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
briandealwis: <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>A</kbd> gives a quick way to access all OSGi & Eclipse-related artifacts: bundles, packages, extension points & extensions.  It's all pulled from what's in the target platform.  There are various views (Window > Show View > Plug-in Development) related to introspecting on the target; most useful there is the _Plug-in Registry_.  And then using the Manifest editor and its tabs.  If you want to go deeper, use _File > Import > Plug-in Development > Plug-ins and Fragments_ to import bundles into your environment so they show up in your project list.

elharo:
Travis is now passing. Everyone PTAL

eclipse/mars/gcp-eclipse-mars.tpd
```diff
@@ -18,6 +21,7 @@ location "http://download.eclipse.org/cbi/updates/license" {
 location "http://download.eclipse.org/releases/mars/" {
 	org.eclipse.sdk.feature.group
 	org.eclipse.jdt.feature.group
+  org.eclipse.platform.feature.group
```
briandealwis: Huh, why did you need to bring this in?  It should be included with org.eclipse.sdk

elharo: It didn't seem to be. This was the change that enabled it to find the relevant classes needed for the wizard.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/XmlTest.java
briandealwis: I'm not sure we can assume the test is run from the project's working directory, but it seems to work and we can fix that later as required.

> Tried it but "The import org.osgi cannot be resolved".

Need to `Import-Package: org.osgi.framework`.

briandealwis:
lgtm

Issue #92, opened 2016-04-21
Dev app server command line argument validation


akerekes:
There is a minimal validation implemented by #93, however we could provide more useful validation to the users for e.g.:
 - path values
 - log level (or in general for enum) values
 - numbers, port numbers
 - hostnames

elharo:
obsolete. This is now part of app-tools-lib-for-java

PR #93, opened 2016-04-21
Fixes #91 - Enable setting command line flags for launch configuration

### LINKED ISSUE #91: Enable setting command line flags for launch configuration
The contains code that creates a second tab for the launch configuration dialog, but it does not work right now. Also server-flags.json's content is pretty outdated. This issue is about updating the dialog to offer the latest supported dev_appserver.py command line arguments (Cloud SDK v105.0) and add the selected arguments to the launch configuration.
However using those arguments in the launch will be covered by #74 
The flags json is updated manually now, but #70 covers the use case to populate the flags automatically is possible.

akerekes:
This PR contains the current list of flags supported by Cloud SDK (except for those used in non-Java runtimes). The user can select a flag, define a value and add it to the launch configuration. There is validation in place to ensure that string arguments are not empty and boolean arguments are in the supported format. More extensive validation is covered by #92

@briandealwis @elharo PTAL


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerFlagSelectionDialog.java
```diff
@@ -78,28 +78,32 @@ protected Control createDialogArea(Composite parent) {
 
       @Override
       public IStatus validate(Object[] selection) {
-        if (selection.length == 1) {
+        if (selection == null || selection.length == 0) {
+          return new Status(OK, Activator.PLUGIN_ID, "");
```
briandealwis: `return Status.OK_STATUS;`

akerekes: I thought about it, but that would use plugin ID as null. Is that OK?

briandealwis: OK Statuses are normally not reported anywhere.  The plugin id is only used in the `!ENTRY` section when reporting to the Eclipse log:

```
!ENTRY org.eclipse.equinox.p2.core 4 0 2016-04-21 11:32:26.091
!MESSAGE Provisioning exception
!STACK 1
…
```

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerFlagSelectionDialog.java
```diff
@@ -129,6 +184,12 @@ private void createArgumentArea(Composite parent) {
     argumentText = new Text(container, SWT.BORDER);
     argumentText.setFont(container.getFont());
     argumentText.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
+    argumentText.addModifyListener(new ModifyListener() {
+      @Override
+      public void modifyText(ModifyEvent e) {
```
elharo: e --> event

elharo:
LGTM

elharo:
I'm going to finish this one up and merge since Andras is now on leave.

PR #98, opened 2016-04-23
Fix #75 - Access debug icons via DebugUITools

### LINKED ISSUE #75: Use own icons instead of Eclipse internal icons
The class `com.google.cloud.tools.eclipse.util.TerminateAction` should use icons from from our bundle, instead of the one found in the internal code of the DebugPlugin.

briandealwis:
PTAL @elharo


elharo:
LGTM

PR #101, opened 2016-04-23
setup wizard UI

### LINKED ISSUE #103: Validate Java package names
In new app engine wizard, show some sort of error when user has typed an illegal package name such as one that contains spaces.

### LINKED ISSUE #104: Validate app engine project IDs
In new app engine wizard, show some sort of error when user has typed an illegal App Engine Project ID such as one that contains spaces. 

First, figure out if App Engine has documented this anywhere. A quick google search doesn't find it so I'm not sure what is and is not illegal.

### LINKED ISSUE #105: Validate Eclipse project location
In new app engine wizard. Make sure directory exists and is writable.

elharo:
@briandealwis 
fix #103 
fix #104 
fix #105


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
```diff
@@ -0,0 +1,45 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
+
+/**
+ * Collects all data needed to create and configure an App Engine Standard Project.
+ */
+class AppEngineStandardProjectConfig {
+
+  private String appEngineProjectId;
```
briandealwis: I wonder if this is better named `projectId`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
```diff
@@ -0,0 +1,45 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
+
+/**
+ * Collects all data needed to create and configure an App Engine Standard Project.
+ */
+class AppEngineStandardProjectConfig {
```
briandealwis: Does this need validation?  Is it valid to not have a java package?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -0,0 +1,172 @@
...
+      }
+    }
+  }
+
+  private final class PageValidator implements ModifyListener {
+    @Override
+    public void modifyText(ModifyEvent event) {
+      // todo add checks for project ID and directory 
+      // todo add error messages
+      boolean complete = JavaPackageValidator.validate(javaPackageField.getText())
```
briandealwis: Is the default package allowed for an AE app?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -0,0 +1,172 @@
...
+import org.eclipse.swt.widgets.DirectoryDialog;
+import org.eclipse.swt.widgets.Group;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Text;
+
+/**
+ * UI to collect all information necessary to create a new App Engine Standard Java Eclipse project.
+ */
+public class AppEngineStandardWizardPage extends WizardPage implements IWizardPage {
```
briandealwis: The general layout differs from other Eclipse project wizards.  Unless there's a reason to deviate, I think we should keep with the standard approach and extend from `WizardNewProjectCreationPage`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
elharo: If the package is the empty string, that's allowed. It's the default package.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
elharo: I wanted to duistinguish between the *Eclipse* project and the *App Engine* project.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -1,28 +1,32 @@
...
   public void init(IWorkbench workbench, IStructuredSelection selection) {
     this.setWindowTitle("New App Engine Standard Project");
-    // see BasicNewResourceProjectWizard which is what we need except it isn't meant to be subclassed
-    IWizardPage page = new StubWizardPage("first page");
+    page = new AppEngineStandardWizardPage("first page");
     this.addPage(page);
   }
   
   @Override
   public boolean canFinish() {
```
briandealwis: Remove this method then — the default iterates over the pages to verify that they are all complete.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -0,0 +1,172 @@
...
+  
+  private final class DirectorySwitcher extends SelectionAdapter {
+    @Override
+    public void widgetSelected(SelectionEvent event) {
+      projectDirectoryBrowseButton.setEnabled(event.widget == customProjectDirectoryButton);
+      projectDirectoryField.setEnabled(event.widget == customProjectDirectoryButton);
+    }
+  }
+
+  AppEngineStandardProjectConfig getAppEngineStandardProjectConfig() {
```
briandealwis: The best practice with wizards is to have the wizard instance host the data, and have the pages pull out and set that data.  It avoids having wizard pages needing to access each other's contents.  For example, when we add support for managed APIs, the ManagedApiWizardPage could need to know the Project Id to enable the API, and the Eclipse project to fetch and install the API jars.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
```diff
@@ -0,0 +1,45 @@
...
+
+  private String appEngineProjectId;
+  private String eclipseProjectName;
+  private String packageName;
+  private String eclipseProjectDirectory;
+
+  // todo does builder pattern make more sense here?
```
briandealwis: Looking forward, I also wondered how we'd integrate AE-Flex, GKE, and GCE here.  And whether a deployment region is needed too (that seems to be specified on project creation).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
briandealwis: Fair enough.  I was thinking about supporting AE-Flex/GKE/GCE — maybe just Google Developer Project (`gdpId`).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
elharo: I expect we'll need more fields as we go forward. I want to get the basics working this quarter though.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
elharo: Looks like I can pull that class in from three different bundles. Any reason to prefer one over the other? (workspace or ide?)

elharo: also, 

Diagnostic of the generation of the target platform.
No installable unit with ID 'org.eclipse.ui.workbench.feature.group' can be found in 'http://download.eclipse.org/releases/mars/'.

Any idea which feature group in which repo this comes from?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
elharo: so to be clear, you're suggesting that we add, for example, getProjectId()/setProjectId() methods in the wizard class which the page class would invoke? I.e. the page class will hold a reference to the wizard it belongs to.

At what point /in what method should the page class set the value in the wizard? In a ModifyListener or when a button is pressed?

briandealwis: Something along that line.  What I've found easiest to manage is for the `Wizard` instance to configure the `WizardPage` with a particular model object (like `AppEngineStandardProjectConfig`), and the `WizardPage` changes that model object.  The `Wizard` knows that the `WizardPage` is in a consistent state via `WizardPage#isPageComplete()`.

So essentially the `WizardPage` is a view, the `Wizard` is the controller.

elharo: So in what method in which class should the data from the UI (wizard page) copy into the model (Wizard/AppEngineStandardProjectConfig)?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
briandealwis: I guess we only need the java package for generated examples?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
briandealwis: `WizardNewProjectCreationPage` is in the `org.eclipse.ui.ide` bundle.  I'm not sure where else you're seeing it?

elharo: org.eclipse.ui.workbench and some rap thing.

What feature do I need to add to the target platform to pull in org.eclipse.ui.ide? More importantly, how is one supposed to figure out what feature contains a certain package? Is there somewhere a list of feature groups and the packages they contain or vice versa?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
briandealwis: Sorry, I wasn't clear.  What I meant was: We'll have different New wizards for AE Flex, GKE, and GCE projects, and we'll want to share as many pieces between them.  So one approach would have them all share this "Create a Google Developers Project" page (configure Eclipse project name and location, and GDP Id), and then have some additional pages for specifics of AE Standard, AE Flex, GKE, GCE.  Minimize churn.

elharo: I'm not optimistic about our ability to predict the future. Whatever we do now will change as we add features. For instance, if there's little enough data to fit into a single page, we may make each environment its own wizard, perhaps sharing some widgets between them rather sharing entire pages. I.e. churn will happen, whether we plan for it or not.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
briandealwis: Ah ok.  Avoid RAP: that's the tooling that makes a web-based Eclipse Workbench.  It's about 90% transparent: if we write nice code that uses APIs, then it can be automatically run under RAP.

Figuring out feature-package is awkward in part because a bundle can be included in many features.  The best approach I've found is to use the [P2 Browser](https://github.com/ifedorenko/p2-browser): you can right-click on a bundle and select _Open Including Features_.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectIdValidator.java
```diff
@@ -0,0 +1,29 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
```
briandealwis: Oh gosh, how does this relate to the Project ID shown on the developer console?  The [Go binding tests](https://github.com/golang/appengine/blob/master/internal/app_id_test.go) have other syntaxes too.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
elharo: org.eclipse.ui.dialogs.WizardNewProjectCreationPage is not found in our current Simply importing WizardNewProjectCreationPage in AppEngineStandardWizardPage, not even referencing it, breaks the compile in eclipse, though it looks like it may still work at the command line. Hmm...

elharo: OK, got WizardNewProjectCreationPage imported. Now I have to figure out how to add our custom pieces to it.

elharo:
Looks like the tests are breaking:

Display must be created on main thread due to Cocoa restrictions.

and lots of messages like

org.eclipse.swt.SWTException: Invalid thread access

briandealwis:
Run with `-XstartOnFirstThread`; we had an ${os-vm-flags} property that was set up for that purpose in GPE.

briandealwis:
Should be fixed by #108

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
elharo: right

elharo:
Passing locally. Still failing on Travis though:

Running com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest
!SESSION 2016-04-26 20:25:02.864 -----------------------------------------------

!MESSAGE An error occurred while automatically activating bundle org.eclipse.ui.ide (6).
!STACK 0
org.osgi.framework.BundleException: Exception in org.eclipse.ui.internal.ide.IDEWorkbenchPlugin.start() of bundle org.eclipse.ui.ide.
	at 
...
Caused by: org.eclipse.swt.SWTError: No more handles [gtk_init_check() failed]
...
om.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest
testOnePage(com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest)  Time elapsed: 0.098 sec  <<< ERROR!
java.lang.NoClassDefFoundError: org/eclipse/ui/dialogs/WizardNewProjectCreationPage

briandealwis:
Looks like we need to explicitly set up X for Travis:

https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectIdValidator.java
elharo: This is the project ID shown on the developer console. It has basically two forms:

word-word
domain.com:word-word

The third form shown in the Go binding tests that adds a partition is an internal form not exposed externally.

elharo:
OK, GUI tests now passing on travis. PTAL and ping if there are any unresolved comments that are still relevant. (Switching to the NewProjectCreationWizard removed a lot of redundant code from the original PR.)

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
```diff
@@ -8,4 +8,5 @@ Bundle-Vendor: Google, Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.newproject
 Require-Bundle: org.junit,
- org.eclipse.equinox.registry
+ org.eclipse.equinox.registry,
```
briandealwis: `.ide` isn't necessary here since it's picked up from the host bundle

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPageTest.java
```diff
@@ -0,0 +1,15 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AppEngineStandardWizardPageTest {
```
briandealwis: Could test various error conditions, and also check that `getErrorMessage() != null`

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizardTest.java
```diff
@@ -16,16 +16,22 @@ public void setUp() {
     } catch (NullPointerException ex) {
       wizard = new StandardProjectWizard();
     }
+    wizard.init(null, null);
```
briandealwis: Shouldn't be needed — see comments at `init()`

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -0,0 +1,94 @@
...
+
+  private boolean validateLocalFields() {
+    String packageName = javaPackageField.getText();
+    if (!JavaPackageValidator.validate(packageName)) {
+      setErrorMessage("Illegal Java package name: " + packageName);
+      return false;
+    }
+    
+    String projectId = projectIdField.getText();
+    if (!AppEngineProjectIdValidator.validate(projectId)) {
```
briandealwis: `validatePage()` will be called on the first keystroke and so it's possible the user will immediately see errors.  A better user experience would differentiate between a missing project id (warning) and an invalid projectId (error).   You can provide a warning message with `setMessage(xxx, WARNING)` (no shortcut like `setErrorMessage`).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -0,0 +1,21 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IWorkspaceRoot;
+import org.eclipse.core.resources.ResourcesPlugin;
+
+/**
+* Utility to make a new Eclipse project with the App Engine Standard Nature in the workspace.
```
briandealwis: Not a nature, but facets.  It's worth describing here the standard facets that we'll be putting into place.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+* Utility to make a new Eclipse project with the App Engine Standard Nature in the workspace.  
+*/
+class EclipseProjectCreator {
+
+  /**
+   * @return true on successful project creation; false otherwise
+   */
+  static boolean makeNewProject(AppEngineStandardProjectConfig config) {
```
briandealwis: A boolean doesn't provide any information on failure.  There are a lot of reasons why project creation can go wrong.  Eclipse APIs return an IStatus object.  Or you could throw an exception.

This should take an `IProgressMonitor`; a dummy can be created with `NullProgressMonitor` (for tests).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -0,0 +1,21 @@
...
+    IWorkspaceRoot workspaceRoot = ResourcesPlugin.getWorkspace().getRoot();
+    IProject project = workspaceRoot.getProject(config.getEclipseProjectName());
```
briandealwis: For the actual creation process, take a look at
- `org.eclipse.ui.wizards.newresource.BasicNewProjectResourceWizard#createNewProject()` for an overview of how the creation is structured
- `org.eclipse.swt.common.project.facet.ui`, particularly the `BasicFacetedProjectWizard` because we'll have to configure the appropriate facets.
- An article on using [Progress Monitors](https://eclipse.org/articles/Article-Progress-Monitors/article.html)

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -1,28 +1,32 @@
...
 import org.eclipse.ui.INewWizard;
 import org.eclipse.ui.IWorkbench;
 
 public class StandardProjectWizard extends Wizard implements INewWizard {
 
+  private AppEngineStandardWizardPage page;
+  private AppEngineStandardProjectConfig config = new AppEngineStandardProjectConfig();
+
   @Override
   public void init(IWorkbench workbench, IStructuredSelection selection) {
```
briandealwis: This method is really only intended to configure the wizard based on context from when invoked (e.g., for modifying a the selected project).

The configuration (`setWindowTitle`) should be done in the constructor.
The page configuration should be done by overriding `addPages`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -1,28 +1,32 @@
...
     this.setWindowTitle("New App Engine Standard Project");
-    // see BasicNewResourceProjectWizard which is what we need except it isn't meant to be subclassed
-    IWizardPage page = new StubWizardPage("first page");
+    page = new AppEngineStandardWizardPage("first page");
```
briandealwis: Not a great page identifier 😄.  Could check that you can fetch the page by id in the tests.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -1,28 +1,32 @@
...
   @Override
   public boolean performFinish() {
+    // todo is this the right time/place to grab these?
+    config.setAppEngineProjectId(page.getAppEngineProjectId());
+   // config.setEclipseProjectDirectory(page.getLocationPath());
+    config.setEclipseProjectName(page.getProjectName());
+    config.setPackageName(page.getPackageName());
+    
+    EclipseProjectCreator.makeNewProject(config);
+    // todo what are we supposed to do if project creation fails?
```
briandealwis: For the todo: Show an error message.  You could either use `setErrorMessage()` or open an error dialog (`org.eclipse.ui.statushandlers.StatusManager`)

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPageTest.java
elharo: Hard to test too much without calling createControl. The error message should initially be null, I thought?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizardTest.java
elharo: seems to be needed:

Running com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest
Tests run: 3, Failures: 3, Errors: 0, Skipped: 0, Time elapsed: 0.202 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest
testOnePage(com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest)  Time elapsed: 0.136 sec  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizardTest.testOnePage(StandardProjectWizardTest.java:34)

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
elharo: Missing project ID is allowed. The validator returns true in that case.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
elharo: done. For now this is just a stub. It's the next piece I'm going to work on once this is in place.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizardTest.java
briandealwis: You're not triggering the normal lifecycle of the wizard.  When a WizardDialog is opened on the Wizard, it calls Wizard.addPages() before rendering the controls on the active page.

You could do something like this:
```
WizardDialog wizardDialog = new WizardDialog(null, wizard);
wizardDialog.setBlockOnOpen(false);
wizardDialog.open();
Assert.assertNotNull("Wizard dialog should be open", wizardDialog.getShell());
// and then other checks

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java (hunk 14 above)
Comment (elharo) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java]: IStatus sounds good. done and done. Will flesh this out further as the EclipseProjectCreator grows.

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java (hunk 15 above)
Comment (elharo) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java]: Thanks

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java (hunk 16 above)
Comment (elharo) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java]: done

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizardTest.java (hunk 11 above)
Comment (elharo) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizardTest.java]: done

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPageTest.java (hunk 10 above)
Comment (briandealwis) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPageTest.java]: I meant (but didn't say) setting some fields to invalid values and then testing.  But I suspect that you're right: it'll error without calling createControl().

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java (hunk 17 above)
Comment (elharo) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java]: Any suggestions for an ID?

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java (hunk 18 above)
Comment (elharo) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java]: todo added

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java (hunk 17 above)
Comment (briandealwis) [inline plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java]: I see a bunch of other wizards use `basicNewProjectPage`.

File: plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java [hunk 19]
```diff
@@ -1,28 +1,32 @@
...
-    return true;
-  }
 
   @Override
   public boolean performFinish() {
+    // todo is this the right time/place to grab these?
+    config.setAppEngineProjectId(page.getAppEngineProjectId());
+   // config.setEclipseProjectDirectory(page.getLocationPath());
+    config.setEclipseProjectName(page.getProjectName());
+    config.setPackageName(page.getPackageName());
```
briandealwis: I meant to mention here: the style I think works a bit better is to provide the config object to the page when defined (addPages()) and use that object here.

So `addPages()` would look like:

```
addPage(new AppEngineStandardWizardPage(config));
```

elharo:
PTAL

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -9,11 +11,14 @@
 
   private AppEngineStandardWizardPage page;
   private AppEngineStandardProjectConfig config = new AppEngineStandardProjectConfig();
-
-  @Override
-  public void init(IWorkbench workbench, IStructuredSelection selection) {
+  
+  StandardProjectWizard() {
```
briandealwis: THis constructor needs to be public:
```
Class org.eclipse.core.internal.registry.osgi.RegistryStrategyOSGI can not access a member of class com.google.cloud.tools.eclipse.appengine.newproject.StandardProjectWizard with modifiers ""
```

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
```diff
@@ -10,5 +10,6 @@ Require-Bundle: org.eclipse.ui,
  javax.servlet;bundle-version="2.5.0",
  javax.servlet.jsp;bundle-version="2.2.0",
  org.eclipse.core.resources,
- org.eclipse.ui.ide
+ org.eclipse.ui.ide,
+ org.eclipse.core.runtime
```
briandealwis: Use `org.eclipse.equinox.common` instead.  Try to avoid `.core.runtime` as it brings in all sorts of dependencies for historical reasons.

briandealwis:
With those two fixes, LGTM

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
elharo: Passing around mutable state like that scares me. Maybe if we need to prefill the fields, and even then I'd want to make a copy.

briandealwis: Well it is the instance created by the wizard instance.  We can use databinding etc to allow reacting to changes.  Managing state across multiple pages quickly becomes complex otherwise. But we can figure that out later.

PR #111, opened 2016-04-28
Create empty project from wizard

elharo:
@briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
```diff
@@ -9,6 +13,9 @@
   private String eclipseProjectName;
   private String packageName;
   private String eclipseProjectDirectory;
+  private URI eclipseProjectLocationUri;
+  private IProject project;
+  private boolean useDefaultLocation;
```
briandealwis: Does `useDefaultLocation` belong here?  `useDefaultLocation` seems like UI-level state, used to decide what's put into `eclipseProjectDirectory`, `eclipseProjectLocationUri`, and `project`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -15,10 +25,57 @@
...
+      AppEngineStandardProjectConfig config, IProgressMonitor monitor,
+      final Shell shell, final IWizardContainer container) {
+    
+    // todo submonitor????
+    
+
+    URI location = null;
+    if (!config.getUseDefaultProjectLocation()) {
+      location = config.getEclipseProjectLocationUri();
+    }
```
briandealwis: I think this if-default-then-use-this-directory should be decided in the wizard page, no?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -15,10 +25,57 @@
   /**
    * @return status of project creation
    */
-  static IStatus makeNewProject(AppEngineStandardProjectConfig config, IProgressMonitor monitor) {
-    IWorkspaceRoot workspaceRoot = ResourcesPlugin.getWorkspace().getRoot();
-    IProject project = workspaceRoot.getProject(config.getEclipseProjectName());
-    return Status.OK_STATUS;
+  static IStatus makeNewProject(
+      AppEngineStandardProjectConfig config, IProgressMonitor monitor,
+      final Shell shell, final IWizardContainer container) {
```
briandealwis: This should be `IRunnableContext` rather than `IWizardContainer`.

But really this class should be headless, with no references to `Shell`, `IRunnableContext`, or any UI stuff.  Rather this method should be called in the context of an `IRunnableWithContext` — really it's a wrapper around the `CreateProjectOperation`.  So the method should take an `AEStandardProjectConfig`, an `IProgressMonitor`, and an `IAdaptable uiInfo` (to keep the workspace operation happy).  This class handles the orchestration of creating the project within the workspace, adding the facets, etc.  It can be called from a variety of contexts, including from a wizard, and called programmatically (e.g., from a test).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -29,10 +29,14 @@ public boolean performFinish() {
    // config.setEclipseProjectDirectory(page.getLocationPath());
     config.setEclipseProjectName(page.getProjectName());
     config.setPackageName(page.getPackageName());
+    config.setEclipseProjectLocationUri(page.getLocationURI());
+    config.setUseDefaultProjectLocation(page.useDefaults());
+    config.setProject(page.getProjectHandle());
     
     // todo set up
     IProgressMonitor monitor = null;
```
briandealwis: So this is where you'd use 
```
getContainer().run(true, true, (monitor) -> { 
   status = EclipseProjectCreator.makeNewProject(config, monitor, WorkspaceUndoUtil.getUIInfoAdapter(getShell()));
});
```

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
elharo: good plan. done.

elharo:
PTAL

elharo:
PTAL. There's now a significant amount of refactoring to get it to create an src folder too. Pretty much all the work has moved into the run method.

briandealwis:
LGTM. If I was going to change anything, it would be to rename the CreateNewProjectOperation to associate it more to AE-Std and WTP, like `CreateAppEngineStandardWtpProject`.  Because the user should be able to create the identical setup using the Dynamic Web Project, and then associating the runtime, and other facets.  This wizard just makes it easier.

briandealwis:
I also suspect the org.eclipse.jst.* faceted project creation classes might simplify some of the creation.  I'd expect there was some operation where you can provide a set of facet + versions ("jst.web" "2.5") or a runtime mapping.  Some classes I came across in a quick search:
- org.eclipse.wst.common.project.facet.core.ProjectFacetsManager: support for creating a faceted project
- org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils: defines Java Servlet facets
- org.eclipse.jst.server.core.FacetUtil: converts between a facet runtime and a server runtime
And checkout the `com.google.cloud.tools.eclipse.appengine.localserver`'s `CloudSdkProjectConfigurator.java` for useful code too

elharo:
Renamed to CreateAppEngineStandardWtpProject. I'll look at the org.eclipse.jst.* faceted project creation classes in the next PR.

PR #113, opened 2016-04-28
Add "Google" to wizard name

elharo:
@briandealwis This is to make the wizard easier to find in autocomplete. I found myself typing "Google" and wondering why the wizard didn't show up.


briandealwis:
LGTM.  You can also add <keywordReference> elements to the wizard description.  Requires adding an extension to the `org.eclipse.ui.keywords` extension point.  Probably worth creating keywords for:

- Google
- AppEngine
- GAE

These keywords don't show up in the UI.  So when we do the AE Flex tooling, we could add _MVM_ for those [who don't like the new naming](https://googlecloud-community.slack.com/archives/app-engine/p1461852716000542).

PR #114, opened 2016-04-28
Start work on a test plan

elharo:
@briandealwis

### DIFF
--- a/TESTPLAN.md
+++ b/TESTPLAN.md
@@ -18,8 +18,24 @@ Manual testing of the GCloud Plugin for Eclipse.
 # Create a project
 
 1. File > New > Project...
-1. In the Wizards: text field begin typing "Goo". You should see 
-   "Google App Engine Standard Project" in a "Google Cloud Platform" category.
+1. In the Wizards: text field begin typing "Goo".  
+   "Google App Engine Standard Project" should 
+   appear in a "Google Cloud Platform" folder.
+   Also check that the App Engine logo (a cartoon plane) also appears.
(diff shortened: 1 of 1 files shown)


TESTPLAN.md
```diff
@@ -0,0 +1,47 @@
+Manual testing of the GCloud Plugin for Eclipse.
+
+# Install the plugin
+
+1. Launch Eclipse 4.5.
+
+2. If necessary, remove the old version of the plugin.
```
briandealwis: Figure out what should happen if we install the new version over top of the old plugin.

TESTPLAN.md
```diff
@@ -0,0 +1,47 @@
...
+  2. Click "Google Cloud Platform for Eclipse 4.5". (If this is not present,
+    move on to step 3 below.)
+  3. Click the Uninstall... button.
+  4. Click the Finish button.
+  5. Click the "Yes" button in the dialog that prompts you to restart Eclipse.
+2. ????
+3. Restart Eclipse
+4. Check installation details. 
+
+# Create a project
```
briandealwis: These are related to single project situations.  We'll need something similar for EAR-style multi-service projects.

elharo: agreed

TESTPLAN.md
elharo: Ideally nothing special. Mostly this is to make sure we actually are using the new version we're testing and that the install works.

elharo:
PTAL

briandealwis:
LGTM.

PR #117, opened 2016-04-29
Templates for new code

elharo:
@briandealwis I want to get your early thoughts on how we might design this. Clearly we need to break the code generation out into a separate class, likely a separate package; and support multiple project templates. Also, let me know if I doing anything too boneheaded in how I create folders and files.

What I don't yet know is whether there's a good generic templating system for eclipse projects, along the lines of maven archetypes. And/or whether we want to depend on something more generic like Freemarker.

We don't necessarily have to answer al these questions right away (unless there's an obvious answer.) We can get away with some hard coding for now and refactor into our eventual generic solution.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
@@ -47,14 +47,38 @@ class CreateAppEngineStandardWtpProject implements IRunnableWithProgress {
         description, "Creating new App Engine Project");
     try {
       operation.execute(monitor, uiInfoAdapter);
-      IFolder folder = newProject.getFolder("src");
-      if (!folder.exists()) {
-        boolean force = true;
-        boolean local = true;
-        folder.create(force, local, monitor);
-      }
+      createCode(monitor, newProject);
     } catch (ExecutionException | CoreException ex) {
       throw new InvocationTargetException(ex);
     }
+  }
+
+  // todo design a template system for this.
+  /**
+   * Set up the sample code.
+   */
+  private void createCode(IProgressMonitor monitor, IProject newProject) throws CoreException {
+    boolean force = true;
+    boolean local = true;
+    IFolder src = newProject.getFolder("src");
+    if (!src.exists()) {
+      src.create(force, local, monitor);
+    }
+    IFolder main = createChildFolder("main", src, monitor);
+    IFolder java = createChildFolder("java", main, monitor);
+    IFolder webapp = createChildFolder("webapp", main, monitor);
+    IFolder webinf = createChildFolder("WEB-INF", webapp, monitor);
+    IFolder test = createChildFolder("test", src, monitor);
+  }
+
+  private IFolder createChildFolder(String name, IFolder parent, IProgressMonitor monitor) 
+      throws CoreException {
+    boolean force = true;
+    boolean local = true;
+    IFolder child = parent.getFolder(name);
+    if (!child.exists()) {
+      child.create(force, local, monitor);
+    }
+    return child;
   }
 }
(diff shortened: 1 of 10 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -47,14 +49,50 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException {
...
+      child.create(force, local, monitor);
+    }
+    return child;
+  }
+  
+  private IFile createChildFile(String name, IFolder parent, IProgressMonitor monitor) 
+      throws CoreException {
+    boolean force = true;
+    IFile child = parent.getFile(name);
+    InputStream in = CreateAppEngineStandardWtpProject.class.getResourceAsStream(name);
```
briandealwis: You don't typically see this kind of resource access in Eclipse/OSGi — we access files relative to the bundle (e.g., in a `templates` folder).  I've personally found it really difficult to keep track of templates alongside classes.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/EclipseProjectCreator.java
```diff
@@ -47,14 +49,50 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException {
...
+    boolean force = true;
+    boolean local = true;
+    IFolder src = newProject.getFolder("src");
+    if (!src.exists()) {
+      src.create(force, local, monitor);
+    }
+    IFolder main = createChildFolder("main", src, monitor);
+    IFolder java = createChildFolder("java", main, monitor);
+    IFolder webapp = createChildFolder("webapp", main, monitor);
+    IFile appengineWebAppXml = createChildFile("appengine-web.xml", webapp, monitor);
```
briandealwis: I think you're missing the `appengine-web.xml` file on this commit

briandealwis:
This is a good start.  So LGTM, modulo the above.

The structure created in `createCode()` isn't the typical structure created by a Dynamic Web Project, which follows more Eclipse-style conventions of:
- `src` (a java root)
- `build/classes`
- `WebContent/WEB-INF` with `web.xml` and `lib/`
- `WebContent/META-INF` with `MANIFEST.MF`

That should be done automatically once we turn this into creating the project as a faceted project and adding the appropriate facets, which will create the appropriate directories and configure natures etc.  We'll also need to set our server runtime, which will add the `appengine-web.xml`.

But this code will need to:
- set the `<application></application>` field (and possibly the `<service></service>` field?) in the `appengine-web.xml`
-  generate and hook up a sample servlet

elharo:
The directory structure is maven standard layout. How important do you think it is to change this to Eclipse standard layout? We're still discussing how much we want this to depend on maven (answers range from completely to not at all.) Your input here would be valued.

I've added some more template code. More will be coming. I've also split it out into a separate package. We may end up splitting it into a separate module as you suggest. I'm not sure what that's going to look like yet. I am pretty sure that the end result will not look like what we have now, but it's easier to start somewhere and iterate. 

Yes, we will need to do some search and replace, at minimum, in the templates. That remains to be done, and probably won't happen until the next PR.

I've added some test code for the file and folder operations. Please check this out too, with a particular eye to whether it's leaving junk behind on the file system I should clean up.

briandealwis:
Re: Maven vs Eclipse WTP layouts (and Gradle, …).  For new projects, I *think* we could just settle on a Maven layout, but then we could (should?) just invoke Maven’s corresponding archetype to make sure things are done in the Right Way and let [the m2e-wtp plugin](https://www.eclipse.org/m2e-wtp/ <https://www.eclipse.org/m2e-wtp/>) do the heavy lifting to pull them in. We may still need to add some support for our GAE server runtime support.

patflynn:
I tend to agree that if you're generating a Maven project, you may as well
use an Archetype to define the template. It's pretty easy to do, and I
believe we have some already.

PR #119, opened 2016-05-01
fix resource path

elharo:


elharo:
@briandealwis

briandealwis:
When I try this, I see two problems:

- don't get the `HelloServlet` copied into place
- several progress monitor dialogs get created, resulting in flickering

elharo:
The HelloServlet hasn't been added in yet. I've seen the flickering too. May involve the status monitor, but I don't want to hang this up on fixing that.

briandealwis:
lgtm then

elharo:
As best I can tell, the flickering seems to be progress  monitors swapping in and out as part of the wizard dialog. Maybe I just need to figure out how to use a single monitor instead? (which is what I thought I was doing, but perhaps the library converts them into separate SubMonitors for each file and folder?

elharo:
Using a screen capture tool I was able to see what's happening with the flicker:

useroperation


Why these dialogs are popping up, or how we make them not pop up I don't know yet.

elharo:
OK, the key seems to be to change

      boolean fork = true;
      boolean cancelable = true;
      getContainer().run(fork, cancelable, runnable);

to

      boolean fork = false;
      boolean cancelable = true;
      getContainer().run(fork, cancelable, runnable);

and then the flicker goes away.

PR #122, opened 2016-05-03
Generate servlet java code from wizard

elharo:
@briandealwis The more I look at this the more I'm convinced we should use FreeMarker. Otherwise we'll just end up reinventing it. However for now this will get us to MVP.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
@@ -34,7 +34,7 @@ public class CodeTemplates {
   // solution such as key-value map or a different design
   // todo what if project isn't empty?
   public static void materialize(IProject project, AppEngineStandardProjectConfig config,
-      IProgressMonitor monitor, String name) throws CoreException {
+      IProgressMonitor monitor) throws CoreException {
     
     String packageName = config.getPackageName();
     createCode(monitor, project, packageName);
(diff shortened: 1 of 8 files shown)


briandealwis:
+1 to templating support.  I thought I had posted a bit about the `org.eclipse.text`/`org.eclipse.jface.text` templating support. But I can't seem to find any mention anywhere.

elharo:
I did see that. I think it was in chat though.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -103,11 +117,53 @@ static IFile createChildFile(String name, IFolder parent, SubMonitor monitor)
...
+    InputStream in = CodeTemplates.class.getResourceAsStream("templates/" + name + ".ftl");
+    if (in == null) {
+      IStatus status = new Status(Status.ERROR, "todo plugin ID", 2, 
+          "Could not load template for " + name, null);
+      throw new CoreException(status);
+    }
+    
+    if (!child.exists()) {
+      // todo total hack; lots of problems with edge conditions and performance;
+      // replace this with FreeMarker or better
```
briandealwis: It would seem worth extracting this into a separate method though.

briandealwis:
I hacked up an example of using the JFace tempting.
```
        // uncertain of importance of this contextTypeId
        String contextTypeId = "com.google.cloud.templating.context";
        TemplateContextType contextType = new TemplateContextType(contextTypeId);
        Template jtemplate = new Template("", "", contextTypeId, template, true);
        IDocument document = new Document();
        TemplateContext context = new DocumentTemplateContext(contextType, document, 0, 0);
        for (Entry<String, String> mapping : values.entrySet()) {
          context.setVariable(mapping.getKey(), mapping.getValue());
        }
        TemplateBuffer buffer = context.evaluate(jtemplate);
        template = buffer.getString();
```
Requires adding an `Import-Package: org.eclipse.jface.text,org.eclipse.jface.text.templates` to the bundle.

briandealwis:
(I was curious myself.  There some additional support in there for dealing with types, with custom variable resolvers, but I'm not sure how that works.)

Anyways, LGTM.

PR #124, opened 2016-05-04
move config files into WEB-INF and HTML file out of WEB-INF

### LINKED ISSUE #123: Where to put webxml and appengine-web.xml
We're generating them into src/main/webapp but maven expects them in src/main/webapp/WEB-INF

See if there's a canonical location for these; and, if we're doing it wrong, fix the location

elharo:
@briandealwis fix #123


briandealwis:
lgtm

Issue #134, opened 2016-05-04
Configure Generated Project to build in Eclipse


elharo:
1. Attach Java nature
2. Set src/main/java in classpath entries
3. Attach servlet jar

Or can we just do this all with facets?

elharo:
ISimpleWebFacetInstallDataModelProperties.CONTENT_DIR, new Path("src/main/webapp")

except that doesn't seem to be quite right.
search for WebContent in WST to find relevant code

elharo:
see also SimpleWebFacetInstallDelegate

elharo:
SimpleWebFacetInstallDelegate.execute uses the config to set up a lot of the project

elharo:
Tentative next step: add cloudSdkDirectory to wizard and add to classpath:

${cloudSdkDirectory}/platform/google_appengine/google/appengine/tools/java/lib/shared/servlet-api.jar"
${cloudSdkDirectory}/platform/google_appengine/google/appengine/tools/java/lib/shared/jsp-api.ja

Probably first figure out how to add to classpath a hard coded path

briandealwis:
There's two parts to this:

1. This is where the `IRuntime`'s come into to play.  These are normally added via an `IRuntimeComponent` associated with the `IRuntime`.  Some of these provide an adapter to `IClasspathProvider` that pulls them from the `IRuntime`.

2. Where does the runtime pull these jars from?  Patrick recommended against using `gcloud` as the canonical place to locate them.  But that could be a later step.

elharo:
I can confirm that when I manually add the Generic Server Framework Runtime to the project that the classpath is configured appropriately; i.e. servlet-api.jar and the other jar are added and it all builds. 

This might be a stopgap until the new runtime is ready. If nothing else, I can figure out how to attach a runtime programmatically when creating a project.

elharo:
FYI org.eclipse.wst.project.facet.ProductManager contains many of the keys we may need to set.

elharo:
FYI IFacetedProject.setTargetedRuntimes/setPrimaryRuntime is how we attach a runtime to a project. Now I need to figure out how to create/point to a runtime. The Dynmaic Web Project wizard provides UI for this. Do we need that?

elharo:
This is probably not quite right but something like this should be in the direction of what we need:

      Set<IRuntime> runtimes = RuntimeManager.getRuntimes();
[...]

Issue #132, opened 2016-05-04
Generate build.gradle


elharo:

rdayal:
Closing this for now. We're not doing this in the near future.

PR #135, opened 2016-05-04
Generate a pom.xml

### LINKED ISSUE #131: Generate pom.xml
For now, just make it all runnable from the CLI.

Still TBD: how/whether we sync back and forth from the pom.xml to Eclipse and whether the Eclipse Build goes through maven

elharo:
@briandealwis @patflynn 

fix #131

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/templates/pom.xml.ftl
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/templates/pom.xml.ftl
@@ -1,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" 
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <packaging>war</packaging>
+  <version>1.0-SNAPSHOT</version>
+
+  <groupId>HelloAppEngine</groupId>
+  <artifactId>HelloAppEngine</artifactId>
+
+  <properties>
+    <app.id>${ProjectID}</app.id>
+    <app.version>1</app.version>
+    <appengine.version>1.9.36</appengine.version>
+    <gcloud.plugin.version>2.0.9.74.v20150814</gcloud.plugin.version>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.showDeprecation>true</maven.compiler.showDeprecation>
+  </properties>
+
+  <prerequisites>
+    <maven>3.1.0</maven>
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/templates/pom.xml.ftl
```diff
@@ -0,0 +1,131 @@
...
+                        <directory>${basedir}/src/main/webapp/WEB-INF</directory>
+                        <filtering>true</filtering>
+                        <targetPath>WEB-INF</targetPath>
+                    </resource>
+                </webResources>
+            </configuration>
+        </plugin>
+
+        <plugin>
+            <groupId>com.google.appengine</groupId>
```
patflynn: you'll eventually want to update this to the new maven plugins that we're going to release to dogfood soon.

briandealwis:
I only gave a very cursory look, but this looks like creating a Maven archetype project.  Rather than reinvent that wheel,  I think we should have different streams for creating a new project:

- Create a native new WTP AppEngine project, with appropriate facets.  This is pretty much the path you're following right now, though it should be transitioned to using faceted projects, and I realize that some of the code generation you're doing should be transitioned into facet actions instead.
- Create a new Maven AppEngine project
   - This should hook into the M2Eclipse support and creates a project using the appropriate archetype
   - Using the new Maven Project wizard is painful as it seemingly doesn't come configured out of the box to use the Maven Central Archetype Catalog, which is baffling
   - and there are several appengine archetypes and it's not clear which is the right one, so there's definite value here
- Create a new Gradle AppEngine project

elharo:
This PR isn't the final build system. The team asked me to throw in a quick pom.xml so we could have a CLI build in place for now. #134 and #118 are about getting more of this working inside Eclipse. I think we're leaning against making the build inside Eclipse maven based, though that could change with a strong enough case in favor or maven. If we do go all the way down the maven road, we'll need to understand how changes in the pom.xml sync back into the Eclipse build.

One question about facets: do these indeed work in regular Eclipse for Java? I.e. not Eclipse for Java EE? From what I'm reading they seem like an EE thing. 

Gradle we're still discussing in #132. How much do you think Eclipse developers use Gradle? Our impression is that it's more common in the IntelliJ space, but it's easy enough to add a build.gradle file in addition to the pom and everything else.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/templates/pom.xml.ftl
elharo: For sure, easy enough to do when they're ready.

briandealwis:
LGTM

briandealwis:
> One question about facets: do these indeed work in regular Eclipse for Java? I.e. not Eclipse for Java EE? From what I'm reading they seem like an EE thing.

Faceted Projects are a WTP feature; they never migrated further into the platform.  They're almost equivalent to natures, but natures don't have the ability to place constraints like pre-reqs and anti-reqs.  There's nothing preventing using a faceted project for a Java project, but I guess it's rare for non-WTP projects to mix different languages and libraries.

PR #144, opened 2016-05-06
Don't generate an extra deployment descriptor web.xml

elharo:
@briandealwis This doesn't actually work. When I call createDataModel it seems to throw a NullPointerException without much of a stack trace. I'm trying to find an example of configuring the jst.web facet

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
@@ -15,6 +15,7 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.jst.j2ee.web
 Import-Package: org.eclipse.core.runtime;version="3.5.0",
  org.eclipse.jst.common.project.facet.core,
+ org.eclipse.jst.j2ee.project.facet,
  org.eclipse.wst.common.frameworks.datamodel,
  org.eclipse.wst.common.project.facet.core
 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
@@ -17,7 +17,6 @@ Import-Package: org.eclipse.core.runtime;version="3.5.0",
  org.eclipse.jst.common.project.facet.core,
  org.eclipse.jst.j2ee.project.facet,
  org.eclipse.wst.common.componentcore.datamodel.properties,
- org.eclipse.wst.common.componentcore.internal.util,
  org.eclipse.wst.common.frameworks.datamodel,
  org.eclipse.wst.common.project.facet.core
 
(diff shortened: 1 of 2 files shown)


briandealwis:
That's NPE is a bug with the definitions in `.localserver`. I'll get some fixes up ASAP.

elharo:
OK, I finally managed to get a config set up for the web facet that the framework accepts. So far all this does is refrain from creating an extra web.xml deployment descriptor but it's a good checkpoint so PTAL.

Next step will be to configure the various paths for the webapp folder and the like.

briandealwis:
> So far all this does is refrain from creating an extra web.xml deployment descriptor but it's a good checkpoint so PTAL

So that's to avoid the web.xml in the `WebContent/WEB-INF` in favour of the one in `src/main/webapp`?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
```diff
@@ -15,6 +15,10 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.jst.j2ee.web
 Import-Package: org.eclipse.core.runtime;version="3.5.0",
  org.eclipse.jst.common.project.facet.core,
+ org.eclipse.jst.j2ee.project.facet,
+ org.eclipse.wst.common.componentcore.datamodel.properties,
+ org.eclipse.wst.common.componentcore.internal.util,
```
briandealwis: `.internal.util` doesn't seem to be used

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -67,12 +71,15 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
-      facetedProject.installProjectFacet(WebFacetUtils.WEB_25, null, monitor);
-
+      
       CodeTemplates.materialize(newProject, config, progress.newChild(40));
       
+      IDataModel webModel = DataModelFactory.createDataModel(new WebFacetInstallDataModelProvider());
+      webModel.setBooleanProperty(IJ2EEModuleFacetInstallDataModelProperties.ADD_TO_EAR, false);
+      webModel.setBooleanProperty(IJ2EEFacetInstallDataModelProperties.GENERATE_DD, false);
+      webModel.setBooleanProperty(IWebFacetInstallDataModelProperties.INSTALL_WEB_LIBRARY, false);
+      facetedProject.installProjectFacet(WebFacetUtils.WEB_25, webModel, monitor);
```
briandealwis: So if we were to make this more Eclipse-WTP'y rather than Maven'y, then we could keep these and instead forcibly overwrite them when we materialize the templates (since we know we're a new project).

briandealwis:
LGTM with the .internal removed

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
elharo: done

PR #145, opened 2016-05-06
Fix AE definitions to specify up to Java 1.7

briandealwis:


plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -131,13 +133,11 @@
...
-                  <requires facet="jst.web" version="2.3"/>
-                  <requires facet="jst.web" version="2.4"/>
-                  <requires facet="jst.web" version="2.5"/>
-              </or>
-              <requires facet="java" version="1.7]"/>
+               <requires
+                     facet="jst.web"
+                     version="[2.2-2.5]">
+               </requires>
+              <requires facet="java" version="[1.4-1.7]"/>
```
elharo: This seems exactly the opposite of the PR description. Maybe [1.7]? I guess the bug was that the opening square bracket was missing?

briandealwis: You're right, but if we pin it to 1.7 then users get errors if they try specifying anything else.  We should be able to run code compiled against 1.3 too, I suppose.

briandealwis: Retitled.  I think this may be a bug in WTP, but the examples don't actually show anything with a cap like "1.7]", just a lower base "(1.3".

elharo:
So why not specify java 1.7 since it's what App Engine supports. I guess 1.7 is mostly backwards compatible with earlier versions. 

LGTM

briandealwis:
AE provides a 1.7 runtime, but it supports executing classes built against other JRE versions.  The `java` facet version seems to be used to set the Execution Environment _providing_ the server runtime instance doesn't explicitly set a JRE.
  - The `com.google.cloud.tools.eclipse.gcloud.runtime` runtime (called _App Engine_ and driven by the `gcloud_serverdef.xml` file) does prompt for an explicit JRE when creating the runtime instance, and so the JRE sticks.
  - The other runtime, `com.google.cloud.tools.eclipse.runtime.id1` (called _GAE Standard_) doesn't prompt for a JRE and seems to set the project JRE from the `java` facet (i.e., changing the `java` facet to 1.5 changes the JRE to JavaSE-1.5). I think this is a good thing.

PR #146, opened 2016-05-10
New Maven-based Project creation

briandealwis:
Use m2eclipse to create a Maven project from the `com.google.appengine:appengine-skeleton-archetype`.  Includes lifecycle-mappings for m2eclipse to ignore the uses of the `versions-maven-plugin` in the archetype.

I've put this functionality into a separate bundle as this ensures that if the m2eclipse or maven are not present then the option won't appear in the New Wizard list.

There are still some polish issues:
- [ ] I'm using the `.newproject`'s wizard starting page, but will likely have to change it to something more like m2eclipse's `MavenProjectWizardArtifactPage` that prompts for the group-id and artifact-id.  
- [x] I wonder if we should remove the generated `nbactions.xml` file as it causes XML validation errors


plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,23 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: New Maven-based App Engine Project Generation
```
elharo: App Engine Standard

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -0,0 +1,154 @@
...
+    config.setPackageName(page.getPackageName());
+
+    if (page.useDefaults()) {
+      config.setEclipseProjectLocationUri(null);
+    } else {
+      config.setEclipseProjectLocationUri(page.getLocationURI());
+    }
+    config.setProject(page.getProjectHandle());
+
+    // todo set up
```
elharo: what's this todo about?

briandealwis: That's a copy-n-paste from the WTP wizard :-)

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -0,0 +1,154 @@
...
+  @Override
+  public void init(IWorkbench workbench, IStructuredSelection selection) {}
+
+  void createProject(IProgressMonitor monitor) throws CoreException {
+    SubMonitor progress = SubMonitor.convert(monitor);
+    monitor.beginTask("Creating Maven AppEngine archetype", 100);
+
+    ProjectImportConfiguration importConfiguration = new ProjectImportConfiguration();
+    IPath creationPath = config.getProject().getParent().getLocation();
+    String groupId = "com.example.appengineapp";
```
elharo: add a todo to pull 6this from wizard, or just set it to package name

elharo:
nbactions.xml is, I think, for NetBeans. What sort of validation errors do you see?

elharo:
There's quite a bit of duplicate code with StandardProjectWizard here. Is there any way we can combine some of that?

briandealwis:
PTAL @elharo.  I subclassed `AppEngineStandardProjectConfig` to add Maven-specific information, and modified the `AppEngineStandardWizardPage` to take and update the `config` object with changes.  The MavenProjectConfig has a default groupId.  We can create a Maven-specific wizard page at some point.

I also added the DTD for appengine-web.xml and add an XML validation rule to exclude the `nbactions.xml` file.  So newly generated projects are clean with no errors or warnings (to fix #150).

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -0,0 +1,65 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jface.operation.IRunnableWithProgress;
+import org.eclipse.jface.viewers.IStructuredSelection;
+import org.eclipse.jface.wizard.Wizard;
+import org.eclipse.ui.INewWizard;
+import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.statushandlers.StatusManager;
+import org.slf4j.Logger;
```
elharo: Any string reason to add this dependency? Google-standard (which doesn't necessarily apply here on Github) is to use java.util.logging

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -0,0 +1,65 @@
...
+
+    IStatus status = Status.OK_STATUS;
+    try {
+      boolean fork = true;
+      boolean cancelable = true;
+      getContainer().run(fork, cancelable, runnable);
+    } catch (InterruptedException ex) {
+      status = Status.CANCEL_STATUS;
+    } catch (InvocationTargetException ex) {
+      logger.error("Unable to create new project", ex);
```
elharo: I tend to think this needs to be surfaced to the end user.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
briandealwis: SLF4j is becoming fairly widely used in Eclipse projects, though it's not a standard.  I'm happy to switch to java.util.logging.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
briandealwis: That's what the StatusManager does.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/plugin.xml
```diff
@@ -0,0 +1,45 @@
...
+   </extension>
+   <extension
+         id="id1"
+         point="org.eclipse.wst.validation.exclude">
+      <validator
+            id="org.eclipse.wst.xml.core.xml">
+         <exclude>
+            <rules>
+               <file
+                     name="nbactions.xml"
```
elharo: Can we just exclude appengine-web.xml as well rather than adding the schema? The schema really isn't necessary. I suppose we might at some point validate it at milestone 3 (the quick fixes and code inspections) but if so we'll want to pull the schema from gcloud rather than bundling it ourselves.

briandealwis: Ok, will do.  I thought given the namespace identifier was 1.0 that it hadn't changed (or in compatible ways), but looking at the [history of the XSD](https://code.google.com/p/googleappengine/source/browse/trunk/java/src/main/com/google/appengine/tools/development/appengine-web.xsd) that doesn't seem to be the case.

The AppEngine Java SDKs include a copy of the XSD in `docs/appengine-web.xsd`, as does `gcloud` in `platform/google_appengine/google/appengine/tools/java/docs/appengine-web.xsd`.

elharo: Namespaces are very confusing, schemas are very confusing, and namespaces*schemas is confusion squared. :-) Usually (though not always) namespace URIs do not change, even when the schema evolves.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
briandealwis: Since we were sending the status message to the StatusManager, I added the LOG flag too to cause it to be logged.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/plugin.xml
briandealwis: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/AppEngineStandardProjectMavenConfig.java
```diff
@@ -0,0 +1,31 @@
+
+package com.google.cloud.tools.eclipse.appengine.newproject.maven;
+
+import com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandardProjectConfig;
+
+public class AppEngineStandardProjectMavenConfig extends AppEngineStandardProjectConfig {
```
elharo: Shouldn't be too hard to add a test for this class.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -0,0 +1,60 @@
...
+import java.lang.reflect.InvocationTargetException;
+
+public class MavenArchetypeProjectWizard extends Wizard implements INewWizard {
+  private AppEngineStandardWizardPage page;
+  private AppEngineStandardProjectMavenConfig config = new AppEngineStandardProjectMavenConfig();
+
+  public MavenArchetypeProjectWizard() {
+    this.setWindowTitle("New Maven-based App Engine Standard Project");
+    page = new AppEngineStandardWizardPage(config);
+    setNeedsProgressMonitor(true);
```
elharo: Do I need to add this to the other wizard too?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -0,0 +1,60 @@
...
+    this.addPage(page);
+  }
+
+  @Override
+  public boolean performFinish() {
+    IRunnableWithProgress runnable = new IRunnableWithProgress() {
+      @Override
+      public void run(IProgressMonitor monitor)
+          throws InvocationTargetException, InterruptedException {
+        new CreateAppEngineStandardMavenProject(config).run(monitor);
```
elharo: This class is almost identical to the other wizard. We should at least add a todo to refactor to share code.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -83,11 +87,21 @@ public void modifyText(ModifyEvent event) {
     }
   }  
 
-  String getAppEngineProjectId() {
+  @Override
+  public void setPageComplete(boolean complete) {
+    if (complete) {
+      config.setAppEngineProjectId(getAppEngineProjectId());
```
elharo: The side effects and exposure of mutable internal state still make me nervous.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.properties
```diff
@@ -2,4 +2,7 @@ google = Google
 Bundle-Vendor = Google Inc.
 Bundle-Name = Wizards for Google Cloud Platform Apps
 
+gcp.category = Google Cloud Platform
+
+wizard.name = Google App Engine Standard Java Project
```
elharo: Tangential note: is there a quick way to read these properties from the code? If so, I should do that in https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/156

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
```diff
@@ -0,0 +1,86 @@
...
+  protected void execute(IProgressMonitor monitor)
+      throws CoreException, InvocationTargetException, InterruptedException {
+    SubMonitor progress = SubMonitor.convert(monitor);
+    monitor.beginTask("Creating Maven AppEngine archetype", 100);
+
+    ProjectImportConfiguration importConfiguration = new ProjectImportConfiguration();
+    IPath creationPath = config.getProject().getParent().getLocation();
+    String groupId = config.getGroupId();
+    String artifactId = config.getArtifactId();
+    String appId = config.getAppEngineProjectId();
```
elharo: projectId

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
```diff
@@ -0,0 +1,86 @@
...
+    if (appId == null || appId.trim().isEmpty()) {
+      appId = artifactId;
+    }
+    String appengineArtifactVersion = resolveLatestReleasedArtifact(progress.newChild(20),
+        "com.google.appengine", "appengine-api-1.0-sdk", "jar", "1.9.36");
```
elharo: 1.9.37

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
```diff
@@ -0,0 +1,86 @@
...
+            config.getPackageName(), properties, importConfiguration, progress.newChild(60));
+
+    Job job = new MappingDiscoveryJob(archetypeProjects);
+    job.schedule();
+  }
+
+  private String resolveLatestReleasedArtifact(SubMonitor progress, String groupId,
+      String artifactId, String type, String defaultVersion) {
+    try {
+      progress.beginTask(MessageFormat.format("Resolving latest version of {0}", artifactId), 10);
```
elharo: I think simple string concatenation works here; not need for MessageFormat

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
```diff
@@ -0,0 +1,86 @@
...
+      Artifact artifact = MavenPlugin.getMaven().resolve(groupId, artifactId, "LATEST", type, null,
```
elharo: Could you make the two null arguments named variables so this is a smidge more readable?

elharo:
What's the status on this work?

briandealwis:
I'll address your comments and extract some of the common pieces of the wizard (UI setup) into a abstract class.  That should avoid some of the difficulties in sharing strings across bundles.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/AppEngineStandardProjectMavenConfig.java
briandealwis: will do

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
briandealwis: good point.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
briandealwis: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
briandealwis: This string can be user facing (shown in the dialog progress monitor) and so should not use concatenation.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
briandealwis: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateAppEngineStandardMavenProject.java
elharo: You're going to have to educate me on this one. Why does the string being  user facing imply a need for a MessageFormat?

briandealwis: For i18n purposes, the ordering of the arguments that make sense for English may be different from the order for the user's language.  Some languages are Right-to-Left, so the actual presentation of the string will be "{0} :zzz yyy xxx".  In a multiple-argument string, {0} and {1} might be re-ordered as "xxx {1} yyy {0} zzz" instead.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
briandealwis: My experience with complex wizards is that they inevitably need to share information between pages.  If state isn't centralized in an object like this config object, then it leads to either (A) spreading the state out amongst the pages, which becomes unwieldy and hard to reason about (and part of the reason that it's necessary to have good identifiers for wizard pages), or (B) hosting the state on the wizard which leads to increased coupling and lots of casting.

All that said, Eclipse is somewhat inconsistent as demonstrated by the `WizardNewProjectCreationPage`, since we can't alter the project location programmatically.

I'll back this out.

briandealwis:
PTAL @elharo.  I didn't find much shared code in the wizards.  I found the config object was adding little value so I just configured the `CreateMavenBasedAppEngineStandardProject` operation class directly.  And added some tests.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateMavenBasedAppEngineStandardProject.java
```diff
@@ -0,0 +1,127 @@
...
+
+
+
+  @Override
+  protected void execute(IProgressMonitor monitor)
+      throws CoreException, InvocationTargetException, InterruptedException {
+    SubMonitor progress = SubMonitor.convert(monitor);
+    monitor.beginTask("Creating Maven AppEngine archetype", 100);
+
+    String appId = appEngineProjectId;
```
elharo: I guess you don't want to write this back into the field?

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateMavenBasedAppEngineStandardProject.java
briandealwis: It's an artefact of the archetype that we _have_ to provide a project id, but we're not requiring it of users.  Not entirely sure that the artifactId is the right choice to default to.

briandealwis:
I'm biting the bullet and creating a Maven-specific page to get the groupId and version, rather than sticking in defaults.  I'll submit that separately.

elharo:
Why create a new page instead of adding the two fields to the existing
page? I think we have room for them.

On Wed, May 18, 2016 at 11:07 AM, Brian de Alwis <notifications@github.com>
wrote:

> I'm biting the bullet and creating a Maven-specific page to get the
> groupId and version, rather than sticking in defaults. I'll submit that
> separately.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/146#issuecomment-220056765>
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

briandealwis:
Oops, didn't submit this yesterday.  Why create a new page?  Two reasons:

1. it's weird seeing group and version but no artifact
2. to avoid confusion with the project location (as mentioned elsewhere) and project name — the Maven archetype generator uses the artifact id as the project name, and it's out of our control.

elharo:
I think we're actually in agreement here. The code you've submitted only has a single page. I thought you proposing a two-page wizard with a second page just for the maven properties.

PR #148, opened 2016-05-10
set error message when project creation fails

elharo:
fix #144

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
@@ -9,6 +9,7 @@ import org.eclipse.jface.wizard.Wizard;
 import org.eclipse.ui.INewWizard;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.ide.undo.WorkspaceUndoUtil;
+import org.eclipse.ui.statushandlers.StatusManager;
 
 import java.lang.reflect.InvocationTargetException;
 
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -58,10 +57,26 @@ public boolean performFinish() {
...
+  void showErrorMessageIfNecessary(IStatus status) {
+    // if fail, display error message in wizard
+    if (status == Status.CANCEL_STATUS) {
+      page.setErrorMessage("User canceled project creation");
+    } else if (!status.isOK()) {
+      String message = "Failed to create project";
+      if (status.getMessage() != null && !status.getMessage().isEmpty()) {
+        message += ": " + status.getMessage();
+      }
+      page.setErrorMessage(message);
```
briandealwis: This doesn't seem right.  The page may not be the active page (for example, if we add more pages).  So at the very least this should use `getCurrentPage()` rather than `page`.  But I think for errors in this process it would be better to use the StatusManager:
```
StatusManager.getManager().show(status, StatusManager.SHOW);
```
That will pop up a dialog with the status details.

Alternatively log it.  My preference is via SLF4j (`org.slf4j.api` is pulled in both by Jetty and M2Eclipse, so is almost always present).

elharo: I definitely want to show this to the user; i.e. not log it.

I believe this puts it in the wizard page's text area up at the top. Do you think that's preferable to opening a separate dialog?

I tried your suggestion of getCurrentPage(), or rather getWizardContainer().getCurrentPage(). Unfortunately that returns the wrong type, IWizardPage instead of WizardNewProjectCreationPage; and IWizardPage doesn't have the setErrorMessage method. I could cast it, but that defeats the point of using the current page. I think we may need to live with using the page field here (unless we pop up an extra dialog) and rely on the tests to catch this if at some point we add a second page, at which time we'll have to rework this method.

briandealwis: JDT's `org.eclipse.jdt.internal.ui.wizards.JavaProjectWizard` opens an error dialog on exception during its performFinished().  I'd follow their lead.  The `WizardPage#setErrorMessage()` is more meant for validation errors.

My real problem with using `page.setErrorMessage()` is that if we have a different page, the setErrorMessage() is ignored with no warning.

elharo: OK, I've changed this to use StatusManager.getManager().handle(status, StatusManager.SHOW);

elharo:
@briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -54,14 +54,24 @@ public boolean performFinish() {
...
-      status = new Status(Status.ERROR, "todo plugin ID", errorCode, ex.getMessage(), null);
+      status = setErrorStatus(ex);
     }
     
-    // todo if fail, call setErrorMessage()
     return status.isOK();
   }
 
+  // visible for testing
+  static IStatus setErrorStatus(Exception ex) {
```
briandealwis: `showError`?

briandealwis:
lgtm

PR #152, opened 2016-05-11
remove project location from config since it's part of the project

elharo:


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -50,8 +50,7 @@
   public void execute(IProgressMonitor monitor) throws InvocationTargetException, CoreException {
     SubMonitor progress = SubMonitor.convert(monitor, 100);
     
-    // todo just use getproject().getLocationUri()
-    URI location = config.getEclipseProjectLocationUri();
+    URI location = config.getProject().getRawLocationURI();
```
briandealwis: Is there a reason you're using `getRawLocationURI()`? From the Javadocs: 

>  Since the returned location may contain unresolved variables, the resulting URI is typically only suitable for display. To access or manipulate the actual resource backing location, clients should obtain the resolved location using getLocationURI().

`getRawLocationURI()` doesn't do any interpolation of variables, etc. E.g., this could be `${workspace}/${work folder}/${blah}/${blah}/${blah}`.

briandealwis:
lgtm otherwise

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
elharo: Because when I used getLocationUri it didn't work (pressing finish errored out, presumably since it didn't close) and when I swtiched to getRawLocationUri it did. Now that we have better error reporting I can dig a little deeper on this.

elharo:
Just tried it  again and this time it worked with setLocationUri. Hmm. I'll use setLocationUri for now, but let's remember this if a bug shows up later.

PR #153, opened 2016-05-11
attach Generic Server Framework Runtime

### LINKED ISSUE #134: Configure Generated Project to build in Eclipse

elharo:
@briandealwis The wizard code is now created without any errors since the runtime supplies the necessary entries in the classpath. (We can easily change this to a different primary runtime later.) 

fix #134


briandealwis:
lgtm

Issue #161, opened 2016-05-12
IllegalArgumentException when defining new project with StandardProjectWizard


briandealwis:
I get an exception if I create a project using the native new AE project wizard when I haven’t defined a runtime using the code on the [wizard branch](dc29c80ef3b01d142b238daab1209001a0a1aa8c).  We need to create a runtime instance if necessary.  `StandardProjectWizard#canFinish()` should return false when there is no valid runtime, and either add a new wizard page or at least prompt at some point in the process.

```
java.lang.IllegalArgumentException: Runtime App Engine has not been defined.
at org.eclipse.wst.common.project.facet.core.runtime.internal.RuntimeManagerImpl.getRuntime(RuntimeManagerImpl.java:210)
at org.eclipse.wst.common.project.facet.core.runtime.RuntimeManager.getRuntime(RuntimeManager.java:120)
at com.google.cloud.tools.eclipse.appengine.newproject.CreateAppEngineStandardWtpProject.installAppEngineRuntime(CreateAppEngineStandardWtpProject.java:114)
at com.google.cloud.tools.eclipse.appengine.newproject.CreateAppEngineStandardWtpProject.execute(CreateAppEngineStandardWtpProject.java:81)
at org.eclipse.ui.actions.WorkspaceModifyOperation$1.run(WorkspaceModifyOperation.java:108)
at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2241)
at org.eclipse.ui.actions.WorkspaceModifyOperation.run(WorkspaceModifyOperation.java:130)
at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:119)
```

elharo:
I've suppressed the IllegalArgumentException. Now we need to figure out how to programmatically attach a runtime if the one we want is not already there.

elharo:
I think we're going to need to set this up in the wizard since even if we can configure it programmatically, we still need to ask the user for the various properties, most importantly the location of the gcloud SDK.

briandealwis:
Note that if we pull the runtime (compile-time) jars from the AppEngine SDK / Maven Central, then the use of GCloud is needed only in the local debug server and deployment story.

rdayal:
Did you guys figure out what you want to do here?

Is this actually an enhancement? If there's no runtime associated with this on import, isn't this a broken project?

elharo:
Yes, if there's no runtime it's a broken project.

[...]

PR #163, opened 2016-05-12
don't die if no runtime is configured

elharo:
@briandealwis This is a quick stopgap for #161 while I figure out how to add a new runtime page to the wizard.


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -49,14 +49,14 @@ public boolean performFinish() {
...
-      status = setErrorStatus(ex);
+      status = setErrorStatus(ex.getCause());
     }
     
     return status.isOK();
   }
 
   // visible for testing
-  static IStatus setErrorStatus(Exception ex) {
+  static IStatus setErrorStatus(Throwable ex) {
```
briandealwis: And if you could change the `StatusManager` to
```
StatusManager.getManager().handle(status, StatusManager.LOG | StatusManager.SHOW);
```
to log the exception too.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -111,10 +110,14 @@ static void installAppEngineRuntime(IFacetedProject project, IProgressMonitor mo
     Set<IProjectFacetVersion> facets = new HashSet<>();
     facets.add(WebFacetUtils.WEB_25);
     Set<IRuntime> runtimes = RuntimeManager.getRuntimes(facets);
-    IRuntime appEngineRuntime = RuntimeManager.getRuntime("App Engine");
-    
     project.setTargetedRuntimes(runtimes, monitor);
-    project.setPrimaryRuntime(appEngineRuntime, monitor);
+    
+    try {
+      IRuntime appEngineRuntime = RuntimeManager.getRuntime("App Engine");
```
briandealwis: `RuntimeManager.isRuntimeDefined("App Engine")`?

Does anything bad happen if we don't have a primary runtime?  I'd guess not, since that's no different from the situation where we didn't set a runtime at all.

elharo: Near as I can tell the worst that happens is that the classpath isn't set up which, as you point out, also happens without a runtime.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
elharo: RuntimeManager.isRuntimeDefined("App Engine")

done

briandealwis:
lgtm

PR #168, opened 2016-05-13
don't pop up error dialog on every keystroke

### LINKED ISSUE #166: Excess validation
1. Create a new GAE standard runtime 
2. Browse to find the SDK directory
3. Delete the SDK directory in the text field

You'll see this dialog:

image

even though we've only gone back to the state we were in.

Similarly editing the path will pop up dialogs for directories that don't exist on every keystroke. This makes typing very hard.

We should not pop up these error dialogs here.

elharo:
fix #166

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
@@ -14,6 +14,8 @@
  *******************************************************************************/
 package com.google.cloud.tools.eclipse.appengine.localserver.ui;
 
+import java.io.File;
+
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
@@ -51,7 +53,6 @@ public final class CloudSdkRuntimeWizardFragment extends WizardFragment {
   private IWizardHandle wizard;
   private CloudSdkRuntime runtime;
   private Text dirTextBox;
-  private int status = IStatus.ERROR;
 
   @Override
   public Composite createComposite(Composite parent, IWizardHandle handle) {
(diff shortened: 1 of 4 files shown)


elharo:
@briandealwis PTAL

briandealwis:
I came across this during as part of the re-org work that I (temporarily) abandoned.  There are a few things I'd improve on this:

  1. The validation job is conflating its validation status with the job status (and why we're getting those errors) — they are independent.  Rather than pass the validation status to the follow-on listener as the job result, we can instead use the `Job.ASYNC_FINISH` result to indicate that the job is being moved to a different thread.
  2. Since we're creating a new job on each keystroke, we could conceivably have many validation jobs running simultaneously (should they prove long-lived).  As jobs are reschedulable, we should have a single job and then if running, cancel and re-schedule it as needed.  We can also schedule it to happen 50 milliseconds after to avoid continual churn.

I'll integrate my patch into yours and will update here in a few minutes.

briandealwis:
PTAL @elharo.  

It's a bit distracting seeing the flickering header especially when I know the path I'm providing is incomplete.  A better user experience might be to have a focus listener and a longer delay, so we only validate if they switch focus to a different field, or after 200 ms.  And remove the _Validating_ message.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
```diff
@@ -163,34 +194,7 @@ private void validate() {
...
-              updateStatus(null, IStatus.OK);
-            }
-            wizard.update();
-          }
-        });
-      }
-    });
-    job.schedule();
+    validateLocationJob.cancel();
+    validateLocationJob.schedule(20); // small wait in case of more keystrokes
```
elharo: That's in milliseconds, right? That seems short. Maybe 100ms?

elharo:
LGTM modulo change from 20ms to 200ms, and fixing the CLA.

elharo:
The CLA problem is because there are commits by multiple authors. If you squash the commits and 

git push -f

that should be fixed.

briandealwis:
I had it at 200ms in an earlier patch but it seemed a bit laggy.  Given that the validation is already a bit laggy, it might feel very slow.  But let's give it a try.

Issue #157, opened 2016-05-11
Point gcloud_serverdef.xml webappDirectory property at src/main/webapp


elharo:
Per https://www.ibm.com/support/knowledgecenter/SSAW57_6.1.0/com.ibm.websphere.nd.doc/info/ae/ae/trun_genericsvr_create.html we might be able to use ${JAVA_HOME} style variables

elharo:
Try path variables: http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.user%2Fconcepts%2Fcpathvars.htm especially PROJECT_LOC

elharo:
obsolete now that GSF is gone

PR #171, opened 2016-05-16
project loc environment variable

elharo:
#157

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/gcloud_serverdef.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/gcloud_serverdef.xml
@@ -21,7 +21,7 @@
       label="Web app directory (parent of WEB-INF)"
       type="directory"
       context="server"
-      default="/usr/local/www/WebContent" />
+      default="${PROJECT_LOC}/src/main/webapp" />
   <property id="autoDeployDirectory"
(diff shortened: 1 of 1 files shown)


briandealwis:
This doesn't do much unfortunately.

screen shot 2016-05-18 at 3 54 28 pm

elharo:
The question is whether it will find the right directory when attached. If so, we can futz with the error detection.

Or is there API somewhere to resolve ${PROJECT_LOC} from inside out code?

relevant info:

https://bugs.eclipse.org/bugs/show_bug.cgi?id=229633
http://help.eclipse.org/mars/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2FresInt_linked.htm

Or maybe what we need to do is allow relative paths here and resolve them ourselves  later during project creation?

briandealwis:
I think we've misunderstood the purpose of this field: this is a directory where our server support should publish the webapps-to-be-run in a format supported by our dev server.  So for a tomcat installation, this might be `/opt/lib/tomcat/webapps/`. Since we're not expecting to have a long-lived deployments, I think these fields are internal and should not be visible to the user.

elharo:
Yes, I need to rethink how, when, and where the exploded war gets passed to dev_appserver.py

elharo:
No longer relevant

Issue #165, opened 2016-05-13
Common resources


elharo:
e.g. app_engine_standard_16x16.png, other icons, and messages and names

We want to move these out of the individual plugins into some central location.

PR #172, opened 2016-05-17
add keywords for easy autocomplete

### LINKED ISSUE #116: add elements to the wizard description for additional autocomplete in new project...
You can also add elements to the wizard description. Requires adding an extension to the org.eclipse.ui.keywords extension point. Probably worth creating keywords for:

Google
AppEngine
GAE
GCP
These keywords don't show up in the UI. So when we do the AE Flex tooling, we could add MVM for those who don't like the new naming.

elharo:
close #116 
@nbashirbello @briandealwis


briandealwis:
I'd suggest that we the keyword definitions into a common `.ui` bundle, that could also hold the images and graphics.  This won't be the only place needing these keywords.

elharo:
To make sure I understand you, this will be the one big bundle with all our UI strings, icons, etc.? I.e. make this part of #165 ?

briandealwis:
Yeah — but I'm not sure about the strings, since sharing strings is kinda hard.

elharo:
PTAL @briandealwis 
Now pulls in the commonly defined keywords.

elharo:
Test failure looks like this:

[INFO] Fetching p2.index from http://download.eclipse.org/releases/neon/201605061000/ (0B at 0B/s)
[ERROR] Failed to resolve target definition /home/travis/build/GoogleCloudPlatform/gcloud-eclipse-tools/eclipse/neon/gcp-eclipse-neon.target: Could not find "org.eclipse.sdk.feature.group/4.6.0.v20160317-0200" in the repositories of the current location -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MavenExecutionException

May not be related, but I'm investigating

plugins/com.google.cloud.tools.eclipse.appengine.newproject/META-INF/MANIFEST.MF
```diff
@@ -14,6 +14,7 @@ Require-Bundle: org.eclipse.ui,
  org.eclipse.ui.ide,
  org.eclipse.jst.j2ee.web
 Import-Package: com.google.cloud.tools.eclipse.appengine.localserver.runtime,
+ com.google.cloud.tools.eclipse.appengine.ui,
```
briandealwis: You don't actually technically need to reference a bundle to pull in these references within the plugin.xml.  But such a reference does ensure that the other bundle will be pulled in as a dependency. But since we're referencing resources within the bundle, rather than classes from a package, then I think this might be better as a Require-Bundle.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
```diff
@@ -11,9 +11,17 @@
             category="com.google.cloud.tools.eclipse.appengine.wizards">
          <description>%wizard.description</description>
          <selection class="org.eclipse.core.resources.IResource"/>
+         <keywordReference id="com.google.cloud.tools.eclipse.appengine.ui.GCPKeyword"/>
```
briandealwis: This would be a great plugin.xml test: ensure the keyword references are found.

briandealwis:
LGTM.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
elharo: Is there API to check that?

briandealwis: (I apparently didn't click the submit button, so this wasn't sent yesterday)

> Is there API to check that?

 Nothing beyond using the IExtensionRegistry directly.

PR #174, opened 2016-05-17
better checking of package names

### LINKED ISSUE #173: Disallow keywords in Java package names
Currently the wizard allows the creation of package names such as 

com.new.int.for

elharo:
@nbashirbello fix #173 and also disallow package names with two or more periods in a row


briandealwis:
Maybe we should just punt to JDT's `org.eclipse.jdt.core.JavaConventions`?

```
JavaConventions.validatePackageName(text, JavaCore.VERSION_1_3, JavaCore.VERSION_1_3);
```

(That's what's used by JDT's `NewPackageWizardPage`.)

elharo:
validatePackageName looks interesting though we'll still need to wrap it. In my tests it allows "com. google" (note the space after the internal period) due to this code:

		while (st.hasMoreTokens()) {
			String typeName = st.nextToken();
			typeName = typeName.trim(); // grammar allows spaces


I have no idea why it does that.

elharo:
PTAL @briandealwis

briandealwis:
Very weird!  But they claim that [spaces at the beginning or end of a package component is valid](https://bugs.eclipse.org/bugs/show_bug.cgi?id=255040#c1).  Which it is, and necessarily so for line wrapping.  
```java
package  foo  . bar  . baz;
```
The JDT code ignores the whitespace and strips in the generated code.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -13,27 +16,26 @@ public static boolean validate(String packageName) {
...
+      // note very weird condition because validatePackageName allows internal white space
+      return false;
     } else {
-      String[] parts = packageName.split("\\.");
-      for (int i = 0; i < parts.length; i++) {
-        if (!isValidJavaName(parts[i])) {
-          return false;
-        }
-      }
+      return JavaConventions.validatePackageName(packageName, "1.4", "1.4").isOK();
```
briandealwis: I think we should use whatever version we're generating in case there are other reserved words that come into play.  And we should the JavaCore constants ``JavaCore#VERSION_1_4``?

briandealwis:
Otherwise LGTM

elharo:
I'm trying to find the section of the JLS that allows spaces in package declarations, but so far I don't see it: https://docs.oracle.com/javase/specs/jls/se7/html/jls-7.html#jls-7.4.1

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
elharo: done

elharo:
If we allow spaces in package names, we'll also need to fix our project generation code to account for that.

Even of that is legal I suspect there's still an Eclipse bug here. They're stripping all characters <= 0x20 and most of those don't count as whitespace in this context.

briandealwis:
The section on _PackageName_ references [§6.7 in the JLS](https://docs.oracle.com/javase/specs/jls/se7/html/jls-6.html#jls-6.7) which talks of a package name being made up of simple identifiers.   [§3.5 on Input Elements and Tokens](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.5) says whitespace can appear between any tokens, and a token may be an identifier.  That's about as far as I want to go :-)

All that said, I'm fine with not allowing spaces.  It seems an esoteric case, and a formatter would strip them out anyways.  The relevant JDT code in `CreatePackageFragmentOperation` eventually calls `org.eclipse.jdt.internal.core.Util#getTrimmedSimpleNames()` to return a whitespace-trimmed package name; it uses `org.eclipse.jdt.core.Signature#getSimpleNames()` and that applies `String#trim()` on the components.

PR #175, opened 2016-05-17
don't create project on top of existing directory

### LINKED ISSUE #130: Test what happens when creating a new project on top of existing directory
New App Engine Wizard
e.g. when deleted old project from eclipse but not file system

elharo:
fix #130 

Open questions:

1. Should we also check in CreateAppEngineStandardWtpProject?
2. Should the user have an option to overwrite an existing directory? (My gut says no. This is just a wizard. They can always do that manually if they want to, but let's not make it easy for the user to shoot themselves in the foot.)

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
@@ -1,6 +1,7 @@
 package com.google.cloud.tools.eclipse.appengine.newproject;
 
 import java.io.File;
+import java.net.URI;
 
 import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.wizard.IWizardPage;
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
@@ -1,5 +1,7 @@
 package com.google.cloud.tools.eclipse.appengine.newproject;
 
+import java.io.File;
+
 import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.wizard.IWizardPage;
 import org.eclipse.ui.dialogs.WizardNewProjectCreationPage;
@@ -82,6 +84,13 @@ class AppEngineStandardWizardPage extends WizardNewProjectCreationPage implement
     String projectId = projectIdField.getText();
     if (!AppEngineProjectIdValidator.validate(projectId)) {
       setErrorMessage("Illegal App Engine Project ID: " + projectId); //$NON-NLS-1$
+      return false;
+    }
+    
+    File parent = getLocationPath().toFile();
+    File projectDirectory = new File(parent, getProjectName());
+    if (projectDirectory.exists()) {
+      setErrorMessage("Project location already exists: " + projectDirectory); 
       return false;
     }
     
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -85,6 +87,13 @@ private boolean validateLocalFields() {
       return false;
     }
     
+    File parent = getLocationPath().toFile();
+    File projectDirectory = new File(parent, getProjectName());
```
briandealwis: So there's a bug here (and I meant to file it last night, except I accidentally installed an OS update and it was lost due to a reboot).  When 'default' is ticked (which is the default), then this code is correct: the project is created in the workspace.  But if the user has unticked 'default' then they specify a location for the *project* content, and the project name is unused in determining the directory name.  (It's a bit confusing, but I guess it's because there's no way other way for a user to say they want a directory different from the project name.)

I'd embed this knowledge into a `getNewProjectPath()` or something like that.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -52,14 +52,13 @@
...
-  public void execute(IProgressMonitor monitor) throws InvocationTargetException, CoreException {
+  public void execute(IProgressMonitor monitor) throws InvocationTargetException, CoreException {    
     SubMonitor progress = SubMonitor.convert(monitor, 100);
     
-    URI location = config.getProject().getLocationURI();
-    
     IWorkspace workspace = ResourcesPlugin.getWorkspace();
     IProject newProject = config.getProject();
-    
+    URI location = newProject.getLocationURI();
```
briandealwis: Following on to the previous comment: when 'defaults' is not ticked, then getLocationURI won't be correct.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
elharo: Agreed there's a bug, but if there's a project name shouldn't we use that in determin where to put the project? I.e. it sounds to me like if the user selects

/usr/local/tmp with a project name of sample, then we should create the project in 

/usr/local/tmp/sample

elharo: Actually, checking this now it looks like we're ignoring the custom location completely and always putting it into the workspace default.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
elharo: I think that relying on the wizard to put the location in the project handle was a mistake. I'm backing that out now.

elharo:
PTAL

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
briandealwis: A project's name is taken from the .project file, and it can differ from the project directory.

briandealwis:
LGTM.

> 1. Should we also check in CreateAppEngineStandardWtpProject?

They'll get an IllegalArgumentException, won't they?

> 2. Should the user have an option to overwrite an existing directory? (My gut says no. This is just a wizard. They can always do that manually if they want to, but let's not make it easy for the user to shoot themselves in the foot.)

I don't see how it would be useful behaviour.

PR #177, opened 2016-05-18
more specific error messages on invalid package names

elharo:
@nbashirbello

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
@@ -16,15 +16,15 @@ public class JavaPackageValidator {
   public static IStatus validate(String packageName) {
     
     if (packageName == null) {
-      return new Status(1, PLUGIN_ID, 45, "null package name", null);
+      return new Status(IStatus.ERROR, PLUGIN_ID, 45, "null package name", null);
     } else if (packageName.isEmpty()) { // default package is allowed
       return Status.OK_STATUS;
     } else if (packageName.endsWith(".")) {
       // todo or allow this and strip the period
-      return new Status(1, PLUGIN_ID, 46, packageName + " ends with a period.", null);
+      return new Status(IStatus.ERROR, PLUGIN_ID, 46, packageName + " ends with a period.", null);
     } else if (containsWhitespace(packageName)) {
       // very weird condition because validatePackageName allows internal white space
-      return new Status(1, PLUGIN_ID, 46, packageName + " contains whitespace.", null);
+      return new Status(IStatus.ERROR, PLUGIN_ID, 46, packageName + " contains whitespace.", null);
     } else {
       return JavaConventions.validatePackageName(
           packageName, JavaCore.VERSION_1_4, JavaCore.VERSION_1_4);
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
@@ -7,21 +7,23 @@ import org.eclipse.jdt.core.JavaCore;
 
 public class JavaPackageValidator {
 
+  private static final String PLUGIN_ID = "com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandard";
+  
   /**
    * Check if a string is a legal Java package name.
    */
   public static IStatus validate(String packageName) {
     
     if (packageName == null) {
-      return new Status(1, "pluginId", 45, "null package name", null);
+      return new Status(1, PLUGIN_ID, 45, "null package name", null);
     } else if (packageName.isEmpty()) { // default package is allowed
       return Status.OK_STATUS;
     } else if (packageName.endsWith(".")) {
       // todo or allow this and strip the period
-      return new Status(1, "pluginId", 46, packageName + " ends with a period.", null);
+      return new Status(1, PLUGIN_ID, 46, packageName + " ends with a period.", null);
     } else if (containsWhitespace(packageName)) {
       // very weird condition because validatePackageName allows internal white space
-      return new Status(1, "pluginId", 46, packageName + " contains whitespace.", null);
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -1,28 +1,33 @@
...
 
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.jdt.core.JavaConventions;
 import org.eclipse.jdt.core.JavaCore;
 
 public class JavaPackageValidator {
 
+  private static final String PLUGIN_ID = 
+      "com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandard";
```
briandealwis: Technically, this isn't a plugin id, but it does work.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -1,28 +1,33 @@
...
-      return false;
-    } else if (packageName.isEmpty()) {
-      return true;
+      return new Status(1, PLUGIN_ID, 45, "null package name", null);
+    } else if (packageName.isEmpty()) { // default package is allowed
+      return Status.OK_STATUS;
     } else if (packageName.endsWith(".")) {
       // todo or allow this and strip the period
-      return false;
+      return new Status(1, PLUGIN_ID, 46, packageName + " ends with a period.", null);
```
briandealwis: These should use `MessageFormat.format()` with positional elements.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -1,28 +1,33 @@
...
    */
-  // todo return an IStatus for better error reporting
-  public static boolean validate(String packageName) {
+  public static IStatus validate(String packageName) {
+    
     if (packageName == null) {
-      return false;
-    } else if (packageName.isEmpty()) {
-      return true;
+      return new Status(1, PLUGIN_ID, 45, "null package name", null);
```
briandealwis: Use the codes on `IStatus`, like `ERROR`, `WARNING`, etc rather than numbers

elharo: Done, I guess this is a pre-enum API.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
elharo: Lets' chat about this in the meeting today.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -16,15 +16,15 @@
   public static IStatus validate(String packageName) {
     
     if (packageName == null) {
-      return new Status(1, PLUGIN_ID, 45, "null package name", null);
+      return new Status(IStatus.ERROR, PLUGIN_ID, 45, "null package name", null);
```
nbashirbello: Where does the 45 and 46 codes come from?

elharo: I made them up, no particular reason except that if one shows up I'd like to be able to find the relevant section of code easily. Now that I think about I should pick a value even less likely to appear by coincidence such as 9423757 or 1543234

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -1,28 +1,33 @@
...
+      "com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandard";
+  
   /**
    * Check if a string is a legal Java package name.
    */
-  // todo return an IStatus for better error reporting
-  public static boolean validate(String packageName) {
+  public static IStatus validate(String packageName) {
+    
     if (packageName == null) {
```
briandealwis: Why don't you let the `JavaConventions` do the heavy lifting and then do any mop-ups, like contains whitespace?  It should already do the package-ends-with-a-dot, and has standard (and localized) messaging

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/JavaPackageValidator.java
```diff
@@ -1,28 +1,33 @@
...
+      return Status.OK_STATUS;
     } else if (packageName.endsWith(".")) {
       // todo or allow this and strip the period
-      return false;
+      return new Status(IStatus.ERROR, PLUGIN_ID, 46, packageName + " ends with a period.", null);
     } else if (containsWhitespace(packageName)) {
-      // note very weird condition because validatePackageName allows internal white space
-      return false;
+      // very weird condition because validatePackageName allows internal white space
+      return new Status(IStatus.ERROR, PLUGIN_ID, 46, packageName + " contains whitespace.", null);
```
briandealwis: MessageFormat.format()

elharo: done

elharo:
@briandealwis / @nbashirbello Any strong thoughts/suggestions about the error codes? Otherwise, PTAL.

nbashirbello:
LGTM

briandealwis:
Those error codes are only useful if you want code to be able to react to handle different error or warning scenarios.  If they're just being reported to the user, it doesn't much matter.

PR #187, opened 2016-05-19
When creating a new App Engine project, creates a new runtime if GAE doesn't exist

nbashirbello:
Fix for #161


nbashirbello:
@elharo @briandealwis

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -116,7 +120,22 @@ static void installAppEngineRuntime(IFacetedProject project, IProgressMonitor mo
       IRuntime appEngineRuntime = RuntimeManager.getRuntime("App Engine");
       project.setPrimaryRuntime(appEngineRuntime, monitor);
     } else {
-      // todo figure out how to create a new App Engine runtime
+      // Create a new App Engine runtime
+      IRuntimeType appEngineRuntimeType =
+        ServerCore.findRuntimeType("com.google.cloud.tools.eclipse.gcloud.runtime");
+      if (appEngineRuntimeType == null) {
+        throw new NullPointerException("Null App Engine server runtime");
```
elharo: How likely is this to happen and what does it indicate? I'm wondering if maybe this should be a checked exception.

Assuming this indicates an unlikely programming problem (in which case a runtime exception is called for) please change the message to "Could not locate App Engine runtime")

elharo:
LGTM

PR #189, opened 2016-05-20
Start sharing the icons between plugins

elharo:
#165 

also checks in eclipse settings for the ui package


elharo:
@briandealwis Please check that this looks like a reasonable way to share icons between plugins.

briandealwis:
Looks sane.  Here are some thoughts:

  - should the new bundle be `com.google.cloud.tools.eclipse.ui` rather than `.appengine.ui`? Hmm, given the growth in products in GCP, maybe not.
  - it's worth adding an images class that holds `ImageDescriptor`s for the various images, so we aren't hardcoding locations in our code. `AppEngineImages`?
  - Eclipse uses a naming convention to separate out images by purpose (and hence size); I'm not suggesting we adopt this name as-is, but the differentiation is useful.
       - elcl16: enabled local item (toolbar or menu), 16x16
       - dlcl16: disabled local item, 16x16
       - clcl16: hot (hovered) local item, 16x16
       - obj16: object item
       - dtool16: disabled tool (view, wizard), 16x16
       - etool16: enabled tool (view, wizard), 16x16
       - wizban: wizard banner items, 75x66
   - oh, and Eclipse Neon has support for @2x style images which is handled transparently by `ImageDescriptor`

elharo:
PTAL @briandealwis now includes an AppEngineImages class as suggested

briandealwis:
Rather than the code knowing what size needed (and 32 pixels is wrong for the wizard — it should be 75 x 66 pixels), it's better to provide a purpose (e.g., for a Wizard Banner) and let the provider figure it out.

But simpler is to just have these be static final fields that hold onto an `ImageDescriptor`.  E.g., `GCP_WIZBANNER`, `GCP_TOOL`, `GCP_OBJ`.  The ImageDescriptors are lightweight.

elharo:
ATM these are bitmaps and these are the sizes we have. Can we use resizable vector graphics for these? Won't happen immediately, but might be worth looking into.

I'm not sure the provider should be responsible for figuring out what size is needed. The caller has that information, not the AppEngineImages class.

briandealwis:
Unfortunately Eclipse doesn't have support for SVG or resizable images.  One of our very artistically talented committers rebuilt the Eclipse icon set in SVG, but they came out to about 100MB in size.  So we instead generate them at various sizes.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -50,11 +51,7 @@ public MavenAppEngineStandardWizardPage() {
     super("basicNewProjectPage"); //$NON-NLS-1$
     setTitle("Maven-based App Engine Standard Project");
     setDescription("Create new Maven-based App Engine Standard Project");
-    
-    // todo: pull this from commons
-    ImageDescriptor descriptor = AbstractUIPlugin.imageDescriptorFromPlugin(
-        "com.google.cloud.tools.eclipse.appengine.localserver", "icons/gcp-32x32.png"); //$NON-NLS-1$ //$NON-NLS-2$
-    this.setImageDescriptor(descriptor);
+    setImageDescriptor(AppEngineImages.googleCloudPlatform(32));
```
briandealwis: > I'm not sure the provider should be responsible for figuring out what size is needed. The caller has that information, not the AppEngineImages class.

This code is requesting the wrong size for this situation.  I think the best we can right now is to describe the context for which the code is requested, and let the provider figure out the best image for that location.

elharo: Again, I do not think that the AppEngineImages class should decide what size is needed based on a context. That putz the  knowledge in the wrong location and tightly couples this package to its clients. For loosest coupling, all AppEngineImages (and the ui package) should know is which images are available. That's it. The caller is the one that knows what size it wants. 

If we can't provide a resolution independent vector graphic, then the caller has to explicitly or implicitly request an image at a particular size.

elharo: One reason to prefer a method over public fields is that the method is more flexible. E.g. we could store vector graphics and autogenerate the relevant size of bitmap when requested.

elharo: Seems switching to fields broke something. Maybe to do with initialization? E.g. images can't be loaded when the static initializer is run?

elharo:
Ping. PTAL.

elharo:
@nbashirbello Brtian's on vacation, so please take over this review.

nbashirbello:
LGTM

Issue #188, opened 2016-05-19
Invalid error when validating App Engine runtime


nbashirbello:
When you add a new App Engine runtime whether through the New Server Runtime Environment wizard or when creating a new project, the following error message appears:

Missing classpath entry /usr/.../google-cloud-sdk/platform/google_appengine/google/appengine/tools/java/lib/shared/servlet-api.jar

even though servlet-api.jar exists in the specified location.

elharo:
The issue here is that when the new runtime is created in the code, it gets the default location for the google-cloud-sdk:

 /usr/local/google/google-cloud-sdk

when you manually configure a runtime  you have the opportunity to pick the location of this SDK.

We need to find the google-cloud-sdk somehow, either by asking the user or perhaps by using the https://github.com/GoogleCloudPlatform/app-tools-lib-for-java library. Then we need to configure it when we set up a new SDK.

elharo:
This value is initially set in gcloud_serverdef.xml like so:

  <property id="cloudSdkDirectory"
      label="Directory where gcloud is installed"
      type="directory"
      context="runtime"
      default="/usr/local/google/google-cloud-sdk" />

I don't think there's any path we can put here that works universally. What we need to figure out is how an alternative value can be supplied in CreateAppEngineStandardWtpProject.installAppEngineRuntime

Then we can write some code to find the correct location on the user's system.

elharo:
believe this is loaded in org.eclipse.jst.server.generic.internal.xml.XmlUtils.readDefinitionFile to produce a org.eclipse.jst.server.generic.servertype.definition.ServerRuntime object. This ServerRuntime has properties and that's what we need to set.

I'm not yet sure how ServerRuntime is connected to our IRuntime.

elharo:
Also see GenericServerRuntime.getServerInstanceProperties

This may be where the value is stored.

elharo:
OK, here's the trick:


	  IRuntimeWorkingCopy appEngineRuntimeWC = appEngineRuntimeType.createRuntime(null, monitor);
	  
	  RuntimeWorkingCopy mutator = (RuntimeWorkingCopy) appEngineRuntimeWC;
	  Map map = mutator.getAttribute("generic_server_instance_properties", new HashMap<>());
	  map.put("cloudSdkDirectory", "/Users/elharo/google-cloud-sdk");
	  mutator.setAttribute("generic_server_instance_properties", map);


[...]

PR #191, opened 2016-05-22
locate cloud SDK

elharo:
@nbashirbello 

start work on #188 
This does not yet actually locate the SDK. That can be done in the next PR, but it does configure it once we find it.


elharo:
PTAL @nbashirbello

nbashirbello:
LGTM

Issue #193, opened 2016-05-23
GSF server cannot load classes


elharo:
To reproduce (after fixing #181)

1. Create a new application using the wizard
2. "Run on Server" from the context menu

probably need to adjust build path or classpath some way.

INFO     2016-05-23 14:20:17,098 devappserver2.py:769] Skipping SDK update check.
INFO     2016-05-23 14:20:17,221 api_server.py:205] Starting API server at: http://localhost:65152
INFO     2016-05-23 14:20:17,250 dispatcher.py:197] Starting module "default" running at: http://localhost:8080
INFO     2016-05-23 14:20:17,253 admin_server.py:116] Starting admin server at: http://localhost:8000
May 23, 2016 2:20:18 PM com.google.appengine.tools.development.AbstractContainerService configure
WARNING: Null value for containerConfigProperties.get(devappserver.portMappingProvider)
May 23, 2016 2:20:19 PM com.google.apphosting.utils.jetty.JettyLogger info
INFO: Logging to JettyLogger(null) via com.google.apphosting.utils.jetty.JettyLogger
May 23, 2016 2:20:19 PM com.google.apphosting.utils.jetty.JettyLogger info
INFO: jetty-6.1.x
May 23, 2016 2:20:19 PM com.google.apphosting.utils.jetty.JettyLogger warn
WARNING: EXCEPTION 
java.lang.ClassNotFoundException: c.f.HelloAppEngine
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at com.google.appengine.tools.development.IsolatedAppClassLoader.loadClass(IsolatedAppClassLoader.java:199)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.mortbay.util.Loader.loadClass(Loader.java:91)
	at org.mortbay.util.Loader.loadClass(Loader.java:71)
	at org.mortbay.jetty.servlet.Holder.doStart(Holder.java:73)
	at org.mortbay.jetty.servlet.ServletHolder.doStart(ServletHolder.java:242)
	at org.mortbay.component.AbstractLifeCycle.start(AbstractLifeCycle.java:50)
	at org.mortbay.jetty.servlet.ServletHandler.initialize(ServletHandler.java:685)
	at org.mortbay.jetty.servlet.Context.startContext(Context.java:140)
[...]

PR #196, opened 2016-05-24
Build war archive

elharo:
@nbashirbello This PR simply builds a war archive from the web application. It does not yet supply it to the dev_appserver at the correct time. 

I may need to add a step to the deploy task that relaunches the dev_appserver to make it pick this up. 

#193


nbashirbello:
LGTM!

PR #197, opened 2016-05-24
remove legacy deploy classes

elharo:
@nbashirbello Code @akerekes  pulled over from GPE that we don't need, at least not yet.


nbashirbello:
lgtm

PR #210, opened 2016-06-01
Update README.md with details to regenerate the .tpd files

briandealwis:
PTAL @elharo

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This project provides an Eclipse plugin for building, debugging, and deploying G
 1. Eclipse 4.5 (Mars) or later
 
 1. The [m2eclipse plugin](http://www.eclipse.org/m2e/) (also called m2e) installed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ This project provides an Eclipse plugin for building, debugging, and deploying G
 1. Eclipse 4.5 (Mars) or later
 
 1. The [m2eclipse plugin](http://www.eclipse.org/m2e/) (also called m2e) installed
-to import the projects into Eclipse.
+to import the projects into Eclipse. It's easiest to use the _Eclipse IDE for Java EE Developers_.
```
elharo: This should be a separate point below the list.

briandealwis: I'll move it to the first point (use Mars or later) as it seems most relevant there.

elharo:
LGTM

PR #212, opened 2016-06-02
standardize font sizes in dialog per user prefs

elharo:
@nbashirbello


elharo:
If you haven't changed your font defaults, this is a no-op but for those of use with less than perfect vision who like to use bigger fonts than normal, this is significant.

elharo:
@nbashirbello

nbashirbello:
LGTM!

PR #215, opened 2016-06-03
Pull in dependencies from Maven Central (#195)

### LINKED ISSUE #195: com.google.gson vs com.google.code.gson
Why are we depending on com.google.gson (which I don't see in Maven central) vs com.google.code.gson? What is the difference between these two?

briandealwis:
Add Maven dependencies for com.google.son and the new app-engine-tools-for-java / com.google.cloud.tools.app.lib.

We add these dependencies to the top pom instead of following the normal Maven approach of putting the dependencies within the requesting poms.  With Tycho 0.24, these dependencies bleed through and require being added to consumers of those bundles and features too.

Fixes #195 
PTAL @elharo


elharo:
LGTM

briandealwis:
I realize the one problem with this approach: it doesn't play well within Eclipse as these bundles aren't found on the target platform, and they're not brought in by m2eclipse.

elharo:
On Fri, Jun 3, 2016 at 1:56 PM, Brian de Alwis <notifications@github.com>
wrote:

> roach: it doesn't play well within Eclipse as these bundles aren't found
> on the target platform, and they're not brought in by m2ec


It still builds at the command line, I assume? Still it would be extremely
helpful to make this work in Eclipse. Any ideas?


-- 
Elliotte Rusty Harold
elharo@macfaq.com

briandealwis:
As an out-of-the-box solution, the only thing I can think of right now is to setup a separate repository and use that for the target.

chanseokoh:
So, is it normal that my Eclipse reports "Bundle 'com.google.json' cannot be resolved"? Is there a workaround to launch a debug session on Eclipse?

elharo:
Not yet, I'm afraid. Here's an idea: what if we added a lib directory with
the relevant jars and pointed Eclipse at them manually? Or maybe a separate
project on the build path that contributed these dependencies? It's a hack,
but if it doesn't clutter our repo or commit lists, it might enable us to
iterate faster.

On Mon, Jun 6, 2016 at 5:32 PM, Chanseok Oh <notifications@github.com>
wrote:

> So, is it normal that my Eclipse reports "Bundle 'com.google.json' cannot
> be resolved"? Is there a workaround to launch a debug session on Eclipse?
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly, view it on GitHub
> <https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/215#issuecomment-224095120>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AA9X6PBIz1jrmVv8ocSZyhVAQV3oBfZJks5qJJH-gaJpZM4ItKyB>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

briandealwis:
It's a total hack, but a necessary one.  I'll get this finally done this morning.

Issue #139, opened 2016-05-05
Analytics Utility for tracking client side events


elharo:
Via GA: likely along the lines of AnalyticsPingManager in the old plugin though hopefully simpler

chanseokoh:
As mentioned in #142, I'm thinking of using [https://github.com/google/cloud-reporting](https://github.com/google/cloud-reporting) since it's what the Concord and the DI Metrics teams offer for client-side Java (according to their docs).

We need to hide an AnalyticsID in the codebase though.

chanseokoh:
If you prefer, we can use AnalyticsPingManager from the old plugin.

However, some modifications are necessary because we will send pings to the Concord GA property (i.e., not to our own GA properties). Unlike our old GPE and CT4IJ, this Concord Analytics ID is set up as a "web" type instead of an "Android app" type (because it was designed to accommodate metrics from all Cloud products).

- Mainly, we need to send a 'pageview' ping ([should be `t=pageview`](https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L73)) instead of an 'event' ping ([currently, it is `t=event` in AnalyticsPingManager](https://github.com/GoogleCloudPlatform/google-plugin-for-eclipse/blob/master/plugins/com.google.gdt.eclipse.suite.update/src/com/google/gdt/eclipse/suite/update/usage/AnalyticsPingManager.java#L183)).
  - For example, this requires a virtual "web" URL and a virtual "webpage" title.
- That said, we need to follow the contract/interface that Concord wants.
  - Some custom dimensions (mostly optional though).
  - A complete example can be found [here](https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L64) (the library that Concord offers).

patflynn:
Why not use the cloud-reporting project for this?

chanseokoh:
We had a very brief discussion, and the general consensus was that

- Making our code depend on an external library, particularly if it is not published into the Maven Central, seems a bad idea. In fact, the library is not actively maintained.

- Nevertheless, a small piece of code (possibly a single class) will do the job, so it is ok in whatever ways this is done even though it may result in duplicate code in IntelliJ and Eclipse.

chanseokoh:
Closed by #216.

Issue #138, opened 2016-05-05
Get user consent for analytics on install


elharo:
Need to explicitly ask user to consent to usage tracking via GA when plugin is installed (or updated if from a pre-analytics version)

chanseokoh:
@briandealwis I am working on displaying a opt-in dialog (pop-up, notification, or whatever) that asks for a user consent on collecting usage metrics. I first tried `org.eclipse.jface.dialogs.PopupDialog`, but this pop-up is way too transient in that it disappears as soon as a user clicks on anywhere else. I am now trying to extend `org.eclipse.jface.dialogs.Dialog` directly to have a custom dialog, but I wonder if there is a better option. Ideally, it should be a small dialog having non-modal, less-intrusive, and notification-like characteristics.

briandealwis:
Hmm, I thought there was a way to prevent the transient-ness of the `InputDialog`, but nothing pops up (hohoho).  Given the importance of the user's decision, wouldn't it be best to use a normal modal dialog?

chanseokoh:
@briandealwis Now, we should decide when and how to show the Analytics opt-in dialog. (It's a matter of calling [a method](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java#L179), which can be called many times.)

@rdayal suggested to do so at the "activation" time of our plugin (doing lazily and in the right context) and ask for advice from you. Should the method call be in `Activator.start()` in each bundle that has an `Activator`? Will this work okay technically? Any opinions/suggestions/concerns?

chanseokoh:
@briandealwis Any ideas about https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/issues/138#issuecomment-227902048?

briandealwis:
Sorry @chanseokoh I missed the above in the noise of Github notifications.  I think we need to both do it from the activator, which should ideally coincide with the first trigger of a Cloud Tools action.  But bundles may be activated in states where we can't prompt the user (below), so we also need to be able to prompt from within the ping code have an unknown user-confirmation state.

[...]

PR #216, opened 2016-06-06
Add AnalyticsPingManager

chanseokoh:
In relation to #139.

- Adds a new OSGi bundle responsible for sending usage metrics.
- Stores user consent to send usage metrics as part of `ConfigurationScope` user preferences.
- Generates and stores a unique client ID (for the purpose of estimating the number of unique users) as a user preference.
- Uses `templating-maven-plugin` to auto-generate a Java source class. The auto-generated class is meant for injecting a Google Analytics tracking ID at compile-time. The purpose is to hide the actual ID from the code.
- Adds `generated-sources` to `.gitignore`


plugins/com.google.cloud.tools.eclipse.usagetracker/.classpath
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="lib" path="lib/google-http-client-1.22.0.jar"/>
```
elharo: we're trying to figure out how to grab these from maven central rather than bundling jars

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+ * Provides methods that report plugin-specific events to Analytics.
+ */
+public class AnalyticsPingManager {
+
+  private static final String ANALYTICS_COLLECTION_URL = "https://ssl.google-analytics.com/collect";
+  // TODO: retrieve tracking ID from a system environment variable.
+  private static final String ANALYTICS_TRACKING_ID = "@ga.tracking.id@";
+  private static final String APPLICATION_NAME = "gcloud-eclipse-tools";
+
+  // TODO: see if there a way to get a unique ID for each Eclipse installation.
```
elharo: Why would we want that? Seems to have PII issues.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static final PercentEscaper ESCAPER = new PercentEscaper(".-_", false);
+
+  private static boolean isTrackingIdDefined() {
+    return ANALYTICS_TRACKING_ID != null && ANALYTICS_TRACKING_ID.startsWith("UA-");
+  }
+
+  public static void sendPing(String eventType, String eventName,
+      String metadataKey, String metadataValue) {
+    Map<String, String> parametersMap = Maps.newHashMap(STANDARD_PARAMETERS);
```
elharo: no longer need to use Maps.newHashMap now that we have type inference. HashMap<>() works these days. Maps.newHashMap may be deprecated/removed one of these days.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+    String parametersString = getParametersString(parametersMap);
+    try {
+      URL url = new URL(ANALYTICS_COLLECTION_URL);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setDoOutput(true);
+      connection.setRequestMethod("POST");
+      connection.setRequestProperty("Content-Length", Integer.toString(parametersString.length()));
+      try {
+        DataOutputStream out = new DataOutputStream(connection.getOutputStream());
```
elharo: danger: do not use DataOutputStream here (or ever for writing strings). Use an OutputStreamWriter with a specified encoding.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+          out.writeBytes(parametersString);
+          out.flush();
+        } finally {
+          out.close();
+        }
+      } finally {
+        connection.disconnect();
+      }
+    } catch (IOException e) {
```
elharo: nit: e --> ex

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+    sendPostRequest(parametersMap);
+  }
+
+  private static void sendPostRequest(Map<String, String> parametersMap) {
+    if (Platform.inDevelopmentMode() || !isTrackingIdDefined()) {
+      return;
+    }
+
+    String parametersString = getParametersString(parametersMap);
+    try {
```
elharo: try-with-resources here and below might make this  a tad cleaner

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+  // TODO: see if there a way to get a unique ID for each Eclipse installation.
+  private static final String CLIENT_ID = UUID.randomUUID().toString();
+
+  // Fixed-value query parameters present in every ping, and their fixed values:
+  //
+  // For the semantics of each parameter, consult the followings:
+  //
+  // https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L183
+  // https://developers.google.com/analytics/devguides/collection/protocol/v1/reference
+  private static final ImmutableMap<String, String> STANDARD_PARAMETERS =
```
elharo: Use Collections.unmodifiableMap on a regular Map and I think you can remove the Guava dependency

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+      } finally {
+        connection.disconnect();
+      }
+    } catch (IOException e) {
+      // Don't try to recover or retry.
+    }
+  }
+
+  private static String getParametersString(Map<String, String> parametersMap) {
```
elharo: this needs good tests (or a library method)

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+    parametersMap.put("cd19", eventType);
+    parametersMap.put("cd20", eventName);
+
+    String virtualPageUrl = "/virtual/" + APPLICATION_NAME + "/" + eventType + "/" + eventName;
+    parametersMap.put("dp", virtualPageUrl);
+
+    if (metadataKey != null) {
+      // Event metadata are passed as a (virtual) page title.
+      String virtualPageTitle =
```
elharo: should metadataValue be escaped here? Would probably be clearer to read if this were pulled apart into several statements with out using ?:

elharo:
@briandealwis please look at how the plugin is configured, particularly the extra dependencies

elharo:
Some unit tests would be very helpful.

elharo:
Test failures look related:

[ERROR]   Software being installed: google-cloud-plugin-eclipse.repo raw:0.1.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):0.1.0-SNAPSHOT
[ERROR]   Missing requirement: com.google.cloud.tools.eclipse.appengine.newproject 0.1.0.qualifier requires 'bundle com.google.cloud.tools.eclipse.usagetracker 0.0.0' but it could not be found
[ERROR]   Cannot satisfy dependency: com.google.cloud.tools.eclipse.suite.e45.feature.feature.group 0.1.0.qualifier depends on: com.google.cloud.tools.eclipse.appengine.newproject 0.0.0
[ERROR]   Cannot satisfy dependency: google-cloud-plugin-eclipse.repo raw:0.1.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):0.1.0-SNAPSHOT depends on: com.google.cloud.tools.eclipse.suite.e45.feature.feature.group [0.1.0,0.1.1)
[ERROR] 
[ERROR] See http://wiki.

chanseokoh:
Question related to the test failure: is there a way to compile/test the entire project and run an Eclipse instance from the command line (i.e., something like `mvn runIdea`?) I think I can check the travis failure ahead by checking compilation/tests on the command line.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: If this UUID initialization seems ok to you, I think it's fine. I asked this because our IntelliJ plugin has an API `getInstallationUUID()`, and we have been using it actually.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static void sendPostRequest(Map<String, String> parametersMap) {
+    if (Platform.inDevelopmentMode() || !isTrackingIdDefined()) {
+      return;
+    }
+
+    String parametersString = getParametersString(parametersMap);
+    try {
+      URL url = new URL(ANALYTICS_COLLECTION_URL);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
briandealwis: Are we actually using the google-http-client?

briandealwis:
> One of them is to pull a GA tracking ID at run-time, which I don't know how this can be achieved best in our new plugin.

Do you mean *compile-time*?  We found the `org.codehaus.mojo:tempting-maven-plugin` worked best.  Take a look at [this pom from GPE](https://github.com/GoogleCloudPlatform/google-plugin-for-eclipse/blob/master/plugins/com.google.gdt.eclipse.suite/pom.xml#L17) for an example.

briandealwis:
> Question related to the test failure: is there a way to compile/test the entire project and run an Eclipse instance from the command line?

Just a plain `mvn verify` will do it.

briandealwis:
It seems we're missing:

- [ ] some event codes
- [ ] tie-ins into the new wizards (there's a separate Maven generator too in `.newproject.maven`)

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: That jar is for `PercentEscaper`. Any alternative or suggestion?

elharo: Maybe https://docs.oracle.com/javase/7/docs/api/index.html?java/net/URLEncoder.html though you need to write tests and be careful here

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+  // TODO: retrieve tracking ID from a system environment variable.
+  private static final String ANALYTICS_TRACKING_ID = "@ga.tracking.id@";
+  private static final String APPLICATION_NAME = "gcloud-eclipse-tools";
+
+  // TODO: see if there a way to get a unique ID for each Eclipse installation.
+  private static final String CLIENT_ID = UUID.randomUUID().toString();
+
+  // Fixed-value query parameters present in every ping, and their fixed values:
+  //
+  // For the semantics of each parameter, consult the followings:
```
elharo: followings --> following

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+  private static final ImmutableMap<String, String> STANDARD_PARAMETERS =
+      ImmutableMap.<String, String>builder()
+          .put("v", "1")  // Google Analytics Measurement Protocol version
+          .put("tid", ANALYTICS_TRACKING_ID) // tracking ID
+          .put("ni", "0")  // Non-interactive? Report as interactive.
+          .put("t", "pageview")  // Hit type
+          .put("cd21", "1")  // Yes, this ping is a virtual "page".
+          .put("cd16", "0")  // Internal user? No.
+          .put("cd17", "0")  // User signed in? We will ignore this.
+          .put("cid",  CLIENT_ID)
```
elharo: nit: extra space after comma

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,116 @@
...
+  // https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L183
+  // https://developers.google.com/analytics/devguides/collection/protocol/v1/reference
+  private static final ImmutableMap<String, String> STANDARD_PARAMETERS =
+      ImmutableMap.<String, String>builder()
+          .put("v", "1")  // Google Analytics Measurement Protocol version
+          .put("tid", ANALYTICS_TRACKING_ID) // tracking ID
+          .put("ni", "0")  // Non-interactive? Report as interactive.
+          .put("t", "pageview")  // Hit type
+          .put("cd21", "1")  // Yes, this ping is a virtual "page".
+          .put("cd16", "0")  // Internal user? No.
```
elharo: where are cd16 and cd17 documented?

chanseokoh: Haven't seen any document, but I can actually see these defined on GA. FYI, I based these off of [this implementation](https://github.com/google/cloud-reporting/blob/master/src/main/java/com/google/cloud/metrics/MetricsUtils.java#L183).

plugins/com.google.cloud.tools.eclipse.usagetracker/.classpath
chanseokoh: Removed google-http-client by replacing `PercentEscaper` with `URLEncoder`.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Now uses `new HashMap<>()`.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Done. `HttpURLConnection` was unfortunately not auto-closable.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Done. Couldn't find a better way than this. With this, I removed the guava dependency.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Added a comment about the Custom Dimension.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Not sure what's really going on, but I added `@SuppressWarnings("serial") suggested by Eclipse. Seems related to the inner anonymous map(?).

chanseokoh:
@briandealwis I added this new bundle to the `feature.xml` you mentioned, but the Travis build still fails with an error. Couldn't figure out why.

chanseokoh:
I address most of the code reviews. I will work on adding some tests.

elharo:
I also wonder if we're doing this work in the right order. This must not be used until #138 "Get user consent for analytics on install" is complete. Perhaps we should addf a class/method somewhere such as 

boolean checkUserConsent() 

and for now have it always return false, along with a strong comment explaining what conditions need to be met to turn it on.

chanseokoh:
Regarding storing user consent and possibly a unique client ID, where do we store plugin-wide user preferences? I don't see a relevant bundle or a class in our project setup.

briandealwis:
Preferences are stored using the Eclipse/OSGi preferences mechanisms:
  - to get/set preferences programmatically, use `InstanceScope.INSTANCE.node(pluginId)`; this returns an `IEclipsePreferences`/`org.osgi.services.prefs.Preferences` object that is very similar to the Java preferences object.
  - to provide UI, create a `PreferencePage`, which is most easily done via WindowBuilder.  There's a `FieldEditorPreferencePage` class that makes setup very easy. It's based on an older interface for preference properties called `IPreferenceStore`, and there's a wrapper called `ScopedPreferenceStore` that will map an `InstanceScope` &rarr; `IPreferenceStore`

chanseokoh:
@briandealwis  Thanks. I used `ConfigurationScope` instead of `InstanceScope`. According to [this](https://wiki.eclipse.org/FAQ_What_is_a_preference_scope%3F), `ConfigurationScope` seems to be the right scope to store opt-in user consent to send usage metrics as well as a unique client ID (for the usage tracking purpose).

briandealwis:
Pushed up a slight tweak to fix compilation.

`ConfigurationScope` makes sense.

chanseokoh:
@briandealwis Thanks for the fix. BTW, yesterday, I removed this new bundle from `feature.xml` you mentioned above, after noticing that this works fine without it.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Now it uses `URLEncoder`.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AnalyticsPingManagerTest {
+
+  private static final String RANDOM_UUID = "bee5d838-c3f8-4940-a944-b56973597e74";
```
elharo: just UUID, it's not random

plugins/com.google.cloud.tools.eclipse.usagetracker/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,10 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: UsageTracker (common)
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.usagetracker;singleton:=true
+Bundle-Version: 0.1.0.qualifier
+Bundle-Vendor: Google, Inc.
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Export-Package: com.google.cloud.tools.eclipse.usagetracker
+Require-Bundle: org.eclipse.core.runtime;bundle-version="3.11.1"
+Import-Package: com.google.common.annotations
```
elharo: why do we need this?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
elharo: I don't think any of your tests exercise the percent escaping functionality. Please add some that do.

plugins/com.google.cloud.tools.eclipse.usagetracker/META-INF/MANIFEST.MF
chanseokoh: That's for `VisibleForTesting`. What do you think of using this?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,158 @@
...
+        resultBuilder.append('&');
+      } else {
+        ampersandNeeded = true;
+      }
+      resultBuilder.append(entry.getKey());
+      resultBuilder.append('=');
+      try {
+        resultBuilder.append(URLEncoder.encode(entry.getValue(), "UTF-8"));
+      } catch (UnsupportedEncodingException uee) {
+        throw new AssertionError("UTF-8 not supported");
```
elharo: Can we set this to a RuntimeException instead? Errors are really nasty, per Effective Java. Alternately, we could fallback to the default charset here and return the encoded value without the character set specified. Though of course this should be unreachable in a properly configured VM.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Added another test for this.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: I used RuntimeException since the Analytics protocol requires UTF-8 and it seems highly unlikely that any VM does not support UTF-8 these days.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
chanseokoh: RANDOM_UUID --> UUID done.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,158 @@
...
+
+    sendPostRequest(parametersMap);
+  }
+
+  private static void sendPostRequest(Map<String, String> parametersMap) {
+    if (Platform.inDevelopmentMode() || !isTrackingIdDefined() || !hasUserOptedIn()) {
+      return;
+    }
+
+    String parametersString = getParametersString(parametersMap);
```
elharo: can you move this into sendPing so a one method adds all the parameters and one actually sends the request?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,158 @@
...
+      }
+
+      parametersMap.put("dt", virtualPageTitle);
+    }
+
+    sendPostRequest(parametersMap);
+  }
+
+  private static void sendPostRequest(Map<String, String> parametersMap) {
+    if (Platform.inDevelopmentMode() || !isTrackingIdDefined() || !hasUserOptedIn()) {
```
elharo: I wonder if this might belong at the beginning of sendPing instead. What do you think?

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+  private static final String EVENT_TYPE = "some-event-type";
+  private static final String EVENT_NAME = "some-event-name";
+
+  private static final String VIRTUAL_DOCUMENT_PAGE =
+      "/virtual/some-application/" + EVENT_TYPE + "/" + EVENT_NAME;
+
+  private static final String METADATA_KEY = "some-custom-key";
+  private static final String METADATA_VALUE = "some-custom-value";
+
+  @SuppressWarnings("serial")
```
elharo: This suppress isn't really necessary. Just turn off this warning in Eclipse.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+          put("cd17", "0");
+          put("cid", UUID);
+          put("cd19", EVENT_TYPE);
+          put("cd20", EVENT_NAME);
+          put("dp", VIRTUAL_DOCUMENT_PAGE);
+          put("dt", METADATA_KEY + "=" + METADATA_VALUE);
+        }
+      });
+
+  @SuppressWarnings("serial")
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -0,0 +1,90 @@
...
+    Assert.assertEquals(keyValuePairs.length, RANDOM_PARAMETERS.size());
+
+    for (String pair : keyValuePairs) {
+      String[] keyValue = pair.split("=");
+      Assert.assertEquals(2, keyValue.length);
+      Assert.assertEquals(keyValue[1], ENCODED_PARAMETERS.get(keyValue[0]));
+    }
+  }
+
+  private static final String NO_ENCODE_TEXT = ".*-_abcXYZ";
```
elharo: These should be local variables, or maybe even just literals if they're only used once.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Nice idea.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
chanseokoh: I like the simplicity of using just literals.

chanseokoh:
Comments addressed.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private static void sendPostRequest(String parametersString) {
+    try {
+      URL url = new URL(ANALYTICS_COLLECTION_URL);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setDoOutput(true);
+      connection.setRequestMethod("POST");
+      connection.setRequestProperty("Content-Length", Integer.toString(parametersString.length()));
+
+      try (
```
elharo: just to be safe, put the first line into the try-with-resources so the writer is closed

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Not sure I understand your point here.

The writer is already in the inner try-with-resources, so I think it is always closed. There is no other resources that can be used with the try-with-resources, I believe. (`url` is not closable, and `connection` is not auto-closable.) Also, it doesn't seem possible to pull the inner try-with-resources to the outer scope or merge with the try in the first line, by the syntax. I may be missing something though.

chanseokoh: Ah, perhaps I should move `connection.disconnect()` to the outer try.

chanseokoh: I wonder if we can forget about closing `writer` entirely, knowing that we always close `connection`.

chanseokoh: I think I need to close the writer, according to the JDK doc. BTW, I pulled out `disconnect()` into the outer try. PTAL.

elharo: You're write. The indentation threw me. Just mover it to the previous line and break somewhere else and it should be clear.

chanseokoh: Done. I'll merge it sooner or later.

PR #219, opened 2016-06-07
Remove Google App Engine Standard Facet

### LINKED ISSUE #199: The Google App Engine Standard facet is not being installed when you do Maven based installation

nbashirbello:
close #199


nbashirbello:
@elharo @briandealwis

elharo:
Looks good. Please patch in https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/220 and test against that.

elharo:
LGTM. Please merge with master and verify it all still works before checking in (though I don't see any reason it shouldn't).

briandealwis:
Do we really want to remove this facet?  Might it be useful to act as a placeholder to distinguish between general Servlet webapps and those written for AppEngine?

elharo:
YAGNI. If we need it we'll write it or restore it. Until then, it's better if we don't clutter the projects with it.

PR #220, opened 2016-06-07
Assemble IDE Target Platform

### LINKED ISSUE #195: com.google.gson vs com.google.code.gson

briandealwis:
Tycho is able to resolve dependencies from Maven-style repositories (`pomDependencies=consider`), but the Eclipse IDE cannot.  This patch creates a new project in `eclipse/ide-target-platform` that assembles the dependencies from the Tycho build into a suitable target platform for the IDE.

Fixes #195

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,12 @@
       <artifactId>app-tools-lib-for-java</artifactId>
       <version>0.1.0-SNAPSHOT</version>
     </dependency>
+    <dependency>
+      <!-- bundle: com.google.guava -->
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>17.0</version>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -69,6 +69,12 @@
       <artifactId>app-tools-lib-for-java</artifactId>
       <version>0.1.0-SNAPSHOT</version>
     </dependency>
+    <dependency>
+      <!-- bundle: com.google.guava -->
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>17.0</version>
```
elharo: why 17 instead of 19?

elharo:
LGTM

elharo:
I can confirm that this PR enables the project to run in side Eclipse again. LGTM.

pom.xml
briandealwis: That's what's used by `app-tools-lib-for-java`

PR #221, opened 2016-06-08
Add missing bundles to the suite feature

### LINKED ISSUE #218: Plugin cannot be installed
The zip built from head cannot be installed, as near as I can tell.

I think we may need to rollback the change that added dependencies on maven central. It may not be only that we can't run from inside eclipse, but that the plugin no longer works, at all.

screen shot 2016-06-06 at 7 31 06 pm

@briandealwis If you can get to this ASAP, please do so. Otherwise I'll experiment with a rollback tomorrow morning. If a rollback will fix this, please do that first, before working on a fix for the issue.

briandealwis:
Fixes #218 

PTAL @nbashirbello (if you're still awake)


nbashirbello:
LGTM

PR #223, opened 2016-06-08
switch sample to use new plugin

elharo:
@briandealwis @loosebazooka This is template for generated code from the new project wizard.


elharo:
ping

briandealwis:
LGTM though we should have an outstanding blocking issue that we need to update to a released version of `app-maven-plugin`.

PR #237, opened 2016-06-13
remove vestigial reference to gcloud-maven-plugin

elharo:
@loosebazooka @nbashirbello


nbashirbello:
LGTM

loosebazooka:
there's a gcloud plugin version parameter in the configuration that you probably don't need anymore
https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/fcf174500744504f30ef937dfdda77b21b9d0390/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/templates/pom.xml.ftl#L17

loosebazooka:
@elharo Can you remove the reference to the version of the gcloud plugin in another PR?

PR #238, opened 2016-06-13
Remove need to include the 3rd party feature in the final build

### LINKED ISSUE #234: When installing from locally-built repo, 3rd-party features item appears under "Uncategorized"
After successfully building the code locally, I installed the plugin from the local archive. I see the following update site info:

screen shot 2016-06-12 at 3 16 47 pm

"Cloud Tools for Eclipse" category - should it be called "Google Cloud Tools for Eclipse", or just "Google Cloud Platform"?

Underneath,  the item should probably be called "Cloud Tools for Eclipse".

I also see "Uncategorized", and an item under there called "3rd party features". Is that going to remain in the final version? Is the installation of these really optional?

briandealwis:
Change the `gcp-repo` project include the `com.google.cloud.tools.eclipse.3rdparty.feature` via an invisible product definition.  This fix ensures that the feature and its referenced bundles are included in the generated repository, but aren't listed as formally published features in the _Install New Software_ dialog.  I've also added a human-meaningful name for the repository, which is shown in the _Preferences > Install/Update > Available Software Sites_ if the user didn't explicitly specify a different label.

Fixes #234

### DIFF
--- a/gcp-repo/category.xml
+++ b/gcp-repo/category.xml
@@ -3,6 +3,5 @@
    <feature id="com.google.cloud.tools.eclipse.suite.e45.feature" version="0.1.0.qualifier">
       <category name="GCP"/>
    </feature>
-   <feature id="com.google.cloud.tools.eclipse.3rdparty.feature" version="0.1.0.qualifier"/>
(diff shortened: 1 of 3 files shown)


elharo:
LGTM

PR #241, opened 2016-06-13
don't generate pom.xml in non-maven project

### LINKED ISSUE #240: Don't generate pom.xml in non-maven project
Per @rdayal in latest meeting. 

We will of course still generate one in the maven-based project wizard.

elharo:
@nbashirbello fix #240


nbashirbello:
LGTM

PR #243, opened 2016-06-13
Add opt in dialog for user usage metric reporting

chanseokoh:
In relation to #138.

Added a new dialog class that recommends opting in for sending client-side usage metrics to Google. The dialog will be displayed at the top-right corner of the active workbench as shown below:

opt-in-pop-up

This PR only adds the class and relevant methods. A few other things need to be added further:
- Display the dialog at the start of Eclipse (if not yet accepted or declined). Not sure how to do this though.
- Create an entry in global settings so that users can later change the opt-in status through the settings UI.


plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
```diff
@@ -0,0 +1,105 @@
...
+    Rectangle parentClientArea = getParentShell().getClientArea();
+
+    int heightCaptionAndUpperBorder =
+        parentBounds.height - parentClientArea.height - getParentShell().getBorderWidth();
+    return new Point(parentBounds.x + parentClientArea.width - initialSize.x,
+        parentBounds.y + heightCaptionAndUpperBorder);
+  }
+
+  /**
+   * Overridden in order to remove the button bar area (and the OK and CANCEL buttons) that the
```
elharo: Let's talk offline about the UI. I think we can add the buttons and make this a lot clearer to the user. We don't want anyone opting in by accident.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
```diff
@@ -0,0 +1,105 @@
...
+    label.setFont(smallerFont);
+
+    Link link = new Link(container, 0);
+    link.setText(Messages.OPT_IN_NOTIFICATION_LINK);
+    link.setFont(smallerFont);
+
+    // Register the opt-in status depending on user selection.
+    link.addSelectionListener(new SelectionAdapter() {
+      @Override
+      public void widgetSelected(SelectionEvent e) {
```
elharo: no abbreviations please; e --> event

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
```diff
@@ -0,0 +1,105 @@
...
+        if ("opted-in".equals(e.text)) {
+          AnalyticsPingManager.registerOptInStatus(true);
+        } else if ("opted-out".equals(e.text)) {
```
elharo: just else, no else if needed here

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -82,6 +84,13 @@ private static boolean hasUserOptedIn() {
     return false;
   }
 
+  public static void registerOptInStatus(boolean optedIn) {
```
elharo: does this need to be public?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/messages.properties
```diff
@@ -0,0 +1,2 @@
+OPT_IN_NOTIFICATION_LINK=Please click <a href="opted-in">I agree</a> if you want to help make Google IntelliJ plugins better or <a href="opted-out">I don't agree</a> otherwise.
```
elharo: This is the IntelliJ language. We'll need to rewrite this.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
chanseokoh: e --> event Done.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/messages.properties
chanseokoh: Ooooops. We can rewrite this.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
chanseokoh: The idea is that if a user didn't explicitly say yes or no (e.g., just canceled the dialog), then we can show this dialog again. If a user really said no, we save this choice explicitly to prevent this. How about that?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Changed: public --> package protected

briandealwis:
> Display the dialog at the start of Eclipse (if not yet accepted or declined). Not sure how to do this though.

It would be preferable to do this on first use of any Cloud Tools "thing".  We shouldn't do this from the bundle's activator: there's no guarantee which thread the activation comes on (may not be the SWT thread), and any other threads attempting to access an object from the bundle will be blocked pending the activation.  Since `sendPing()` should be lightweight and not block the caller, we could queue any events whilst in the neither-yes-nor-no state, put up the dialog in an `asyncExec()`, and then send or discard events once the user makes a decision.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
chanseokoh: Sounds good. We can include @patflynn in the discussion. FYI, below is what we do in IntelliJ.

intellij-opt-in

Also a corresponding entry in settings:

intellij-opt-in-settings

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
elharo: Opt-in means that we don't send data without an explicit yes. As a user, I hate being bugged by incessant pop-ups. If the user closes the dialog without pressing a button, we should take that as "no".

chanseokoh:
@briandealwis Thanks for the idea. Still not sure how to trigger this on the first use of any Cloud thing, but I guess we can talk about this in detail soon. Triggering this on the first use of `sendPing()` may work, but I'm not sure if this is desired. This seems quite late.

About the queuing, I prefer simple approaches: if a user has not made a decision yet, `sendPing()` just won't do anything. I think there is no need to queue the events. It's fine to lose some ping events.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
briandealwis: The wording come across as overly broad: we're just trying to send anonymous usage data *of Google Plugins*. Not the IDE generally.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInPopup.java
chanseokoh: Agreed. I will take the action of closing the dialog as "no" (and save it).

briandealwis:
We don't want to annoy users by putting this on startup:
  - It contributes to longer launch times.
  - It's disconnected from their actual use:
     - They may be running a corporate build that has a pile of pre-installed plugins that they never use.
    - Tt contributes to the feeling that we're watching everything they're doing since they weren't using anything from Google at that point.

chanseokoh:
I converted it to a dialog with buttons and a caption. The actual UI will change ultimately with many iterations, but you can start reviewing the code at least.

chanseokoh:
eclipse-opt-in-dialog

eclipse-opt-in-dialog2

Without the caption, it will look like this:

eclipse-opt-in-dialog3

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/messages.properties
```diff
@@ -0,0 +1,4 @@
+OPT_IN_DIALOG_TITLE=Google Eclipse Plugin
+OPT_IN_DIALOG_TEXT=Do you agree to send anonymous usage statistics of Google Eclipse Plugin to Google?
```
elharo: Google Cloud Plugin

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/messages.properties
```diff
@@ -0,0 +1,4 @@
...
+OPT_IN_BUTTON=Send
+OPT_OUT_BUTTON=Don't send
```
elharo: Don't Send (capital S)

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/messages.properties
briandealwis: This could be used by other plugins, like the Google Dataflow plugin too. Right?

chanseokoh: Okay, I changed it to Google Cloud Plugins (plural).

elharo:
LGTM

Issue #242, opened 2016-06-13
Local server run should rely on app-tools-lib-for-java to execute DevAppServer


rdayal:
Right now, there are three different ways that the DevAppServer could get executed

1) App Engine - Custom Runtime
2) App Engine - Generic Server Framework
3) Pure Maven. Defined as:
---->Run As > Run Configurations
---->Maven Build > DevAppServer
---->Click the Run button
---->Load http://localhost:8080 in browser.

In cases 1) and 2) , we are running the dev_appserver.py ourselves by locating the script and shelling out to it. 

In case 3), the user is running Maven directly.

We should really collapse 1) and 2) down so that they use the app-tools-lib-for-java to invoke the DevAppServer (may require changes to app-tools-lib-for-java). 3) will remain as-is, as the user is always free to execute Maven goals directly from the IDE if they wish to.

nbashirbello:
Done in #273 and #308

PR #244, opened 2016-06-14
Start sketching out the app-tools-lib-for-java based server

elharo:
@nbashirbello @loosebazooka @meltsufin start planning out for #242 to switch over to app-tools-lib-for-java


plugins/com.google.cloud.tools.eclipse.appengine.localserver/META-INF/MANIFEST.MF
```diff
@@ -33,4 +34,5 @@ Require-Bundle: org.eclipse.wst.server.core;bundle-version="1.7.0",
  org.eclipse.ui.console;bundle-version="3.6.100",
  org.eclipse.ui.ide;bundle-version="3.11.0",
  com.google.cloud.tools.eclipse.appengine.ui;bundle-version="0.1.0"
-Export-Package: com.google.cloud.tools.eclipse.appengine.localserver.runtime
+Export-Package: com.google.cloud.tools.eclipse.appengine.localserver.runtime,
```
nbashirbello: Why do we need to export these packages?

elharo: It's an autofix suggested by Eclipse when the relevant classes couldn't be found by the test plugin. I think if we don't export them, the test plugin can't see them, even though they'e in the same package.

nbashirbello:
LGTM

PR #247, opened 2016-06-14
Change debug launch to use socket-listen instead of socket-connect

### LINKED ISSUE #227: debug maven based devappserver
we can run it. let's make sure we can debug it.

It didn't work when last I tried it. This may be an issue in the eclipse launch configuration supplied by the archetype.

briandealwis:
Launching the dev_appserver in debug mode often fails due to timing
issues.  Although we have a 500ms wait, it seems insufficient
to ensure the remote JVM has gotten to a debuggable state and
so the debug connection fails.  We instead use the socket-listen
launch, such that the remote JVM connects to the debugger.

Fix #227


briandealwis:
Build fails on `testCreateAppRunCommand_validDebugCommand` which compares the generated command line for `dev_appserver.py` to ensure it contains `server=y`.  The rest of the code is ok.

nbashirbello:
LGTM

PR #252, opened 2016-06-15
Fake launch a server

elharo:
@nbashirbello This pull request doesn't really launch a server but it pretends to by printing a message in the console and setting the server state to started. This avoids an error message when Run As > Run on Server in the target platform.


elharo:
ping @nbashirbello

nbashirbello:
LGTM

PR #257, opened 2016-06-16
Interconnect preferences and AnalyticsPingManager

chanseokoh:
It's time to integrate `CloudToolsPreferencePage` with `AnalyticsPingManager`'.

I defined two more preference strings: `ANALYTICS_OPT_IN_REGISTERED` and `ANALYTICS_CLIENT_ID`

`ANALYTICS_OPT_IN_REGISTERED` is to determine if the opt-in dialog was ever presented to the user. (Simply checking if the opt-in status has ever been set does not serve this purpose, because the status can be set (with a default value) and saved to a file just by navigating through the Eclipse preference page).

One thing that annoyed me was the lack of a genuine way to replace the underlying `InstanceScope` preference store of `Activator`. I ended up declaring a separate `preferenceStore` field in the `Activator`.


plugins/com.google.cloud.tools.eclipse.preferences/src/com/google/cloud/tools/eclipse/preferences/Activator.java
```diff
@@ -41,4 +46,20 @@ public static Activator getDefault() {
...
+  /**
+   * This methods is not supposed to be overridden, but we couldn't find any alternative
+   * to replace the Instance-scoped {@link AbstractUIPlugin#preferenceStore} in the base
+   * class with a Configuration-scoped {@link ScopedPreferenceStore}. As such, this class
+   * defines its own {@link preferenceStore} field that is distinct from the field by the
+   * same name in the base class.
+   */
+  @Override
+  public IPreferenceStore getPreferenceStore() {
+    if (preferenceStore == null) {
```
elharo: Is there a reason to lazy initialize this? That tends to be error prone in multithreaded environments. It would be nicer to make this a final field.

chanseokoh: Right, we are mostly likely good with one final field of ours. I just copy-pasted the code from the base class. I will take off the lazy initialization.

elharo:
LGTM

Issue #258, opened 2016-06-16
Disconnect GSF


elharo:
i.e. remove it from plugin.xml so it no longer appears in the UI. We'll delete the actual code a little bit later.

elharo:
The next piece of this is complex. CreateAppEngineStandardWtpProject attaches this runtime to new projects and uses it to set up the classpath.


We'll need to attach a different runtime instead (easy enough) and figure out how to configure the classpath (not so easy). Maybe this should be a facet?

PR #260, opened 2016-06-16
turn off GSF server

elharo:
@nbashirbello #258


nbashirbello:
The GSF runtime stuff is still here.

elharo:
I've removed some more pieces. There are some surprising connections between the pieces that are left and the classpath for wizard generated projects I don't yet follow.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -21,25 +21,6 @@
...
-            launchModes="run,debug"
-            launchConfigId="org.eclipse.jst.server.generic.core.ExternalLaunchConfigurationType"
-            name="%gsfServerTypeName"
-            runtime="true"
-            runtimeTypeId="com.google.cloud.tools.eclipse.gcloud.runtime"
-            startBeforePublish="true"
-            startTimeout="120000"
-            stopTimeout="15000">
-      </serverType>
-   </extension>
```
nbashirbello: If we just want to hide the GSF server, should we just comment this out since we are not removing the classes associated with it?

elharo: The classes are all part of Eclipse so we're not removing them.

Commenting out code you want to get rid of is too wishy washy. If we need it back, that's what git is for.

elharo:
@nbashirbello I figured out why I couldn't delete the GSF runtime completely without breaking things. The new application wizard associates that runtime with its project, and uses that runtime to locate classes in CreateAppEngineStandardWtpProject. This can and should be fixed. However it's going to take some work to do so. Meanwhile I'd like to check in this much so that the GSF option is removed from the Run As > Run on Server dialog even if it still shows up in the runtimes preference.

nbashirbello:
LGTM

PR #261, opened 2016-06-17
Workaround javax.annotation-related class resolution problems

briandealwis:
Add a new `com.google.cloud.tools.eclipse.javax.annotations` bundle to
subsume the org.jsr-305 and javax.annotation bundles.  The bundle:
  - explicitly pulls in the JRE's javax.annotation classes (via `Require-Bundle: system.bundle`) to ensure that any classes it provides are first looked up and resolved in the JRE;
  - includes the FindBugs org.jsr-305 jar and exports its packages with a bumped version number of 3.0.2 to ensure it 'wins' over the FindBugs org.jsr-306 bundle;
  - imports its own packages so that if it has won, its other packages win too.

This should work with both the `org.jsr-305` and `javax.annotation` bundles loaded.  To test, you should be able to open the _OSGi Host Console_ from the Console view (in your Eclipse being debugged) and type `packages javax.annotation` to see what bundles are wired to the different exporters of `javax.annotation` and see that org.jsr-305 is not wired to any bundles:
```
osgi> packages javax.annotation
osgi.wiring.package; bundle-symbolic-name:List<String>="org.eclipse.osgi,system.bundle"; bundle-version:Version="3.10.102.v20160118-1700"; version:Version="0.0.0"; osgi.wiring.package="javax.annotation"<org.eclipse.osgi_3.10.102.v20160118-1700 [0]>
  org.eclipse.wst.jsdt.manipulation_1.0.600.v201505071819 [617] imports
  [… etc …]
osgi.wiring.package; bundle-symbolic-name="com.google.cloud.tools.eclipse.javax.annotations"; bundle-version:Version="0.1.0.qualifier"; jsr305="split"; version:Version="3.0.2"; osgi.wiring.package="javax.annotation"; uses:="javax.annotation.meta"<com.google.cloud.tools.eclipse.javax.annotations_0.1.0.qualifier [13]>
  org.eclipse.e4.ui.workbench.renderers.swt_0.13.0.v20150901-1347 [248] imports
  org.eclipse.ui.trace_1.0.300.v20150220-1252 [565] imports
  [… etc …]
osgi.wiring.package; bundle-symbolic-name="org.jsr-305"; bundle-version:Version="3.0.1"; version:Version="3.0.1"; osgi.wiring.package="javax.annotation"; uses:="javax.annotation.meta"<org.jsr-305_3.0.1 [677]>
osgi> 
```
PTAL @nbashirbello @chanseokoh


plugins/com.google.cloud.tools.eclipse.javax.annotations/.classpath
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry exported="true" kind="lib" path="lib/org.jsr-305_3.0.1.jar"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
```
chanseokoh: I'm using JavaSE-1.7, so I think this is in conflict. (The [README.md](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/README.md) says to use JavaSE-1.7.)

plugins/com.google.cloud.tools.eclipse.javax.annotations/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,7 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=1.8
+org.eclipse.jdt.core.compiler.compliance=1.8
+org.eclipse.jdt.core.compiler.problem.assertIdentifier=error
+org.eclipse.jdt.core.compiler.problem.enumIdentifier=error
+org.eclipse.jdt.core.compiler.source=1.8
```
chanseokoh: Is 1.8 intended? After opening this javax.annotation project (module, bundle, or whatever), my eclipse seems to change these to 1.7. (Maybe after I set to use JavaSE-1.7?)

briandealwis:
Good catch: I'll fix them.  I manually changed the Bundle's Required-ExecutionEnvironment to 1.7 but forgot to change the project settings.

chanseokoh:
Thanks!

Not sure what's going on, but I was able to launch the IDE and checked things are working properly. However, I had to clear out some minor issues, such as specifying to use JavaSE-1.7 instead of JavaSE-1.8.

Also, I am not sure why my Eclipse cannot compile `plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java` saying that `TargetPlatform` cannot be resolved. (Strangely, doing `mvn verify` in the root directory on the comman d line works ok.) I took them out those references to `TargetPlatform` to launch the IDE.

plugins/com.google.cloud.tools.eclipse.javax.annotations/.classpath
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry exported="true" kind="lib" path="lib/org.jsr-305_3.0.1.jar"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src"/>
```
chanseokoh: How about removing this `src` line? When I checkout this project fresh, the folder doesn't exist, so my Eclipse reports an error that I have to fix. (By either creating an empty directory or removing this line.)

chanseokoh:
Ok, forget about `TargetPlatform`. They were introduced in #252.

chanseokoh:
LGTM after addressing the missing `src` directory problem.

This also feels like a hack, so I think it's a good idea to put a README explaining the purpose of this bundle (just in a few sentences with easy words).

briandealwis:
Abandoning this change; problem solved in [`app-tools-lib-for-java`](https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/126) instead.

PR #264, opened 2016-06-19
Fix finding of GSF runtime in new app engine wizard

elharo:
@briandealwis We do want to get rid of this runtime, but while working on that I found a bug we should fix that will also help us change the runtime going forward.


elharo:
Ping

briandealwis:
LGTM

PR #265, opened 2016-06-20
Define a facet for AppEngine

elharo:
@nbashirbello For now this is an empty facet that we can hang functionality off of later. 

Please consider in particular which plugin should define this facet. I put it in newproject but perhaps it belongs in localserver or a completely new plugin.


nbashirbello:
Are we still moving this to its own plugin?

briandealwis:
This approach doesn't seem right to me: we should be using the approach we used for the _Generic Server Runtime_ but with better names, and split across two bundles:

  1. one bundle to define the WTP Server Runtime definition for `appengine` ('appengine.runtime'?)
     - defines a `runtime-component` and associates it to the runtime type
     - this runtime component pulls out the `servlet-api.jar` and JSP jars from the Google Cloud SDK, but that's an implementation detail; the bundle could just ship with the servlet and JSP jars or procure them from elsewhere
  2. one bundle to contain the DevAppServer server implementation for the `appengine` WTP Server Runtime; depends on Bundle 1 ('appengine.devserver'?)

We could conceptually have a 3rd bundle that provided a WTP Server implementation that talks directly to the real AppEngine, and that is independent of the DevAppServer implementation.  We'll need something like this for the debug snapshots support, I imagine.

We'd also have a separate `appengine.apis` bundle that defines a facet to provide access to the _Advanced App Engine APIs_ ([link](https://cloud.google.com/appengine/docs/java/javadoc/)).  Although we could integrate this directly in the runtime-component, I think there may be some advantage to allow developers to control what APIs they bind to.

elharo:
Now moved to its own plugin.

The runtime is a separate issue. The only assumption here is that there's something we'll want to put into a facet. Earlier we talked about putting the app engine API jars here and the servlet-api.jar and jsp-api.jar in the runtime. There may be more we want to put here (e.g. appengine-web.xml).

I'm not sure whether we want to split appengine.runtime and appengine.devserver into two plugins as you suggest, but that decision doesn't need to be made in this PR.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
```diff
@@ -23,5 +23,36 @@
           id="com.google.cloud.tools.eclipse.appengine.wizards" 
           name="%gcp.category" />
    </extension>
+   
+   <extension point="org.eclipse.wst.common.project.facet.core.facets">
```
nbashirbello: Do we still need this facet definition?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/FacetUninstallDelegate.java
```diff
@@ -0,0 +1,18 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
```
nbashirbello: Do we still need this too?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/FacetInstallDelegate.java
```diff
@@ -0,0 +1,18 @@
+package com.google.cloud.tools.eclipse.appengine.newproject;
```
nbashirbello: And this?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/FacetUninstallDelegate.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/FacetInstallDelegate.java
elharo: no, gone

plugins/com.google.cloud.tools.eclipse.appengine.newproject/plugin.xml
elharo: no. gone

briandealwis:
> Earlier we talked about putting the app engine API jars here and the servlet-api.jar and jsp-api.jar in the runtime.

I think the use of _Basic_ is what threw me: I thought that was referencing the Servlet and JSP jars.

> There may be more we want to put here (e.g. appengine-web.xml).

That seems like a very good idea.

LGTM.

elharo:
Good point. removed the word "Basic"

PR #266, opened 2016-06-20
use custom runtime

elharo:
@nbashirbello @briandealwis Any idea what I'm doing wrong here? The classes are not found and as best I can tell that's because the runtimeClasspathProvider is never actually consulted. (I.e. debugger won't stop in its constructor)

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
@@ -1,7 +1,6 @@
 package com.google.cloud.tools.eclipse.appengine.localserver;
 
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.JavaCore;
@@ -16,7 +15,6 @@ import com.google.cloud.tools.appengine.cloudsdk.PathResolver;
 public class ServletClasspathProvider extends RuntimeClasspathProviderDelegate {
 
   public ServletClasspathProvider() {
-    System.err.println("setting up classpath provider");
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Override
+  public IClasspathEntry[] resolveClasspathContainer(IProject project, IRuntime runtime) {
+    return resolveClasspathContainer(runtime);
+  }
+
+  @Override
+  public IClasspathEntry[] resolveClasspathContainer(IRuntime runtime) {
+    String cloudSdkPath = PathResolver.INSTANCE.getCloudSdkPath().toString();
```
briandealwis: Remember that `PathResolver` may return `null` when no SDK is found.

briandealwis:
> The classes are not found and as best I can tell that's because the runtimeClasspathProvider is never actually consulted. (I.e. debugger won't stop in its constructor)

I'm having _déja vu_…  Trying to remember…

briandealwis:
You need to provide an adapter for the runtime-component to `IClasspathProvider`. The only provided adapter is an internal class `org.eclipse.jst.server.core.internal.RuntimeClasspathProvider$Factory`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
elharo: good catch. done.

elharo:
PTAL @briandealwis 

Thanks. The adapter did the trick. This now successfully loads the classes from the custom runtime.

Next things to consider here:

1. Is there a non-internal factory class we should use in the adapter, or should we write our own?
2. How to improve the relative path resolution against the cloud SDK. As written I'm not confident this will work on Windows. 

Not sure if that should be part of this PR or the next batch of work.

nbashirbello:
LGTM

PR #268, opened 2016-06-21
require facets bundle

elharo:
@nbashirbello this corrects dependencies to unbreak the creation of an appengine project


nbashirbello:
LGTM

PR #269, opened 2016-06-21
Remove GSF elements from plugin.xml

### LINKED ISSUE #258: Disconnect GSF

elharo:
@nbashirbello Remove Generic Server Framework
fix #258


plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -12,68 +12,6 @@
       </image>
```
nbashirbello: Remove com.google.cloud.tools.eclipse.gcloud.runtime from typedefs

nbashirbello:
LGTM once typeIds are updated.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
elharo: done

PR #273, opened 2016-06-21
Run the app engine dev server via the app tools lib

nbashirbello:
@elharo @briandealwis 

This PR is afftected by this bug: https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/issues/124


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineOutputLineListener.java
```diff
@@ -0,0 +1,18 @@
+package com.google.cloud.tools.eclipse.appengine.localserver.server;
+
+import org.eclipse.ui.console.MessageConsoleStream;
+
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+
+public class LocalAppEngineOutputLineListener implements ProcessOutputLineListener {
+	MessageConsoleStream stream;
+
+	public LocalAppEngineOutputLineListener (MessageConsoleStream stream) {
```
elharo: seems to be over indented

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineOutputLineListener.java
```diff
@@ -0,0 +1,18 @@
+package com.google.cloud.tools.eclipse.appengine.localserver.server;
+
+import org.eclipse.ui.console.MessageConsoleStream;
+
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+
+public class LocalAppEngineOutputLineListener implements ProcessOutputLineListener {
+	MessageConsoleStream stream;
```
elharo: private

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,269 @@
...
+import org.eclipse.wst.server.core.model.PublishOperation;
+import org.eclipse.wst.server.core.model.ServerBehaviourDelegate;
+import org.eclipse.wst.server.core.util.PublishHelper;
+
+import com.google.common.collect.Lists;
+
+public class LocalAppEnginePublishOperation extends PublishOperation {
+	  private static final String PLUGIN_ID = LocalAppEnginePublishOperation.class.getName();
+
+	  /**
```
elharo: please detab all your files

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,269 @@
...
+	   * Throws new {@link CoreException} if status list is not empty.
```
elharo: @throws

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,269 @@
...
+	   */
+	  private static void checkStatuses(List<IStatus> statusList) throws CoreException {
+	    if (statusList == null || statusList.size() == 0) {
```
elharo: statusList.isEmpty() preferred over size() == 0

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
     setServerState(IServer.STATE_STOPPED);    
   }
 
-  void setStarted() {
-    setServerState(IServer.STATE_STARTED);    
+  /**
+   * Convenience method allowing access to protected method in superclass.
+   */
+  @Override
+  protected IModuleResourceDelta[] getPublishedResourceDelta(IModule[] module) {
```
elharo: why do we need this is both are protected?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
   }
-  
-  void setStarting() {
-    setServerState(IServer.STATE_STARTING);    
+
+  /**
+   * Convenience method allowing access to protected method in superclass.
+   */
+  @Override
+  protected IModuleResource[] getResources(IModule[] module) {
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
+      // Create dev app server instance
+	  CloudSdk cloudSdk = new CloudSdk.Builder()
+			  .addStdOutLineListener(outputListener)
+			  .addStdErrLineListener(outputListener)
+			  .startListener(localAppEngineStartListener)
+			  .exitListener(localAppEngineExitListener)
+			  .build();
+	  devServer = new CloudSdkAppEngineDevServer(cloudSdk);
+	  DefaultRunConfiguration devServerRunConfiguration = new DefaultRunConfiguration();
+	  devServerRunConfiguration.setAppYamls(runnables);
```
elharo: we don't have app.yaml?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
+
+  void startDevServer(List<File> runnables, MessageConsoleStream stream) {
+	  setServerState(IServer.STATE_STARTING);
+
+	  LocalAppEngineOutputLineListener outputListener =
+			  new LocalAppEngineOutputLineListener(stream);
+
+      // Create dev app server instance
+	  CloudSdk cloudSdk = new CloudSdk.Builder()
+			  .addStdOutLineListener(outputListener)
```
elharo: This is a very weird API. We should not need to parse the output streams like this.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
   }
   
+  private void terminate() {
+	  if (devServer != null) {
+		  DefaultStopConfiguration stopConfig = new DefaultStopConfiguration();
+		  try {
+			  devServer.stop(stopConfig);
+		  } catch (AppEngineException ex) {
+			  // TODO what do we need to do here
+			  System.out.println("Exception from terminate(): " + ex.getMessage());
```
elharo: prerfer logging

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
+		  devServer = null;
+	  }
+  }
+
+  /**
+   * A {@link ProcessExitListener} for the App Engine server.
+   */
+  public class LocalAppEngineExitListener implements ProcessExitListener {
+	@Override
+	public void onExit(int arg0) {
```
elharo: better name for arg0 needed

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,145 @@
...
+		stop(true);
+	}
+  }
+
+  /**
+   * A {@link ProcessStarteListener} for the App Engine server.
+   */
+  public class LocalAppEngineStartListener implements ProcessStartListener {
+	@Override
+	public void onStart(Process arg0) {
```
elharo: better name for arg0 needed

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -1,37 +1,105 @@
...
+@SuppressWarnings("restriction") // For FacetUtil
 public class LocalAppEngineServerDelegate extends ServerDelegate {
+  private static final String SERVLET_MODULE_FACET = "jst.web";
+  private static final String ATTR_CLOUD_SDK_SERVER_MODULES = "cloudsdk-server-modules-list";
+	
+  public static LocalAppEngineServerDelegate getAppEngineServer(IServer server) {
+	LocalAppEngineServerDelegate appEngineServer = server.getAdapter(LocalAppEngineServerDelegate.class);
+    if (appEngineServer == null) {
+    	appEngineServer = (LocalAppEngineServerDelegate) server.loadAdapter(LocalAppEngineServerDelegate.class,
+                                                           new NullProgressMonitor());
```
elharo: why a null progress monitor here?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -16,25 +21,30 @@
...
-    serverBehaviour.setStarted();
-    
-    MessageConsole console = TargetPlatform.findConsole("debugging");
+    List<File> runnables = new ArrayList<File>();
+    for (IModule module : modules) {
+      IPath deployPath = serverBehaviour.getModuleDeployDirectory(module);
+      runnables.add(deployPath.toFile());
+    }
+
+    MessageConsole console = TargetPlatform.findConsole(configuration.getName());
```
elharo: these lines are no longer needed

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: We need both for LocalAppEnginePublishOperation.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: Created issue for this: https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/issues/132

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: I must be missing something. It looks tlike if you simply deltete these methods it should just work.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
nbashirbello: Because we don't need to get any reports on this task. This server delegate is set in the plugin.xml so this call is very unlikely to fail.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineOutputLineListener.java
```diff
@@ -0,0 +1,18 @@
...
+public class LocalAppEngineOutputLineListener implements ProcessOutputLineListener {
+  private MessageConsoleStream stream;
+
+  public LocalAppEngineOutputLineListener (MessageConsoleStream stream) {
+    this.stream = stream;
+  }
+
+  @Override
+  public void onOutputLine(String arg0) {
+    stream.println(arg0);
```
elharo: better argument name please

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineOutputLineListener.java
```diff
@@ -0,0 +1,18 @@
+package com.google.cloud.tools.eclipse.appengine.localserver.server;
+
+import org.eclipse.ui.console.MessageConsoleStream;
+
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+
+public class LocalAppEngineOutputLineListener implements ProcessOutputLineListener {
+  private MessageConsoleStream stream;
+
+  public LocalAppEngineOutputLineListener (MessageConsoleStream stream) {
```
elharo: nit: no space before parenthesis

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+import org.eclipse.wst.server.core.model.ServerBehaviourDelegate;
+import org.eclipse.wst.server.core.util.PublishHelper;
+
+import com.google.common.collect.Lists;
+
+public class LocalAppEnginePublishOperation extends PublishOperation {
+  private static final String PLUGIN_ID = LocalAppEnginePublishOperation.class.getName();
+
+  /**
+   * @throws {@link CoreException} if status list is not empty.
```
elharo: nit: no period except on complete sentences

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+import org.eclipse.wst.server.core.IServer;
+import org.eclipse.wst.server.core.model.IModuleResource;
+import org.eclipse.wst.server.core.model.IModuleResourceDelta;
+import org.eclipse.wst.server.core.model.PublishOperation;
+import org.eclipse.wst.server.core.model.ServerBehaviourDelegate;
+import org.eclipse.wst.server.core.util.PublishHelper;
+
+import com.google.common.collect.Lists;
+
+public class LocalAppEnginePublishOperation extends PublishOperation {
```
elharo: JavaDoc would be very helpful on this class.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * Construct the operation object to publish the specified modules(s) to the
+   * specified server.
+   */
+  public LocalAppEnginePublishOperation(LocalAppEngineServerBehaviour server, int kind, IModule[] modules,
+      int deltaKind) {
+    super("Publish to server", "Publish modules to App Engine Development Server");
+    this.server = server;
+    this.modules = modules;
```
elharo: in general, be wary of passing in mutable state. Should probably make a copy.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+          deployPath = deployPath.append(uri);
+        } else {
+          // no uri is OK for removed modules
+          if (deltaKind != ServerBehaviourDelegate.REMOVED) {
+            statusList.add(newErrorStatus("Cannot get URI for module: " + modules[i + 1].getName()));
+            return;
+          }
+        }
+      }
+      // TODO: Disabled saving and loading module locations below (marked
```
elharo: is this commented out code still relevant?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+      } else {
+        publishDir(deployPath, moduleUrls, statusList, monitor, childModule);
+      }
+      // XXX: server.saveModulePublishLocations(moduleUrls);
+    }
+    checkStatuses(statusList);
+    server.setModulePublishState2(modules, IServer.PUBLISH_STATE_NONE);
+  }
+
+  private IStatus newErrorStatus(String message) {
```
elharo: static

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+
+  /**
+   * Publish as binary modules.
+   */
+  private void publishArchiveModule(IPath path, Properties mapping, List<IStatus> statusList,
+      IProgressMonitor monitor, IModule childModule) {
+    boolean isMoving = false;
+    // check older publish
+    String oldUri = (String) mapping.get(childModule.getId());
+    String jarUri = path.toOSString();
```
elharo: looks like this is a path, not a URI

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  private IStatus newErrorStatus(String message) {
+    return new Status(IStatus.ERROR, getClass().getName(), message);
+  }
+
+  /**
+   * Publish as binary modules.
+   */
+  private void publishArchiveModule(IPath path, Properties mapping, List<IStatus> statusList,
```
elharo: please pick a more explicit name for "mapping"

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  /**
+   * Publish child modules as directory if not binary.
+   */
+  private void publishDir(IPath path, Properties mapping, List<IStatus> statusList, IProgressMonitor monitor,
+      IModule childModule) throws CoreException {
+    boolean isMoving = false;
+    // check older publish
+    String oldUri = (String) mapping.get(childModule.getId());
```
elharo: again, I think these are more paths than URIs

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,270 @@
...
+    for (IModuleResourceDelta delta : deltas) {
+      IStatus[] publishStatus = helper.publishDelta(delta, path, monitor);
+      statusList.addAll(Arrays.asList(publishStatus));
+    }
+  }
+
+  /**
+   * Publish child modules as directory if not binary.
+   */
+  private void publishDir(IPath path, Properties mapping, List<IStatus> statusList, IProgressMonitor monitor,
```
elharo: Is the monitor actually used here? I don't see it, but if it is, should probably convert to a submonitor.

nbashirbello:
Still trying to fix error with stopping server caused by some server state still being set to "starting" and not "started".

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,148 @@
 package com.google.cloud.tools.eclipse.appengine.localserver.server;
 
+import java.io.File;
+import java.util.Arrays;
+import java.util.List;
+
```
elharo: nit: extra blank line

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,148 @@
...
+   * @return the directory at which module will be published.
+   */
+  public IPath getModuleDeployDirectory(IModule module) {
+    return getRuntimeBaseDirectory().append(module.getName());
+  }
+
+  /**
+   * Convenience accessor to protected member in superclass.
+   */
+  public void setModulePublishState2(IModule[] module, int state) {
```
elharo: why "2"?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,148 @@
...
+    setModulePublishState(module, state);
   }
-  
+
+  void startDevServer(List<File> runnables, MessageConsoleStream stream) {
```
elharo: please document what the runnable are. war files? directories containing what exactly?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,148 @@
...
+      devServer.run(devServerRunConfiguration);
+      //setServerState(IServer.STATE_STARTED);
+    } catch (AppEngineException ex) {
+      stop(true);
+    }
+  }
+
+  private void terminate() {
+    if (devServer != null) {
+      DefaultStopConfiguration stopConfig = new DefaultStopConfiguration();
```
elharo: Is default what we want? This might want to pull the admin host/port from the existing server.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,148 @@
...
+    devServerRunConfiguration.setAppYamls(runnables);
+
+    // FIXME: workaround bug when running on a Java8 JVM
+    // https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/issues/181
+    devServerRunConfiguration.setJvmFlags(Arrays.asList("-Dappengine.user.timezone=UTC"));
+
+    // Run server
+    try {
+      devServer.run(devServerRunConfiguration);
+      //setServerState(IServer.STATE_STARTED);
```
elharo: Why is this commented out? My understanding is that this is critical for eclipse to tell that we've started.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,148 @@
...
+  public LocalAppEngineServerBehaviour () {
+    localAppEngineStartListener = new LocalAppEngineStartListener();
+    localAppEngineExitListener = new LocalAppEngineExitListener();
+  }
 
   @Override
   public void stop(boolean force) {
     setServerState(IServer.STATE_STOPPING);
-    // TODO stop server
+    // stop server
```
elharo: I think you can remove this comment now

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -1,40 +1,50 @@
...
     TargetPlatform.showConsole(console);
-    MessageConsoleStream out = console.newMessageStream();
-    out.println("Server started (not really)");
-  }
+    console.clearConsole();
```
elharo: are you sure we want to clear the console here? Might there be useful information, possibly completely unrelated to this run, left there that the user will want?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: The method getPublishedResourceDelta(IModule[]) and getResources(IModule[])  from the type ServerBehaviourDelegate is not visible in LocalAppEnginePublishOperation.java

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: The method getPublishedResourceDelta(IModule[]) and getResources(IModule[]) from the type ServerBehaviourDelegate is not visible in LocalAppEnginePublishOperation.java

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: ah, you can't override protected with package. That's the trick.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
elharo: ProgressMonitors, if I understand them, are for reporting completion percentage to the user, not for reporting success/failure. Even if it's guaranteed to succeed, we still want to report long-running progress to the end user.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
nbashirbello: getClass() is non-static. Its a private method and dont think its worth it to hard code the name

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
nbashirbello: Yes, it gets cluttered quickly makes it harder to see the details of the current running server.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: The superclass, ServerBehaviourDelegate,  has a  final setModulePublishState method

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: We currently only use the default ports

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: The started state of the server is configured in LocalAppEngineStartListener#onStart

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: Seems strange that we need to access all these classes we weren't intended to access. Maybe just poor design in WTP.

Anyway, mark this method final too then.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: In that case remove the commented out code.

nbashirbello:
TODO: add test (#287)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,265 @@
...
+          }
+        }
+      }
+
+      // modules given as parent-child chain
+      // get last one, the prior modules should already be published
+      IModule childModule = modules[modules.length - 1];
+      Properties moduleUrls = new Properties();
+
+      // get as j2ee
```
elharo: we don't support JEE, do we need this?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,265 @@
...
+
+  @Override
+  public void execute(IProgressMonitor monitor, IAdaptable info) throws CoreException {
+    List<IStatus> statusList = Lists.newArrayList();
+    IPath deployPath = server.getModuleDeployDirectory(modules[0]);
+    if (modules.length == 1) {
+      // root modules
+      publishDirectory(deployPath, statusList, monitor);
+    } else {
+      // TODO: Why and when does this code happen?
```
elharo: maybe for multiple module runs? but how does it get published?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,117 @@
...
+import com.google.common.collect.Lists;
+
+/**
+ * Handles the publishing operations for the App Engine development server.
+ */
+public class LocalAppEnginePublishOperation extends PublishOperation {
+  private static final String PLUGIN_ID = LocalAppEnginePublishOperation.class.getName();
+
+  /**
+   * @throws {@link CoreException} if status list is not empty
```
elharo: or has size 1.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,117 @@
...
+   */
+  private static void checkStatuses(List<IStatus> statusList) throws CoreException {
+    if (statusList == null || statusList.isEmpty()) {
+      return;
+    }
+    if (statusList.size() == 1) {
+      throw new CoreException(statusList.get(0));
```
elharo: why is this a problem?

elharo: what if there's one status and it's Status.OK?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,117 @@
...
+    IPath deployPath = server.getModuleDeployDirectory(modules[0]);
+    publishDirectory(deployPath, statusList, monitor);
+    checkStatuses(statusList);
+    server.setModulePublishState2(modules, IServer.PUBLISH_STATE_NONE);
+  }
+
+  /**
+   * Publish modules as directory.
+   */
+  private void publishDirectory(IPath path, List<IStatus> statusList, IProgressMonitor monitor) throws CoreException {
```
elharo: This method should return a status list rather than using a mutable argument.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,117 @@
...
+      return;
+    }
+    // publish changes only
+    IModuleResourceDelta[] deltas = server.getPublishedResourceDelta(modules);
+    for (IModuleResourceDelta delta : deltas) {
+      IStatus[] publishStatus = helper.publishDelta(delta, path, monitor);
+      statusList.addAll(Arrays.asList(publishStatus));
+    }
+  }
+
```
elharo: here check for non-OK statuses and throw CoreException if you find any.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -1,23 +1,152 @@
...
+  /**
+   * Convenience accessor to protected member in superclass.
+   */
+  public final void setModulePublishState2(IModule[] module, int state) {
+    setModulePublishState(module, state);
   }
-  
+
+  /**
+   * Starts the development server.
```
elharo: nit: add blank line before first @param

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: Is this going to work when we start using other ports?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
briandealwis: The `statusList` accumulates statuses as this method (and others) may be called recursively.  I did look at using a `MultiStatus` object, but it required more invasive changes.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: Yes, this has get and set methods and uses the defaults if not specified.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
nbashirbello: Only errors or warnings will be added to statusList

elharo: Does MultiStatus work with a single status? If so, we could just drop the if block

briandealwis: `MultiStatus` works with a single status.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: Done

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
elharo: OK. Agreed we shouldn't make that change here then. Would it make sense to add a TODO/issue to move this to a MultisStatus instead of a List in another PR?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: Extra work is going to be required then to pull the port from the existing server. At least add a todo here and file an issue on it. I'm worried that in 6 weeks or so when we do add extra ports we're going to be scratching our heads wondering why the server doesn't stop. :-)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/CloudSdkServer.java
```diff
@@ -70,8 +70,7 @@
   public static CloudSdkServer getCloudSdkServer(IServer server) {
     CloudSdkServer cloudSdkServer = server.getAdapter(CloudSdkServer.class);
     if (cloudSdkServer == null) {
-      cloudSdkServer = (CloudSdkServer) server.loadAdapter(CloudSdkServer.class,
-                                                           new NullProgressMonitor());
+      cloudSdkServer = (CloudSdkServer) server.loadAdapter(CloudSdkServer.class, null);
```
akerekes: Is it possible that the `loadAdapter` method returns `null`? If so, we should guard against it in the caller methods.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Handles the publishing operations for the App Engine development server.
+ */
+public class LocalAppEnginePublishOperation extends PublishOperation {
+  private static final String PLUGIN_ID = LocalAppEnginePublishOperation.class.getName();
+
+  /**
+   * @throws {@link CoreException} if status list is not empty
+   */
+  private static void checkStatuses(List<IStatus> statusList) throws CoreException {
```
akerekes: Could this method be called something more expressive? `failOnError` or similar?

elharo: +1 Andras's suggested name is clearer. When I first read this PR, I had to work through the code to understand what it was doing.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -1,37 +1,99 @@
...
+  /**
+   * If the module is a web module returns the utility modules contained within its
+   * WAR, otherwise returns an empty list.
+   */
   @Override
   public IModule[] getChildModules(IModule[] module) {
-    // TODO Auto-generated method stub
-    return null;
+    if ((module != null) && (module[0] != null) && (module[0].getModuleType() != null)) {
+      IModule thisModule = module[module.length - 1];
```
akerekes: The `if` on the previous line checks for the item in the 0th position and this one works with the last item in the array, is this consistent?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
akerekes: How about clearing the console first and then showing it? I'd prefer not to be distracted by some blink of a text that I cannot read/parse in that millisecond but my brain keeps trying to figure out what it was.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/CloudSdkServer.java
nbashirbello: Null will occur is the serverType is not defined in the plugin.xml. This would cause a compile error.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
nbashirbello: #299

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
nbashirbello: The check is done on right after this function is called (line 79)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: #300

elharo:
LGTM

PR #279, opened 2016-06-23
use groupId as default package name if necessary

### LINKED ISSUE #277: maven project without package name fails
creates this:

package ;

import java.io.IOException;

import javax.servlet.http.HttpServlet;
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletResponse;

public class HelloAppEngine extends HttpServlet {

  public void doGet(HttpServletRequest request, HttpServletResponse response) 
      throws IOException {
      
    response.setContentType("text/plain");
    response.getWriter().println("Hello App Engine!");

  }
}

either archetype needs to be changed or package made required

elharo:
@akerekes fix #277


briandealwis:
LGTM

PR #280, opened 2016-06-23
app-tools-lib-for-java to appengine-plugins-core

### LINKED ISSUE #278: Pull non-snapshot version of app-tools-lib-for-java from maven central
We've changed the artifact name of the common library from app-tools-lib-for-java to appengine-plugins-core before publishing.

We may want to be on snapshot anyway, at least for now, but we do need to change the plugin name either way.

elharo:
@akerekes @meltsufin fix #278


meltsufin:
Currently, you should use 0.1.1-SNAPSHOT version, or pin to 0.1.0 that was released yesterday night.

meltsufin:
LGTM

PR #286, opened 2016-06-24
Run app engine dev server in debug mode via the app tools lib.

nbashirbello:


nbashirbello:
@elharo @briandealwis

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -103,9 +104,20 @@ void startDevServer(List<File> runnables, MessageConsoleStream stream) {
...
 
+    List<String> jvmFlags = new ArrayList<String>();
     // FIXME: workaround bug when running on a Java8 JVM
     // https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/issues/181
-    devServerRunConfiguration.setJvmFlags(Arrays.asList("-Dappengine.user.timezone=UTC"));
+    jvmFlags.add("-Dappengine.user.timezone=UTC");
+
+    if ((mode != null) && mode.equals(ILaunchManager.DEBUG_MODE)) {
+      if (debugPort <= 0 || debugPort > 65535) {
+        throw new IllegalStateException("Debug port is set to " + debugPort
```
elharo: IllegalArgumentException fits better here

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -83,10 +82,12 @@ public final void setModulePublishState2(IModule[] module, int state) {
...
    * Starts the development server.
-   * @param runnables The path to directories that contain configuration files like appengine-web.xml
-   * @param stream
+   * @param runnables the path to configuration files and/or directories that contain configuration files 
+   * @param stream the stream to send development server process output to
+   * @param mode the mode of the server launch defined by ILaunchManager - RUN_MODE or DEBUG_MODE
+   * @param debugPort the port to attach a debugger to if launch is in debug mode
    */
-  void startDevServer(List<File> runnables, MessageConsoleStream stream) {
+  void startDevServer(List<File> runnables, MessageConsoleStream stream, String mode, int debugPort) {
```
elharo: Feels strange to have aguments that sometimes don't apply. Almost tempted to suggest two methods, one for launching debug mode and one for launching regular mode.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -44,7 +56,64 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
     TargetPlatform.showConsole(console);
     console.clearConsole();
 
+    LocalAppEngineServerDelegate serverDelegate = LocalAppEngineServerDelegate.getAppEngineServer(server);
+    int debugPort = -1;
+    if (ILaunchManager.DEBUG_MODE.equals(mode)) {
+      final int port = getDebugPort();
```
elharo: is there some reason we need to introduce the port variable her einstead of just using debugPort?

briandealwis: This code has previously been in an `IServerChangeListener` (or something to that effect).  So it can be dropped

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
briandealwis: There are other modes, like PROFILE.  We should have a check somewhere, I suppose?

briandealwis:
Something that seems to be missing: although dev_appserver can run multiple modules, only a single module can be debugged as each module is launched in a separate JVM.  I had a test previously for that case.

(Though since we switched to the socket-listen mode on the debugger, I wonder if it's possible to have multiple JVMs connect back…)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
nbashirbello: Removed the profile launch mode and added issue (#293) to decide if we want to support it and how.

elharo: I still don't like a debugPort argument that doesn't always apply. Can we refactor this into runDevServer and debugDevServer methods that share the common code?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -44,7 +56,63 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
...
+  private void runDebugTarget(LocalAppEngineServerDelegate serverDelegate, String projectName, int port)
+      throws CoreException {
+    ILaunchConfigurationWorkingCopy remoteDebugLaunchConfig = createRemoteDebugLaunchConfiguration(serverDelegate,
+        projectName,
+        Integer.toString(port));
+    remoteDebugLaunchConfig.launch(ILaunchManager.DEBUG_MODE, null);
+  }
+
+  private ILaunchConfigurationWorkingCopy createRemoteDebugLaunchConfiguration(LocalAppEngineServerDelegate serverDelegate,
+      final String projectName, final String port) throws CoreException {
```
elharo: why are these arguments final? Google style is to omit final on arguments and local variables unless it's required (e.g. by use in an inner class)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -44,7 +56,63 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
...
+    ILaunchManager manager = DebugPlugin.getDefault().getLaunchManager();
+    ILaunchConfigurationType type = manager.getLaunchConfigurationType(IJavaLaunchConfigurationConstants.ID_REMOTE_JAVA_APPLICATION);
+
+    final ILaunchConfigurationWorkingCopy remoteDebugConfig = type.newInstance(null,
```
elharo: why final?

PR #292, opened 2016-06-24
install google cloud sdk on travis

### LINKED ISSUE #267: Set up gcloud on Travis
so that everything that depends on it including PathResolver works

elharo:
fix #267

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,16 @@ sudo: false
 language: java
 jdk:
   - openjdk7
-install: true
+install: 
+    # download Cloud SDK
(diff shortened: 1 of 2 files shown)


briandealwis:
How do we pass in the location so `PathResolver` and friends find it?

elharo:
@briandealwis @meltsufin PathResolver starts by looking in the path, and here we export this dIrectory into the local PATH:

export PATH=$PWD/google-cloud-sdk/bin:$PATH

briandealwis:
Ah right.  LGTM.

elharo:
Hmm, another of those weird situations where the pr/ fails and push succeeds. Actually that makes sense here. The test I'm putting back in won't pass with the old setup. I'm going to merge, but holler if things start breaking.

PR #298, opened 2016-06-27
Attempt to fix tests on travis

elharo:
To preempt a question, I tried Mockito here. It didn't work. We can figure out how to use Mockito later, but that's not enough reason to hold back this PR now if it does indeed fix the tests (which is still an open question)

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
@@ -23,6 +23,9 @@ import com.google.cloud.tools.appengine.api.AppEngineException;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.eclipse.sdk.CloudSdkProvider;
 
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.util.IPropertyChangeListener;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.io.File;
@@ -30,14 +33,21 @@ import java.nio.file.Path;
 
 public class CloudSdkProviderTest {
 
+  private IPreferenceStore preferences;
+  
+  @Before
+  public void setUp() {
+    preferences = new MockPreferences();
+  }
+  
   /** Verify that the preference overrides PathResolver. */
   @Test
   public void testSetPreferenceInvalid() throws Exception {
     // A path that almost certainly does not contain the SDK
     File root = File.listRoots()[0];
-    PreferenceInitializer.getPreferenceStore().putValue(PreferenceConstants.CLOUDSDK_PATH,
-        root.toString());
-    CloudSdk.Builder builder = CloudSdkProvider.createBuilder(null);
+    
+    CloudSdk.Builder builder = new CloudSdkProvider(preferences).createBuilder(null);
+    // todo we shouldn't need reflection here; use visible for testing if we must
     assertEquals(root.toPath(), ReflectionUtil.getField(builder, "sdkPath", Path.class));
     CloudSdk instance = builder.build();
     assertEquals(root.toPath(), ReflectionUtil.invoke(instance, "getSdkPath", Path.class));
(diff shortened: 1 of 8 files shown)


elharo:
Adding @nbashirbello @briandealwis Travis is passing. I'd like to submit this ASAP so other work is unblocked. If you think this PR is _wrong_, please comment. However if you think this is PR merely not ideal (e.g. should use Mockito) let's fix that up later.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
```diff
@@ -48,4 +58,153 @@ public void testSetPreferenceInvalid() throws Exception {
...
+    }
+
+    @Override
+    public String getString(String name) {
+      return File.listRoots()[0].toString();
+    }
+
+    @Override
+    public boolean isDefault(String name) {
+      // TODO Auto-generated method stub
```
akerekes: can be deleted

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
```diff
@@ -48,4 +58,153 @@ public void testSetPreferenceInvalid() throws Exception {
...
+    }
+
+    @Override
+    public int getInt(String name) {
+      return 0;
+    }
+
+    @Override
+    public long getLong(String name) {
+      // TODO Auto-generated method stub
```
akerekes: can be deleted

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -28,17 +28,26 @@
...
 /**
- * Utililty to find the Google Cloud SDK either at locations configured by the user or in standard
+ * Utility to find the Google Cloud SDK either at locations configured by the user or in standard
  * locations on the system.
  */
 public class CloudSdkProvider extends ContextFunction {
 
+  private final IPreferenceStore preferences;
+
+   public CloudSdkProvider(IPreferenceStore preferences) {
```
akerekes: I'd prefer to have an argumentless constructor that works as the `null` case instead of having constructor calls with `null`s and then probably adding a comment with the argument name.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
```diff
@@ -48,4 +58,153 @@ public void testSetPreferenceInvalid() throws Exception {
       // ignore
     }
   }
+  
+  private static class MockPreferences implements IPreferenceStore{
```
akerekes: If Mockito works, it'd be much cleaner to use a `@Mock` instead of this lengthy default implementation.

akerekes:
LGTM

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
elharo: or we remove this and pass in actual non-null preferences from everywhere. Haven't yet decided. May depend on the dependencies that would introduce into other plug-ins.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkProviderTest.java
elharo: Mockito didn't work. We cab probably make it work, but for now this gets us moving again.

PR #301, opened 2016-06-28
do nothing skeleton for the login plugin

### LINKED ISSUE #285: Login plugin skeleton
Add a new do-nothing plugin for login. We still need to figure out the UX here and implementation here, but I think it's safe to imagine that it will be a new plugin of its own. We will need to  include a basic XmlTest for the plugin.xml.

elharo:
@nbashirbello close #285


elharo:
ping

nbashirbello:
lgtm

PR #302, opened 2016-06-28
Follow the convention /virtual/eventType/eventName

chanseokoh:
The backend team says we should follow the convention of

/virtual/<event type>/<event name>

Per discussion in https://github.com/GoogleCloudPlatform/gcloud-intellij/pull/754#discussion_r68245423, we set the event type to our application name.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
@@ -87,7 +87,14 @@ public class AnalyticsPingManagerTest {
   public void testEventTypeEventNameConvention() {
     Map<String, String> parameters =
         AnalyticsPingManager.buildParametersMap("some.event-name", null, null);
-    Assert.assertEquals("/virtual/" + AnalyticsPingManager.APPLICATION_NAME + "/some.event-name",
+    Assert.assertEquals("/virtual/gcloud-eclipse-tools/some.event-name",
         parameters.get("dp"));
+  }
+
+  @Test
+  public void testMetadataConvention() {
+    Map<String, String> parameters =
+        AnalyticsPingManager.buildParametersMap("some.event-name", "times-happened", "1234");
+    Assert.assertEquals("times-happened=1234", parameters.get("dt"));
   }
 }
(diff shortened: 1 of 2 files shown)


elharo:
I think you can test this by pulling out the code that builds the map into a helper method that does not send the post request.

chanseokoh:
Added a test to check the backend convention.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -82,4 +82,12 @@ public void testGetParametersString_percentEscaping() {
         + "%24%25%5E%26%28%29%3F%3C%3E%7B%7D%5D%5B%7C%3A%3B%2F%5C%27%22",
         AnalyticsPingManager.getParametersString(escape));
   }
+
+  @Test
+  public void testEventTypeEventNameConvention() {
+    Map<String, String> parameters =
+        AnalyticsPingManager.buildParametersMap("some.event-name", null, null);
+    Assert.assertEquals("/virtual/" + AnalyticsPingManager.APPLICATION_NAME + "/some.event-name",
```
elharo: I try not to use constants like this in unit tests. If you just use a string literal, then we're testing the constant too.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -37,7 +37,8 @@
   private static final String ANALYTICS_COLLECTION_URL = "https://ssl.google-analytics.com/collect";
 
   // This name will be recorded as an originating app on Google Analytics.
-  private static final String APPLICATION_NAME = "gcloud-eclipse-tools";
+  @VisibleForTesting
```
elharo: definitely don't need @VisibleForTesting here. The string literal suffices.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -156,6 +137,33 @@ private static void sendPostRequest(String parametersString) {
...
+    Map<String, String> parametersMap = new HashMap<>(STANDARD_PARAMETERS);
+    parametersMap.put("cid", getAnonymizedClientId());
+    parametersMap.put("cd19", APPLICATION_NAME);  // cd19: "event type"
+    parametersMap.put("cd20", eventName);
+
+    String virtualPageUrl = "/virtual/" + APPLICATION_NAME + "/" + eventName;
+    parametersMap.put("dp", virtualPageUrl);
+    parametersMap.put("dh", "virtual.eclipse");
+
+    if (metadataKey != null) {
```
elharo: It would now be easy to add a test for this block as well.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
chanseokoh: Oh, right, I like that we can test the constant too this way.

chanseokoh:
Comments addressed. Added another test for the metadata.

elharo:
LGTM

PR #303, opened 2016-06-28
Make a message posted / set connection timeout

chanseokoh:
All of a sudden, I noticed that AnalyticsPingManager is not sending a HTTP request, despite the connection being made and closed actually. (No idea why.)

Calling `getResponseCode()` ensures the event message is sent.

I also set a timeout for the extremely unlikely event of the backend server holding a connection while not sending any output back. (This will never happen actually.)

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -13,9 +13,7 @@ import org.osgi.service.prefs.BackingStoreException;
 
 import java.io.IOException;
 import java.io.OutputStream;
-import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
-import java.io.Writer;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -140,10 +140,16 @@ public class AnalyticsPingManager {
       connection.setDoOutput(true);
       connection.setRequestMethod("POST");
       connection.setRequestProperty("Content-Length", Integer.toString(parametersString.length()));
+      connection.setReadTimeout(3000);  // milliseconds
 
       try (Writer writer = new OutputStreamWriter(connection.getOutputStream(), "UTF-8")) {
         writer.write(parametersString);
         writer.flush();
+      }
+
+      int responseCode = connection.getResponseCode();
+      if (responseCode != 200) {
+        logger.log(Level.WARNING, "Response code not 200: " + responseCode);
       }
     } catch (IOException ex) {
       // Don't try to recover or retry.
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -12,6 +12,7 @@ import org.eclipse.ui.PlatformUI;
 import org.osgi.service.prefs.BackingStoreException;
 
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.UnsupportedEncodingException;
 import java.io.Writer;
@@ -122,8 +123,11 @@ public class AnalyticsPingManager {
       connection.setReadTimeout(3000);  // milliseconds
       byte[] bytesToWrite = parametersString.getBytes("UTF-8");
       connection.setFixedLengthStreamingMode(bytesToWrite.length);
-      connection.getOutputStream().write(bytesToWrite);
-      connection.getOutputStream().flush();
+
+      try (OutputStream out = connection.getOutputStream()) {
+        out.write(bytesToWrite);
+        out.flush();
+      }
     } catch (IOException ex) {
       // Don't try to recover or retry.
       logger.log(Level.WARNING, "Failed to send a POST request", ex);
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -140,11 +140,17 @@ private static void sendPostRequest(String parametersString) {
...
+      connection.setReadTimeout(3000);  // milliseconds
 
       try (Writer writer = new OutputStreamWriter(connection.getOutputStream(), "UTF-8")) {
         writer.write(parametersString);
         writer.flush();
       }
+
+      int responseCode = connection.getResponseCode();
+      if (responseCode != 200) {
+        logger.log(Level.WARNING, "Response code not 200: " + responseCode);
```
elharo: I'm not sure we really need to log this, but if we do specify where this is from. E.g.

Analytics response code:

chanseokoh:
It now skips logging the Analytics response code.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -140,11 +140,14 @@ private static void sendPostRequest(String parametersString) {
...
       connection.setRequestMethod("POST");
       connection.setRequestProperty("Content-Length", Integer.toString(parametersString.length()));
+      connection.setReadTimeout(3000);  // milliseconds
 
       try (Writer writer = new OutputStreamWriter(connection.getOutputStream(), "UTF-8")) {
         writer.write(parametersString);
         writer.flush();
       }
+
+      connection.getResponseCode();  // Important: ensure this request posted.
```
elharo: I think you can get rid of this comment. File a task/testing issue if you want to remind yourself to do this.

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Sorry for the confusion. I'll change this comment to

"This call is important to ensure posting of the request."

I added this comment because we are not checking the response code.

elharo: Might want to reference the issue URL in the code then. And I suggest language such as "Until we read, HTTP data may not be sent"

elharo: Actually, now that I look at it the reason you didn't get a response is you didn't close the output stream.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -140,11 +140,14 @@ private static void sendPostRequest(String parametersString) {
...
       connection.setRequestMethod("POST");
       connection.setRequestProperty("Content-Length", Integer.toString(parametersString.length()));
+      connection.setReadTimeout(3000);  // milliseconds
 
       try (Writer writer = new OutputStreamWriter(connection.getOutputStream(), "UTF-8")) {
         writer.write(parametersString);
         writer.flush();
       }
+
+      connection.getResponseCode();  // This call is important to ensure posting the request.
```
elharo: you need to close the output stream, not just flush it

reading response code may be irrelevant

chanseokoh: I tried closing it before, and I remember it didn't work. I'll test it again, but isn't the try-resources block takes care of it anyway?

chanseokoh: Yeah, I tried closing it, but it doesn't work. Looks like we need getResponseCode().

chanseokoh: Oh, I found a way to make it work without `getResponseCode()`: calling `setFixedLengthStreamingMode()`. We know the content-length beforehand, and by setting this fixed length, the HTTP connection knows exactly when to send the message.

Some info: http://stackoverflow.com/questions/26808136/why-httpurlconnection-does-not-send-data-unless-i-try-to-receive-something

chanseokoh: Actual testing confirms that the Content-Length header is correctly set with the method and things are working properly.

chanseokoh:
So, finally, I removed `getResponseCode()` and instead added `setFixedLengthStreamingMode()` which seems to be a nicer way.

chanseokoh:
Does the last change look good?

briandealwis:
It's not quite right: in UTF-8, the string length may not be the content length (in bytes).

chanseokoh:
Ah, good catch. That's really important. If the content length doesn't match precisely, we may lose many event pings not noticing it.

I pushed a change. How about this?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -119,7 +119,7 @@ private static void sendPostRequest(String parametersString) {
       connection = (HttpURLConnection) url.openConnection();
       connection.setDoOutput(true);
       connection.setRequestMethod("POST");
-      connection.setFixedLengthStreamingMode(parametersString.length());
+      connection.setFixedLengthStreamingMode(parametersString.getBytes("UTF-8").length);
```
briandealwis: You could save the bytes and write it out directly into the `connection.getOutputStream()` rather than re-encoding it again. No doubt this will reap huge performance gains!

Otherwise LGTM.

chanseokoh: Good idea. Now I write directly to OutputStream. I will merge this PR after the test passes.

PR #309, opened 2016-06-29
sendPing() triggers showing opt-in dialog

chanseokoh:
Two enhancements:

- `sendPing()` will show the opt-in dialog (only once) if needed.
- Allows a `sendPing()` caller to set the parent shell of the opt-in dialog.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -62,7 +62,7 @@ public class AnalyticsPingManager {
 
   private static AnalyticsPingManager instance;
 
-  // Preference store (should be ConfigurationScope-d) from which we get UUID, opt-in status, etc.
+  // Preference store (should be configuration scoped) from which we get UUID, opt-in status, etc.
   private IEclipsePreferences preferences;
 
   private OptInDialogCreator optInDialogCreator;
@@ -74,11 +74,11 @@ public class AnalyticsPingManager {
     this.optInDialogCreator = optInDialogCreator;
   }
 
-  public static AnalyticsPingManager getInstance() {
+  public static synchronized AnalyticsPingManager getInstance() {
     if (instance == null) {
       IEclipsePreferences preferences = ConfigurationScope.INSTANCE.getNode(PREFERENCES_PLUGIN_ID);
       if (preferences == null) {
-        throw new RuntimeException("Preference store cannot be null.");
+        throw new NullPointerException("Preference store cannot be null.");
       }
       instance = new AnalyticsPingManager(preferences, new OptInDialogCreator());
     }
@@ -91,9 +91,7 @@ public class AnalyticsPingManager {
   }
 
   private String getAnonymizedClientId() {
-    String clientId = "0";  // For the extremely unlikely event of getNode() failure.
-
-    clientId = preferences.get(CloudToolsPreferencePage.ANALYTICS_CLIENT_ID, null);
+    String clientId = preferences.get(CloudToolsPreferencePage.ANALYTICS_CLIENT_ID, null);
     if (clientId == null) {
       clientId = UUID.randomUUID().toString();
       preferences.put(CloudToolsPreferencePage.ANALYTICS_CLIENT_ID, clientId);
@@ -102,7 +100,7 @@ public class AnalyticsPingManager {
     return clientId;
   }
 
-  private boolean hasUserOptedIn() {
+  private boolean userHasOptedIn() {
     return preferences.getBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN, false);
   }
 
@@ -112,10 +110,19 @@ public class AnalyticsPingManager {
    * opting out. (Therefore, this method essentially returns true if the opt-in dialog was ever
    * presented before.)
    */
-  private boolean hasUserRegisteredOptInStatus() {
+  private boolean userHasRegisteredOptInStatus() {
     return preferences.getBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN_REGISTERED, false);
   }
 
+  /**
+   * "Registered" here means a user made a decision through the opt-in dialog. (That is, it
+   * essentially means if the opt-in dialog was ever presented in our current setting.) Note
+   * that it is possible that a user may have already opted in/out explicitly through the
+   * preference page without ever seeing the opt-in dialog. It would have been ideal if we
+   * defined a three-valued settings (e.g., OPTED_IN | OPTED_OUT | UNDEF). However, this would
+   * require implementing a custom UI editor field class for the preference page, so here we
+   * take a simple approach to use two Boolean settings.
+   */
   void registerOptInStatus(boolean optedIn) {
     preferences.putBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN, optedIn);
     preferences.putBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN_REGISTERED, true);
@@ -124,7 +131,7 @@ public class AnalyticsPingManager {
 
   /**
    * Sends a usage metric to Google Analytics.
-   * <p/>
+   *
    * If the user has never seen the opt-in dialog or set the opt-in preference beforehand,
    * this method can potentially present an opt-in dialog at the top workbench level. If you
    * are calling this method inside another modal dialog, consider using {@link #sendPing(
@@ -141,7 +148,7 @@ public class AnalyticsPingManager {
     // sendPing() may drop this event.
     showOptInDialog(parentShell);
 
-    if (Platform.inDevelopmentMode() || !isTrackingIdDefined() || !hasUserOptedIn()) {
+    if (Platform.inDevelopmentMode() || !isTrackingIdDefined() || !userHasOptedIn()) {
       return;
     }
 
@@ -228,7 +235,7 @@ public class AnalyticsPingManager {
    * @param parentShell if null, tries to show the dialog at the workbench level.
    */
   public void showOptInDialog(Shell parentShell) {
-    if (!hasUserOptedIn() && !hasUserRegisteredOptInStatus()) {
+    if (!userHasOptedIn() && !userHasRegisteredOptInStatus()) {
       optInDialogCreator.create(findShell(parentShell)).open();
     }
   }
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -7,6 +7,7 @@ import com.google.common.annotations.VisibleForTesting;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.ConfigurationScope;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PlatformUI;
 import org.osgi.service.prefs.BackingStoreException;
@@ -102,10 +103,26 @@ public class AnalyticsPingManager {
     }
   }
 
+  /**
+   * Sends a usage metric to Google Analytics.
+   * <p/>
+   * If the user has never seen the opt-in dialog or set the opt-in preference beforehand,
+   * this method can potentially present an opt-in dialog at the top workbench level. If you
+   * are calling this method inside another modal dialog, consider using {@link #sendPing(
+   * String, String, String, Shell)} and pass the {@Shell} of currently open modal dialog.
+   * (Otherwise, the opt-in dialog won't be able to get input until the workbench can get input.)
+   */
   public static void sendPing(String eventName, String metadataKey, String metadataValue) {
+    sendPing(eventName, metadataKey, metadataValue, null);
+  }
+
+  public static void sendPing(String eventName, String metadataKey, String metadataValue,
+      Shell parentShell) {
     if (Platform.inDevelopmentMode() || !isTrackingIdDefined() || !hasUserOptedIn()) {
       return;
     }
+
+    showOptInDialog(parentShell);
 
     Map<String, String> parametersMap = buildParametersMap(eventName, metadataKey, metadataValue);
     sendPostRequest(getParametersString(parametersMap));
@@ -183,13 +200,21 @@ public class AnalyticsPingManager {
     return resultBuilder.toString();
   }
 
-  public static void showOptInDialog() {
+  /**
+   * @param parentShell if null, shows the dialog at the workbench level.
+   */
+  public static void showOptInDialog(Shell parentShell) {
     if (!hasUserOptedIn() && !hasUserRegisteredOptInStatus()) {
-      IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
-      if (window != null) {
-        new OptInDialog(window.getShell()).open();
+      if (parentShell != null) {
+        new OptInDialog(parentShell).open();
       } else {
-        logger.log(Level.WARNING, "No active workbench window found.");
+        // Show the dialog at the workbench level, if it exists.
+        IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+        if (window != null) {
+          new OptInDialog(window.getShell()).open();
+        } else {
+          logger.log(Level.WARNING, "No active workbench window found.");
+        }
       }
     }
   }
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -232,6 +232,9 @@ public class AnalyticsPingManager {
     }
   }
 
+  /**
+   * May return null. (However, dialogs can have null as a parent shell.)
+   */
   private Shell findShell(Shell parentShell) {
     if (parentShell != null) {
       return parentShell;
@@ -239,10 +242,15 @@ public class AnalyticsPingManager {
 
     try {
       IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
-      return window == null ? null : window.getShell();
+      if (window != null) {
+        return window.getShell();
+      }
     } catch (IllegalStateException ise) {  // getWorkbench() might throw this.
-      return null;
+      // Fall through.
     }
+
+    Display display = Display.getCurrent();
+    return display != null ? display.getActiveShell() : null;
   }
 
   private void flushPreferences() {
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -60,28 +60,50 @@ public class AnalyticsPingManager {
         }
       });
 
+  private static AnalyticsPingManager instance;
+
+  // Preference store (should be ConfigurationScope-d) from which we get UUID, opt-in status, etc.
+  private IEclipsePreferences preferences;
+
+  private OptInDialogCreator optInDialogCreator;
+
+  @VisibleForTesting
+  AnalyticsPingManager(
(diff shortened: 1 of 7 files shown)


elharo:
Is there anyway we could test any of this?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -102,11 +103,27 @@ static void registerOptInStatus(boolean optedIn) {
...
+   * String, String, String, Shell)} and pass the {@Shell} of currently open modal dialog.
+   * (Otherwise, the opt-in dialog won't be able to get input until the workbench can get input.)
+   */
   public static void sendPing(String eventName, String metadataKey, String metadataValue) {
+    sendPing(eventName, metadataKey, metadataValue, null);
+  }
+
+  public static void sendPing(String eventName, String metadataKey, String metadataValue,
+      Shell parentShell) {
     if (Platform.inDevelopmentMode() || !isTrackingIdDefined() || !hasUserOptedIn()) {
```
elharo: this line is now confusing because there are three states: opted in, opted out, not chosen. In the not yet chosen state, the user has not opted in, but we don't want to stop either.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -102,11 +103,27 @@ static void registerOptInStatus(boolean optedIn) {
...
       return;
     }
 
+    showOptInDialog(parentShell);
```
elharo: We should show the opt in dialog here, but we should not send the ping if the user doesn't agree. This looks like it sends the ping no matter what they say here.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -102,11 +103,27 @@ static void registerOptInStatus(boolean optedIn) {
...
+
     Map<String, String> parametersMap = buildParametersMap(eventName, metadataKey, metadataValue);
     sendPostRequest(getParametersString(parametersMap));
```
elharo: probably need to wrap this in 

if (hasUserOptedIn())

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Oh, the bear got me yesterday.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -183,13 +200,21 @@ static String getParametersString(Map<String, String> parametersMap) {
...
-  public static void showOptInDialog() {
+  /**
+   * @param parentShell if null, shows the dialog at the workbench level.
+   */
+  public static void showOptInDialog(Shell parentShell) {
     if (!hasUserOptedIn() && !hasUserRegisteredOptInStatus()) {
-      IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
-      if (window != null) {
-        new OptInDialog(window.getShell()).open();
+      if (parentShell != null) {
```
briandealwis: I would move this logic into `findShell(parentShell)`.  `findShell(parentShell)` would:
  - return `parentShell` if not null
  - look for the active workbench window and return its window
  - look for `Display.getCurrent().getActiveShell()` providing there's a Display
  - return null

`null` is a valid parent shell: it just means the dialog is un-parented.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
briandealwis: This code implies the sendPing will block…  Rather than pass a shell, I wonder if we should instead have a flag to say _it's ok to query the user_ vs _drop event if user hasn't yet agreed_.

chanseokoh: The dialog is non-modal, and the code doesn't block. (I verified it.) This implies that the very first encounter of `sendPing()` will drop the event, but I think that's ok. Therefore, it would be better to also place `showOptInDialog()` in more prominent places before `sendPing()` is likely to happen.

briandealwis: Great — could you add a comment to this effect?  General conventions for dialogs are they are blocking, so it's good to note the deviation

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Sounds like a good idea. For the third bullet, how can I know there's a Display? You said we shouldn't call `Display.getCurrent()` if not in the main thread.

chanseokoh:
To add unit tests, I made structural changes. Some prominent changes are

- `ShowOptInDialog` is called at the beginning of `sendPing()`.
- `AnalyticsPingManager` now gets `PreferenceStore` and `OptInDialogCreator` for dependency injection. (Not sure what I did here is ideal.)
- Many static methods are now instance methods. In fact, I think this change makes more sense; previously, the behaviors of those static methods depended on external states, did not show consistent (i.e., pure function-like) results, and had side effects.
- Now it assumes that the `PreferenceStore` is always there (i.e., not null). The store is defined in the Activator, so I believe this is a safe assumption.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: For now, I left out the `Display` logic in the new `findShell()` method. I also catch `IllegalStateException` that `getWorkbench()` might throw. (This exception will be very usual in practice, but I realized it can happen for example when the IDE is being launched.)

briandealwis: `Display.getCurrent()` is ok (it looks to see if there's a `Display` for the current thread); `Display.getDefault()` is not (it will create a new `Display` if one is not found, which is bad).

chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -59,28 +61,50 @@
...
+      IEclipsePreferences preferences, OptInDialogCreator optInDialogCreator) {
+    this.preferences = preferences;
+    this.optInDialogCreator = optInDialogCreator;
+  }
+
+  public static AnalyticsPingManager getInstance() {
+    if (instance == null) {
+      IEclipsePreferences preferences = ConfigurationScope.INSTANCE.getNode(PREFERENCES_PLUGIN_ID);
+      if (preferences == null) {
+        throw new RuntimeException("Preference store cannot be null.");
```
elharo: NullPointerException

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -59,28 +61,50 @@
...
+  private OptInDialogCreator optInDialogCreator;
+
+  @VisibleForTesting
+  AnalyticsPingManager(
+      IEclipsePreferences preferences, OptInDialogCreator optInDialogCreator) {
+    this.preferences = preferences;
+    this.optInDialogCreator = optInDialogCreator;
+  }
+
+  public static AnalyticsPingManager getInstance() {
```
elharo: should this method be synchronized?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -59,28 +61,50 @@
         }
       });
 
+  private static AnalyticsPingManager instance;
+
+  // Preference store (should be ConfigurationScope-d) from which we get UUID, opt-in status, etc.
```
elharo: configuration scoped

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -59,28 +61,50 @@
...
+  }
+
   private static boolean isTrackingIdDefined() {
     return Constants.ANALYTICS_TRACKING_ID != null
         && Constants.ANALYTICS_TRACKING_ID.startsWith("UA-");
   }
 
-  private static String getAnonymizedClientId() {
+  private String getAnonymizedClientId() {
     String clientId = "0";  // For the extremely unlikely event of getNode() failure.
```
elharo: I think this value will get overwritten no matter what. combine with next line

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -59,28 +61,50 @@
...
+      clientId = UUID.randomUUID().toString();
+      preferences.put(CloudToolsPreferencePage.ANALYTICS_CLIENT_ID, clientId);
+      flushPreferences();
     }
     return clientId;
   }
 
-  private static boolean hasUserOptedIn() {
-    return getBooleanPreference(CloudToolsPreferencePage.ANALYTICS_OPT_IN, false);
+  private boolean hasUserOptedIn() {
```
elharo: userHasOptedIn

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -89,29 +113,45 @@ private static boolean hasUserOptedIn() {
    * opting out. (Therefore, this method essentially returns true if the opt-in dialog was ever
    * presented before.)
    */
-  private static boolean hasUserRegisteredOptInStatus() {
-    return getBooleanPreference(CloudToolsPreferencePage.ANALYTICS_OPT_IN_REGISTERED, false);
+  private boolean hasUserRegisteredOptInStatus() {
```
elharo: userHasRegisteredOptInStatus

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -89,29 +113,45 @@ private static boolean hasUserOptedIn() {
...
   }
 
-  static void registerOptInStatus(boolean optedIn) {
-    IEclipsePreferences prefs = ConfigurationScope.INSTANCE.getNode(PREFERENCES_PLUGIN_ID);
-    if (prefs != null) {
-      prefs.putBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN, optedIn);
-      prefs.putBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN_REGISTERED, true);
-      flushPreferences(prefs);
-    }
+  void registerOptInStatus(boolean optedIn) {
```
elharo: I'm surprised this is a preference of its own. It feels like this should be conditioned on other values.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -89,29 +113,45 @@ private static boolean hasUserOptedIn() {
...
+    preferences.putBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN, optedIn);
+    preferences.putBoolean(CloudToolsPreferencePage.ANALYTICS_OPT_IN_REGISTERED, true);
+    flushPreferences();
+  }
+
+  /**
+   * Sends a usage metric to Google Analytics.
+   * <p/>
```
elharo: don't use <p/> Either <p></p> or just drop the tags completely.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: That's a good question. I thought it is tolerable to create two and overwrite the previous one without synchronization. However, if we are really concerned about absolute correctness in concurrency, I think not only this instance but also every access to the preference store (which is sort of a shared repository) might need to be synchronized. But then, I think concurrent access to the preference store is also tolerable. Your thoughts? I will follow what we finally have consensus on.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialog.java
```diff
@@ -26,13 +28,28 @@ public OptInDialog(Shell parentShell) {
...
+    } catch (IllegalStateException ise) {  // getWorkbench() might throw this.
+      // No special handling. targetShell will be null.
+    }
 
-    int heightCaptionAndUpperBorder =
-        parentBounds.height - parentClientArea.height - getParentShell().getBorderWidth();
-    return new Point(parentBounds.x + parentClientArea.width - initialSize.x,
-        parentBounds.y + heightCaptionAndUpperBorder);
+    if (targetShell == null) {
+      return super.getInitialLocation(initialSize);
```
elharo: I think this case belongs in the catch block; and the rest of this code probably in the try block

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
elharo: The preference store should be as thread safe as it needs to be and no extra synchronization should be needed there; but check the docs and ask @briandealwis in case it isn't.

elharo: You do not want to create two and overwrite the previous one. That way lies madness.

elharo: also see here:

http://stackoverflow.com/questions/11165852/java-singleton-and-synchronization

singletons are much trickier than they look in multihtreaded environments

chanseokoh: Thanks for the link. I'll first go with the simplest approach: `public static synchronized`

For the preference store, I can't find any relevant info regarding thready-safety (which is typical), but there isn't any code example on the Internet that does any sort of sync, so I assume we are good with accessing it concurrently.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: I agree this can be confusing. OPT_IN and OPT_IN_REGISTERED are certainly related. However, they are also independent in some sense. I added some more comments for clarification.

Details:
It would have been ideal if I had a three-valued preference setting: OPTED_IN, OPTED_OUT, and UNDEF. However, such setting will require implementing a custom UI field editor for the preference page, so I took a simple approach to use two Boolean settings (so that I can just use a single `BooleanFieldEditor` in the UI).

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialog.java
chanseokoh: The targetShell logic was not perfect anyway. I refactored this portion out, so hopefully it's more succinct.

chanseokoh:
Comments addressed. ptal.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialog.java
elharo: I stand by my original comment here. The if-else should be removed and combined into the try and catch blocks.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
elharo: My concern is that OPT_IN_REGISTERED is not a preference at all. I.e. the user does not set this.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialog.java
chanseokoh: Hmm, I could push the code into the try-catch block, but I think we still need the if-else. You don't want something like below, do you?

```java
  @Override
  protected Point getInitialLocation(Point initialSize) {
    try {
      IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
      if (window != null) {
        return getTopRightLocation(window.getShell(), initialSize);
      }
      if (getParentShell() != null) {
        return getTopRightLocation(getParentShell(), initialSize);
      }
      return super.getInitialLocation(initialSize);
    } catch (IllegalStateException ise) {  // getWorkbench() might throw this.
      if (getParentShell() != null) {
        return getTopRightLocation(getParentShell(), initialSize);
      }
      return super.getInitialLocation(initialSize);
    }
  }

  private Point getTopRightLocation(Shell targetShell, Point initialSize) {
    // Position the dialog at the top-right corner of the targetShell window.
    Rectangle parentBounds = targetShell.getBounds();
    Rectangle parentClientArea = targetShell.getClientArea();

    int heightCaptionAndUpperBorder =
        parentBounds.height - parentClientArea.height - targetShell.getBorderWidth();
    return new Point(parentBounds.x + parentClientArea.width - initialSize.x,
        parentBounds.y + heightCaptionAndUpperBorder);
  }
```

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Ah, I get your point. Currently, the OPT_IN_REGISTERED cannot be conditioned on OPT_IN itself. The only way we can get rid of OPT_IN_REGISTERED is to use a three-valued setting like I said: OPTED_IN, OPTED_OUT, and UNDEF. I admit that employing a three-valued setting is conceptually ideal, so I will create an issue for this and fix it in another PR.

elharo:
LGTM

PR #321, opened 2016-07-01
Don't add entries to classpath in maven project

### LINKED ISSUE #312: Should a maven project use the facet classpath?
image

elharo:
From either runtimes or facets. Let maven handle it

@nbashirbello @akerekes 

close #312

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegateTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegateTest.java
@@ -1,0 +1,21 @@
+package com.google.cloud.tools.eclipse.appengine.facets;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class FacetInstallDelegateTest {
+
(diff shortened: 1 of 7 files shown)


akerekes:
LGTM

elharo:
I think the push failure is flakiness, but yell if any other tests start failing after this.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegateTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+  private FacetInstallDelegate delegate = new FacetInstallDelegate();
+  
+  @Test
+  public void testMavenNature() throws CoreException {
+    IProject project = Mockito.mock(IProject.class);
+    Mockito.when(project.hasNature("org.eclipse.m2e.core.maven2Nature")).thenReturn(true);
+    Mockito.when(project.isAccessible()).thenReturn(true);
+    
+    delegate.execute(project, null, null, null);
```
akerekes: Ugly and not sure we should do it, but we could assert on project.getType() being called which is invoked during the JavaCore.create(...) method execution and this way we could verify if the body of the `if` statement is executed. Otherwise this test is not useful.

elharo: I think this will fail with an exception if a maven nature is not present.

PR #332, opened 2016-07-07
Use 0.1.3-SNAPHSOT for appengine-plugins-core

chanseokoh:
The 0.1.1-SHAPSHOT doesn't seem to exist anymore in the Maven Central.


akerekes:
0.1.1-SNAPSHOT is indeed  missing from the snapshot repo, but 0.1.2-SNAPSHOT and 0.1.3-SNAPSHOT are available. I think as long as we are in development mode, we can depend on the latest SNAPSHOT version.

chanseokoh:
Ah, I get that. Let's use 0.1.3-SNAPSHOT then.

akerekes:
LGTM

Issue #283, opened 2016-06-23
Minimal Deploy


elharo:
Given how app-tools-lib-for-java is designed, I think we can kick off deployment without solving login first. Specifically, I think we can do these tasks:

1. New deploy plugin. This does nothing but does not include a basic XmlTest for the plugin.xml.
1. Make the plugin add a "Deploy to AppEngine" context menu item on all projects.
1. Limit the menu item to appearing only on projects with the app engine facet.
1. Export the exploded WAR file to a staging directory
1. Call staging operation via CloudSDK
1. Use app-tools-lib-for-java to deploy the project, assuming the user has logged in at the command line in the gcloud cli.
1. Cleanup previous deploy work directories.

Note that as tasked out here this is at least 4 separate PRs.

Later (or in parallel) we can add in IDE login that will be used instead.

Thoughts?

patflynn:
seems reasonable.

akerekes:
All tasks have been implemented.

PR #333, opened 2016-07-08
Write the exploded WAR file to a staging directory

akerekes:
Related to #283 , implements item 4)


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
```diff
@@ -3,17 +3,94 @@
...
+  
+  public AppEngineStandardDeployCommandHandler() {
+    this(new ProjectToStagingExporter(), new DialogStagingDirectoryProvider(), new FacetedProjectHelper());
+  }
+  
+  @VisibleForTesting
+  AppEngineStandardDeployCommandHandler(ProjectToStagingExporter exporter, 
+                                        StagingDirectoryProvider stagingDirectoryProvider,
+                                        FacetedProjectHelper facetedProjectHelper) {
+    if (exporter == null) {
```
briandealwis: This seems kinda grotty.  Mightn't it be a bit cleaner to just make these package protected and have the tests set those variables explicitly?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
```diff
@@ -3,17 +3,94 @@
...
   public Object execute(ExecutionEvent event) throws ExecutionException {
-    new MessageDialog(HandlerUtil.getActiveShell(event), Messages.getString("deploy.standard.dialog.title"), null, //$NON-NLS-1$
-      Messages.getString("deploy.standard.dialog.message"), //$NON-NLS-1$
-      MessageDialog.INFORMATION, new String[]{Messages.getString("button.ok.label")}, 0).open(); //$NON-NLS-1$
-    return null;
+    try {
+      IEvaluationContext applicationContext = (IEvaluationContext)event.getApplicationContext();
+      //TODO should we use the approach from http://www.programcreek.com/2013/02/eclipse-rcp-tutorial-add-a-menu-item-to-the-popup-menu-when-right-click-a-file/
+      IStructuredSelection selection = 
+          (IStructuredSelection) applicationContext.getVariable(ISources.ACTIVE_CURRENT_SELECTION_NAME);
```
briandealwis: Use `HandlerUtil.getCurrentSelectionChecked(event)`

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
```diff
@@ -3,17 +3,94 @@
...
+      if (selection.getFirstElement() instanceof IProject) {
+        IProject project = (IProject) selection.getFirstElement();
```
briandealwis: You should check that the selection has exactly 1 element (could be empty, could be more)

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
```diff
@@ -3,17 +3,94 @@
...
+      // must be null, reserved for future use
+      return null;
+    } catch (CoreException coreException) {
+      throw new ExecutionException("Failed to export the project as exploded WAR", coreException);
+    }
+  }
+
+  private void exportProjectToStageDir(ExecutionEvent event, IProject iProject) throws CoreException {
+    String stageDir = getStagingDirectory(event);
+    if (stageDir != null) {
```
briandealwis: Just make `getStagingDirectory` thrown an `ExecutionException` on `null`.  And why not just pass in the staging directory directly?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
```diff
@@ -3,17 +3,94 @@
...
+      verifyProjectInStageDir(stageDir);
+    }
+  }
+
+  private void writeProjectToStageDir(IProject project, String stageDir) throws CoreException {
+    projectToStagingExporter.writeProjectToStageDir(project, stageDir);
+  }
+
+  private String getStagingDirectory(ExecutionEvent event) {
+    return stagingDirectoryProvider.getStagingDirectory(event);
```
briandealwis: `ExecutionEvent` objects aren't meant to be passed around; your objects independent of the Commands framework (`ExecutionEvent`)

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
```diff
@@ -3,17 +3,94 @@
...
+      IEvaluationContext applicationContext = (IEvaluationContext)event.getApplicationContext();
+      //TODO should we use the approach from http://www.programcreek.com/2013/02/eclipse-rcp-tutorial-add-a-menu-item-to-the-popup-menu-when-right-click-a-file/
+      IStructuredSelection selection = 
+          (IStructuredSelection) applicationContext.getVariable(ISources.ACTIVE_CURRENT_SELECTION_NAME);
+      if (selection.getFirstElement() instanceof IProject) {
+        IProject project = (IProject) selection.getFirstElement();
+        IFacetedProject facetedProject = facetedProjectHelper.getFacetedProject(project);
+        // TODO replace with constant from AppEngineFacet (after possibly relocating that class)
+        if (facetedProjectHelper.projectHasFacet(facetedProject, 
+                                                 "com.google.cloud.tools.eclipse.appengine.facet")) {
```
briandealwis: Don't we have a constant somewhere for this string?  We reference it from multiple places.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
briandealwis: Or maybe we pull in the JSR308 `@NonNull` here?

briandealwis:
A few comments:

I don't think there's any value in extracting `Preconditions` strings out for i18n: they're developer messages and will never be shown to the user.  The `Messages.getString()` approach is especially verbose due to the requirement to add a ``//$NON-NLS-X$`` comment.

To be honest I found this code very hard to grok due to the contortions to test the individual units.  Which is unfortunate as the code itself is actually very simple.  Perhaps separate out the get-the-provided-facet-project-from-an-ISelection so that can be tested independently?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
akerekes: `ExecutionException` on null -> done.

Re the second half of the comment:
The level I'm planning to implement execute() on would be higher than obtaining the staging directory. It would call methods to copy to staging, do staging and then do deploy. The details of these steps would be implemented by these methods.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineStandardDeployCommandHandler.java
akerekes: I just removed the null checks on the assumption that this method will only be used in tests and if they are not set, then we'll have test failures, but not production code would break.

akerekes:
@elharo @briandealwis PTAL I have refactored the code, please disregard test classes for now, I just wanted to put this out ASAP to see if it is an improved approach.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineDeployJob.java
```diff
@@ -0,0 +1,11 @@
+package com.google.cloud.tools.eclipse.appengine.deploy;
+
+import org.eclipse.core.resources.IWorkspaceRunnable;
+
+/**
+ * Tagger interface for App Engine Deploy Jobs
```
elharo: Tagger --> Marker (standard terminology, I think)

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJobRunner.java
```diff
@@ -0,0 +1,25 @@
+package com.google.cloud.tools.eclipse.appengine.deploy;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+
+/**
+ * Runs a {@link AppEngineDeployJob} in an {@link IWorkspace} using a lock to prevent any changes to the project while
```
elharo: a --> an

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJobRunner.java
```diff
@@ -0,0 +1,25 @@
...
+ * the operation is in progress.
+ *
+ */
+// Main purpose of this class is to separate Eclipse' job execution from other logic to enhance testability
```
elharo: Eclipse's

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJobRunner.java
```diff
@@ -0,0 +1,25 @@
...
+public class DeployJobRunner {
+
+  public void runJob(AppEngineDeployJob deployJob, IProject project, IProgressMonitor monitor) throws CoreException {
```
elharo: Is there a superinterface of Job we could use here? It seems not to depend on this being an app engine job speciifcally

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -0,0 +1,46 @@
...
+                           StagingDirectoryProvider stagingDirectoryProvider,
+                           IProject project) {
+    projectToStagingExporter = exporter;
+    this.stagingDirectoryProvider = stagingDirectoryProvider;
+    this.project = project;
+  }
+
+  @Override
+  public void run(IProgressMonitor monitor) throws CoreException {
+    String stageDir = stagingDirectoryProvider.getStagingDirectory();
```
elharo: I think you can just use Files.createTempDirectory here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -0,0 +1,46 @@
...
+    if (stageDir == null) {
```
elharo: if so, this shouldn't be needed

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -0,0 +1,46 @@
...
+    writeProjectToStageDir(project, stageDir);
+    verifyProjectInStageDir(stageDir);
+    //TODO run stage and deploy operations
+  }
+  
+  void writeProjectToStageDir(IProject project, String stageDir) throws CoreException {
+    projectToStagingExporter.writeProjectToStageDir(project, stageDir);
+  }
+
+  void verifyProjectInStageDir(String stageDir) {
```
elharo: why would we expect it not to be?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJobRunner.java
akerekes: Removed `AppEngineDeployJob` marker interface, and replaced it with Eclipse's `IWorkspaceRunnable`, renamed the class to indicate its more general applicability.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
akerekes: For now I'll leave it as it is until in the next steps we implement a better way to configure it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
akerekes: Because the staging directory might be created on a different file system as the Eclipse workspace and then the files are not guaranteed to be copied over during publish.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/plugin.xml
```diff
@@ -25,7 +25,7 @@
   
   <extension point="org.eclipse.ui.handlers">
     <handler
-      class="com.google.cloud.tools.eclipse.appengine.deploy.AppEngineStandardDeployCommandHandler"
+      class="com.google.cloud.tools.eclipse.appengine.deploy.standard.StandardDeployCommandHandler"
       commandId="com.google.cloud.tools.eclipse.appengine.deploy.standard">
```
briandealwis: Typically we'd have a single command ('com.google.cloud.tools.eclipse.appengine.deploy') that would have different Standard and Flex handlers that use `<enabledWhen>` expressions to determine where they're applicable.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/ExclusiveProjectAccessJobRunner.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+// Main purpose of this class is to separate Eclipse' job execution from other logic to enhance testability
+// TODO move to common bundle to reuse across plugin
+public class ExclusiveProjectAccessJobRunner {
+
+  public void runJob(IWorkspaceRunnable job, IProject project, IProgressMonitor monitor) throws CoreException {
+    IWorkspace workspace = project.getWorkspace();
+    workspace.run(
+        job,
+        workspace.getRuleFactory().createRule(project),
```
briandealwis: You can simplify this to provide just the project as `IProject` is an `ISchedulingRule`

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/DialogStagingDirectoryProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.eclipse.swt.widgets.DirectoryDialog;
+import org.eclipse.swt.widgets.Shell;
+
+import com.google.cloud.tools.eclipse.appengine.deploy.Messages;
+
+/**
+ * Prompts the user to select a directory for staging.
+ */
+// TODO should be moved to separate UI bundle (if kept at all in deploy's final implementation)
```
briandealwis: If this is to be moved to a separate bundle, then the interface should just be `DirectoryProvider`, and this should be a `DialogDirectoryProvider` with the labels provided as arguments.  After all, how many places are we going to need a staging directory?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/DialogStagingDirectoryProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+  private Shell parentShell;
+
+  public DialogStagingDirectoryProvider(Shell shell) {
+    this.parentShell = shell;
+  }
+
+  @Override
+  public String getStagingDirectory() {
+    DirectoryDialog directoryDialog = new DirectoryDialog(parentShell);
```
briandealwis: We should probably use `DirectoryDialog(shell, SWT.APPLICATION_MODAL | SWT.SHEET)` to get nice appearance on OS X and Unity.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StagingDirectoryProvider.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.eclipse.core.resources.IProject;
+
+/**
+ * Common interface class to obtain the staging directory for App Engine Standard deploy.
+ */
+public interface StagingDirectoryProvider {
+
+  /**
+   * @param event the event that was triggered to run the App Engine Standard deploy
```
briandealwis: These comments are now out of date

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StagingDirectoryProvider.java
```diff
@@ -0,0 +1,18 @@
...
+   * @return path to the directory to be used for staging
+   * @see com.google.cloud.tools.eclipse.appengine.deploy.standard.ProjectToStagingExporter#writeProjectToStageDir(IProject, String)
+   * @see com.google.cloud.tools.eclipse.appengine.deploy.standard.StandardDeployCommandHandler
+   */
+  String getStagingDirectory();
```
briandealwis: If this is truly a library class, perhaps `getDirectory()` or just `get()` (like Java8's `java.util.function.Supplier`).

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -0,0 +1,70 @@
...
+    jobRunner.runJob(
+        new StandardDeployJob(new ProjectToStagingExporter(),
+                              new DialogStagingDirectoryProvider(HandlerUtil.getActiveShell(event)),
+                              project),
+        project,
+        new NullProgressMonitor());
+  }
+
+  @Override
+  public boolean isEnabled() {
```
briandealwis: Remove this `isEnabled()`.  If we need to do anything, then we call `setEnabled()`.  Generally we should control everything via the `<enabledWhen>` expression.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -0,0 +1,50 @@
...
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IWorkspaceRunnable;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+
+import com.google.common.io.Files;
+
+public class StandardDeployJob implements IWorkspaceRunnable {
```
briandealwis: I'm not a fan of this name since it's not actually a Job.  Which is weird since it otherwise looks like a Job.

Why doesn't this just subclass `WorkspaceJob` and call `setRule(project)` in its constructor?  Then you just use normal Eclipse mechanisms to `schedule()` the job.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/DialogStagingDirectoryProvider.java
akerekes: I'm not sure if this will be kept in the future. Now it helps working around the lack of more sophisticated way to configure the staging directory. For now I just wanted an interface that the deploy can depend on obtaining the staging location and a quick implementation via this class. I don't expect this class to be hanging around for long.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StagingDirectoryProvider.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StagingDirectoryProvider.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
akerekes: done

briandealwis:
lgtm

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -16,9 +16,9 @@
   private IProject project;
 
   public StandardDeployJob(ProjectToStagingExporter exporter,
-                        StagingDirectoryProvider stagingDirectoryProvider,
-                        IProject project) {
-    super(Messages.getString("deploy.standard.runnable.name"));
+                           StagingDirectoryProvider stagingDirectoryProvider,
```
briandealwis: It seems strange to me that we'd pass in a StagingDirectoryProvider here rather than obtaining the staging directory and configuring the job with it.  I could see a provider being passed in if we *might* need the directory

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/messages.properties
```diff
@@ -1,3 +1,7 @@
+dialog.staging.directory.provider.title=Select staging directory
+dialog.staging.directory.provider.message=Select a directory that will be used to export the exploded war file to run stage and deploy operations
+
 deploy.standard.dialog.title=Deploying to App Engine Standard
 deploy.standard.dialog.message=The project is being deployed to App Engine Standard (not really)
+deploy.standard.runnable.name=Deploying to App Engine Standard
```
elharo: This is the same as dialog.title above. Would one do?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/messages.properties
```diff
@@ -1,3 +1,7 @@
+dialog.staging.directory.provider.title=Select staging directory
+dialog.staging.directory.provider.message=Select a directory that will be used to export the exploded war file to run stage and deploy operations
```
elharo: we shouldn't need this

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StagingDirectoryProvider.java
```diff
@@ -0,0 +1,17 @@
...
+ * Common interface class to obtain the staging directory for App Engine Standard deploy.
+ */
+public interface StagingDirectoryProvider {
+
+  /**
+   * @return path to the directory to be used for staging
+   * @see com.google.cloud.tools.eclipse.appengine.deploy.standard.ProjectToStagingExporter#writeProjectToStageDir(IProject, String)
+   * @see com.google.cloud.tools.eclipse.appengine.deploy.standard.StandardDeployCommandHandler
+   */
+  String get();
```
elharo: why String and not Path?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -0,0 +1,49 @@
...
+                           IProject project) {
+    super(Messages.getString("deploy.standard.runnable.name")); //$NON-NLS-1$
+    setRule(project);
+    projectToStagingExporter = exporter;
+    this.stagingDirectoryProvider = stagingDirectoryProvider;
+    this.project = project;
+  }
+
+  @Override
+  public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
```
elharo: instead of throwing CoreException, should this method return an error status?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -0,0 +1,49 @@
...
+    String stageDir = stagingDirectoryProvider.get();
```
elharo: It's simpler to just call Files.createTempDirectory here. At the least, at a TODO for this.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
elharo: so what do we do if they're not? Is there a more reliable way we can copy them?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/messages.properties
akerekes: `deploy.standard.dialog.title` and `deploy.standard.dialog.message` removed as they are not used anymore.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -3,47 +3,45 @@
...
-    if (stageDir != null) {
-      writeProjectToStageDir(project, stageDir);
-      verifyProjectInStageDir(stageDir);
+    if (stagingDir.toFile().exists()) {
+      projectToStagingExporter.writeProjectToStageDir(project, stagingDir);
       //TODO run stage and deploy operations
       return Status.OK_STATUS;
     } else {
-      return Status.CANCEL_STATUS;
+      return new Status(IStatus.ERROR,
```
elharo: actually I thought cancel was right here, assuming the user did cancel. I was suggesting IStatus.ERROR instead of the CoreException

akerekes: Yeah, cancel was appropriate until we tried to get the staging directory here from the provider. Since removing that and using a fix temp dir, if that does not exits, it's an error not a cancel event.
The `throws CoreException` is part of the API and it also contains a `Status` object which used by the caller method to report the error.

elharo:
LGTM

PR #334, opened 2016-07-08
Add Archetype selection page to new project wizard

### LINKED ISSUE #182: Let user choose archetype
When using the new maven project wizard, the user should be able to select an archetype.

As a temporary stopgap, we may want to pick a slightly moe complex archetype such as guestbook that contains actual Java code. the appengine-skeleton-archetype doesn't seem to have any source code.

chanseokoh:
Fixes #182.

My take here is to add a second wizard page where a user selects an archetype to use.

I didn't add two Endpoints archetypes since we don't support Endpoints yet. I checked that templates are properly generated for the current two archetypes: skeleton (hello-world) and guestbook.

archetype-wizard-page


elharo:
LGTM

PR #335, opened 2016-07-08
Uses common lib for Cloud SDK detection.

joaoandremartins:
@elharo @briandealwis PTAL


pom.xml
```diff
@@ -79,7 +79,7 @@
       <!-- bundle: com.google.cloud.tools.app.lib -->
       <groupId>com.google.cloud.tools</groupId>
       <artifactId>appengine-plugins-core</artifactId>
-      <version>0.1.1-SNAPSHOT</version>
+      <version>0.1.2-SNAPSHOT</version>
```
chanseokoh: #332 was just merged.

elharo:
Please resolve conflicts and repush.

joaoandremartins:
@elharo I think the conflicts are resolved. Let me know if you want me to walk you through what I've done so far. Might be faster that way!

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
```diff
@@ -146,7 +146,7 @@ public void widgetSelected(SelectionEvent event) {
       }
     });
 
-    File location = new CloudSdkProvider(null).getCloudSdkLocation();
+    File location = new CloudSdkProvider().getCloudSdk().getJavaAppEngineSdkPath().toFile();
```
elharo: Why do we need toFile here? If getJavaAppEngineSdkPath returns a Path we can just declare location to be that instead, I expect.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -36,8 +33,8 @@ public String getDescription() {
 
   @Override
   public IClasspathEntry[] getClasspathEntries() {
-    if (cloudSdkLocation != null) {
-      File jarFile = cloudSdkLocation.toPath().resolve(SDK_JAR).toFile();
+    if (CLOUD_SDK.getJavaAppEngineSdkPath() != null) {
+      File jarFile = CLOUD_SDK.getJavaToolsJar().toFile();
```
elharo: I think you can drop toFile and just use a Path here

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineFacet.java
```diff
@@ -49,7 +49,7 @@ public static void installAppEngineRuntime(IFacetedProject project, IProgressMon
   
       IRuntimeWorkingCopy appEngineRuntimeWorkingCopy 
           = appEngineRuntimeType.createRuntime(null, monitor);
-      File sdkLocation = new CloudSdkProvider(null).getCloudSdkLocation();
+      File sdkLocation = new CloudSdkProvider().getCloudSdk().getJavaAppEngineSdkPath().toFile();
```
elharo: drop toFile and use Path

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
```diff
@@ -59,13 +61,12 @@ public static CloudSdk getCloudSdk(Shell shell) {
    * @return the Cloud SDK, or {@code null} if unspecified
    */
   public static CloudSdk getCloudSdk(IShellProvider shellProvider) {
-    CloudSdkProvider cloudSdkProvider = new CloudSdkProvider(null);
-    CloudSdk sdk = cloudSdkProvider.getCloudSdk();
-    if (sdk != null) {
-      return sdk;
+    // Not sure I understand what's going on here...
```
elharo: delete comment

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -30,7 +30,7 @@ public ServletClasspathProvider() {
 
   @Override
   public IClasspathEntry[] resolveClasspathContainer(IRuntime runtime) {
-    java.nio.file.Path cloudSdkPath = new CloudSdkProvider(null).getCloudSdkLocation().toPath();
+    java.nio.file.Path cloudSdkPath = new CloudSdkProvider().getCloudSdk().getJavaAppEngineSdkPath();
```
briandealwis: Shouldn't this be the `CLOUD_SDK` instance?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
briandealwis: Shouldn't this be the `CLOUD_SDK` instance?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineFacet.java
briandealwis: Shouldn't this be `CLOUD_SDK`?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
```diff
@@ -61,11 +61,9 @@ public Object compute(IEclipseContext context, String contextKey) {
...
     Object path = context.get(PreferenceConstants.CLOUDSDK_PATH);
     File location = toFile(path);
-    CloudSdk.Builder builder = new CloudSdkProvider(null).createBuilder(location);
-    if (builder == null) {
-      return NOT_A_VALUE;
-    }
-    CloudSdk instance = builder.build();
+    // TODO(joaomartins): When can the SDK location in the context be different from the one in the
+    // preference store?
+    CloudSdk instance = new CloudSdkProvider().getCloudSdk();
```
briandealwis: It shouldn't be different — but we shouldn't assume that.

briandealwis:
@joaoandremartins my thought was that we should ideally make the Cloud SDK discovery more pluggable, so this preference-based code can be just another resolver that complements the existing PATH and fixed-location resolvers.  Perhaps configured with the JDK's IServiceLocator.

I don't think we should treat the resolution as being a cheap.  Certainly doing lots of filesystem checks in the PATH isn't that cheap for non-SSD, and many corporate developers are still on underpowered hardware.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -8,17 +8,14 @@
...
 
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.eclipse.sdk.CloudSdkProvider;
 
 public final class AppEngineSdkClasspathContainer implements IClasspathContainer {
 
-  // TODO should be changed once app-tools-lib can provide the directory
-  // https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/issues/134
-  private static final String SDK_JAR = "platform/google_appengine/google/appengine/tools/java/lib/appengine-tools-api.jar";
+  private static final CloudSdk CLOUD_SDK = new CloudSdkProvider().getCloudSdk();
```
briandealwis: This shouldn't be a static.  Actually, `AppEngineSdkClasspathContainer` should accept being provided with a `CloudSdk` instance.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -26,80 +26,33 @@
...
-    Path discovered = PathResolver.INSTANCE.getCloudSdkPath();
-    if (discovered != null) {
-      return discovered.toFile();
+    CloudSdk.Builder sdkBuilder = new CloudSdk.Builder();
+    
+    String configuredPath = preferences.getString(PreferenceConstants.CLOUDSDK_PATH);
+    
+    // Let's use the configured path, if there is one. Otherwise, the lib will try to discover the path.
+    if (configuredPath != null && !configuredPath.isEmpty()) {
+    	// TODO(joaomartins): What happens when the provided path is invalid? Tools lib isn't
```
briandealwis: Note that the previous code checked that the path exists and returned `null` if not found (and looks like there was a possibility of an NPE).

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
joaoandremartins: Because there is already a Path class imported, this was my shortcut to having less readable java.nio.file.Path. I'll change to Path, though.

@briandealwis because we're only using this once in this class, I didn't factor this to a static variable. Let me know if there is anything I'm missing.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: @briandealwis this is (or should be) all done in the library. I don't like the pattern where the library is surfacing a resolver and want to turn that off. Even in my original implementation of the commlib, I just used the PathResolver because we were using some legacy code from the GAE SDK to stage.

It should be as simple as Eclipse requesting a CloudSdk instance from the library and the library returning a valid one, or error. Let me know if you see any issues with this.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
joaoandremartins: It would help me to understand why we shouldn't assume it. If here I should be getting the SDK location from the context instead of the Preference Store, I'd need to create a setter in CloudSdkProvider or add something to its constructor.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
joaoandremartins: Removed static.

I suspect you want to pass in a CloudSdk instance so we always get the freshest Cloud SDK we have. In that case, could it make sense to just call CloudSdkProvider every time we want it? Or are you worried this will incur in unnecessary FS checks?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
briandealwis: Namely: singletons are evil.  Look at your patch where you have several uses of `new CloudSdkProvider().getCloudSdk()` — you're assuming that subsequent invocations produce an equivalent instance.  We should avoid baking in assumptions that CloudSdk instances are always from the same path.

In this case, the Context Functions form part of the Eclipse 4 Dependency Injection system.  The Eclipse Contexts are a bit like [Guice modules on steroids](https://www.eclipsecon.org/na2014/sites/default/files/slides/2014-EclipseCon-DI.pdf): they're very dynamic, and accesses to variables can be computed based on other elements found within the context (like what's happening here).  The contexts remember what variables are looked up from a context and reruns the computations whenever one of those variables change.

Put differently, we need some variable access to trigger recompilation on a change, and the sdkPath seems a logical element to trigger it.

elharo:
Please put this on hold for the moment. It's starting to look like we may need to back out some of the changes this depends on first.

elharo:
This can move forward again. We've resolved the issues that were blocking it. Though I'm afraid you have to merge again. :-(

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -17,89 +17,40 @@
...
-    if (discovered != null) {
-      return discovered.toFile();
+    CloudSdk.Builder sdkBuilder = new CloudSdk.Builder();
+    
+    String configuredPath = preferences.getString(PreferenceConstants.CLOUDSDK_PATH);
+    
+    // Let's use the configured path, if there is one. Otherwise, the lib will try to discover the
+    // path.
+    if (configuredPath != null && !configuredPath.isEmpty()) {
+    	// TODO(joaomartins): What happens when the provided path is invalid? Tools lib isn't
```
briandealwis: We allow it in case the SDK distribution has changed so the validate() code is now out-of-date.  We show a warning in the preference page.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -17,89 +17,40 @@
...
-  /**
-   * Return a {@link CloudSdk.Builder} instance, suitable for creating a {@link CloudSdk} instance.
-   * {@code location}, if not null, is used as the location to the Cloud SDK, otherwise the
-   * configured or discovered Cloud SDK will be used instead.
-   * 
-   * @param location if not {@code null}, overrides the default location for the SDK; can be a
-   *        {@linkplain String}, {@linkplain File}, or {@linkplain Path}.
-   * @return a builder, or {@code null} if the Google Cloud SDK cannot be located
-   */
-  public CloudSdk.Builder createBuilder(File location) {
```
briandealwis: The `createBuilder()` is required for consumers that want to set additional information before creating the instance.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -17,89 +17,40 @@
...
-     if (preferences == null) {
-       preferences = PreferenceInitializer.getPreferenceStore();
-     }
-     this.preferences = preferences;
-   }
-   
-   public CloudSdkProvider() {
-     this(PreferenceInitializer.getPreferenceStore());
-   }
+  private final IPreferenceStore preferences = PreferenceInitializer.getPreferenceStore();
```
briandealwis: The preference store was passed in for testing

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
```diff
@@ -33,6 +33,8 @@
  * found. Must be called from the SWT UI Thread.
  */
 public class CloudSdkPrompter {
+	
+  private static CloudSdk CLOUD_SDK = new CloudSdkProvider().getCloudSdk();
```
briandealwis: This shouldn't be static as the user may change the SDK location preference and this instance will become out of date.

briandealwis:
`CloudSdkProvider` was only needed because the Cloud SDK resolving mechanism is not sufficiently flexible.  We could happily get rid of `CloudSdkProvider` entirely if we could instead configure the CloudSdk.Builder with an alternative `PathResolver`-like instance that knew to first look at the relevant Eclipse preference.  We could do this via `java.util.ServiceLoader`, for example.

I don't think we can completely get rid of `CloudSdkPrompter`: I don't think it's safe to push UI-prompting into the path resolving code since we don't really know the UI situation.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: I was thinking that if we do need to add more information to the SDK, we can just overload the getCloudSdk method and add more arguments to it.

If we don't need the CloudSdkProvider class, the CloudSdk clients should even directly call the CloudSdk builder and set the desired arguments. (but this might go against making the least possible FS checks, so we might need the provider after all to use in DI)

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: Adding a new constructor visible for testing.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
```diff
@@ -61,11 +61,9 @@ public Object compute(IEclipseContext context, String contextKey) {
 
     Object path = context.get(PreferenceConstants.CLOUDSDK_PATH);
     File location = toFile(path);
```
chanseokoh: `location` is no longer used. Remove it?

briandealwis: I don't think location should be unused, I think there's value in being able to specify an alternative path this way — it certainly doesn't hurt.

The background here is that context functions are only re-evaluated (which we'll want to do if the preference changed) when some input used by the function is changed.  We trigger the re-evaluation here by retrieving the SDK Path from the referencing contexts, and then setting the SDK Path on those contexts when the preference changes (see further above to the `referencedContexts`, which is used by the `Activator`'s `IPreferenceChangeListener` and `CloudSdkContextFunction#sdkPathChanged()`).

That said, there's a logic error here in that a path value explicitly set in the context will be overwritten when the preference is next changed, and `compute()` should be:
```
    Object path = context.get(PreferenceConstants.CLOUDSDK_PATH);
    if(path == null) {
      // record this context as using the preference value
      referencedContexts.add(context);
    }
```

But if you strongly believe there should only be a single SDK Path, then we should use a different context key name than `PreferenceConstants.CLOUDSDK_PATH`, like `"sdkPathChanged"`.

joaoandremartins: CloudSdkProvider now takes an additional location that can come from the context. If there is no context location, it uses the one in the preference store. If none there either, it then delegates SDK discovery to the library. If the library can't find it, then null is returned and later on the user is prompted to set a valid one.

joaoandremartins:
Travis is failing because this depends on changes to the common lib (https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/164). As soon as those changes are pushed, it should go green.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
```diff
@@ -57,15 +57,14 @@ static void sdkPathChanged(String newPath) {
...
-    if (builder == null) {
-      return NOT_A_VALUE;
+    if (path == null) {
+      // record this context as using the preference value
+      referencedContexts.add(context);
     }
-    CloudSdk instance = builder.build();
+    
+    Path location = toFile(path).toPath();
+    CloudSdk instance = new CloudSdkProvider(location).getCloudSdk();
```
chanseokoh: Line 61 assumes that `path` can be null. If it is, then `toFile()` will return null, leading to an NPE. So, looks like either the assumption is wrong, or we need to check null.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -16,90 +16,58 @@
...
-   */
-  public CloudSdk getCloudSdk() {
-    return createBuilder(null).build();
+  private IPreferenceStore preferences = PreferenceInitializer.getPreferenceStore();
+  private Path location;
+  
+  public CloudSdkProvider() {}
+  
+  @VisibleForTesting
+  public CloudSdkProvider(IPreferenceStore preferences) {
```
chanseokoh: If this is only for testing, how about making it package-private?

BTW, to use `@VisibleForTesting`, I think guava needs to be added in `MANIFEST.MF`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -25,17 +25,18 @@ public ServletClasspathProvider() {
...
   public IClasspathEntry[] resolveClasspathContainer(IRuntime runtime) {
-    java.nio.file.Path cloudSdkPath = new CloudSdkProvider(null).getCloudSdkLocation().toPath();
+    java.nio.file.Path cloudSdkPath =
+        new CloudSdkProvider().getCloudSdk().getJavaAppEngineSdkPath();
     if (cloudSdkPath == null) {
       return new IClasspathEntry[0];
     };
-    String servletJar = cloudSdkPath + "/platform/google_appengine/google/appengine/tools/java/lib/shared/servlet-api.jar";
-    String jspJar = cloudSdkPath + "/platform/google_appengine/google/appengine/tools/java/lib/shared/jsp-api.jar";
+    String servletJar = cloudSdkPath + "/shared/servlet-api.jar";
```
elharo: why is /shared not included in the part returned by getJavaAppEngineSdkPath()?

also, we should use the resolve() method in Path instead of string concatenation here

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
```diff
@@ -25,7 +25,6 @@
 import org.junit.Test;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
-import org.w3c.dom.Node;
```
elharo: I have a separate pull request on this

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineFacet.java
```diff
@@ -49,9 +48,10 @@ public static void installAppEngineRuntime(IFacetedProject project, IProgressMon
   
       IRuntimeWorkingCopy appEngineRuntimeWorkingCopy 
           = appEngineRuntimeType.createRuntime(null, monitor);
-      File sdkLocation = new CloudSdkProvider(null).getCloudSdkLocation();
+      java.nio.file.Path sdkLocation =
+          new CloudSdkProvider().getCloudSdk().getJavaAppEngineSdkPath();
       if (sdkLocation != null) {
-        IPath sdkPath = Path.fromOSString(sdkLocation.getAbsolutePath());
+        IPath sdkPath = Path.fromOSString(sdkLocation.toString());
```
elharo: why eliminate the absolute path here?

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
```diff
@@ -95,13 +92,13 @@ public static File getCloudSdkLocation(Shell shell) {
    * @return the Cloud SDK location, or {@code null} if unspecified
    */
   public static File getCloudSdkLocation(IShellProvider shellProvider) {
-    CloudSdkProvider cloudSdkProvider = new CloudSdkProvider(null);
-    File location = cloudSdkProvider.getCloudSdkLocation();
+    CloudSdk cloudSdk = new CloudSdkProvider().getCloudSdk();
+    File location = cloudSdk.getJavaAppEngineSdkPath().toFile();
```
elharo: possible nullpointerexception here. toFile won't be null, but getJavaAppEngineSdkPath might be. suggest not calling toFile until line 98

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
```diff
@@ -57,15 +57,14 @@ static void sdkPathChanged(String newPath) {
...
-    CloudSdk.Builder builder = new CloudSdkProvider(null).createBuilder(location);
-    if (builder == null) {
-      return NOT_A_VALUE;
+    if (path == null) {
+      // record this context as using the preference value
+      referencedContexts.add(context);
     }
-    CloudSdk instance = builder.build();
+    
+    Path location = toFile(path).toPath();
```
elharo: Perhaps we should just have toPath instead of toFile

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
```diff
@@ -59,13 +59,10 @@ public static CloudSdk getCloudSdk(Shell shell) {
...
-    CloudSdk sdk = cloudSdkProvider.getCloudSdk();
-    if (sdk != null) {
-      return sdk;
-    }
-    if (promptForSdk(shellProvider)) {
-      return cloudSdkProvider.getCloudSdk();
+    CloudSdk cloudSdk = new CloudSdkProvider().getCloudSdk();
+    
+    if (cloudSdk != null || promptForSdk(shellProvider)) {
+      return cloudSdk;
```
briandealwis: This isn't right: if `cloudSdk == null` (no SDK found), we prompt the user and if they configure a location we return `cloudSdk` which is `null`.  We instead need to re-resolve the SDK.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
briandealwis: > I was thinking that if we do need to add more information to the SDK, we can just overload the getCloudSdk method and add more arguments to it.

Didn't you introduce the `CloudSdk.Builder` because of that problem?  But yes, this goes away if we can just get everything to instantiate `CloudSdk` the normal way.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
joaoandremartins: There are other jars in the location returned by getJavaAppEngineSdkPath().

Changed to use resolve(), though @briandealwis raised a good point: should the library provide methods to get the locations for all needed JARs or should we assume some client knowledge of the SDK structure?

I can't make my mind up on this. I prefer the first, but if there are many JARs the library must discover, the API might get a bit ugly?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineFacet.java
joaoandremartins: sdkLocation is now a Path, not a File.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
joaoandremartins: Will remove it from this PR.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
joaoandremartins: Fixed.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
joaoandremartins: Changed this up a bit, PTAL.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: Good point, made it package-private.
I had added the dependency to MANIFEST.MF, but forgot to add it here. :)

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: Yes, though since in this plugin we're using CloudSdkProvider, I think it comes out prettier if we overload. We're not likely going to need many more settings.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkContextFunction.java
joaoandremartins: Fixed.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -36,12 +32,12 @@ public String getDescription() {
 
   @Override
   public IClasspathEntry[] getClasspathEntries() {
-    if (cloudSdkLocation != null) {
-      File jarFile = cloudSdkLocation.toPath().resolve(SDK_JAR).toFile();
-      if (jarFile.exists()) {
-        String appEngineToolsApiJar = jarFile.getPath();
+    CloudSdk cloudSdk = new CloudSdkProvider().getCloudSdk();
+    if (cloudSdk.getJavaAppEngineSdkPath() != null) {
```
elharo: This is weird. You're checking to see if getJavaAppEngineSdkPath is null but then calling a different method.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
elharo: Here's an option: cloudSdk.getLibrary("servlet-api.jar")
cloudSdk.getLibrary("jsp-api.jar")

One method call and the client does not need to know the structure of the cloud SDK

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineFacet.java
elharo: yes, but you may still want sdkLocation.toAbsolutePath().toString()

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPrompter.java
elharo: nope, same problem (unless you just need to push the changes I'm not seeing here)

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
joaoandremartins: Yes, I fixed this locally but didn't push yet.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
joaoandremartins: Then the library would need a map of these names to their locations, unless we want to risk potentially expensive file system scans for those jars. But I like it, will try it in the lib.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: It turns out that this is tested from another package (src/com/google/cloud/tools/eclipse/sdk/internal), so it needs to be public.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java
```diff
@@ -146,18 +145,23 @@ public void widgetSelected(SelectionEvent event) {
       }
     });
 
-    File location = new CloudSdkProvider(null).getCloudSdkLocation();
-    if (location != null) {
-      dirTextBox.setText(location.toString());
+    CloudSdk sdk = new CloudSdkProvider().getCloudSdk();
+    if (sdk != null) {
+      java.nio.file.Path location = sdk.getJavaAppEngineSdkPath();
+      if (location != null) {
```
joaoandremartins: @elharo location is never null because the library never returns null. But should we have the null check regardless?

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/messages.properties
```diff
@@ -6,5 +6,5 @@ CloudSdkPreferencePage_4=Cannot launch a browser
 CloudSdkPreferencePage_5=&SDK location:
 CloudSdkPreferencePage_6=No SDK found: {0}
 CloudSdkPrompter_0=Google Cloud SDK Not Configured
-CloudSdkPrompter_1=The Google Cloud SDK cannot be found\!  Would you like to configure it now?
+CloudSdkPrompter_1=Invalid or no Google Cloud SDK specified. Do you want to configure Google Cloud SDK now?
```
briandealwis: I don't think this is an improvement: the first part sounds awkward.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -16,90 +16,70 @@
...
 
-import java.io.File;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 
 /**
  * Utility to find the Google Cloud SDK either at locations configured by the user or in standard
  * locations on the system.
  */
 public class CloudSdkProvider extends ContextFunction {
```
briandealwis: Oh there's a bogon here: CloudSdkProvider should NOT extend ContextFunction.  It was an initial design choice that I abandoned, but forgot to fully clean up.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
```diff
@@ -16,90 +16,70 @@
...
    * 
-   * @return the configured {@link CloudSdk} or {@code null} if no SDK could be found
-   */
-  public CloudSdk getCloudSdk() {
-    return createBuilder(null).build();
-  }
-
-  /**
-   * Return the location of the configured or discovered Cloud SDK.
+   * <p>The used location is the one from IEclipseContext. If there is none, then it is fetched from
```
briandealwis: This reference to the IEclipseContext is not applicable.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/messages.properties
joaoandremartins: I wanted a message that enforces the user must either specify a valid Cloud SDK or the operation will fail, which the first message didn't convey very well IMO.
I agree the first part doesn't sound great, feel free to provide an alternative.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProvider.java
joaoandremartins: Removed extension and import.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/messages.properties
akerekes: How about:

The Google Cloud SDK is required but cannot be found\!  Would you like to configure it now?

briandealwis:
LGTM

joaoandremartins:
Waiting for changes from https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/164 to be pushed to Maven central so tests can succeed.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProviderTest.java
```diff
@@ -1,24 +1,45 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
 import org.mockito.Mockito;
+import org.mockito.runners.MockitoJUnitRunner;
 
-// Must Run as JUnit Plug-in Test in Eclipse.
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
+import com.google.cloud.tools.eclipse.sdk.CloudSdkProvider;
+
+@RunWith(MockitoJUnitRunner.class)
```
joaoandremartins: @elharo @briandealwis PTAL. I made a few more changes so the unit/integration tests pass. After these changes, I think this is ready for submission.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -28,7 +27,8 @@
  *  <li>stage project for deploy</li>
  *  <li>deploy staged project</li>
  * </ol>
- * It uses a work directory where it will create separate directory for the exploded WAR and the staging results.
+ * It uses a work directory where it will create separate directory for the exploded WAR and the
```
elharo: separate directories

elharo:
LGTM

PR #336, opened 2016-07-11
Auto-set package name as groupId in project wizard

### LINKED ISSUE #229: The maven wizard should autofill the package name as groupId.artifactId
The maven wizard should autofill the package name as groupId.artifactId though this can be overridden

chanseokoh:
Fixes #229.

#229 mentions "groupId.artifactId", but the label in the wizard page says "Java package: (defaults to group ID)". For now, I just set it to groupId, since I believe artifactId usually contains hyphens.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
@@ -1,7 +1,5 @@
 package com.google.cloud.tools.eclipse.appengine.newproject.maven;
 
-import static com.google.cloud.tools.eclipse.appengine.newproject.maven.MavenAppEngineStandardWizardPage.suggestPackageName;
-
 import static org.junit.Assert.assertNotNull;
 
 import org.eclipse.swt.widgets.Display;
@@ -65,22 +63,31 @@ public class MavenArchetypeProjectWizardTest {
 
   @Test
   public void testSuggestPackageName() {
-    Assert.assertEquals("aa.bb", suggestPackageName("aa.bb", ""));
-    Assert.assertEquals("aa.bb.cc.dd", suggestPackageName("aa.bb", "cc.dd"));
+    Assert.assertEquals("aa.bb",
+        MavenAppEngineStandardWizardPage.suggestPackageName("aa.bb", ""));
+    Assert.assertEquals("aa.bb.cc.dd",
+        MavenAppEngineStandardWizardPage.suggestPackageName("aa.bb", "cc.dd"));
 
-    Assert.assertEquals("aa.bb", suggestPackageName("aA.Bb", ""));
-    Assert.assertEquals("aa.bb.cc.dd", suggestPackageName("Aa.Bb", "Cc.DD"));
+    Assert.assertEquals("aA.Bb",
+        MavenAppEngineStandardWizardPage.suggestPackageName("aA.Bb", ""));
+    Assert.assertEquals("Aa.Bb.Cc.DD",
+        MavenAppEngineStandardWizardPage.suggestPackageName("Aa.Bb", "Cc.DD"));
 
-    Assert.assertEquals("aa.bb", suggestPackageName(" a  a\t . b\r b \n", " \t \r  "));
-    Assert.assertEquals("aa.bb.cc.dd", suggestPackageName("  A  a.\tBb", " C  c . D D  "));
+    Assert.assertEquals("aa.bb",
+        MavenAppEngineStandardWizardPage.suggestPackageName(" a  a\t . b\r b \n", " \t \r  "));
+    Assert.assertEquals("Aa.Bb.Cc.DD",
+        MavenAppEngineStandardWizardPage.suggestPackageName("  A  a.\tBb", " C  c . D D  "));
 
-    Assert.assertEquals("aa.bb", suggestPackageName("....aa....bb...", "......"));
-    Assert.assertEquals("aa.bb.cc.dd", suggestPackageName("....aa....bb...", "..cc....dd.."));
+    Assert.assertEquals("aa.bb",
+        MavenAppEngineStandardWizardPage.suggestPackageName("....aa....bb...", "......"));
+    Assert.assertEquals("aa.bb.cc.dd",
+        MavenAppEngineStandardWizardPage.suggestPackageName("....aa....bb...", "..cc....dd.."));
 
-    Assert.assertEquals("aa._01234bb",
-        suggestPackageName("aa`~!@#$%^&*()-+=[]{}<>\\|:;'\",?/._01234bb", ""));
+    Assert.assertEquals("aa._01234bb", MavenAppEngineStandardWizardPage.suggestPackageName(
+        "aa`~!@#$%^&*()-+=[]{}<>\\|:;'\",?/._01234bb", ""));
     Assert.assertEquals("aa._01234bb._c_c_0.dd01234_",
-        suggestPackageName("aa`~!@#$%^&*()-+=[]{}<>\\|:;'\",?/._01234bb",
+        MavenAppEngineStandardWizardPage.suggestPackageName(
+            "aa`~!@#$%^&*()-+=[]{}<>\\|:;'\",?/._01234bb",
             "_c_c_0.dd01234_`~!@#$%^&*()-+=[]{}<>\\|:;'\",?/"));
   }
 }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
@@ -1,5 +1,7 @@
 package com.google.cloud.tools.eclipse.appengine.newproject.maven;
 
+import static com.google.cloud.tools.eclipse.appengine.newproject.maven.MavenAppEngineStandardWizardPage.suggestPackageName;
+
 import static org.junit.Assert.assertNotNull;
 
 import org.eclipse.swt.widgets.Display;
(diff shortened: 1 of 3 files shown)


elharo:
I do want to use the artifact ID here, though I might be wrong about that. I don't think artifact IDs usually contain hyphens. 

This is more complicated than what you have so far. We need to allow the user to manually set a different package and not override that choice with the groupId.artifactId.

chanseokoh:
Ok, so I think a check box button like 'Override (suggested) Java package name' can work out here.

briandealwis:
Check if the javaPackageField is empty or if it contains our generated value.

elharo:
No checkboxes please. Only autofill if the user hasn't typed something here.

chanseokoh:
Now I do it more intelligently:

Auto-fills javaPackageField as "groupId.artifactId" (or "groupId" if artifactId is empty), only when 1) javaPackageField is empty; or 2) the field matches previous auto-fill before ID modification.

So, if a user changes the package name manually, it will stop auto-setting. It's working nice from my point of view.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +328,45 @@ public void modifyText(ModifyEvent event) {
...
+   * Auto-fills javaPackageField as "groupId.artifactId" (or "groupId" if artifactId is empty),
+   * only when 1) javaPackageField is empty; or 2) the field matches previous auto-fill before
+   * ID modification.
+   */
+  private final class AutoPackageNameSetter implements VerifyListener {
+    @Override
+    public void verifyText(VerifyEvent event) {
+      Text textField = (Text) event.widget;
+      String oldText = textField.getText().trim();
+      // Below explains how to get text after modification:
```
elharo: useful comment

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +328,45 @@ public void modifyText(ModifyEvent event) {
...
+      // http://stackoverflow.com/questions/32872249/get-text-of-swt-text-component-before-modification
+      String newText =
+          oldText.substring(0, event.start) + event.text + oldText.substring(event.end);
+
+      String oldPackageName;
+      String newPackageName;
+      if (textField == groupIdField) {
+        oldPackageName = suggestPackageName(oldText, getArtifactId());
+        newPackageName = suggestPackageName(newText, getArtifactId());
+      } else {
```
elharo: Would it be clearer to have separate listeners for the groupId and the artifactId fields? It took me a while to parse this.

chanseokoh: I can do that.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +327,57 @@ public void modifyText(ModifyEvent event) {
...
+      String oldPackageName = suggestPackageName(getGroupId(), getArtifactId());
+      String newPackageName = suggestPackageName(getGroupId(), newArtifactId);
+      AutoSetPackageNameIfNecessary(oldPackageName, newPackageName);
+    }
+  }
+
+  /**
+   * See {@link AutoPackageNameSetterOnGroupIdChange#verifyText(VerifyEvent)}.
+   */
+  private void AutoSetPackageNameIfNecessary(String oldPackageName, String newPackageName) {
```
briandealwis: Method name looks more like C# than Java!  `adjustPackageNameIfNecessary()`?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +327,57 @@ public void modifyText(ModifyEvent event) {
...
+    if (getPackageName().isEmpty() || getPackageName().equals(oldPackageName)) {
+      javaPackageField.setText(newPackageName);
+    }
+  }
+
+  /**
+   * Helper function returning a suggested package name based on groupId and artifactId.
+   */
+  private static String suggestPackageName(String groupId, String artifactId) {
+    if (!artifactId.isEmpty()) {
```
briandealwis: Should filter out invalid characters (e.g., dashes) and lower-case everything.  Maven has very few rules for group and artifact IDs.

elharo: There's probably a utility method somewhere, open source or in google3, to do this

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
elharo: I don't think the "IfNecessary" part is necessary. :-)

chanseokoh:
I've added some basic package name sanitation (along with tests) such as filtering out non-word characters, making lowercase, etc. I wasn't able to find a library for this. This is not perfect of course, but users will be alerted of invalid pack names by the dialog validation.

I've done some tests, the whole functionality is still working fine.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
```diff
@@ -1,5 +1,7 @@
 package com.google.cloud.tools.eclipse.appengine.newproject.maven;
 
+import static com.google.cloud.tools.eclipse.appengine.newproject.maven.MavenAppEngineStandardWizardPage.suggestPackageName;
```
elharo: I would not import a rare method statically

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/META-INF/MANIFEST.MF
```diff
@@ -17,7 +17,8 @@ Require-Bundle: com.google.cloud.tools.eclipse.appengine.newproject;bundle-versi
  org.eclipse.core.jobs,
  org.eclipse.m2e.core;bundle-version="1.6.2",
  org.eclipse.m2e.core.ui;bundle-version="1.6.2",
- com.google.guava
+ com.google.guava,
+ org.apache.commons.lang
```
elharo: I'd rather not iontroduce this dependency

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +329,69 @@ public void modifyText(ModifyEvent event) {
...
+   * does basic string filtering/manipulation for valid Java package names, which is not perfect.
+   * However, users will be alerted of any slipping errors in naming by {@link #validatePage}.
+   */
+  @VisibleForTesting
+  protected static String suggestPackageName(String groupId, String artifactId) {
+    // 1) Remove leading and trailing dots.
+    // 2) Keep only word characters ([a-zA-Z_0-9]) and dots (escaping inside [] not necessary).
+    // 3) Replace consecutive dots with a single dot.
+    // 4) Lower-case.
+    groupId = StringUtils.strip(groupId, ".")
```
elharo: You probably don't need this class. Either Guava or String should suffice.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +329,69 @@ public void modifyText(ModifyEvent event) {
...
+    }
+  }
+
+  /**
+   * Helper function returning a suggested package name based on groupId and artifactId. It
+   * does basic string filtering/manipulation for valid Java package names, which is not perfect.
+   * However, users will be alerted of any slipping errors in naming by {@link #validatePage}.
+   */
+  @VisibleForTesting
+  protected static String suggestPackageName(String groupId, String artifactId) {
```
elharo: white listing would be safer than blacklisting, but I already wrote this in the new project wizard, let me look. Yep, I did. see com.google.cloud.tools.eclipse.appengine.newproject.JavaPackageValidator.

chanseokoh:
For the record, what @elharo and I discussed is

- Find a way to reuse `JavaPackageValidator` in the `newproject` plugin. However, it's not meant for sanitation/transform but rather only for validation, so we will need to change its implementation.
- If this kind of re-purposing doesn't look feasible, we may just do basic sanitation only when `JavaPckageValidator.validate()` returns false.

chanseokoh:
Ok, re-purposing `JavaPackageValidator.validate()` to suggest a valid name when there are naming problems does seem like really feasible, since it is delegating bulk of the validation work to another 3rd-party library. The validation logic inside the library is already very complicated, and a logic to suggest a valid name, if existed, would be highly complex.

chanseokoh:
Hence the plan:

To go safe with blacklisting, I'll sanitize the name with the current `suggestPackageName` logic only when `validate()` returns an error. (However, the sanitation by `suggestPackageName` is a subset of the validation constraint, so there is no risk with blacklisting.)

If after sanitation there still remains a name problem, a user has to manually set a valid package name.

briandealwis:
Since it's just to generate a package name when the user hasn't specified something, couldn't you just use something simple like '_string_.replace("\W+","")`?

chanseokoh:
@briandealwis I like such a simple idea. (However, "\W+" will remove dots as well, so I think the current implementation seems good.) `toLowerCase()` still seems a good idea too. @elharo, thoughts?

briandealwis:
+1 to lower casing. You could also consider stripping common prefix from the artifactId. For example, a lot of our POMs use `com.google.cloud.tools.` in both group & artifact IDs.

chanseokoh:
Actually, I found that `validate()` doesn't complain uppercase letters, so I left it out.

elharo:
Upper case letters are legal in package names, just uncommon.

elharo:
You don't need to use JavaPackageValidator but you should use 

JavaConventions.validatePackageName(
          packageName, JavaCore.VERSION_1_4, JavaCore.VERSION_1_4);

(i.e. the class from Eclipse JDT)

to check both pre and post package names.

chanseokoh:
Okay, so I added calling `validate()` before doing basic sanitation. If `validate()` fails, we do sanitation. If a problem persists after sanitation, we can't help it. But the dialog will immediately call `validate()` and complain about it, so users will have to fix it themselves. (This situation will be rare though.)

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -323,4 +330,79 @@ public void modifyText(ModifyEvent event) {
...
+    }
+  }
+
+  /**
+   * See {@link AutoPackageNameSetterOnGroupIdChange}.
+   */
+  private final class AutoPackageNameSetterOnArtifactIdChange implements VerifyListener {
+    @Override
+    public void verifyText(VerifyEvent event) {
+      String newArtifactId = getArtifactId().substring(0, event.start)
```
elharo: where do we use this variable?

elharo:
LGTM

PR #338, opened 2016-07-11
Fix NPE from Cloud Sdk directory field validation

### LINKED ISSUE #337: NPE from CloudSdkRuntimeWizardFragment due to uninitialized validateLocationJob
During dialog init, [this line](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkRuntimeWizardFragment.java#L151) can trigger `validate()` immediately, but `validateLocaltionJob` can be null (race condition).

I'll fix this right now.

chanseokoh:
Fixes #337.

`configureValidationJob` also depends on the `dirTextBox` field that's triggering the NPE, but `configureValidationJob` checks if the field is initialized currently, so no more problem.


briandealwis:
lgtm, but could you add a comment with the above (``createContents() may set the dirTextField, triggering validation``).

Issue #342, opened 2016-07-11
Enable login to Google


chanseokoh:
This issue is a top-level issue to track adding a login feature to our plugins.

Like in #283, this can be done in several steps.

- [x] Create a plugin skeleton
- [x] Add a no-op login menu item and a toolbar button
- [x] Implement `IElementUpdater` to update corresponding UI elements with the login status (e.g., if logged in, then label should be _Logout as XXX_, and change icon to indicate login status)
- [x] Check in code that actually tests if we are properly authorized (not unit tests, but e.g., an actual menu item on Eclipse)
- [x] Create a valid credential that passes the above test
- [x] Implement GPE-style login screen (i.e., instead of users entering verification code manually) (#398)
- [x] Inject client ID and client secret at compile tile (e.g., using `org.codehaus.mojo`)
- [x] Create OAuth client ID and client secret (#425)
- [x] Use the ide-login library (#480)
- [x] Use an external browser for login (#480)
- [x] Logout (#462)
- [x] Enable persisting and reusing credentials (or access/refresh tokens?) (#397)
- [x] Convert `GoogleLoginService` as an OSGi service (#446)
- [x] OAuth scopes documentation (#461)

Other related issues:

- Refactor: from `.eclipse.appengine.login` to `.eclipse.login` and remove all `appengine` naming (#396)
- Refactor: E4 Dependency Injection (#395)
- Refactor: UI-safe class and UI-inducing class (along the similar line of `CloudSdkProvider` vs `CloudSdkPrompter`) (#394)
- Clear login credentials on exit from Eclipse (#459)
- Establish lifetime for login credentials (#460)
- Enhanced UI for login (#488)

briandealwis:
Added a TODO for IElementUpdater to reflect login status

chanseokoh:
Added a TODO; we decided the next plan is to add manual test code (into production which will of course eventually removed)

briandealwis:
> Implement GPE-style login screen (i.e., instead of users entering verification code manually), and log in via local server

You don't need a local server — you just watch for redirects to your registered callback URL or the `urn:ietf:wg:oauth:2.0:oob`-style URNs.  That callback URL can be `https://localhost`, but it doesn't actually require setting up a web server.

chanseokoh:
For future reference:

So far, we've created the following issues against Google Java OAuth libraries to generate OGSi metadata. (Thanks, @elharo.)

[...]

PR #323, opened 2016-07-01
Add no-op context menu item for deploy

akerekes:
Related to #283 , implements item 2)

PR #343, opened 2016-07-11
Add no-op login menu

chanseokoh:
For #342.

As the next step to the skeleton work (#285), we now add an no-op menu.

This PR adds a top-level menu "App Engine" and attaches a sub-menu "Login to Google...". It also adds a tool bar button bound to this sub-menu. (So, this menu is global, unlike the deploy context menu in #323.)

Most of the work was done automatically by the Eclipse Extension Wizard. The icons are basically placeholders at the moment. (32x32 is not used; not sure how to choose automatically between 16x16 and 32x32  depending on resolution).


elharo:
We probably shouldn't have a hierarchical menu for only one menu item.

plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.properties
```diff
@@ -1,2 +1,6 @@
 pluginName=Google Cloud Platform Login Support
 providerName=Google, Inc.
+
+appEngineTopLevelMenuLabel=App Engine
+loginMenuItemLabel=Login to Google...
+loginToolbarTooltip=Click to login to Google
```
elharo: tooltips should provide additional info. How about "Authorize Eclipse to operate on Google Cloud Platform resources."

elharo:
can you add a screenshot of this?

chanseokoh:
I moved the menu item "Login to Google..." under the existing top-level "Project" menu. Unfortunately, I couldn't take a screenshot of this, as pressing a screenshot button make the menu disappear.

@nbashirbello, feel free to take this over from now on.

Tooltip screenshot:
login_toolbar_button

elharo:
There are screenshot programs that can take timed captures. e.g. snagit on Mac or Windows.

chanseokoh:
I remember the name sangit. I should install and try it. Thanks!

chanseokoh:
Here's the current menu item:

login_menu

elharo:
Aha! You put it in the Project menu. I was thinking context menu but I'm not sure this is actually wrong. Do you have a strong preference?

chanseokoh:
No strong preference. We could do both.

Just thought if we do this on a context menu, we should check what kind of item against which the context menu is shown and show/hide accordingly? (For example, it doesn't make sense to show this item when right-clicking in a console window.)

plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.xml
```diff
@@ -1,5 +1,48 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
+   <extension
+         point="org.eclipse.ui.commands">
+      <category
+            id="com.google.cloud.tools.eclipse.appengine.login.commands.category"
+            name="Google Login Category">
```
nbashirbello: Do we need "appengine" in the id? Login is for the GCP not only App Engine right?

nbashirbello:
Can we make the login menu right justified like IntelliJ or gmail?

plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.xml
chanseokoh: I've updated the TODO list in #342, since this also requires refactoring of the bundle name itself.

briandealwis:
Unless we're enabling login on a per-project basis, I don't think the _Project_ menu is the right place.  It's certainly not where I'd expect to find such a beast.  I'd put it either under _File_ or _Window_, but maybe it's best put into the Preferences dialog.

Accounts support is something the Eclipse Foundation wants to tackle.  They're establishing an initiative called the User Storage Service to store and share user data between workspaces, and it requires having a user login.  We haven't figured out where that login lives yet.

elharo:
On Mac OS X there's a third option: the Eclipse menu, though I'm not sure that's any better than File or Window. I think Brian's argument against Project is strong, though. Let's put it under File for now. We can move it later, though we'll want a final decision before we start taking screenshots for docs.

chanseokoh:
After hours of Internet search, I was able to move the toolbar to the top-right:
login_toolbar_button

I also moved the menu under "File".
login_menu

briandealwis:
> After hours of Internet search, I was able to move the toolbar to the top-right: 

Hmm, I'm not sure that's a great place.  If you were really wanting to make it look distinguished, I'd move it into the status area at the bottom, similar to GPE.

nbashirbello:
In Andriod Studio/IntelliJ, the decision to have the login icon/action button on the top left in the IDE was a very big push from the PMs so I figured we should have it in the same place for Eclipse. I am thinking we can get input from the PMs after we put a design doc together and change the login icon/action button later if the decision is different.

plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.xml
```diff
@@ -1,5 +1,50 @@
...
+   </extension>
+   <extension
+         point="org.eclipse.ui.menus">
+      <menuContribution
+            locationURI="menu:file?after=additions">
+         <command
+               commandId="com.google.cloud.tools.eclipse.appengine.login.commands.loginCommand"
+               id="com.google.cloud.tools.eclipse.appengine.login.menus.loginCommand"
+               label="%loginMenuItemLabel"
+               mnemonic="L">
```
elharo: are we sure about L? Does this conflict with anything else in that menu?

nbashirbello:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.xml
```diff
@@ -1,5 +1,50 @@
...
+         </command>
+      </menuContribution>
+      <!-- The following blog lists some useful eclipse built-in location URIs:
```
elharo: This comment is not needed here

elharo:
I'd guess the difference is that Android Studio is a Google product for
developing for Google products. In Eclipse, we're a much smaller, optional
piece.

On Wed, Jul 13, 2016 at 11:20 AM, nbashirbello <notifications@github.com>
wrote:

> In Andriod Studio/IntelliJ, the decision to have the login icon/action
> button on the top left in the IDE was a very big push from the PMs so I
> figured we should have it in the same place for Eclipse. I am thinking we
> can get input from the PMs after we put a design doc together and change
> the login icon/action button later if the decision is different.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/343#issuecomment-232389404>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AA9X6IaUZUuOS4r-nEeXnWbpXEvL1jmTks5qVQJUgaJpZM4JJ3ab>
> .
>



-- 
Elliotte Rusty Harold
elharo@macfaq.com

plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.xml
chanseokoh: It conflicts with 'Close All' (although a conflict itself won't be a problem).

Let me then go with G? It doesn't conflict.

chanseokoh:
Okay, since GPE had it at the bottom, l think it's safe to do so. We will see if PMs want it top-right. (The IntelliJ plugin has it at the top-right.)

login_toolbar_button_bottom

I also changed the menu shortcut key to 'G'.

chanseokoh:
I wonder if this button will look fine on high DPI, since we are always using the 16x16 icon.

elharo:
This probably won't work on High DPI. I don't think Mars works on High DPI at all though Neon should. This is one reason (though hardly the most important) to prefer text over icons. I wonder if we could just make this button say "Login to Google"? 

The difficulties we're having figuring out where and how to present this makes me wonder if we're going about this wrong. Perhaps instead of a separate action we should simply present login as a step in any flow that requires the user to be logged in? E.g. if you're not logged in and you try to deploy, you'll see a dialog asking you to login? That might be simpler and more discoverable.

briandealwis:
> I wonder if this button will look fine on high DPI, since we are always using the 16x16 icon.

Since you're providing a @2x icon, it will work out of the box on 4.6/Neon.

chanseokoh:
@elharo I agree a login step, e.g., in a deploy flow is necessary. There's actually more than that. For example, users may want to add multiple accounts and switch between them (like in Gmail, our IntelliJ plugin, and Andrio Studio). For this purpose, a toolbar button seems useful.

@briandealwis I just added a file only. Can the `@2x` version be picked up automatically? (Currently, `plugin.xml` has `icon="icons/googleFavicon.png"` only.)

briandealwis:
Yup, Eclipse will automatically look for the @2x.  So you just need to ensure the icon or its directory is included in the bundle (specified in `build.properties`) which… um, which it doesn't seem to be.  So that needs to be fixed.

chanseokoh:
`icons/` added to `build.properties`.

Showing a label in addition to the icon seems a good idea to me. How about this?

login_toolbar_button_label

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -0,0 +1,21 @@
...
+import org.eclipse.core.commands.AbstractHandler;
+import org.eclipse.core.commands.ExecutionEvent;
+import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.handlers.HandlerUtil;
+import org.eclipse.jface.dialogs.MessageDialog;
+
+public class GoogleLoginCommandHandler extends AbstractHandler {
+
+	public GoogleLoginCommandHandler() {
```
briandealwis: Don't need this constructor

briandealwis:
@chanseokoh the label doesn't add any information and will likely cause the status bar to overflow to two lines on default-sized windows.  I recommend using the icon and tooltips to show useful information: use a desaturated icon to show not-logged-in, and use the tooltip to provide details.

chanseokoh:
Sounds good.

For the icons, we may show users' profile pictures when logged in, which other tools do. Let's worry about the actual icons and tooltips later. (In any case, I believe a desaturated version should be created by ourselves, right?)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
chanseokoh: Removed the constructor.

chanseokoh:
@elharo and I had a discussion. We are in a good shape to move forward to the next step with this. Merging in.

PR #345, opened 2016-07-12
Do not include maven metafiles in Tycho-generated artifacts

### LINKED ISSUE #341: maven directory in jar
surprisingly our jar files in the update site include the maven files like pom.xml:

     0 Mon Jul 11 17:33:26 EDT 2016 META-INF/maven/
     0 Mon Jul 11 17:33:26 EDT 2016 META-INF/maven/com.google.cloud.tools.eclipse/
     0 Mon Jul 11 17:33:26 EDT 2016 META-INF/maven/com.google.cloud.tools.eclipse/com.google.cloud.tools.eclipse.appengine.localserver/
   703 Fri Apr 08 13:31:08 EDT 2016 META-INF/maven/com.google.cloud.tools.eclipse/com.google.cloud.tools.eclipse.appengine.localserver/pom.xml
   176 Mon Jul 11 17:33:26 EDT 2016 META-INF/maven/com.google.cloud.tools.eclipse/com.google.cloud.tools.eclipse.appengine.localserver/pom.properties

Is this actually needed or can/should we strip it out?

briandealwis:
Here's a patch to avoid including the Maven metafiles in the generated bundles.  These metafiles are normally included in most Maven-produced jars, including our own `app-tools-lib-for-java`, so I'm not sure this is a real win.

Fixes #341

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,6 +200,17 @@
         </plugin>
 
         <plugin>
+          <groupId>org.eclipse.tycho</groupId>
+          <artifactId>tycho-packaging-plugin</artifactId>
+          <version>${tycho.version}</version>
+          <configuration>
+            <archive>
+              <addMavenDescriptor>false</addMavenDescriptor>
+            </archive>
+          </configuration>
+        </plugin>
+
+        <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-enforcer-plugin</artifactId>
           <version>1.4.1</version>


elharo:
Why are these metafiles normally included in most Maven-produced jars? If there's a reason we can keep them. Otherwise, it's just code litter. (Hey, I just coined a new term! :-) )

briandealwis:
My guess is to be able to recover information about the artifact, like its artifact and group IDs and version, that would be lost if the jar is renamed.  I don't think it hurts us to leave the metafiles.

briandealwis:
I'll close this as a WONTFIX.  We can re-evaluate later if needed.

PR #376, opened 2016-07-15
make app-tools-lib-for-java the default runtime

elharo:
@nbashirbello


nbashirbello:
To clarify, this change implies that when you create an App Engine project (non-maven and maven) via the New Project Wizard, a new app-tools-lib runtime will be created if it didn't exist?

elharo:
yes

nbashirbello:
LGTM

PR #379, opened 2016-07-18
Fixes #350 - Create core bundle

### LINKED ISSUE #350: Create core bundle
Create a core bundle that can contain common/utility classes that are used throughout the plugin, e.g.: common helper classes to wrap Eclipse's SDK to allow unit testing

akerekes:
This is just the skeleton of the bundle and the test bundle, reloacting classes here will be part of subsequent PRs.


briandealwis:
LGTM.  I'm not sure including an empty `plugin.xml` is required: Eclipse complains about empty `plugin.xml`, and the Manifest editor will create it (and add it to `build.properties`) as needed.

elharo:
LGTM

PR #369, opened 2016-07-15
Add a manual login integration test

chanseokoh:
A step in #342 to add a manual integration test for login during development.

The test runs "gcloud projects lists --credential-override=<json credential>". If a proper credential file is provided, the command exits with code 0.

The bundle embeds a jar.

PR #386, opened 2016-07-18
Create valid credential

chanseokoh:
For #342.

Provides a very basic log in flow: login is done through a browser, and it eventually creates and returns a valid credential that passes the manual integration test we added in #369.

However, in reality, testing will always fail because of the undefined client id and secret. In my own local test, I plugged in the client id and secret from our IntelliJ plugin as a hack. @nbashirbello, you said you will create a new project for this? Let me have these values once they are ready.

@briandealwis One question: I have embedded `jackson-core-2.1.3.jar`, which is on the Maven Central and does have proper OGSi metadata (unlike other jars). This jar is not needed for compilation but needs to be included in the bundle binary. Is there a way to automatically download this jar from the Maven Central and not embed it under `lib/`?


chanseokoh:
I figured out how to force opening an _external_ browser, so no need to change browser preferences.

briandealwis:
> @briandealwis One question: I have embedded jackson-core-2.1.3.jar, which is on the Maven Central and does have proper OGSi metadata (unlike other jars).

The process:
  - add a Maven-style `<dependency>` in the top-level `pom.xml`
  - add a reference to the bundle's symbolic name in `eclipse/ide-target-platform/category.xml`
  - add a reference to the bundle in either `features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml` (if we depend on exactly this version) or `features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml` (if it's something that we need to have available)
This is how we bring in `com.google.cloud.tools.app.lib` (aka `app-tools-lib-for-java`/`appengine-plugins-core`), `com.google.gson`, and `com.google.guava`.

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
```diff
@@ -10,7 +10,12 @@ Bundle-ActivationPolicy: lazy
 Require-Bundle: com.google.cloud.tools.app.lib;bundle-version="0.1.0",
  com.google.gson,
  com.google.guava;bundle-version="15.0.0",
+ org.eclipse.core.runtime,
```
briandealwis: If you can, avoid including `org.eclipse.core.runtime`.  Most of its functionality has been split out to other bundles like `org.eclipse.equinox.common`, `org.eclipse.equinox.preferences`, `org.eclipse.equinox.registry`, and more.  The reason is that it brings in the kitchen sink and does a few things badly but can't be changed to preserve backwards compatibility.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
+    return createCredential(authResponse);
+  }
+
+  private GoogleTokenResponse logIn() throws IOException {
+    // 1. Open a browser that takes care of the login.
+    GoogleAuthorizationCodeRequestUrl requestUrl = new GoogleAuthorizationCodeRequestUrl(
+        OAUTH_CLIENT_ID, GoogleOAuthConstants.OOB_REDIRECT_URI, OAUTH_SCOPES);
+
+    try {
+      IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
```
briandealwis: I'll encourage you to look at GPE's `LoginBrowser`; I've used this same approach in the [Eclipse OAuth support](https://git.eclipse.org/r/#/c/77473/) — it's much nicer than requiring the user to copy over verification codes.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
+      browserSupport.getExternalBrowser().openURL(requestUrl.toURL());
+    } catch (PartInitException pie) {
+      // TODO(chanseok): display error message to user
+      return null;
+    }
+
+    // 2. Show a dialog to get a verification code returned from successful browser login.
+    InputDialog dialog = new InputDialog(Display.getCurrent().getActiveShell(),
```
briandealwis: `Display.getCurrent()` can return null if executed on non-SWT thread.

briandealwis:
I think we'll need to carefully consider the APIs here, using something along the lines of the `CloudSdkProvider` (safe for headless use) and `CloudSdkPrompter` (may result in UI prompting; methods are provided a `Shell` or `IShellProvider`).

We should also include support for the E4 Dependency Injection.  The easiest way to do this is to expose the Google Login providers as OSGi services; the E4 DI can inject OSGi services too.  For people who use multiple windows (like myself) and multiple accounts, we could also consider setting a per-window login; this could be useful if we had a _deploy_ or _admin_ perspective.

elharo:
We may eventually use  E4 Dependency Injection here, but we're not going to make that leap in one PR. This is a good step forward, not the final product.

plugins/com.google.cloud.tools.eclipse.appengine.login/.classpath
```diff
@@ -4,5 +4,7 @@
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src/"/>
 	<classpathentry kind="lib" path="lib/google-oauth-client-1.22.0.jar"/>
+	<classpathentry kind="lib" path="lib/google-api-client-1.22.0.jar"/>
+	<classpathentry kind="lib" path="lib/google-http-client-1.22.0.jar"/>
```
elharo: https://github.com/google/google-http-java-client/issues/328

plugins/com.google.cloud.tools.eclipse.appengine.login/.classpath
```diff
@@ -4,5 +4,7 @@
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src/"/>
 	<classpathentry kind="lib" path="lib/google-oauth-client-1.22.0.jar"/>
+	<classpathentry kind="lib" path="lib/google-api-client-1.22.0.jar"/>
```
elharo: https://github.com/google/google-api-java-client/issues/1018

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
+  private static final List<String> OAUTH_SCOPES = Collections.unmodifiableList(Arrays.asList(
+      "email",
+      "https://www.googleapis.com/auth/cloud-platform"
   ));
 
   /**
    * Returns the credential of the active user. If there is no active user, returns {@code null}.
    */
   // Should probably be synchronized properly.
   // TODO(chanseok): consider returning a String JSON (i.e., hide Credential)
```
elharo: delete todo, pretty sure we want to return credential

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Thanks. Copying verification codes should certainly go away. I'll also follow the `LoginBrowser` flow, which I think I need setup a local web server to listen for a verification code. This will require a fairly large amount of change, so I put another TODO item in #342.

chanseokoh:
I have added two more TODO items in #342:

- Refactor: E4 Dependency Injection
- Refactor: UI-safe class and UI-inducing class (along the similar line of `CloudSdkProvider` vs `CloudSdkPrompter`)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
-  public Credential getActiveCredential() {
-    return null;
+  public Credential getActiveCredential() throws IOException {
+    GoogleTokenResponse authResponse = logIn();
+    return createCredential(authResponse);
+  }
+
+  private GoogleTokenResponse logIn() throws IOException {
+    // 1. Open a browser that takes care of the login.
```
akerekes: unnecessary comment

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
+
+    try {
+      IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
+      browserSupport.getExternalBrowser().openURL(requestUrl.toURL());
+    } catch (PartInitException pie) {
+      // TODO(chanseok): display error message to user
+      return null;
+    }
+
+    // 2. Show a dialog to get a verification code returned from successful browser login.
```
akerekes: unnecessary comment

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
+    InputDialog dialog = new InputDialog(Display.getCurrent().getActiveShell(),
+        "Enter Verification Code", "Enter verification code from the browser login.", null, null);
+    if (dialog.open() != InputDialog.OK) {
+      return null;
+    }
+
+    String verificationCode = dialog.getValue();
+    // 3. Authorize the user with the verification code via Google Login API.
```
akerekes: unnecessary comment

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -34,17 +47,60 @@
...
+    GoogleAuthorizationCodeTokenRequest authRequest = new GoogleAuthorizationCodeTokenRequest(
```
akerekes: Breaking after `=`, keeping the first parameter on the same line with the method and aligning subsequent parameters with the first one is easier to read (for me at least)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
elharo: Issues more discoverable and trackable than TODOs. Please open issues for this or add the relevant details to the existing issue.

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
chanseokoh: That's good to know. Thanks. I now use `org.eclipse.equinox.common`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Now checking if `Display.getCurrent()` is null. We will get away from this input verfication dialog soon anyway.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Unnecessary comments removed.

chanseokoh:
Removing jackson-core-2.1.3.jar still pending, but all the other review comments have been addressed.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: A good approach is to take an `org.eclipse.jface.window.IShellProvider` rather than requiring an explicit Shell.  There's a wrapper IShellProvider class `org.eclipse.jface.window.SameShellProvider` that returns its provided shell.

chanseokoh:
Okay, I was able to bring in jackson-core-2.1.3.jar as @briandealwis [explained above](url), but the problem was that it in turn depended on google-http-client-jackson2-1.22.0.jar, which is not OGSi-ready. It results in NoClassDefFoundError.

```
java.lang.NoClassDefFoundError: Could not initialize class com.google.api.client.googleapis.util.Utils$JsonFactoryInstanceHolder
	at com.google.api.client.googleapis.util.Utils.getDefaultJsonFactory(Utils.java:36)
	at com.google.cloud.tools.eclipse.appengine.login.GoogleLoginService.logIn(GoogleLoginService.java:85)
```

So, I think removing jackson-core-2.1.3.jar is not possible unfortunately.

briandealwis:
Jackson is in Orbit (an Eclipse-hosted set of OSGi'fied jars used by Eclipse projects):
```
com.fasterxml.jackson.core.jackson-annotations 2.5.0.v201504151636
com.fasterxml.jackson.core.jackson-core 2.5.0.v201504151636
com.fasterxml.jackson.core.jackson-databind 2.5.0.v201504151636
com.fasterxml.jackson.dataformat.jackson-dataformat-yaml 2.5.0.v201504151636
com.fasterxml.jackson.datatype.jackson-datatype-guava 2.5.0.v201504151636
com.fasterxml.jackson.jaxrs.jackson-jaxrs-base 2.5.0.v201504171603
com.fasterxml.jackson.jaxrs.jackson-jaxrs-json-provider 2.5.0.v201504171603
com.fasterxml.jackson.module.jackson-module-jaxb-annotations 2.5.0.v201504151636
```
We already pull in dependencies from Orbit so it's should only be necessary to add them to `features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml`.

But that doesn't help with `google-http-client-jackson2-1.22.0.jar` and other bundles.

chanseokoh:
As suggested by @briandealwis, I made `getActiveCredential()` accept `IShellProvider`. That makes sense, because attempting to get a credential may involve signing in.

chanseokoh:
> But that doesn't help with google-http-client-jackson2-1.22.0.jar and other bundles.

Yeah, since jackson-core pulled from Maven Central or Orbit doesn't seem like it can load classes from the embedded `google-http-client-jackson2-1.22.0.jar` (I think), I'll embed the jackson-core jar as I did at the beginning.

BTW, I tried adding Jackson only in `features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml` as you said for testing, but `mvn -Pide-target-platform package` doesn't seem it can find the dependency.

```
[INFO] Adding repository http://download.eclipse.org/tools/orbit/downloads/drops/R20160221192158/repository
[INFO] Resolving dependencies of MavenProject: com.google.cloud.tools.eclipse:google-cloud-plugin-eclipse.repo:0.1.0-SNAPSHOT @ /home/chanseok/workspace/gcloud-eclipse-tools/gcp-repo/pom.xml
[INFO] {osgi.ws=gtk, osgi.os=linux, osgi.arch=x86, org.eclipse.update.install.features=true}
[ERROR]   Software being installed: com.google.cloud.tools.eclipse.dist 1.0.0.qualifier
[ERROR]   Missing requirement: com.google.cloud.tools.eclipse.3rdparty.feature.feature.group 0.1.0.qualifier requires 'com.fasterxml.jackson.core.jackson-core 0.0.0' but it could not be found
[ERROR]   Cannot satisfy dependency: com.google.cloud.tools.eclipse.dist 1.0.0.qualifier depends on: com.google.cloud.tools.eclipse.3rdparty.feature.feature.group 0.0.0
```

briandealwis:
> Yeah, since jackson-core pulled from Maven Central or Orbit doesn't seem like it can load classes from the embedded google-http-client-jackson2-1.22.0.jar (I think), I'll embed the jackson-core jar as I did at the beginning.

Welcome to OSGi :-) Managed class paths are simultaneously great and infuriating!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -56,34 +56,30 @@
...
-    // 1. Open a browser that takes care of the login.
+  private GoogleTokenResponse logIn(IShellProvider shellProvider) throws IOException {
     GoogleAuthorizationCodeRequestUrl requestUrl = new GoogleAuthorizationCodeRequestUrl(
         OAUTH_CLIENT_ID, GoogleOAuthConstants.OOB_REDIRECT_URI, OAUTH_SCOPES);
 
     try {
       IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
       browserSupport.getExternalBrowser().openURL(requestUrl.toURL());
-    } catch (PartInitException pie) {
+    } catch (PartInitException pie) {  // requires org.eclipse.equinox.common OSGi bundle.
```
akerekes: Is this comment necessary? Would not the compiler give an error is the given bundle is not added to the dependencies?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -56,34 +56,30 @@
...
       // TODO(chanseok): display error message to user
       return null;
```
akerekes: We could use `org.eclipse.jface.dialogs.ErrorDialog.openError(Shell, String, String, IStatus)` for quick error display.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: That's right. I expect that `PartInitException` is removed when I create a new login screen, so this is not to forget to remove this dependency later. Sounds fair?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Thanks. I used `MessageDialog.openError` since it seemed more convenient.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -57,19 +52,21 @@
...
+                                              OAUTH_SCOPES);
 
     try {
       IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
       browserSupport.getExternalBrowser().openURL(requestUrl.toURL());
-    } catch (PartInitException pie) {  // requires org.eclipse.equinox.common OSGi bundle.
-      // TODO(chanseok): display error message to user
+    } catch (PartInitException pie) {
+      MessageDialog.openError(shellProvider.getShell(),
+          "Error launching external browser", pie.getMessage());
```
akerekes: This is visible to the user so it should go into a message bundle.

chanseokoh: Done. I took this opportunity to set up the message bundle.

chanseokoh:
Any new comments?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -6,15 +6,18 @@
...
 import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.jface.window.SameShellProvider;
 
 public class GoogleLoginCommandHandler extends AbstractHandler {
 
-	public Object execute(ExecutionEvent event) throws ExecutionException {
-	  boolean success = new GoogleLoginTemporaryTester().testLogin();
-
+  public Object execute(ExecutionEvent event) throws ExecutionException {
     IWorkbenchWindow window = HandlerUtil.getActiveWorkbenchWindowChecked(event);
```
briandealwis: You could consider using `HandlerUtil.getActiveShell()` — in case this is triggered from a dialog.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -6,15 +6,18 @@
...
+
+    SameShellProvider shellProvider = new SameShellProvider(window.getShell());
```
briandealwis: I think it's better to push this kind of thing onto your test class (e.g., offer `testLogin(Shell)` and `#testLogin(IShellProvider)` with the former just calling the latter with `SameShellSupplier`)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +42,59 @@
  */
 public class GoogleLoginService {
 
-  // TODO(chanseok): this will be needed later.
-  public static final List<String> OAUTH_SCOPES = Collections.unmodifiableList(Arrays.asList(
-      "https://www.googleapis.com/auth/userinfo#email"
+  private static final List<String> OAUTH_SCOPES = Collections.unmodifiableList(Arrays.asList(
```
briandealwis: Perhaps this functionality should be in `com.google.cloud.tools.eclipse.`_appengine_`.login` instead of `com.google.cloud.tools.eclipse.login` — and if so, this list shouldn't be hardcoded here.  Or maybe renamed like `APPENGINE_SCOPES`?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +42,59 @@
...
+      "email", //$NON-NLS-1$
+      "https://www.googleapis.com/auth/cloud-platform" //$NON-NLS-1$
   ));
 
   /**
    * Returns the credential of the active user. If there is no active user, returns {@code null}.
```
briandealwis: Rather than _active user_, _assigned user associated with the provided `Shell`_?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +42,59 @@
...
    */
   // Should probably be synchronized properly.
-  // TODO(chanseok): consider returning a String JSON (i.e., hide Credential)
-  public Credential getActiveCredential() {
-    return null;
+  public Credential getActiveCredential(IShellProvider shellProvider) throws IOException {
```
briandealwis: Hmm, now this is where it gets interesting and we have to be careful.  I wouldn't suggest associating credentials with `Shell`s but with `IWorkbenchWindow`/`MWindow`.  This code would look up the `Shell` and its parent `Shell`s (in case the `Shell` is a dialog), and try to find the IWW/MW.  But rather than try to embed that map-Shell-to-MW/IWW logic here, I'd push it onto the caller and just have this method be `getAssignedCredential(IWorkbenchWindow)` and we can expand others later as required.

briandealwis: Oh, now that I think a bit more, I think this shell was just for hanging the login browser.  So belay the above ;)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Not sure our plan is to allow incremental login, i.e., starting with virtually no scope and then incrementally adding new scopes by presenting a consent screen whenever a user requires additional scope. That sounds complex, so I think at least for now we will allow every scope we are going to use in our tools, including scopes for, e.g., Stackdriver Debugger, Source Repository, App Engine, etc. So, it's not just for App Engine.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
chanseokoh: That's handy. However, this line is auto-generated by a wizard for a menu command, and it gets a window based on the `event` argument. so I feel like I don't need to touch it?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +42,59 @@
...
+    }
+
+    String verificationCode = dialog.getValue();
+    GoogleAuthorizationCodeTokenRequest authRequest = new GoogleAuthorizationCodeTokenRequest(
+        Utils.getDefaultTransport(), Utils.getDefaultJsonFactory(),
+        Constants.getOAuthClientId(), Constants.getOAuthClientSecret(),
+        verificationCode,
+        GoogleOAuthConstants.OOB_REDIRECT_URI);
+
+    return createCredential(authRequest.execute());
```
briandealwis: To enable @akerekes, you could stash the credential on the IWorkbench/MApplication. Something like:
```
MApplication app = PlatformUI.getWorkbench().getService(MApplication.class);
app.getTransientData().put("OAUTH_CRED", credential);
```
That value will be tossed on restart.

(Just in case: `MApplication` comes from an intermediate layer between the Workbench layer and the SWT widgets, called the _Modelled Workbench_.)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Just to confirm, you're saying getting `IShellProvider` is okay, right? I'll add javadoc for this argument.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
briandealwis: Those are just canned examples.  But we can always do this later if necessary.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: As you noticed, at least for now, user credentials are not associated with `Shell` / workbench / window. We can figure out later how we can deal with login per workbench in the multiple account scenario. (Although I think per workbench login and multiple accounts would be orthogonal matters not interfering each other.)

I'll update the comment.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
chanseokoh: Okay, that's good. I'll go with `HandlerUtil.getActiveShell()`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: It's fine by me.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Thanks. This is great as a stop-gap. It's working great!

chanseokoh:
Comments addressed. ptal.

chanseokoh:
Since we added a new login browser (#417), I integrated it in my last commit.

chanseokoh:
I also refactored the code that converts a credential to JSON into `GoogleLoginService` as a helper method. ptal.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +44,97 @@
...
+   * may be presented, e.g., if no user is logged in or login is required due to an expired
+   * credential.
+   *
+   * @param shellProvider provides a shell for the login screen if login is necessary
+   * @throws IOException can be thrown by the underlying Login API request
+   */
+  // TODO(chanseok); synchronize properly
+  public Credential getActiveCredential(IShellProvider shellProvider) throws IOException {
+    MApplication app = PlatformUI.getWorkbench().getService(MApplication.class);
+    Credential credential = (Credential) app.getTransientData().get(STASH_OAUTH_CRED_KEY);
```
briandealwis: I think this looks fine.  Going in the future, it might be better to turn `GoogleLoginService` into an OSGi-level or Workbench-level service — just as you're fetching the `MApplication` here.  The OSGi Services and Eclipse Contexts (`IEclipseContext`) provide a nifty way to access and store information simplifying the maintenance of per-window and per-application state.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLogicServiceTest.java
```diff
@@ -0,0 +1,52 @@
...
+    String jsonCredential = GoogleLoginService.getJsonCredentialHelper("fake_refresh_token");
+
+    CredentialType credentialType = new Gson().fromJson(jsonCredential, CredentialType.class);
+    Assert.assertEquals(Constants.getOAuthClientId(), credentialType.client_id);
+    Assert.assertEquals(Constants.getOAuthClientSecret(), credentialType.client_secret);
+    Assert.assertEquals("fake_refresh_token", credentialType.refresh_token);
+    Assert.assertEquals("authorized_user", credentialType.type);
+  }
+
+  public class CredentialType {
```
elharo: class probably doesn't need to be public

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLogicServiceTest.java
```diff
@@ -0,0 +1,52 @@
...
+    public String client_id;
```
elharo: ditto the fields

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +44,97 @@
...
   ));
 
   /**
-   * Returns the credential of the active user. If there is no active user, returns {@code null}.
+   * Returns the credential of an active user (among multiple logged-in users). A login screen
+   * may be presented, e.g., if no user is logged in or login is required due to an expired
+   * credential.
+   *
+   * @param shellProvider provides a shell for the login screen if login is necessary
+   * @throws IOException can be thrown by the underlying Login API request
```
elharo: please rewrite the comment to specify under what circumstances this is thrown; i.e. what might go wrong here

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +44,97 @@
...
+   */
+  // TODO(chanseok); synchronize properly
+  public Credential getActiveCredential(IShellProvider shellProvider) throws IOException {
+    MApplication app = PlatformUI.getWorkbench().getService(MApplication.class);
```
elharo: app --> application

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +44,97 @@
...
   /**
-   * Returns the credential of the active user. If there is no active user, returns {@code null}.
+   * Returns the credential of an active user (among multiple logged-in users). A login screen
+   * may be presented, e.g., if no user is logged in or login is required due to an expired
+   * credential.
+   *
+   * @param shellProvider provides a shell for the login screen if login is necessary
+   * @throws IOException can be thrown by the underlying Login API request
+   */
+  // TODO(chanseok); synchronize properly
```
elharo: This needs to be done in this PR, not as a TODO

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +44,97 @@
...
+  public Credential getActiveCredential(IShellProvider shellProvider) throws IOException {
+    MApplication app = PlatformUI.getWorkbench().getService(MApplication.class);
+    Credential credential = (Credential) app.getTransientData().get(STASH_OAUTH_CRED_KEY);
+
+    if (credential == null) {
+      credential = logIn(shellProvider);
+      app.getTransientData().put(STASH_OAUTH_CRED_KEY, credential);
+    }
+    return credential;
```
elharo: This can still be null, right?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -28,17 +44,97 @@
...
+  private static final List<String> OAUTH_SCOPES = Collections.unmodifiableList(Arrays.asList(
+      "email", //$NON-NLS-1$
+      "https://www.googleapis.com/auth/cloud-platform" //$NON-NLS-1$
   ));
 
   /**
-   * Returns the credential of the active user. If there is no active user, returns {@code null}.
+   * Returns the credential of an active user (among multiple logged-in users). A login screen
+   * may be presented, e.g., if no user is logged in or login is required due to an expired
+   * credential.
```
elharo: You need to document carefully that this method may return null, what that means (e.g. user cacnelled), and how the client code should handle this case

briandealwis:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLogicServiceTest.java
chanseokoh: Interesting. I see GSON can work fine with private fields.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Yeah, it can still return null. I forgot to document it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: This has always bothered me. It seems like virtually all Eclipse (and Intellij too) code doesn't care about synchronization. For example, looking at their code, I think none of `getWorkbench()`, `getService()`, `getTransientData()`,`get()`, and `put()` is thread-safe. I think it's mostly acceptable given that Eclipse is not a mission-critical system, the chances of contention is very low, and the consequences from unsafe concurrent operations are tolerable.

In this regard, I think I'll drop the comment that `GoogleLoginService` is thread-safe and follow the general practice that we don't synchronize most of the time. Although `put()` is not atomic and thus concurrent puts may corrupt the transient map in theory, I think in practice the possibility is zero. Even if the map has been corrupted, users can easily recover from malfunction, e.g., by restarting Eclipse.

elharo: Please let's not do anything that will make them restart eclipse. It's easy to use a thread-safe map.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: For now, I made it return null. If a checked exception makes more sense later, we can change it to so. I added comments about when it returns null and how the caller should react to it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: OSGi Services and Eclipse Contexts sound like a rescue. I'll probably bug you when I get there to use them. Thanks!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -53,19 +51,26 @@
   /**
    * Returns the credential of an active user (among multiple logged-in users). A login screen
    * may be presented, e.g., if no user is logged in or login is required due to an expired
-   * credential.
+   * credential. This method can still return {@code null} if a user cancels the login process.
```
elharo: can still return --> returns

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -53,19 +51,26 @@
...
+   * Under normal circumstances, {@code null} is returned only when a user cancelled login.
```
elharo: delete this sentence because it's redundant

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -53,19 +51,26 @@
...
+   * For this reason, if {@code null} is return, the caller is advised to cancel the current
```
elharo: If null is returned, the caller should cancel...

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -53,19 +51,26 @@
...
+   * @throws IOException can be thrown by the underlying Login API request (e.g., network
+   *     error from the transport layer while sending/receiving a HTTP request/response.)
    */
-  // TODO(chanseok); synchronize properly
   public Credential getActiveCredential(IShellProvider shellProvider) throws IOException {
-    MApplication app = PlatformUI.getWorkbench().getService(MApplication.class);
-    Credential credential = (Credential) app.getTransientData().get(STASH_OAUTH_CRED_KEY);
+    MApplication application = PlatformUI.getWorkbench().getService(MApplication.class);
+
+    // get() and put() are not transactional, but we do our best to improve thread-safety.
```
elharo: I don't understand what your concern is here now. get() and put() are now thread safe. I don't think we need to worry about a get followed by a put being locked together, but I often get thread stuff wrong

chanseokoh: I agree, and I don't worry about a get followed by a put not locked together here. It's tolerable. (I saw FindBugs generates a warning in this case not being an atomic transaction.) Just wanted to make a note, as someone else who reads the code later might otherwise wonder why the original code writer didn't consider it while making the map thread-safe.

elharo: I suspect this comment is more confusing than not having it at all.

chanseokoh: Okay, I'll remove it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: > This has always bothered me. It seems like virtually all Eclipse (and Intellij too) code doesn't care about synchronization.

There are other avenues for synchronization than just the `synchronize` keyword.
  * A lot of the code we use are implicitly synchronized by only being able to be called from the UI/SWT thread. That includes most of the Eclipse Workbench (IWorkbench and friends) and the E4 Modelled Workbench (`getTransientData()` etc).  They can sometimes be read from other threads, but they can only be modified on the UI thread.  
  * `getService()` is backed by `IEclipseContext`s which are thread-safe.
  * Other things are assumed to be protected by running in Jobs which has scheduling/locking support.

chanseokoh: Very interesting. These implicit synchronizations are almost impossible to discover if not knowing the underlying architectural design.

> (getTransientData() etc). They can sometimes be read from other threads, but they can only be modified on the UI thread.

So, we are currently on the UI thread in this class because we are using `getTransientData()` and modify the data, right? It then means that any thread calling `getActiveCredential()` will be a UI thread too?

> Other things are assumed to be protected by running in Jobs which has scheduling/locking support.

Does this mean that at any point, only one Job will be running in Eclipse? That's cool.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: Maybe we should bring in the Findbugs/JSR305 javax.annotations.{Nonnull,Nullable} or use the JDT equivalent annotations?  The JDT annotations allow exploiting some the type use features introduced in Java 8.

chanseokoh: I like the idea of using annotations. :+1:

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: One aspect that's missing here are the calling conventions — is this safe to call from outside of the UI thread?  Will this invoke UI actions?  We should document that, and we probably should have a non-prompting query-only variant that will obtain the current user but return null if none.

chanseokoh: I actually thought about adding the query-only variant recently. I'll do it now.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: > Does this mean that at any point, only one Job will be running in Eclipse? That's cool.

No, but when you create a `Job` you can add _scheduling rules_, which can be compound rules, and the Jobs framework will interleave jobs so that no two jobs with conflicts are run at the same time.  There's support for resolving deadlocks too.
  * [Eclipse Help: Scheduling Jobs](http://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fruntime_jobs_rules.htm)
  * [On the Job: The Eclipse 3.0 Jobs API](https://eclipse.org/articles/Article-Concurrency/jobs-api.html)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
elharo: I like these annotations in theory too. In practice, they are so incredibly useful that they've been reinvented at least 4 times, and constantly conflict with each other. :-)

I.e. we need to be careful about which one we use. Since we're an eclipse plugin we may want the eclipse versions. Brian, are these documented somewhere?

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: They are.  I'll open an issue.  The Eclipse null analysis supports multiple annotations.

PR #388, opened 2016-07-18
Start removing the custom runtime

elharo:
@nbashirbello @akerekes still a lot of code to delete, but these changes remove the cusotm runtime from the UI leaving only the app-tools-lib-fpr-java runtime


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/CloudSdkLaunchConfigurationDelegate.java
```diff
@@ -1,213 +0,0 @@
...
-  }
-
-  /*
-   * Workaround for https://bugs.eclipse.org/bugs/show_bug.cgi?id=38016 The
-   * custom IProcessFactory creates an IProcess object that calls the /quit
-   * handler on the dev app server
-   */
-  private void addProcessFactoryToLaunchConfiguration(ILaunchConfiguration configuration) throws CoreException {
-    ILaunchConfigurationWorkingCopy workingCopy = configuration.getWorkingCopy();
-    workingCopy.setAttribute(DebugPlugin.ATTR_PROCESS_FACTORY_ID, DevAppServerProcessFactory.ID);
```
akerekes: If we don't have the same issue with the app-tools-lib based dev server as here, then we can also get rid of the `DevAppServerProcessFactory` and `DevAppServerRuntimeProcess` classes in the subsequent PRs. Is there a list of classes marked for deletion?

akerekes:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/CloudSdkLaunchConfigurationDelegate.java
elharo: Not yet. I'm going to have go through and cut them out. I did try removing a few more but they had connections to code that's still used, so it may take some more careful surgery to extract them.

PR #353, opened 2016-07-12
Verify that our feature can be installed with no missing dependencies

briandealwis:
Add new test for Travis build that attempts to install the
com.google.cloud.tools.eclipse.suite.e45.feature and fails on any
missing dependencies.

Enabled for Travis builds only (i.e., `mvn -Ptravis package`).

Fixes #351

PR #391, opened 2016-07-18
Ignore .project and .settings/ under build/

### LINKED ISSUE #387: Add verify-feature-completeness settings
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	build/verify-feature-completeness/.project
	build/verify-feature-completeness/.settings

chanseokoh:
Fixes #387.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ bin/
 proxy_bin/
 target/
 generated-sources/
+build/verify-feature-completeness/.project
+build/verify-feature-completeness/.settings/


briandealwis:
Hmm, generally these files should be checked in rather than ignored.

chanseokoh:
@briandealwis So you forgot to check these in in #353? If you believe these should be checked in, let's do so.

elharo:
We talked about this before you joined the team, I think, and made an explicit decision to check in these files which is captured in our technical design doc: https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/wiki/Cloud-Tools-for-Eclipse-Technical-Design

PR #400, opened 2016-07-19
use CloudSdk path instead of App Engine SDK path when setting preferences

akerekes:


elharo:
We should think about a test for this but LGTM for now

PR #362, opened 2016-07-14
Remove CloudSdkProvider in favour of new CloudSdkResolver strategy

briandealwis:
Turn `CloudSdkProvider`, a wrapper class that looked for the Google Cloud SDK in a user-specified location, into [a CloudSdk resolver](https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/168).  This allows us to get rid of `CloudSdkProvider` and use the normal CloudSdk.Builder pattern.  `CloudSdkPrompter` must continue to exist, however.

PR #401, opened 2016-07-19
Restore CloudSdkProvider#createBuilder()

briandealwis:
Until https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/pull/168 and #362 and fixed, we must continue to use `CloudSdkProvider#createBuilder()` to create and configure `Cloudsdk.Builder` instances.


akerekes:
lgtm (but we should avoid import reordering and comment reformatting)

joaoandremartins:
This PR isn't needed if 1) you don't need to verify if the generated Cloud SDK is valid (which we weren't doing before in the places the builder was being directly called) and 2) we're moving to get rid of CloudSdkProvider.

briandealwis:
Agreed @joaoandremartins except that the replacement isn't yet in place :-/

PR #403, opened 2016-07-19
Continue cutting out the custom runtime

elharo:
@akerekes

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.properties
@@ -4,10 +4,8 @@ pluginName=Google App Engine Standard WTP Tools
 providerName=Google, Inc.
 
 runtimeTypeDescription=Google App Engine Standard Runtime
-runtimeTypeName=App Engine Standard
 runtimeTypeVendor=Google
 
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.properties
```diff
@@ -3,14 +3,10 @@ cloudsdkdeploy=App Engine Project Deploy
 pluginName=Google App Engine Standard WTP Tools
 providerName=Google, Inc.
 
-runtimeTypeDescription=Google App Engine Standard Runtime (Custom)
-runtimeTypeName=App Engine Standard (Custom)
+runtimeTypeDescription=Google App Engine Standard Runtime
 runtimeTypeVendor=Google
 
-apptoolsTypeDescription=Google App Engine Standard Runtime
 apptoolsTypeName=App Engine Standard
```
akerekes: I'm wondering if the `apptools` prefix is descriptive enough... Does not seem to provide any useful.

akerekes:
lgtm

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.properties
elharo: we can change them easily enough now that there's only one

PR #413, opened 2016-07-20
remove the word custom from runtime ID

elharo:
@akerekes


briandealwis:
LGTM.  I think we should come up with more semantically meaningful identifiers (e.g., `com.google.appengine.runtime` or something like it), that corresponds more closely to their assigned labels.

PR #415, opened 2016-07-20
deploy menu is now in deploy plugin

elharo:
@akerekes


briandealwis:
LGTM: this forgotten deploy command was never shown as it had a `<visibleWhen checkEnabled="true">`, and there are no handlers and so never enabled.

(You accidentally included commit bae0be2 from #413.)

PR #416, opened 2016-07-20
Stop computing the default Cloud SDK location preference

### LINKED ISSUE #404: Change CloudSdk default preference initialization and representation in preference page
> @akerekes: it is confusing how cloudsdkprovider.getcloudsdk() works
> @akerekes: It calls to the preference store, which if does not have a value for it, will call the default scope preference, whose initializer calls cloudsdkprovider.getcloudsdk() 
which in turns call to the preference store and ….
> @briandealwis: That's true.  My thinking was that the pref page should show the Cloud SDK that's found and that's used as the truth
> @briandealwis: But that's currently misleading as if the user doesn't hit ok, then the value isn't actually set and it will be recomputed.  Perhaps it should be greyed to show it's what's found and remove the default initializer

briandealwis:
But do show the computed location in the Cloud SDK preference page.

Fixes #404


plugins/com.google.cloud.tools.eclipse.sdk/plugin.xml
```diff
@@ -1,10 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
```
elharo: I thought Eclipse didn't like empty plugin.xml?

elharo:
LGTM

Issue #414, opened 2016-07-20
Move com.google.cloud.tools.eclipse.util from localserver to util package


elharo:
This is MavenUtils and NatureUtils.

akerekes:
Fixed by #418

PR #418, opened 2016-07-21
Fixes #371 - LocalAppEngineOutputLineListener should be renamed, moved to the util bundle and #414 - Move com.google.cloud.tools.eclipse.util from localserver to util package

### LINKED ISSUE #371: LocalAppEngineOutputLineListener should be renamed, moved to the util bundle
This class looks like a good base implementation to use in other bundles as well, so it should be moved to the util bundle.

akerekes:


plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
```diff
@@ -7,6 +7,10 @@ Bundle-Vendor: Google, Inc.
 Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-ActivationPolicy: lazy
-Export-Package: com.google.cloud.tools.eclipse.util.io,
+Export-Package: com.google.cloud.tools.eclipse.util,
+ com.google.cloud.tools.eclipse.util.io,
  com.google.cloud.tools.eclipse.util.status
 Require-Bundle: org.eclipse.core.runtime
+Import-Package: com.google.cloud.tools.appengine.cloudsdk.process,
```
elharo: why do we need to import this? I would expect the util package not to have any dependencies on our other non-util packages.

elharo:
If you're having trouble with this, try moving one class at a time, starting with the simplest.

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
akerekes: relocated to sdk.ui bundle

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/DefaultOutputLineListener.java
```diff
@@ -10,16 +10,16 @@
...
  *******************************************************************************/
-package com.google.cloud.tools.eclipse.appengine.localserver.server;
+package com.google.cloud.tools.eclipse.sdk.ui;
 
 import org.eclipse.ui.console.MessageConsoleStream;
 
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
 
-public class LocalAppEngineOutputLineListener implements ProcessOutputLineListener {
+public class DefaultOutputLineListener implements ProcessOutputLineListener {
```
briandealwis: This name doesn't convey what it does.  `OutputStreamLineWriter`?  `MessageConsoleLineWriter`?

briandealwis:
LGTM modulo the `DefaultOutputLineListener` class name

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
elharo: do you still need to import this?

PR #423, opened 2016-07-21
Fix new project setup to reference appengine-api.jar, and point to the online javadocs.

briandealwis:
`appengine-tools-api.jar` isn't very helpful for writing servlets!

This should be exposed by the runtime component, I think: the library is currently marked for being exported.  And there must be something further required to index the jar as the classes aren't being found by content assist.


akerekes:
lgtm

PR #426, opened 2016-07-21
Fixes #373 - Honor cancel request from user during deploy

### LINKED ISSUE #373: Honor cancel request from user during deploy
#283

akerekes:
I was experimenting with cancellation for long running calls like deploy, but I believe rather the app-tools-lib should provide some hooks for this as this might be useful for other libraries building on top of app-tools-lib as well.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardProjectStaging.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardProjectStaging.java
@@ -2,6 +2,7 @@ package com.google.cloud.tools.eclipse.appengine.deploy.standard;
 
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.SubMonitor;
 
 import com.google.cloud.tools.appengine.api.deploy.DefaultStageStandardConfiguration;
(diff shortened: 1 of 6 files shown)


chanseokoh:
This is good, but I guess users cannot immediately cancel out if, e.g., deployment already started?

akerekes:
@chanseokoh right, that's why we need some hooks in app-tools-lib to support that.

elharo:
Have you filed an issue on app-tools-lib-for-java to support this?

briandealwis:
You could add a `ProcessStartListener` via `CloudSdk.Builder#startListener()`, record the process and then call `Process#destroy()`?

akerekes:
After a discussion with @meltsufin it seems that cancelling the deploy is not desirable, as the result is undefined.

@briandealwis I thought about that, but at that time I did not realize the async flag on CloudSdk.Builder, so I was launching the thread and with all the other code it just seemed something that would be better handled in app-tools-lib (which in fact is the case if I use async).

I can add the cancellation to the staging operation, but we probably shouldn't allow the user to cancel the deploy operation once it has started.

chanseokoh:
I agree we don't actually need to cancel the deployment thread, but I think at least it is necessary to display the message that once deploy started, deployment will NOT be cancelled and the app will be deployed.

briandealwis:
> but we probably shouldn't allow the user to cancel the deploy operation once it has started.

Please allow me to humbly disagree: when (not _if_) I've accidentally mis-deployed, I want to stop the deploy RIGHT NOW — kill that process and abort.  Humans will make errors, and we can only try to help them recover as best we can.

> it just seemed something that would be better handled in app-tools-lib (which in fact is the case if I use async

I don't disagree that it should be handled by app-tools-lib.  In the meantime…

akerekes:
#439 is filed to implement cancelling app-tools-lib operations during deploy. The corresponding app-tools-lib issue is https://github.com/GoogleCloudPlatform/app-tools-lib-for-java/issues/180

This PR will implement only cancellation only before the given step in deploy is executed, and the fix for #439 should provide a more fine grained control for the user.

@briandealwis @chanseokoh @elharo in light of this, please take another look.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardProjectStaging.java
```diff
@@ -16,16 +17,20 @@
 
   /**
    * @param explodedWarDirectory the input of the staging operation
-   * @param stagingDir where the result of the staging operation will be written to
+   * @param stagingDirectory where the result of the staging operation will be written to
```
elharo: delete "to" at end of sentence

elharo:
LGTM

PR #434, opened 2016-07-22
Note feature is for 4.5 and 4.6

### LINKED ISSUE #427: Plugin name contains "...for Eclipse 4.5"
Currently, the plugin name appears as "Google Cloud Platform for Eclipse 4.5". This is misleading because the plugin is designed to work with Eclipse 4.6 as well.

briandealwis:
Fix #427


elharo:
LGTM.
Will this work as is with a hypothetical 4.7 release or are we going to  need to add a new target platform for each release?

briandealwis:
It will work with 4.7.  But we will need to add a new target for 4.7/Oxygen for testing.

Issue #422, opened 2016-07-21
Unable to find feature.xml in directory: /Users/elharo/gcloud-eclipse-tools/eclipse/ide-target-platform/target/repository/features/


elharo:
Not sure how significant this is, but just noticed that when running inside the IDE with the plugin we get a couple of dozen error log entries like this:

Unable to find feature.xml in directory: /Users/elharo/gcloud-eclipse-tools/eclipse/ide-target-platform/target/repository/features/org.eclipse.platform_4.5.2.v20160212-1500.jar

This is repeated for various other jars we're importing.

briandealwis:
It's not significant.  The _Directory_-based target platform that we've switched to with `ide-target-platform` is using a p2 repo as an installation.  Works fine for bundles, but features are normally exploded on disk.

We may be able to use one of the other types (_Installation_ perhaps).

PR #438, opened 2016-07-22
Create a .target definition for the IDE

briandealwis:
I realized we can create a `.target` that references a directory using an Eclipse variable.  No more need to manually create a target, and we can avoid pulling in the features and do away with the warning messages about features.

PTAL @chanseokoh

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,9 +133,18 @@ described below.
      previous steps to mark it as compatible with the `JavaSE-1.7` execution
      environment.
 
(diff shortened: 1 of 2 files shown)


elharo:
LGTM 
Does this fix #422?

chanseokoh:
Sweet. This does seem to fix #422 on my local testing.

briandealwis:
Note: the travis build failed as I'd merged and deleted the branch already.

PR #444, opened 2016-07-25
Update UI (login menu and button) based on login state / enable logout

chanseokoh:
- Updates the UI message of the login menu and the tooltip of the login toolbar button based on the current login state.
- Now users can log out, with a logout confirmation dialog.

One gripe is that the same messages are duplicated both in `plugin.properties` and `messages.properties` for some localized messages (for the initial menu name and tooltip). It would be nice if `plugin.xml` could pull values directly from `messages.properties`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
@@ -1,21 +1,63 @@
 package com.google.cloud.tools.eclipse.appengine.login;
 
+import java.io.IOException;
+import java.util.Map;
+
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
-import org.eclipse.ui.handlers.HandlerUtil;
 import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.jface.window.SameShellProvider;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.commands.ICommandService;
+import org.eclipse.ui.commands.IElementUpdater;
+import org.eclipse.ui.handlers.HandlerUtil;
+import org.eclipse.ui.internal.WorkbenchWindow;
+import org.eclipse.ui.menus.UIElement;
 
-public class GoogleLoginCommandHandler extends AbstractHandler {
+import com.google.api.client.auth.oauth2.Credential;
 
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -1,21 +1,63 @@
...
+        "com.google.cloud.tools.eclipse.appengine.login.commands.loginCommand", null); //$NON-NLS-1$
 
-    MessageDialog.openInformation(shell,
-        "TESTING AUTH", success ? "SUCCESS" : "FAILURE (to be implemented)");
     return null;
   }
+
+  @Override
+  public void updateElement(UIElement element, @SuppressWarnings("rawtypes") Map parameters) {
+    boolean loggedIn = new GoogleLoginService().getCachedActiveCredential() != null;
```
briandealwis: When you get to the point of implementing an OSGi service, you can use `UIElement#getServiceLocator()`

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -127,6 +127,14 @@ Credential createCredentialHelper(String accessToken, String refreshToken) {
     return credential;
   }
 
+  public void clearCredential() {
+    MApplication application = PlatformUI.getWorkbench().getService(MApplication.class);
+
+    Map<String, Object> synchronizedMap =
```
briandealwis: Hmm, using a synchronizeMap doesn't do anything here — other accessors won't be going through your map, and other callers to this method will get a new synchronizedMap.

If you're really concerned, put an `AtomicReference`

chanseokoh: > other callers to this method will get a new synchronizedMap.

Ah, I see. I may have misunderstood how sychronizedMap can work. Then I should re-work this thing in the other methods too. (Not in this PR.) I think for now I'll remove the synchronizedMap here, and do something about the re-work in another PR.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
chanseokoh: Note taken. Thanks.

briandealwis:
LGTM.  We should have some tests for the login state, but that can wait for the integration tests.

PR #445, opened 2016-07-25
Regenerate .classpath files for .login and .login.tests

briandealwis:
`com.google.cloud.tools.eclipse.appengine.login.tests` is pulling in two jars from `.login`, `google-api-client-1.22.0.jar` and `google-oauth-client-1.22.0.jar`, relating to OAuth.  `GoogleLoginServiceTest.java` is referencing `com.google.api.client.auth.oauth2.Credential`, whose definition references classes in `google-http-client-1.22.0.jar`:

```java
public class com.google.api.client.auth.oauth2.Credential implements 
    com.google.api.client.http.HttpExecuteInterceptor,
    com.google.api.client.http.HttpRequestInitializer,
    com.google.api.client.http.HttpUnsuccessfulResponseHandler {
```  

Because this jar is not in `.login.test`'s classpath, `GoogleLoginServiceTest` is generated with errors.

I used PDE's Regenerate Classpath and all is now well.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.login.test/.classpath
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.login.test/.classpath
@@ -3,7 +3,5 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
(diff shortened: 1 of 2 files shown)


briandealwis:
Tycho doesn't have a problem as it uses the OSGi metadata to resolve dependencies and `.login.test` is a fragment of `.login` and so shares its classpath, and `.login`'s classpath has all 5 libraries.

chanseokoh:
After all, I guess all the five jars need to be exported? That makes sense. LGTM

briandealwis:
PDE is responsible for taking the OSGi metadata and setting up JDT classpaths.

chanseokoh:
:rocket:

PR #447, opened 2016-07-26
Synchronize access to transient data map

chanseokoh:
It was pointed out in #444 that `Collections.synchronizedMap()` does not work in the way I imagined. (My bad.)

`application.getTransientData()` looks like it returns a map that is not thread-safe, so I am still inclined to do our part to synchronize access to the map, as we will allow concurrent access to `GoogleLoginService`.


briandealwis:
This change won't do much in practice as no other code attempts to obtain a lock on the map.  It only protects against itself.

It's worth stepping back and looking at the bigger picture.  The Eclipse Modelled Workbench (e.g., `MApplication` and its `transientData` map, and the other `M*` interfaces) are defined using EMF (Eclipse Modelling Framework).  They're not synchronized but writes are effectively serialized by having to be done on the UI thread. Admittedly that's a convention, but since the Modelled Workbench is a model of the UI, it makes sense.  And writes from other threads quickly result in raising SWT thread errors.

A roundabout way to say: make sure your `put()` to transientData is only done from the UI thread and you should be ok.

You could instead use the `MApplication`/`IWorkbench`'s `IEclipseContext` instead.  They are synchronized, and they are a pretty amazing data structure.  You can get the workbench's context with:
`PlatformUI.getWorkbench().getService(IEclipseContext.class)`.  There's a bit of doc on the [Eclipse 4 wiki](https://wiki.eclipse.org/Eclipse4/RCP/Contexts)

chanseokoh:
Thanks for the amazing explanation. You help me get a better picture.

> It only protects against itself.

Agreed. That's why I said "do our part."

> A roundabout way to say: make sure your put() to transientData is only done from the UI thread and you should be ok.

Then by putting a convention that `getActiveCredential()` and `clearCredential()` can/should only be called from a UI thread (both modify the map), we are okay, right? However, I think `IEclipseContext` seems like an ideal solution, so I will study it a bit and try using it.

chanseokoh:
Now uses `IEcipseContext` and things have been simplified. Did some manual login/logout testing; working fine.

briandealwis:
LGTM.

PR #451, opened 2016-07-26
correct URL

elharo:
@chanseokoh


chanseokoh:
The href value is actually being ignored. The URL being used is always GOOGLE_PRIVACY_POLICY_URL below. I think putting an actual URL might perplex others who later find changing the href value doesn't work. Still want to put a valid URL?

elharo:
The URL can be copied out of the dialog and pasted into a browser, in which case it's incorrect.

chanseokoh:
Make sense. LGTM

Issue #424, opened 2016-07-21
Preference panel UI prettification


elharo:
1. Make font size consistent.
2. Align labels with each other (Share and Sharing should have same left align.

![image](https://cloud.githubusercontent.com/assets/1005544/17035330/bbe2a6e8-4f54-11e6-82cf-0ac404961c15.png)

PR #452, opened 2016-07-26
unify font sizes

elharo:
@chanseokoh  @briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.preferences/src/com/google/cloud/tools/eclipse/preferences/AnalyticsOptInArea.java
+++ b/plugins/com.google.cloud.tools.eclipse.preferences/src/com/google/cloud/tools/eclipse/preferences/AnalyticsOptInArea.java
@@ -4,7 +4,6 @@ import com.google.cloud.tools.eclipse.preferences.areas.PreferenceArea;
 
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.preferences/src/com/google/cloud/tools/eclipse/preferences/AnalyticsOptInArea.java
```diff
@@ -4,6 +4,7 @@
 
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.jface.dialogs.Dialog;
```
chanseokoh: This is not used.

elharo: removed

chanseokoh:
Looks good. I just wonder why this isn't happening on my Ubuntu and why the fonts of the two controls are different. I'll let @briandealwis to give LGTM.

briandealwis:
Kinda fascinating.  `Link` widgets on OS X have a smaller default font of 12 points vs the default for other widgets of 13 points.  This is the same approach taken with the JFace `FieldEditor`s, so LGTM.

briandealwis:
#424

PR #454, opened 2016-07-26
Do manual login test only in dev mode

chanseokoh:
... so that end users aren't affected.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.login.test/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.login.test/META-INF/MANIFEST.MF
@@ -6,7 +6,6 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.login.test;singleton:=true
 Bundle-Vendor: Google, Inc.
 Bundle-Version: 0.1.0.qualifier
-Bundle-ActivationPolicy: lazy
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.login
 Require-Bundle: org.hamcrest;bundle-version="1.1.0",
  org.junit;bundle-version="4.12.0"
--- a/plugins/com.google.cloud.tools.eclipse.appengine.login.test/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.login.test/META-INF/MANIFEST.MF
@@ -3,9 +3,10 @@ Bundle-ClassPath: .
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for GCloud for Eclipse Login Support
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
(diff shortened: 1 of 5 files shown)


briandealwis:
A different way of accomplishing this is to define a new handler in your test bundle for `com.google.cloud.tools.eclipse.appengine.login.commands.loginCommand` that uses an `<activeWhen>` expression to become more-_applicable_ than your other handler.  When a command is triggered, the Eclipse workbench examines the enabled handlers and chooses a winner based on the complexity of their activeWhen expressions — the idea being that the more complex the expression, then the more tailored the handler is to particular situations.  Your `activeWhen` doesn't have to do anything real.

chanseokoh:
> define a new handler in your test bundle

Do you mean adding `plugin.xml` to the _test_ bundle and define a new handler there? So far, I couldn't make it use the handler defined in the test bundle.

briandealwis:
> Do you mean adding plugin.xml to the test bundle and define a new handler there? So far, I couldn't make it use the handler defined in the test bundle.

That's exactly what I meant.  You need to define an `<activeWhen>` expression.  Something like:
```
<activeWhen>
<with variable="selection" />
</activeWhen>
```
should be sufficient.

chanseokoh:
I tried adding `plugin.xml` first and then `fragment.xml` in the test bundle. None of the extensions defined there take effect. Only after I removed `Fragment-Host:` in the `MANIFEST.MF` (which I believe converts it to a regular plugin which in turn required me to add some `Required-Bundle:` entries) those extensions started to take effect. Is it the right approach to make the test fragment a standalone bundle to do this?

chanseokoh:
Aha, using `fragment.xml` and putting `singleton="true"` to `Bundle-SymbolicName` makes it work. I'll continue working on this.

chanseokoh:
Moved manual test code from `.login` to `.login.test`. Now `.login` contains only production code and is clear of any test code.

Some notable changes to `.login.test`:
- Added `fragment.xml` that defines a login command handler that overrides the handler defined in `.login`.
- The fragment is now a singleton. (Otherwise, the extension doesn't take effect.)

The only gripe is that when I launch a debug session inside IDE, I see the following messages:

`!MESSAGE Conflicting handlers for com.google.cloud.tools.eclipse.appengine.login.commands.loginCommand: {com.google.cloud.tools.eclipse.appengine.login.GoogleLoginCommandTestHandler@753133b0} vs {com.google.cloud.tools.eclipse.appengine.login.GoogleLoginCommandHandler}`

However, I think it is nice that we removed all the test code from `.login`.

briandealwis:
> The only gripe is that when I launch a debug session inside IDE, I see the following messages:

That means there's been a tie in the complexity of the two `activeWhen` expressions.  I thought the `<with>` should have boosted the score; you might need to add something like `instanceof java.lang.Object`.

elharo:
I don't think we need this. There are no end users at this point.

chanseokoh:
On second thought, we could just add another button for the test rather than overloading the login command.

Anyway, as @elharo said, let's first have a discussion about if this PR is needed.

PR #466, opened 2016-07-27
add server launch tab

### LINKED ISSUE #448: Run from run configuration
When you press the run button in the run configuration dialog, nothing happens:

image

We should probably make this run on local server.

elharo:
@briandealwis fix #448


briandealwis:
LGTM.  I think we should remove 'profile' though.

elharo:
on removing profile: maybe, but that should be a separate PR since it's unrelated to the purpose of this PR. Perhaps file an issue.

briandealwis:
Filed #471 for fixing/removing `profile`

PR #468, opened 2016-07-27
Enable server-side usage tracking

### LINKED ISSUE #137: Set user agent string for gcloud
for analytics tracking:

1. Figure out what user agent string we want (e.g. gcloud-eclipse)
2. Set the environment variable "CLOUDSDK_METRICS_ENVIRONMENT" to "gcloud-eclipse" or whatever) when invoking gcloud

see https://github.com/GoogleCloudPlatform/gcloud-intellij/blob/master/google-cloud-tools-plugin/src/com/google/cloud/tools/intellij/appengine/cloud/AppEngineAction.java#L89

chanseokoh:
Fixes #137.

I believe, for now, there is only one place in code that requires setting these constants when creating a `CloudSdk` instance for server interaction.

I created a new `CloudToolsInfo` class under `.util`. Let me know if you have better suggestions or better names.

FYI, `getToolsVersion()` currently returns `0.1.0.qualifier`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
@@ -15,6 +15,7 @@ import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.eclipse.appengine.deploy.AppEngineProjectDeployer;
 import com.google.cloud.tools.eclipse.appengine.deploy.Messages;
 import com.google.cloud.tools.eclipse.sdk.ui.MessageConsoleWriterOutputLineListener;
+import com.google.cloud.tools.eclipse.util.CloudToolsInfo;
 import com.google.cloud.tools.eclipse.util.MessageConsoleUtilities;
 import com.google.common.base.Preconditions;
 
@@ -88,6 +89,8 @@ public class StandardDeployJob extends WorkspaceJob {
     CloudSdk cloudSdk = new CloudSdk.Builder()
                           .addStdOutLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
                           .addStdErrLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
+                          .appCommandMetricsEnvironment(CloudToolsInfo.TOOLS_NAME_FOR_METRICS)
+                          .appCommandMetricsEnvironmentVersion(CloudToolsInfo.getToolsVersion())
                           .build();
     return cloudSdk;
   }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
@@ -89,7 +89,7 @@ public class StandardDeployJob extends WorkspaceJob {
     CloudSdk cloudSdk = new CloudSdk.Builder()
                           .addStdOutLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
                           .addStdErrLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
-                          .appCommandMetricsEnvironment(CloudToolsInfo.TOOLS_NAME_FOR_METRICS)
+                          .appCommandMetricsEnvironment(CloudToolsInfo.METRICS_NAME)
                           .appCommandMetricsEnvironmentVersion(CloudToolsInfo.getToolsVersion())
                           .build();
     return cloudSdk;
--- a/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
+++ b/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
@@ -1,0 +1,17 @@
+package com.google.cloud.tools.eclipse.util;
+
+import org.osgi.framework.FrameworkUtil;
+
+/**
+ * Provides generic information about our collective tools, such as a name to be used for usage
+ * reporting and the current version, etc.
+ */
+public class CloudToolsInfo {
+  
+  // Don't change the value; this name is used as an originating "application" of usage metrics.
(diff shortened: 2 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
```diff
@@ -0,0 +1,32 @@
...
+import org.osgi.framework.FrameworkUtil;
+
+/**
+ * Provides generic information about our collective tools, such as a name to be used for usage
+ * reporting and the current version, etc.
+ */
+public class CloudToolsInfo {
+
+  // Don't change the value; this name is used as an originating "application" of usage metrics.
+  public static String TOOLS_NAME_FOR_METRICS = "gcloud-eclipse-tools";
```
elharo: ANALYTICS_NAME or METRICS_NAME

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
```diff
@@ -0,0 +1,32 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *******************************************************************************/
+
+package com.google.cloud.tools.eclipse.util;
+
+import org.osgi.framework.FrameworkUtil;
+
+/**
+ * Provides generic information about our collective tools, such as a name to be used for usage
```
elharo: "our collective tools" --> the plug-in

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
```diff
@@ -0,0 +1,32 @@
...
+ * reporting and the current version, etc.
+ */
+public class CloudToolsInfo {
+
+  // Don't change the value; this name is used as an originating "application" of usage metrics.
+  public static String TOOLS_NAME_FOR_METRICS = "gcloud-eclipse-tools";
+
+  public static String getToolsVersion() {
```
elharo: can this be tested without excessive effort or does it pull in too much eclipse junk?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -88,6 +89,8 @@ private CloudSdk getCloudSdk() {
     CloudSdk cloudSdk = new CloudSdk.Builder()
                           .addStdOutLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
                           .addStdErrLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
+                          .appCommandMetricsEnvironment(CloudToolsInfo.TOOLS_NAME_FOR_METRICS)
```
elharo: this API is very peculiar

chanseokoh: I thought the same way. I guess it tries to say something like "set an environment variable for metrics and this is a command-line thing". There are other things like

`appCommandCredentialFile`
`appCommandOutputFormat`

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
chanseokoh: There was no problem calling this method directly from `.test`. However, it returns `0.1.0.201607280330` in the test despite that `MANIFEST.MF` of the `.test` bundle has `Bundle-Version: 0.1.0.qualifier`. (The method returns `0.1.0.qualifier` in the production envrionment as expected.) Not sure why. @briandealwis Any thoughts?

chanseokoh: I remembered the chat we don't need to do anything about versioning when releasing because our version is automatically augmented with time. I think then it's just fine to check if it begins with `"0.1.0."`.

briandealwis: `.qualifier` is replaced at build time with a timestamp.  You can use the OSGi `Version` to assemble a variant to your liking.

chanseokoh: I think I will use the current date-replaced versioning as-is, since we don't want to change this method every time we roll out a new release (unless we set up a "branding version" scheme for release or something like that.)

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -88,6 +89,8 @@ private CloudSdk getCloudSdk() {
     CloudSdk cloudSdk = new CloudSdk.Builder()
                           .addStdOutLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
                           .addStdErrLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
+                          .appCommandMetricsEnvironment(CloudToolsInfo.METRICS_NAME)
```
akerekes: For now we have only this location, but I can imagine a helper/factory that always sets these two properties so that we don't need to repeat, and we won't forget to do so in future instances.
We can work on that when it comes to implementing a feature requiring CloudSdk.Builder again.

akerekes:
LGTM

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
briandealwis: I think that makes sense.  Most of us will be in development mode and so it won't show up in any analytics.  And including the qualifier will identify fishfooders and dogfooders.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
chanseokoh: +1

PR #469, opened 2016-07-27
Fixes #431 - Use login credential for deployment

### LINKED ISSUE #431: Use login credential for deployment
How-to:

1. Get a credential by `new GoogleLoginService().getActiveCredential()`.
 1. This may require exporting classes in the `.login` bundle (by changing `MANIFEST.MF`) and adding a couple jars that live in `.login/lib` to the `.deploy` project setting.
 1. This can open a login browser. You should be in a UI-thread. (If this is not desired, use `getCachedActiveCredential()`.)
1. Check if the returned credential is null.
1. Write the credential to a file.
 1. Use the `GoogleLoginService.getJsonCredential()` helper method first to get a JSON in String.
 1. `GoogleLoginTemporaryTester` has an example of writing it to a file.
1. When calling gcloud, give the option `--credential-file-override=<credential file>`
 1. In app-tools-lib, this is done by calling `CloudSdk.Builder.appCommandCredentialFile()`.

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -69,6 +80,18 @@ public void done(IJobChangeEvent event) {
     deploy.schedule();
   }
 
+  private IPath createWorkDirectory() throws IOException {
+    String now = Long.toString(System.currentTimeMillis());
+    IPath workDirectory = getTempDir().append(now);
+    Files.createDirectories(workDirectory.toFile().toPath(), new FileAttribute<?>[0]);
```
elharo: I think you can drop the second argument, new FileAttribute<?>[0]

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -69,6 +80,18 @@ public void done(IJobChangeEvent event) {
...
+    return workDirectory;
+  }
+
+  private void ensureLoggedInAndSaveCredential(IPath workDirectory, IShellProvider shellProvider) throws IOException,
```
elharo: this name is weird. Maybe just logInAndSaveCredential? better yet, how about just logIn? I think the saving of credential is understood

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployer.java
```diff
@@ -52,13 +52,13 @@ public void deploy(IPath stagingDirectory, CloudSdk cloudSdk, IProgressMonitor m
...
 
       String projectId = deployInfo.getProjectId();
-      if (projectId == null) {
+      if (projectId == null || projectId.isEmpty()) {
         throw new CoreException(new Status(IStatus.ERROR, PLUGIN_ID, Messages.getString("project.id.missing"))); //$NON-NLS-1$
       }
 
       String projectVersion = deployInfo.getProjectVersion();
-      if (projectVersion == null) {
-        throw new CoreException(new Status(IStatus.ERROR, PLUGIN_ID, Messages.getString("project.version.missing"))); //$NON-NLS-1$
```
chanseokoh: Can we now remove `"project.version.missing"` from `messages.properties`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -82,12 +93,21 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
+    MessageConsole messageConsole = MessageConsoleUtilities.getMessageConsole(CONSOLE_NAME, null, true /* show */);
     final MessageConsoleStream outputStream = messageConsole.newMessageStream();
     CloudSdk cloudSdk = new CloudSdk.Builder()
                           .addStdOutLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
                           .addStdErrLineListener(new MessageConsoleWriterOutputLineListener(outputStream))
+                          .appCommandCredentialFile(LoginCredentialExporter.getCredentialFilePath(workDirectory).toFile())
+                          .exitListener(new ProcessExitListener() {
+                            // temporary way of error handling, after #439 is fixed, it'll be cleaner
+                            @Override
+                            public void onExit(int exitCode) {
```
elharo: does it really need to be indented this much? If so, maybe pull out to an inner class

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporterTest.java
```diff
@@ -0,0 +1,72 @@
...
+  @Mock private IShellProvider shellProvider;
+  @Mock private Credential credential;
+
+  @Test(expected = CoreException.class)
+  public void testLogInAndSaveCredential_exceptionOnFailedLogin() throws IOException, CoreException {
+    when(loginService.getActiveCredential(shellProvider)).thenReturn(null);
+    
+    LoginCredentialExporter exporter = new LoginCredentialExporter(loginService);
+    
+    Path workDirectory = Files.createTempDirectory(null, new FileAttribute<?>[0]);
```
elharo: probably don't need second argument

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporterTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Test
+  public void testLogInAndSaveCredential_successful() throws IOException, CoreException {
+    Credential credential = GoogleLoginService.createCredentialHelper(
+        FAKE_ACCESS_TOKEN, FAKE_REFRESH_TOKEN);
+    when(loginService.getActiveCredential(shellProvider)).thenReturn(credential);
+    
+    LoginCredentialExporter exporter = new LoginCredentialExporter(loginService);
+    
+    Path workDirectory = Files.createTempDirectory(null, new FileAttribute<?>[0]);
```
elharo: probably don't need second argument

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -114,7 +114,7 @@ private Credential createCredential(GoogleTokenResponse tokenResponse) {
    * are {@code final}, so Mockito can't mock them.)
    */
   @VisibleForTesting
```
elharo: still only visible for testing?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/status/StatusUtil.java
```diff
@@ -2,23 +2,36 @@
...
-    return new Status(IStatus.ERROR, FrameworkUtil.getBundle(origin).getSymbolicName(), message);
+    return error(origin, message, null);
   }
 
   public static IStatus error(Class<?> origin, String message, Throwable error) {
-    return new Status(IStatus.ERROR, FrameworkUtil.getBundle(origin).getSymbolicName(), message, error);
+    String bundleOrClassname = null;
+    
+    Bundle bundle = FrameworkUtil.getBundle(origin);
+    if (bundle ==  null) {
```
elharo: nit: extra space after ==

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/status/StatusUtil.java
```diff
@@ -32,4 +45,12 @@ public static IStatus error(Object origin, String message, Throwable error) {
     }
   }
 
+  private static IStatus errorInternal(String bundleOrClassname, String message, Throwable error) {
```
elharo: internalError? (adjectives usually come before nouns)

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -69,6 +80,18 @@ public void done(IJobChangeEvent event) {
...
 
+  private IPath createWorkDirectory() throws IOException {
+    String now = Long.toString(System.currentTimeMillis());
+    IPath workDirectory = getTempDir().append(now);
+    Files.createDirectories(workDirectory.toFile().toPath(), new FileAttribute<?>[0]);
+    return workDirectory;
+  }
+
+  private void ensureLoggedInAndSaveCredential(IPath workDirectory, IShellProvider shellProvider) throws IOException, 
+                                                                                                         CoreException {
```
chanseokoh: This line is over 100 columns.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
akerekes: I thought so too, but there's no method without this parameter and using null leads to NullPointerException.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
```diff
@@ -0,0 +1,48 @@
...
+                                                                                               CoreException {
+    Credential credential = loginService.getActiveCredential(shellProvider);
+    if (credential == null) {
+      throw new CoreException(StatusUtil.error(LoginCredentialExporter.class, "Login failed"));
+    }
+
+    String jsonCredential = GoogleLoginService.getJsonCredential(credential);
+    Files.write(workDirectory.append(CREDENTIAL_FILENAME).toFile().toPath(), 
+                jsonCredential.getBytes(Charsets.UTF_8), 
+                new OpenOption[0]);
```
chanseokoh: Perhaps you can omit the third argument. (Should be tested though.)

```java
Files.write(workDirectory.append(CREDENTIAL_FILENAME).toFile().toPath(), 
                 jsonCredential.getBytes(Charsets.UTF_8));
```

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
akerekes: The reason I wanted to emphasize saving the credentials is that it's saved to the work directory and will be used in the subsequent steps, and this is not part of the standard login process. I'll drop `ensure`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
akerekes: I'll extract it to an inner class (but I expect it to be short lived and replaced when CloudSdk operations will return `Future`s

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
elharo: Huh. I must be missing something about how varargs work then. I thought you could simply pass no arguments for varargs. The Javadoc says it's optional:

https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createDirectories(java.nio.file.Path,%20java.nio.file.attribute.FileAttribute...)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
akerekes: Yes, I need it in a test from a different package (`...login.ui`). But it would make sense to extract it to its own class.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployer.java
akerekes: +1

chanseokoh:
One thing we could decide is whether we bring up a login screen if a user is not logged when trying to deploy or simply complain with a message that "you need to sign in first to deploy." This PR goes with the first choice.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
chanseokoh: Do you see a value in adding `.toFile().deleteOnExit()`?

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
```diff
@@ -19,3 +19,6 @@ Bundle-Classpath: lib/google-api-client-1.22.0.jar,
  lib/google-oauth-client-1.22.0.jar,
  lib/jackson-core-2.1.3.jar,
  .
+Export-Package: com.google.api.client.auth.oauth2;x-friends:="com.google.cloud.tools.eclipse.appengine.deploy",
+ com.google.cloud.tools.eclipse.appengine.login.ui;x-friends:="com.google.cloud.tools.eclipse.appengine.deploy"
```
chanseokoh: Question: now I see 4 instances of `x-friends` (including these two) across all the bundles. Should we always use this directive?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
```diff
@@ -0,0 +1,48 @@
...
+  @VisibleForTesting
+  LoginCredentialExporter(GoogleLoginService googleLoginService) {
+    this.loginService = googleLoginService;
+  }
+
+  public void logInAndSaveCredential(IPath workDirectory, IShellProvider shellProvider) throws IOException, 
+                                                                                               CoreException {
+    Credential credential = loginService.getActiveCredential(shellProvider);
+    if (credential == null) {
+      throw new CoreException(StatusUtil.error(LoginCredentialExporter.class, "Login failed"));
```
chanseokoh: I'm just thinking if it makes sense for a helper class in the `.login.ui` bundle to throw `CoreException` and why it has to be `CoreException`. Would be nice if this helper class is decoupled with this Eclipse specific exception.

Perhaps, it's just enough to make this method accept a Credential and just write it to a file rather than inducing the login process? I don't know.

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
akerekes: @briandealwis what do you think?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/status/StatusUtil.java
akerekes: I could not find a better name for this method which is basically where all the public method calls end up. It is intended to be used only by this class. Anyway, I remove  `Internal` from the name.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
akerekes: You're right, I've missed this. Fixed here and at the other locations in the PR.

akerekes:
I'd rather not make the deploy action fail is the user is not logged in. However I agree that a message explaining it to the user would be better before showing the browser to the user -> #472

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
akerekes: My assumption is that the user will more often deploy than restart Eclipse, and considering the `CleanupOldDeploysJob` will cleanup all but the last two deploy directories on each deploy, I think `deleteOnExit()` would not add much value. On the other hand we could remove the credential file as soon as the deploy process is done if we're concerned about it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
```diff
@@ -0,0 +1,49 @@
...
+    this(new GoogleLoginService(), new CredentialHelper());
+  }
+  
+  @VisibleForTesting
+  LoginCredentialExporter(GoogleLoginService googleLoginService, CredentialHelper credentialHelper) {
+    this.loginService = googleLoginService;
+    this.credentialHelper = credentialHelper;
+  }
+
+  public void logInAndSaveCredential(IPath workDirectory, IShellProvider shellProvider) throws IOException,
```
chanseokoh: Let's add a comment

`Must be called from a UI context.`

because this is a public-facing utility class being exported to other bundles.

akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
briandealwis: It's a way to emphasize that something is definitely not intended unintentionally.

We should look at our bundles at some point, nearer to a 1.0, and decide whether we want to expose an API and move stuff around.

chanseokoh:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
```diff
@@ -0,0 +1,77 @@
...
+   * <p>
+   * <b>This method must be called from the UI thread.</b>
+   *
+   * @param workDirectory the destination for the json file containing the received OAuth credential
+   * @param shellProvider used if a web-based logn process is needed
+   * @throws IOException if communication with the login service fails or the json file with the credential
+   * cannot be created
+   * @throws CoreException if the login attempt was unsuccessful
+   */
+  public void logInAndSaveCredential(IPath workDirectory, IShellProvider shellProvider) throws IOException,
```
briandealwis: It seems strange for a `LoginCredentialExporter` to initiate UI and affect login state.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
briandealwis: I agree.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -41,22 +49,25 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
     try {
       IProject project = helper.getProject(event);
       if (project != null) {
-        launchDeployJob(project);
+        launchDeployJob(project, new SameShellProvider(HandlerUtil.getActiveShell(event)));
```
briandealwis: I think the deploy job should be written as a headless job and all parameters should be obtained here, in this handler, and configured on the job.  So that includes obtaining the login credential, the console stream.  The job itself should not change any state (e.g., the login state) as that's a UI thing.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginCredentialExporter.java
akerekes: This class grew out of a helper method. I agree the naming is not fully consistent with it's role. As @chanseokoh was also commenting on this, I'm thinking about moving the login step out of the class keeping only the write/read methods here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
akerekes: #476

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -41,22 +51,25 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
+      throw new ExecutionException(Messages.getString("deploy.failed.error.message"), exception); //$NON-NLS-1$
     }
   }
 
-  private void launchDeployJob(IProject project) {
-    String now = Long.toString(System.currentTimeMillis());
+  private void launchDeployJob(IProject project, IShellProvider shellProvider) throws IOException, CoreException {
+    IPath workDirectory = createWorkDirectory();
+    
+    logInAndSaveCredential(workDirectory, shellProvider);
```
briandealwis: So the only thing I'd change is to explicitly pass along the credentials filename to the deploy job rather than having it pick up the credential from some fixed-name file.  

Even better would move responsibility for the writing the credentials to disk to DeployJob — it has more control over the lifecycle of the credentials file (i.e., can delete it as soon as possible).  And gcloud could hypothetically take the credentials as an envvar.  (My earlier issue was with the possibility of deploy triggering UI for login.)

Having DeployJob write out to disk also avoids the possibility of multiple deploy jobs writing out conflicting credentials.

briandealwis:
Otherwise LGTM.

briandealwis:
LGTM

PR #475, opened 2016-07-28
Fix Fragment-Host of .util.test

chanseokoh:


plugins/com.google.cloud.tools.eclipse.util.test/META-INF/MANIFEST.MF
```diff
@@ -6,8 +6,9 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-SymbolicName: com.google.cloud.tools.eclipse.util.test
 Bundle-Vendor: Google, Inc.
 Bundle-Version: 0.1.0.qualifier
-Fragment-Host: com.google.cloud.tools.eclipse.appengine.deploy
-Require-Bundle: org.hamcrest;bundle-version="1.1.0",
+Fragment-Host: com.google.cloud.tools.eclipse.util
+Require-Bundle: com.google.guava;bundle-version="15.0.0",
```
elharo: where possible, we should use Import-Package in preference to Require-Bundle ( I got this exactly backwards in a lot of existing manifest files.)

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.util.test/META-INF/MANIFEST.MF
chanseokoh: Got it. (guava doesn't seem like working with Import-Package at the moment.)

PR #483, opened 2016-07-29
Export additional API jars

### LINKED ISSUE #385: Add same GAE SDK to the classpath than GPE was adding before
Right now, we're only adding three jars (appengine-tools-api.jar, servlet-api.jar and jsp-api.jar) to the classpath, so that newly created projects compile without errors (I suspect).

We should add the same jars as GPE did so non-Maven or non-Gradle projects don't have to manually import missing libraries (e.g., Datastore, etc.).

We should also start providing the gcloud-java SDK for Flex apps, either from the Eclipse new software page or through gcloud.

briandealwis:
Discovered the `AppEngineSdkClasspathContainer#getKind()` was returning an `IClasspathEntry` kind rather than an `IClasspathContainer` kind.  Type proposals work!

Added the other libraries shown in #385 except I can't find `libservice.jar`.

We may want to consider exposing a the library path via a variable so we don't have the full path to the SDK?

Fixes #385

PTAL @nbashirbello


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -38,21 +48,27 @@ public String getDescription() {
...
+                null /* sourceAttachmentPath */,
+                null /* sourceAttachmentRootPath */,
+                new IAccessRule[] { allAccessible  } /* accessRules */,
+                new IClasspathAttribute[] { javadocAttribute },
+                false /* isExported */);
+            //@formatter:on
+            entries.add(jarEntry);
+          }
         }
+        return entries.toArray(new IClasspathEntry[entries.size()]);
```
chanseokoh: Looking at the javadoc, there is no practical difference from `entries.toArray()`, isn't it?

briandealwis: entries.toArray() is an `Object[]` — you at least need to provide the class for Java to figure it out.  Java's generic types are compile-time only (mostly) and are _erased_ from what's put out in the `.class` files.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -38,21 +48,27 @@ public String getDescription() {
...
-              null /* sourceAttachmentPath */,
-              null /* sourceAttachmentRootPath */,
-              null /* accessRules */,
-              new IClasspathAttribute[] { javadocAttribute },
-              true /* isExported */);
-          //@formatter:on
-          return new IClasspathEntry[] {appEngineApisEntry};
+        List<IClasspathEntry> entries = new ArrayList<>(INCLUDED_JARS.length);
+        for (String jarLocation : INCLUDED_JARS) {
+          java.nio.file.Path jarFile = cloudSdk.getJavaAppEngineSdkPath().resolve(jarLocation);
```
joaoandremartins: I'd prefer we didn't assume knowledge of the SDK, but rather use `getJarPath()` with the jar name, on a first go.
Ideally, commlib will have a method that gives us all the jars we need to get an app running.

briandealwis: We'll need both the set of jars, as well as corresponding javadocs. And potentially source jars?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -38,21 +48,27 @@ public String getDescription() {
...
+            IClasspathAttribute javadocAttribute = JavaCore.newClasspathAttribute(
+                IClasspathAttribute.JAVADOC_LOCATION_ATTRIBUTE_NAME, APPENGINE_API_JAVADOC_URL);
+            IAccessRule allAccessible =
+                JavaCore.newAccessRule(new Path("**"), IAccessRule.K_ACCESSIBLE);
+            //@formatter:off
+            IClasspathEntry jarEntry = JavaCore.newLibraryEntry(
+                new Path(jarFile.toString()),
+                null /* sourceAttachmentPath */,
+                null /* sourceAttachmentRootPath */,
+                new IAccessRule[] { allAccessible  } /* accessRules */,
```
akerekes: you can remove the comment, it's not a `null` value passed in anymore

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
joaoandremartins: Yes, that sounds good. I'm thinking we should provide three methods for jars, javadocs and sources respectively. Let me know if you have a better idea.

briandealwis: https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/186

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
briandealwis: I'll revert this: I originally thought this might be the culprit before discovering the problem with `getKind()`

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
```diff
@@ -38,21 +47,27 @@ public String getDescription() {
...
-          IClasspathEntry appEngineApisEntry = JavaCore.newLibraryEntry(
-              new Path(jarFile.toString()),
-              null /* sourceAttachmentPath */,
-              null /* sourceAttachmentRootPath */,
-              null /* accessRules */,
-              new IClasspathAttribute[] { javadocAttribute },
-              true /* isExported */);
-          //@formatter:on
-          return new IClasspathEntry[] {appEngineApisEntry};
+        // todo: INCLUDED_JARS should be pulled from appengine-plugins-core
```
akerekes: Is there a corresponding issue for this repository?

akerekes:
LGTM

nbashirbello:
lgtm

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineSdkClasspathContainer.java
briandealwis: #495

PR #484, opened 2016-08-01
Remove eclipse org.eclipse.core.runtime and org.eclipse.ui from required bundles, replace them with more specific bundles

akerekes:


briandealwis:
lgtm

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
```diff
@@ -11,5 +11,7 @@ Export-Package: com.google.cloud.tools.eclipse.util,
  com.google.cloud.tools.eclipse.util.io,
  com.google.cloud.tools.eclipse.util.status
 Require-Bundle: org.eclipse.core.runtime
```
briandealwis: Oops, `org.eclipse.core.runtime` is still here.

PR #485, opened 2016-08-01
Fixes #482 - Cannot deploy

### LINKED ISSUE #482: Cannot deploy
I'm on the master branch. Not sure if this is happens on my machine only.

Steps to reproduce:

1. Create a new project (either Maven-based or plain)
1. Do deploy
   * Nothing happens

I did some investigation: `null` is returned [here](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java#L51). Inside `getProject()`, it seems like `structuredSelection.getFirstElement()` doesn't return an instance of `IProject`.

akerekes:
When user runs the deploy command the type of the selection depends on the current ~~perspective~~ explorer (e.g. Project Explorer (default in Java EE perspective) vs Package Explorer (default in Java perspective). If it is ~~Java~~ Package Explorer, then the selection will be an IJavaProject otherwise an IProject.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/util/ProjectFromSelectionHelper.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/util/ProjectFromSelectionHelper.java
@@ -4,6 +4,7 @@ import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.ui.handlers.HandlerUtil;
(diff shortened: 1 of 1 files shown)


chanseokoh:
Tested and verified it works fine. You said this takes care of the Java perspective. Are we then covered in all cases, e.g., can there be another perspective that this won't work?

Otherwise, LGTM.

akerekes:
I've updated the description: it's not about the perspective, it's about the default explorer associated with the perspective. We had similar issues earlier, so we should look out for this type of different behavior between Package and Project Explorer.

chanseokoh:
:rocket:

briandealwis:
Late to the party, but the better way is to use the _Adapter_ framework.  Something like:
```
IProject project = adapt(structuredSelection.getFirstElement(), IProject.class);

private <T> T adapt(Object o, Class<T> cls) {
   if(cls.isInstance(o)) { return cls.cast(o); }
   if(o instanceof IAdapter) {
     Object instance = ((IAdapter<T>)o).getAdapter(cls);
     if(instance != null) { return cls.cast(instance); }
   }
   Object adapted = AdapterManager.getDefault().loadAdapter(o, cls.getName());
   if(adapted != null) { return cls.cast(adapted); }
   return null;
}
```
In Neon / 4.6 we introduced a new helper class, `Adapters`, to do this but we need to do it long-form in Mars.

The above corresponds to the `<adapts>` expression in the Core Expressions language used in `<activeWhen>` and `<enabledWhen>` expressions.

akerekes:
Thanks @briandealwis , I've just submitted a PR to implement this in a more Eclipse-y way with adapters.

Issue #446, opened 2016-07-25
Convert GoogleLoginService as an OSGi service


chanseokoh:
Could be either OSGi-level or workbench-level. Perhaps starting at the OSGi level.

Note: UIElement#getServiceLocator() can be used to update UI.

elharo:
Is this related to #395 ?

chanseokoh:
From what I've learned (could be wrong), this is related to #395 but seems like a different matter.

chanseokoh:
Closed by #486.

Issue #472, opened 2016-07-27
Let user know that he/she needs to log in for deploy


akerekes:
When the user is not logged in and wants to deploy to App Engine, the plugin simply pops up a browser with the authentication page without any explanation. We should tell the user the reason for the login page.

PR #486, opened 2016-08-02
Use ide login library

### LINKED ISSUE #480: Use ide-login library
https://github.com/GoogleCloudPlatform/ide-login

The library basically has one core class that is very lightweight, so I anticipate this will be somewhat like a refactoring task applying a new design pattern.

chanseokoh:
Fixes #480.

- Uses an external browser
- `GoogleLoginService` creates and maintains one instance of `GoogleLoginState`, which is the core class of the login library
- Made `GoogleLoginService` singleton to do unit testing


plugins/com.google.cloud.tools.eclipse.appengine.login/.classpath
```diff
@@ -4,10 +4,12 @@
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="src" path="generated-sources"/>
+	<classpathentry exported="true" kind="lib" path="lib/com.google.cloud.tools.ide.login-0.0.0-SNAPSHOT.jar"/>
```
briandealwis: I thought this was a bundle already…?  We should be able to bring it in like we do app-tools-lib

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -1,58 +1,44 @@
...
         loginService.clearCredential();
       }
     }
 
-    ICommandService commandService =
-        (ICommandService) PlatformUI.getWorkbench().getService(ICommandService.class);
-    commandService.refreshElements(
-        "com.google.cloud.tools.eclipse.appengine.login.commands.loginCommand", null); //$NON-NLS-1$
-
+    if (credential != null) {
```
briandealwis: I thought you were able to move away from this code?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -16,36 +16,54 @@
...
+public final class GoogleLoginService {
 
   // For the detailed info about each scope, see
   // https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/wiki/Cloud-Tools-for-Eclipse-Technical-Design#oauth-20-scopes-requested
-  private static final List<String> OAUTH_SCOPES = Collections.unmodifiableList(Arrays.asList(
-      "https://www.googleapis.com/auth/cloud-platform" //$NON-NLS-1$
-  ));
+  private static final SortedSet<String> OAUTH_SCOPES = Collections.unmodifiableSortedSet(
+      new TreeSet<>(Arrays.asList(
+          "email", //$NON-NLS-1$
```
briandealwis: We need `email` to show their name on the login widgets too, don't we.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -1,4 +1,6 @@
-LOGIN_BROWSER_TITLE=Sign in to Google
+LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requires an open browser.
+LOGIN_ERROR_DIALOG_MESSAGE=An error occurred while trying to sign in.
+LOGIN_ERROR_DIALOG_TITLE=Signing In Error
```
akerekes: `Sign in failed`?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -0,0 +1,87 @@
...
+      browserSupport.getExternalBrowser().openURL(authCodeRequestUrl.toURL());
+
+      // TODO(chanseok): remove InputDialog and set up a web server to receive authorization code.
+      InputDialog dialog = new InputDialog(null,
+          "Enter Authorization Code", "Enter authorization code from the browser.", null, null);
+      dialog.open();
+      return dialog.getValue();
+
+    } catch (PartInitException pie) {
+      Shell shell = PlatformUI.getWorkbench().getDisplay().getActiveShell();
```
briandealwis: Use SWT's `Program.launch(urlString)` — it takes URLs and will use the system's default browser.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -0,0 +1,87 @@
...
+import org.eclipse.jface.dialogs.InputDialog;
+import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
+import org.eclipse.ui.commands.ICommandService;
+
+public class LoginServiceUi implements UiFacade {
```
briandealwis: `LoginServiceUi` should have a constructor that takes the relevant `Shell`.  Then you can remove all `PlatformUI` references

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStore.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.cloud.tools.ide.login.OAuthDataStore;
+import com.google.common.annotations.VisibleForTesting;
+
+import org.eclipse.e4.core.contexts.IEclipseContext;
+import org.eclipse.ui.PlatformUI;
+
+/**
+ * Provides a transient store for saving and loading {@link OAuthData} (a user credential).
+ */
+public class TransientOAuthDataStore implements OAuthDataStore {
```
briandealwis: I think a better approach here is to have the `IEclipseContext` be provided in the constructor and let the creator decide which context should be used.  You could even mark the Constructor with `@Inject` so that it can be created and injected (cf. `ContextInjectionFactory#make(Class,IEclipseContext)`)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
chanseokoh: I noticed a slight problem here. There should be a single instance of `GoogleLoginState`, which by its design takes an instance of `UiFacade`. That instance of `UiFacade` lives in the lifetime of `GoogleLoginState`, so clients can't dynamically pass a new instance of `UiFacade` accepting a shell that they see fit. Thoughts?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
chanseokoh: It was suggested that we just forget about #454, not putting any further effort to make it work.

plugins/com.google.cloud.tools.eclipse.appengine.login/.classpath
chanseokoh: Yeah, this is already a bundle. I couldn't find it in the Maven Central though. I guess someone (us?) need to publish it to the Maven Central first?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Yeah, eventually we need a fancier UI. The UI... doesn't exist yet. I will create an issue for having an enhanced UI.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
chanseokoh: Thanks!

plugins/com.google.cloud.tools.eclipse.appengine.login.test/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,8 @@ Bundle-Vendor: Google, Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.login
 Require-Bundle: org.hamcrest;bundle-version="1.1.0",
- org.junit;bundle-version="4.12.0"
+ org.junit;bundle-version="4.12.0",
+ com.google.cloud.tools.eclipse.test.dependencies
```
akerekes: Is this required? We import packages below and if something is missing, shouldn't we follow the same pattern to use Import-Package?

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+  GoogleLoginService loginService;
+
+  private static final SortedSet<String> OAUTH_SCOPES = Collections.unmodifiableSortedSet(
+      new TreeSet<>(Arrays.asList(
+          "email",
+          "https://www.googleapis.com/auth/cloud-platform"
+      )));
+
+  @Before
+  public void Setup() {
```
akerekes: setUp

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class GoogleLoginServiceTest {
+
+  @Mock OAuthDataStore dataStore;
+  @Mock OAuthData savedOAuthData;
+  @Mock UiFacade uiFacade;
+  @Mock LoggerFacade loggerFacade;
+
+  GoogleLoginService loginService;
```
akerekes: Is this used? All methods seem to create their on local instances.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+    // Persisted credential in the data store has an out-dated scopes.
+    SortedSet<String> newScope = new TreeSet<String>(Arrays.asList("new scope"));
+    when(savedOAuthData.getStoredScopes()).thenReturn(newScope);
+    when(savedOAuthData.getRefreshToken()).thenReturn("fake_refresh_token");
+
+    GoogleLoginService loginService = new GoogleLoginService(dataStore, uiFacade, loggerFacade);
+    Assert.assertNull(loginService.getCachedActiveCredential());
+  }
+
+  OAuthData singleStorageForDataStore;
```
akerekes: is this used?

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+          "https://www.googleapis.com/auth/cloud-platform"
+      )));
+
+  @Before
+  public void Setup() {
+    when(dataStore.loadOAuthData()).thenReturn(savedOAuthData);
+  }
+
+  @Test
+  public void testClearSavedCredentialIfNullRefreshToken() {
```
akerekes: use `_` to separate method name and test case

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -0,0 +1,97 @@
...
+    when(savedOAuthData.getRefreshToken()).thenReturn(null);
+
+    GoogleLoginService loginService = new GoogleLoginService(dataStore, uiFacade, loggerFacade);
+    Assert.assertNull(loginService.getCachedActiveCredential());
+  }
+
+  @Test
+  public void testClearSavedCredentialIfScopesChanged() {
```
akerekes: use `_` to separate method name and test case

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
chanseokoh: Good catch.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStoreTest.java
```diff
@@ -0,0 +1,77 @@
...
+  @Test
+  public void testSaveAndLoadOAuthData() {
+    // Set up IEclipseContext so that it returns OAuthData that was passed and saved.
+    doAnswer(new Answer<Void>() {
+      @Override
+      public Void answer(InvocationOnMock invocation) {
+        singleStorageForIEclipseContext = invocation.getArgumentAt(1, OAuthData.class);
+        return null;
+      }
+    }).doThrow(Exception.class).when(eclipseContext).set(anyString(), any());
```
akerekes: Is the `doThrow()` invocation used in this test?

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStoreTest.java
```diff
@@ -0,0 +1,77 @@
...
+    Assert.assertEquals(null, oAuthData.getAccessToken());
+    Assert.assertEquals(null, oAuthData.getRefreshToken());
+    Assert.assertEquals(null, oAuthData.getStoredEmail());
+    Assert.assertEquals(0, oAuthData.getAccessTokenExpiryTime());
+  }
+
+  OAuthData singleStorageForIEclipseContext;
+
+  @Test
+  public void testSaveAndLoadOAuthData() {
```
akerekes: What's the goal of this test? If it's to ensure that `loadAuthData()` calls `eclipseContext.get()` then we can verify that without the `answer` setup.

chanseokoh: It's a test that `loadAuthData()` returns the `OAuthData` object set through `saveOAuthData()`. It took me some thing to make this work with Mockito. I couldn't find a better than the `Answer` setup.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -54,21 +72,16 @@
...
-  public Credential getActiveCredential(IShellProvider shellProvider) throws IOException {
-    Credential credential = getCachedActiveCredential();
-
-    if (credential == null) {
-      credential = logIn(shellProvider);
-
-      IEclipseContext eclipseContext = PlatformUI.getWorkbench().getService(IEclipseContext.class);
-      eclipseContext.set(STASH_OAUTH_CRED_KEY, credential);
+  public Credential getActiveCredential() {
+    // TODO: holding a lock for a long period of time (especially when waiting for UI events)
```
akerekes: I'm a bit worried that we'll lock up here some action and the user will have no feedback on what's going on.
How about using timeouts when trying to get the lock and throw and exception to notify other threads trying to login, that one's already in progress?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStore.java
akerekes: I agree with @briandealwis I was also looking into this code, but wasn't sure how to solve it. I guess `PlatformUI.getWorkbench().getService(IEclipseContext.class)` should always return the same instance in all reasonable cases?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Good idea, but I think that'd be of no use and in fact not necessary, because this method is called only from the UI-thread. The UI-thread will never be blocked. (I made it that other methods that can be called from non-UI threads and holds a conflicting lock are always terminating quickly.) It's just that, if the UI-thread is holding a lock for a long time for taking care of login, other headless-jobs are blocked. Ideally, I think we should fix the ide-login library itself.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
patflynn: I believe you need https://www.googleapis.com/auth/appengine.admin for app engine deployment.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: On second thought, I wonder if we have two workbenches and both try to login. If two workbenches share a single class loader (i.e., share a single `GoogleLoginService` instance), then I think there is a problem. Not sure how I can actually test this.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: So far, we didn't need it for standard app engine deployment.

patflynn: that is...odd.  Perhaps they've finally sorted this out. I'll check with
the cloud SDK team.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Seems like workbenches are per-JVM basis, so I think we are covered and okay.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStoreTest.java
chanseokoh: It is to fail if `set()` is called twice. (If called twice, this test setup won't work as intended.)

plugins/com.google.cloud.tools.eclipse.appengine.login.test/META-INF/MANIFEST.MF
chanseokoh: Removed `com.google.cloud.tools.eclipse.test.dependencies` and added ` org.mockito.invocation;provider=google;version="1.10.19"` under `Import-Package`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStore.java
chanseokoh: I changed the constructor to accept `IEclipseContext`. Didn't use the `@Inject`. (Wasn't sure if you meant injecting `IEcipseContext` into `TransientOAuthDataStore`, or making `TransientOAuthDataStore` injectable by putting it in `EclipseContext`). Maybe next time, after I study the E4 Dependency Injection in depth.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: Not in a RAP situation.  We don't have that as a requirement, but it's nicer not to make assumptions (and better code) — I'll add a bit about this separately.

briandealwis:
@chanseok I'm not particularly enamoured by turning GoogleLoginService into a singleton — I've just seen far too many problems with singletons.  I wondered what you thought about instead changing it to be an OSGi Service, and plus we can replace the instance for testing by defining a higher ranking (`org.osgi.framework.Constants#SERVICE_RANKING`).  The consumers should be able to use the `IServiceLocator` interface, implemented by most workbench components, to obtain the service, but we can turn `getInstance()` into a helper `getInstance(Class)` that obtains the service from the class' bundle if necessary.

chanseokoh:
@briandealwis I see. Let's avoid making a singleton. Finally, it's time for me to get to know about creating an OSGi service!

briandealwis:
The easiest way to do it is via Declarative Services, which uses an XML file to define the service.  Take a look at the `gcloud.xml` in the `.sdk` bundle, and how it's referenced from the `MANIFEST.MF`.  Your bundle has to be marked with 'activate on class load' (= `Bundle-ActivationPolicy: lazy` in the `MANIFEST.MF`).

chanseokoh:
@briandealwis wow, thanks!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: In that case, here's my idea: check if the login is currently in process (I think a `volatile boolean` flag would work) in `getActiveCredential()`, and if it is, it shows an error message dialog saying "Login already in progress. Please retry later."

akerekes: That's why I was thinking about locks. If you use the flag, and it's false (meaning no login is in progress), by the time you get to the next line and start the synchronized block, it can be changed to true by a concurrent thread. Possibly a very small chance, but it's still there.

briandealwis: Sounds like we want to be able to chain a set of actions together.  Login could be done in a UIJob, hands off to a XXXJob, etc.  With some ability to handle an OK status, cancelled status (dialog: deploy cancelled), or error status (error dialog).

chanseokoh: @akerekes I believe atomic-get-and-set should work. If it is set, error and return. If it is not set, set it and grab a lock.

chanseokoh: @briandealwis Yeah, that'd be cool. The problem here is that the entire stream of login work is done by the ide-login library we can't control and the library mixes UI flow and non-UI flow in itself.

chanseokoh:
I'm still working on making it an OSGi service.

chanseokoh:
I made it an OSGi service. I think the way it is registered and retrieved is a bit different from `.sdk`, but it's working anyway.

- Defined `IGoogleLoginService`.
- Added `Activator`, which registers a service instance.
- Added `OGSI-INF/login.xml` and a corresponding entry in `MANIFEST.MF`.
- Consumers call `PlatformUI.getWorkbench().getService(IGoogleLoginService.class)` to get the service instance.

I assume here that the lifetime of a service instance aligns with the lifetime of Eclipse (or workbench), essentially making it a singleton. If it isn't the case, we'll be in slight trouble.

One warning message I see when starting up:

```
!ENTRY com.google.cloud.tools.eclipse.appengine.login 4 0 2016-08-03 11:54:24.234
!MESSAGE [SCR] Component definition XMLs not found in bundle com.google.cloud.tools.eclipse.appengine.login. The component header value is OGSI-INF/login.xml
```

I don't understand this, since the file is there and everything is working.

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
```diff
@@ -6,20 +6,21 @@ Bundle-Version: 0.1.0.qualifier
 Bundle-Vendor: Google, Inc.
 Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-Activator: com.google.cloud.tools.eclipse.appengine.login.Activator
 Bundle-ActivationPolicy: lazy
-Require-Bundle: com.google.cloud.tools.app.lib;bundle-version="0.1.0",
- com.google.gson,
+Service-Component: OGSI-INF/login.xml
```
akerekes: Should be OSGI-INF/...

chanseokoh:
I realized I was doing the OGSi service stuff in a convoluted way. The last commit removes the `Activator`. You can start taking a look.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -88,8 +86,10 @@ private IPath createWorkDirectory() throws IOException {
     return workDirectory;
   }
 
-  private Credential login(IShellProvider shellProvider) throws IOException, CoreException {
-    Credential credential = new GoogleLoginService().getActiveCredential(shellProvider);
+  private Credential login() throws CoreException {
+    IGoogleLoginService loginService =
+        PlatformUI.getWorkbench().getService(IGoogleLoginService.class);
```
briandealwis: We want to avoid using `PlatformUI.getWorkbench()` wherever possible.  In this case, we're provided the `ExecutionEvent` which provides access to a bunch of objects that implement `IServiceLocator`.  Here's what I'd do:
  - pass in the `IGoogleLoginService` instance into `login()`
  - create a `getService(ExecutionEvent,Class<?>)` method that calls the following sequence of methods to find an `IServiceLocator`:
     - `HandlerUtil.getActiveSite()` (incl views and editors)
     - `HandlerUtil.getActiveEditor()`
     - `HandlerUtil.getActiveWorkbenchWindow()`
     - and if none of them work, then `PlatformUI.getWorkbench()`

plugins/com.google.cloud.tools.eclipse.appengine.login/build.properties
```diff
@@ -6,6 +6,7 @@ bin.includes = META-INF/,\
                plugin.xml,\
                plugin.properties,\
                icons/,\
-               lib/
+               lib/,
+               OSGI-INF/login.xml
```
briandealwis: Could just make this `OSGI-INF/` in case any other services are added

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -19,40 +15,33 @@
 
 public class GoogleLoginCommandHandler extends AbstractHandler implements IElementUpdater {
 
+  private IGoogleLoginService getGoogleLoginService() {
```
briandealwis: Ditto about avoiding _PlatformUI.getWorkbench()_ made in [_StandardDeployCommandHandler_](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/486/files#r73461563)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -16,96 +16,108 @@
...
-   * Returns {@code null} if no credential has been cached.
-   *
-   * Safe to call from non-UI contexts.
-   */
-  public Credential getCachedActiveCredential() {
-    IEclipseContext eclipseContext = PlatformUI.getWorkbench().getService(IEclipseContext.class);
-    return (Credential) eclipseContext.get(STASH_OAUTH_CRED_KEY);
+  public GoogleLoginService() {
+    this(new TransientOAuthDataStore(PlatformUI.getWorkbench().getService(IEclipseContext.class)),
+        new LoginServiceUi(), new LoginServiceLogger());
```
briandealwis: This is where I'd pass in the _Display_ or _Shell_ to `LoginServiceUi`

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -16,96 +16,108 @@
...
 
-  private Credential logIn(IShellProvider shellProvider) throws IOException {
-    GoogleLoginBrowser loginBrowser = new GoogleLoginBrowser(
-        shellProvider.getShell(), Constants.getOAuthClientId(), OAUTH_SCOPES);
-    if (loginBrowser.open() != GoogleLoginBrowser.OK) {
-      return null;
+  @Override
+  public Credential getActiveCredential() {
+    if (!loginInProgress.compareAndSet(false, true)) {
+      LoginServiceUi.showErrorDialogHelper(
```
briandealwis: Why explicit use of LoginServiceUi instead of the uiFacade passed into the `GoogleLoginState`

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -16,96 +16,108 @@
...
     }
 
-    GoogleAuthorizationCodeTokenRequest authRequest = new GoogleAuthorizationCodeTokenRequest(
-        Utils.getDefaultTransport(), Utils.getDefaultJsonFactory(),
-        Constants.getOAuthClientId(), Constants.getOAuthClientSecret(),
-        loginBrowser.getAuthorizationCode(),
-        GoogleLoginBrowser.REDIRECT_URI);
-
-    return createCredential(authRequest.execute());
+    // TODO: holding a lock for a long period of time (especially when waiting for UI events)
```
briandealwis: Should this method return a Future?

chanseokoh: Are you suggesting making `getActiveCredential()` return a Future or making this portion use a Future? If the latter case, that won't really help anything, because (if I'm not mistaken) this method still needs to block on `Future.get()` while holding a lock while the Future computation (i.e., login through UI) is on-going.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: This method is static, so using an instance instead of calling statically actually generates a warning. You'are saying something else?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Ah, sorry. You explicitly said returning a Future. I'll go with that.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+@RunWith(MockitoJUnitRunner.class)
+public class GoogleLoginServiceTest {
+
+  @Mock OAuthDataStore dataStore;
```
elharo: nit: these fields can probably be private

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStoreTest.java
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class TransientOAuthDataStoreTest {
+
+  @Mock IEclipseContext eclipseContext;
```
elharo: private

plugins/com.google.cloud.tools.eclipse.appengine.login/.classpath
elharo: Please file a high-priority issue to do this

plugins/com.google.cloud.tools.eclipse.appengine.login/META-INF/MANIFEST.MF
```diff
@@ -7,19 +7,19 @@ Bundle-Vendor: Google, Inc.
...
-Bundle-Classpath: lib/google-api-client-1.22.0.jar,
+Bundle-Classpath: lib/com.google.cloud.tools.ide.login-0.0.0-SNAPSHOT.jar,
+ lib/google-api-client-1.22.0.jar,
+ lib/google-api-services-oauth2-v2-rev114-1.22.0.jar,
  lib/google-http-client-1.22.0.jar,
- lib/google-http-client-jackson2-1.22.0.jar,
+ lib/google-http-client-jackson-1.22.0.jar,
  lib/google-oauth-client-1.22.0.jar,
- lib/jackson-core-2.1.3.jar,
+ lib/jackson-core-asl-1.9.13.jar,
```
elharo: I'm surprised we can't get this one from maven

elharo:
The non-google jackson jars should be pulled form maven: https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core/2.1.3

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  @Override
+  public void showErrorDialog(String title, String message) {
+    // Ignore "title" and "message", as they are non-localized hard-coded strings in the library.
+    showErrorDialogHelper(Messages.LOGIN_ERROR_DIALOG_TITLE, Messages.LOGIN_ERROR_DIALOG_MESSAGE);
+  }
+
+  @Override
+  // TODO(chanseok): the login library doesn't call this method for logging in. Fix the library.
```
elharo: This would be better as an issue rather than a TODO, since no change to this code is needed here. Please file one.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
nbashirbello: Maybe something like:

GoogleLoginServer.getInstance() returning IGoogleLoginService 

public static IGoogleLoginService getInstance() {
  if (instance == null) {
    instance = new GoogleLoginService();
  }
  return instance;
}

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -0,0 +1,85 @@
...
+  public String obtainVerificationCodeFromUserInteraction(
+      String title, GoogleAuthorizationCodeRequestUrl authCodeRequestUrl) {
+    if (!Program.launch(authCodeRequestUrl.toString())) {
+      Shell shell = PlatformUI.getWorkbench().getDisplay().getActiveShell();
+      MessageDialog.openError(shell,
+          Messages.LOGIN_ERROR_DIALOG_TITLE, Messages.LOGIN_ERROR_CANNOT_OPEN_BROWSER);
+      return null;
+    }
+
+    // TODO(chanseok): remove InputDialog and set up a web server to receive authorization code.
```
elharo: This one should probable be an issue and a TODO. Please file an enhancement issue that would be blocked on this PR.

chanseokoh:
I'll try pulling the jackson jars from the Maven Central again. (I failed it last time.)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: I see now the fully parameterized `GoogleLoginServer` constructor is `@VisibleForTesting`.  Was thinking that since much of the UI interaction has been made pluggable with the `UiFacade`, we may as well use it here too.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -88,8 +86,10 @@ private IPath createWorkDirectory() throws IOException {
...
+    Credential credential = loginService.getActiveCredential();
```
nbashirbello: I think thee GoogleLoginService should probably have and isLoggedIn() method and that can be called here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -1,7 +1,14 @@
 package com.google.cloud.tools.eclipse.appengine.deploy.standard;
```
nbashirbello: License header?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: The UI interaction is not flexibly pluggable unfortunately. `GoogleLoginState` should be constructed only once as a singleton (it maintains a login-related state) and the `UiFacade` is fixed at initialization.

chanseokoh: BTW, `showErrorDialogHelper` is actually just a helper method to display an error dialog. Probably it should go into the `.ui.util` bundle.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
chanseokoh: @nbashirbello It was like that before, but it was suggested we avoid doing a singleton and rather make this an OSGi service.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/TransientOAuthDataStore.java
```diff
@@ -0,0 +1,54 @@
...
+package com.google.cloud.tools.eclipse.appengine.login;
+
+import com.google.cloud.tools.ide.login.OAuthData;
+import com.google.cloud.tools.ide.login.OAuthDataStore;
+
+import org.eclipse.e4.core.contexts.IEclipseContext;
+
+/**
+ * Provides a transient store for saving and loading {@link OAuthData} (a user credential).
+ */
```
nbashirbello: Just to confirm, this is a temporary solution to storing the credentials right?

Does this support Eclipse restarting?

chanseokoh: You're right. It's almost like providing an empty implementation to the `OAuthDataStore`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
chanseokoh: My mistake. I looked at the ide-login code again, and it ends up calling this method. Problem solved!

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
chanseokoh: @akerekes and I discussed this before: #472

chanseokoh:
This PR fixes #446 too.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -0,0 +1,85 @@
...
+    MessageDialog.openError(shell, title, message);
+  }
+
+  @Override
+  public boolean askYesOrNo(String title, String message) {
+    throw new RuntimeException("Not allowed to ensure non-UI threads don't prompt."); //$NON-NLS-1$
+  }
+
+  @Override
+  public void showErrorDialog(String title, String message) {
```
nbashirbello: The title and message are supposed to be passed by the feature trying to log in. For instance, if a user calls deploy without logging in, deploy would make an attempt to log in. If there is an error and we pop up a dialog, the title should say "Deploying project ...", so that the user is clear why logging was called.

chanseokoh: I agree. That should be the case, but the ide-login library is [broken in this regard](https://github.com/GoogleCloudPlatform/ide-login/blob/master/src/main/java/com/google/cloud/tools/ide/login/GoogleLoginState.java#L297). Both the title and the message are hard-coded (not localized either).

I created an issue against the login library: https://github.com/GoogleCloudPlatform/ide-login/issues/16

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
chanseokoh: Created #503.

chanseokoh:
Now pulling `jackson-core-asl-1.9.13.jar` from the Maven Central. Builds fine locally, and the login is working properly. PTAL.

elharo:
Can you also remove plugins/com.google.cloud.tools.eclipse.appengine.login/lib/jackson-core-2.1.3.jar? It looks like you're not actually referencing it.

chanseokoh:
@elharo The jar is being removed as part of this PR. (You can find it in the PR if you look really hard.)

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
briandealwis: Hmm, I don't see how it's more helpful than just calling `MessageDialog.openError(shell, title, message)`.  Regardless of the pluggability of the UiFacade, you have a UiFacade and I don't understand why you wouldn't just use its `UiFacade#showErrorDialog(title,message)`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
elharo: Are you sure about this? I thought login was the result of an explicit user selection of a login button/menu item. What am I missing here?

chanseokoh: We may enforce the policy that login is done only through explicit user selection of a button or menu. Currently, the deployment automatically triggers login rather than simply failing with a message saying that the user needs to login first. Here is the relevant issue: #472. We could make a decision soon.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: No problem. I'll use `UiFacade#showErrorDialog(title,message)`.

chanseokoh: Ah, the problem was that, because of [this](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/486/files/535f3b5b0506cf3fab62fd9015202df008d4e085#r73581984), `title` and `message` are completely ignored in `UiFacade#showErrorDialog(title,message)`. The method is not usable.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -0,0 +1,86 @@
...
+    showErrorDialogHelper(Messages.LOGIN_ERROR_DIALOG_TITLE, Messages.LOGIN_ERROR_DIALOG_MESSAGE);
+  }
+
+  @Override
+  public void notifyStatusIndicator() {
+    // Update and refresh the menu, toolbar button, and tooltip.
+    display.asyncExec(new Runnable() {
+      @Override
+      public void run() {
+        PlatformUI.getWorkbench().getService(ICommandService.class).refreshElements(
```
briandealwis: Here you're using `IWorkbench` as an IServiceLocator.  Rather than hard-code the workbench here, pass in `IServiceLocator` in the constructor.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ServiceUtils.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.ui.IEditorSite;
+import org.eclipse.ui.IWorkbenchPartSite;
+import org.eclipse.ui.IWorkbenchSite;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.handlers.HandlerUtil;
+
+public class ServiceUtils {
+
+  public static <T> T getService(ExecutionEvent event, Class<T> api) {
```
briandealwis: We should have a javadoc here

chanseokoh:
Had a chat with @briandealwis and learned that intrinsic references/access to workbenches is bad.

The last commit does the following:

  - `LoginServiceUi` is now free of any workbench, accepting `IServiceLocator` and `IShellProvider`.
  - `GoogleLoginService` is initialized by `activate()` (which is called by OSGi DS). Before, it was done by a 0-arg constructor. @briandealwis, does this look good? (I hope it is not initialized multiple times for an instance.)
  - Added javadoc to `ServiceUtils`.

briandealwis:
LGTM.  But if you want to keep `GoogleLoginService` as a POJO, you could:
  - drop the `ComponentContext` argument on `activate`: it's optional, and only required to access metadata set in the `OSGI-INF/XXX.xml` file or to access the OSGi framework.
  - keep the 3-arg constructor or add setters for the datastore, UI facade, and logging facade
  - merge your `activate` and `initialize` methods to set defaults for the datastore, UI and logging facades as needed.

Then your OSGI-DS created service will be no different from running:
```
   GoogleLoginService service = new GoogleLoginService();
   service.activate();
```
OSGi DS is then just an activation and service access mechanism.

chanseokoh:
Removed the `ComponentContext` argument. Thanks.

> - keep the 3-arg constructor or add setters for the datastore, UI facade, and logging facade
> - merge your activate and initialize methods to set defaults for the datastore, UI and logging facades as needed.

While I was trying these, I encountered some clunkiness. Setters don't make much sense, as these three arguments are basically meant for creating `GoogleLoginState`. `initialize()` (and the 3-arg constructor) does some extra setup too. So, to reduce code duplication, you anyway end up defining a common init method if you want to keep a constructor. The constructor was only for unit testing, so I think it's okay if we don't have it. The reason for a separate `initialize()` is also for unit testing.

Anyway, without the last two bullets, we can still achieve

```
   GoogleLoginService service = new GoogleLoginService();
   service.activate();
```

briandealwis:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -60,7 +60,7 @@
    * Called by OSGi Declarative Services Runtime when the {@link GoogleLoginService} is activated
    * as an OSGi service.
    */
-  protected void activate(ComponentContext context) {
+  protected void activate() {
     IWorkbench workbench = PlatformUI.getWorkbench();
     IEclipseContext eclipseContext = workbench.getService(IEclipseContext.class);
     IShellProvider shellProvider = new SameShellProvider(workbench.getDisplay().getActiveShell());
```
briandealwis: Oops, that's not what you want: that will fix the service with the shell of the active window / dialog _at creation time_.  I think you want:
```
IShellProvider shellProvider = new IShellProvider() {
  public Shell getShell() {
    return workbench.getDisplay().getActiveShell();
  }
}
```

briandealwis:
LEBTM (looks even better…)

briandealwis:
LGTM

PR #487, opened 2016-08-02
Add AdapterUtil and use the adapter pattern to get the IProject from selection in ProjectFromSelectionHelper

akerekes:
Based on the comment from @briandealwis :

https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/485#issuecomment-236762625


briandealwis:
lgtm

PR #496, opened 2016-08-03
Ui utils

akerekes:


chanseokoh:
It's the first time we put the `.ui` hierarchy at the top of our namespace (i.e., right after `com.google.cloud.tools.eclipse`). Did you mean `.util.ui`?

akerekes:
It was conscious decision as think this class will be pretty general and used in several other bundles. Putting `ui` at the end would suggest to me that this bundle is a UI bundle for `c.g.c.t.e.util`, which is not.

chanseokoh:
I see. So this is a utility bundle for UI stuff.

BTW, looks like an empty test bundle makes the build fail.

chanseokoh:
LGTM

PR #497, opened 2016-08-03
Fixes #366 - Highlight deployed version URL as link

### LINKED ISSUE #366: Highlight deployed version URL as link
When the user deploys a project to App Engine, the console output from CloudSdk will show the URL the new version is accessible on. It would be more user-friendly if the console output would be decorated so that the URL is a clickable link that opens in a browser for the user to test the new version.

akerekes:
WIP: tests to come, feedback is welcome on the approach.

Screenshot from the decorated deploy output:

deploy_url


chanseokoh:
This feature is nice. The approach seems good.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/console/BrowserSupportBasedHyperlink.java
```diff
@@ -0,0 +1,45 @@
...
+  public void linkEntered() {
+  }
+
+  @Override
+  public void linkActivated() {
+    try {
+      IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
+      browserSupport.createBrowser(null).openURL(new URL(url));
+    } catch (PartInitException | MalformedURLException exception) {
+      logger.log(Level.SEVERE, "Cannot open hyperlink", exception);
```
briandealwis: I'd try SWT's `Program.launch()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/plugin.xml
```diff
@@ -43,4 +43,28 @@
...
+  <extension
+    point="org.eclipse.ui.console.consolePatternMatchListeners">
+    <consolePatternMatchListener
+      class="com.google.cloud.tools.eclipse.ui.util.console.PatternToHyperlinkConverter"
+      id="com.google.cloud.tools.eclipse.appengine.deploy.consolePatternMatchListener"
+      regex="https://.+\.appspot\.com">
+      <enablement>
+        <test
+          property="com.google.cloud.tools.eclipse.ui.util.console.name"
+          value="App Engine Deploy">
```
briandealwis: I think this would be better testing the console _type_.  I'd expect the console _name_ to include the projectId + version, and maybe date and time?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/console/ConsolePropertyTester.java
```diff
@@ -0,0 +1,25 @@
...
+ * <i>Currently only the </i><code>name</code><i> property is supported for matching.</i>
+ */
+public class ConsolePropertyTester extends PropertyTester {
+
+  private static final String NAME = "name";
+
+  @Override
+  public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+    IConsole console = (IConsole) receiver;
+    if (NAME.equals(property)) {
```
briandealwis: Console _type_? The name may be translated.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/plugin.xml
```diff
@@ -43,4 +43,28 @@
     </handler>
   </extension>
 
+  <extension
+    point="org.eclipse.ui.console.consolePatternMatchListeners">
+    <consolePatternMatchListener
+      class="com.google.cloud.tools.eclipse.ui.util.console.PatternToHyperlinkConverter"
+      id="com.google.cloud.tools.eclipse.appengine.deploy.consolePatternMatchListener"
+      regex="https://.+\.appspot\.com">
```
briandealwis: This seems broadly useful beyond the deploy console but includes the debugging console (e.g., the output from `dev_appserver` with URL links).  I suppose the `.localserver.ui` bundle could define its own pattern match listener and use the same class.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
```diff
@@ -21,8 +21,8 @@
 import com.google.cloud.tools.eclipse.appengine.deploy.CleanupOldDeploysJob;
 import com.google.cloud.tools.eclipse.appengine.deploy.Messages;
 import com.google.cloud.tools.eclipse.appengine.login.GoogleLoginService;
+import com.google.cloud.tools.eclipse.ui.util.ProjectFromSelectionHelper;
```
briandealwis: Spurious change

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -24,8 +24,8 @@
 import com.google.cloud.tools.eclipse.appengine.deploy.Messages;
 import com.google.cloud.tools.eclipse.appengine.login.CredentialHelper;
 import com.google.cloud.tools.eclipse.sdk.ui.MessageConsoleWriterOutputLineListener;
+import com.google.cloud.tools.eclipse.ui.util.MessageConsoleUtilities;
```
briandealwis: Spurious change

briandealwis:
Nice!  LGTM.  I'm a bit surprised that there aren't implementations of _PatternToHyperlinkConverter_ and _BrowserSupportBasedHyperlink_.

briandealwis:
Ah, I see Mylyn does include a _UrlHyperlink_ and a _UrlPatternMatchListener_ but they will bring in all sorts of dependencies.

elharo:
org.apache.maven.InternalErrorException: Internal error: java.lang.RuntimeException: org.osgi.framework.BundleException: Bundle com.google.cloud.tools.eclipse.ui.util cannot be resolved
Resolution errors:
   Bundle com.google.cloud.tools.eclipse.ui.util - Missing Constraint: Import-Package: org.eclipse.ui; version="0.0.0"
   Bundle com.google.cloud.tools.eclipse.ui.util - Missing Constraint: Import-Package: org.eclipse.ui.browser; version="0.0.0"

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/console/BrowserSupportBasedHyperlink.java
akerekes: My understanding is that users can configure whether to use internal or external (and which external) browser to use. With `Program.launch()` the user cannot open it in an internal browser (maybe there's some hacky way to do so). I like the flexibility of the `IWorkbenchBrowserSupport`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/plugin.xml
akerekes: I agree. This comes from the fact that the `MessageConsole` for deploy comes from the `MessageConsoleUtilities` getConsole() method which uses the name.
For now I'd leave this as is in this PR, file an issue to use type to obtain the console and then use type in the property tester and fix it ASAP.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployCommandHandler.java
akerekes: Rather a bad practice: I've moved this class to the `ui.util` bundle as part of this PR.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
akerekes: Rather a bad practice: I've moved this class to the `ui.util` bundle as part of this PR.

akerekes:
I like those names more, I was struggling with naming these classes and ended up with these weird names: `PatternToHyperlinkConverter` and `BrowserSupportBasedHyperlink`.
I wouldn't mind using the same or similar names as Mylyn uses.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/console/BrowserSupportBasedHyperlink.java
briandealwis: I only meant in the `PartInitException` handler: the normal way failed, so we can at least use the system default browser.

akerekes: Gotcha! I'll add it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/plugin.xml
akerekes: #502

PR #506, opened 2016-08-04
add environment tab to run configuration

### LINKED ISSUE #429: Allow easy way to provide environment variables for Cloud emulators
Local emulators for  gcloud-java (https://github.com/GoogleCloudPlatform/gcloud-java)  for datastore  and pubsub sets environment variable in the current shell.

that is, if i run
gcloud beta emulators datastore start
it will set
export DATASTORE_LOCAL_HOST=localhost:8087
and the gcloud-java datastore client will use that to connect locally.

Its not clear how to set that variable in the eclipse tools (i tried setting it in the launch config manually but also saw the following:
launch_config_tabs

Bonus to find some automagic or easy way to read in these variables without manually adding them in to the runtime every time.   NOTE: I dont' think the dev_appserver allows for dynamic env variables (you need to declare them directly in appengine-web.xml or app.yaml
see:
https://cloud.google.com/appengine/docs/java/config/appref

elharo:
Mostly fix #429


briandealwis:
LGTM

briandealwis:
Note: was removed in #1139.

Issue #492, opened 2016-08-02
Add deployment configuration page


akerekes:
The deployment configuration page should be shown in the dialog of project properties.
This can be broken down into several smaller tasks and implemented incrementally:

- [x] 1. display a Google property item and and App Engine Standard subitem displaying an empty panel

- [x] 2. display the App Engine item only if the project has the App Engine Standard facet

- [x] ~~3. display the Google item only is there is at least one visible child item~~ #562

- [x] 4. add editors corresponding to the App Engine Standard deploy configuration

- [x] 5. persist the values set in the editors

- [x] 6. move UI code to c.g.c.t.appengine.deploy.ui

- [x] 7. use the values set in the deploy configuration when launching the deploy process

- [x] 8. implement Restore Defaults button (restore to the gcloud defaults?) see also https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/554#discussion_r76263074

PR #508, opened 2016-08-05
Add a top level Google and a child App Engine Deployment property page

akerekes:
The Google property page is visible always (should it be restricted to JavaProjects?), the deploy page is only visible for projects having the App Engine Standard facet.

Implements the first task in #492


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,16 @@
...
+Bundle-Name: GCloud for Eclipse App Engine Deployment Support UI
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.deploy.ui;singleton:=true
+Bundle-Version: 0.1.0.qualifier
+Bundle-Vendor: Google, Inc.
+Bundle-Localization: plugin
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-ActivationPolicy: lazy
+Require-Bundle: org.eclipse.ui.workbench
+Import-Package: org.eclipse.core.resources,
+ org.eclipse.core.runtime;bundle-symbolic-name:="org.eclipse.equinox.common",
```
briandealwis: Rather than use the `bundle-symbolic-name` directive, you should be able to use `common=split`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
```diff
@@ -0,0 +1,32 @@
...
+    </page>
+  </extension>
+
+  <extension point="org.eclipse.ui.propertyPages">
+    <page
+        id="com.google.cloud.tools.eclipse.appengine.deploy"
+        name="App Engine Deployment"
+        category="com.google.cloud.tools.eclipse"
+        class="com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployPropertyPage">
+      <enabledWhen> <!-- this will probably be too strict when App Engine Flex becomes supported -->
```
briandealwis: We should [define some of our own standard expressions](common=split) that can be referenced here.  I think it should be safe to not use `forcePluginActivation` since the facets bundle should be started if we're dealing with faceted projects.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -0,0 +1,2 @@
+preferences.title=App Engine Deployment Settings
+google.preferences=Google Cloud platform related settings will live under this page
```
briandealwis: Active voice seems more appropriate here: _Google Cloud Platform-related settings_?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/META-INF/MANIFEST.MF
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: broken link

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
akerekes: done

briandealwis:
LGTM.  It may make sense to implement the contents of the `DeployPropertyPage` as a standalone widget that could also be hosted in a dialog.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: #509

PR #510, opened 2016-08-05
Include usagetracker plugin to e45 feature

chanseokoh:
I'm sure we will need this.

BTW, is it correct that `com.google.cloud.tools.app.lib` is in both `e45` and `3rdparty`?

### DIFF
--- a/features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
+++ b/features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
@@ -117,6 +117,13 @@
          unpack="false"/>
 
   <plugin
+         id="com.google.cloud.tools.eclipse.usagetracker"
+         download-size="0"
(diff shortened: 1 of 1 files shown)


briandealwis:
> @chanseokoh: BTW, is it correct that com.google.cloud.tools.app.lib is in both e45 and 3rdparty?

The `3rdparty` feature is for distributing bundles that aren't available from the usual places (namely bundles from Maven Central).  The feature isn't listed in the `category.xml`, but the contents are visible to p2 when it looks to satisfy the requirements of our main bundle.  The actual feature is never installed.

We could decide not to include `.app.lib` in our `e45.feature`.  But `.app.lib` is somewhat special in central to our functionality. Since feature installation is atomic, we can be sure that CTE was installed with its tested version of `.app.lib`.

briandealwis:
Oh, and LGTM.

Issue #503, opened 2016-08-04
Login - use local web server to receive authorization code


chanseokoh:
After getting #486 in, set up a web sever that receives an authorization code for the OAuth2 login flow.

chanseokoh:
Almost certainly, we want two landing pages each for successful and unsuccessful login. Example

https://cloud.google.com/tools/intellij/auth_success
https://cloud.google.com/tools/intellij/auth_failure

@arouzrokh Who can work on these docs?

chanseokoh:
For the landing pages, #557 is created. This issue is only for using a local web server.

chanseokoh:
Closed by #513.

PR #513, opened 2016-08-07
Login - run local web server to receive authorization code

chanseokoh:
Runs a local web server and receives authorization code (or error) after login through an external browser. Also shows a cancelable modal dialog while the login is in progress.

~~`AuthorizationCodeReceiver` is a class that does most of the job by running Jetty. However, the class depends only on Jetty and Servlet — no OAuth or Google dependency for example.~~

~~Everything works... **except** updating/refreshing the button and menu items right after successful login. The reason?~~

~~`workbench.getDisplay().getActiveShell()` in [here](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java#L68) returns null (i.e., no active shell), when I am looking at an external browser, which is the normal scenario of login using an external browser. On the other hand, if I quickly switch the focus from an external browser to Eclipse after completing the login, `getActiveShell()` returns a usable shell. @briandealwis, do you have some ideas?~~

Update: The issues above are fixed or no longer relevant.


briandealwis:
I think this is an issue because [`LoginServiceUi#notifyStatusIndicator()`](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java#L58) tries to get the `Display` for an `asyncExec()`, right?

```
  @Override
  public void notifyStatusIndicator() {
    // Update and refresh the menu, toolbar button, and tooltip.
    shellProvider.getShell().getDisplay().asyncExec(new Runnable() {
```

You can do one of two things:
  1. Pass the `Display` to `LoginServiceUi` separately (maybe not a bad idea)
  2. Change your shell provider so it tries everything to return a `Shell`. It can check the following until it finds a non-null Shell:
    1. `Display#getActiveShell()` (already done)
    2. the workbench's active workbench window
    3. iterate through `IWorkbench#getWorkbenchWindows()` and return the first one whose Shell `isVisible()`
    4. iterate through `Display#getShells()` and return the first one with `isVisible()`

If you go for the second, that `ShellProvider` would definitely be a candidate for `.ui.util`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Starts a local server listening through an ephemeral port.
+   *
+   * @return "http://127.0.0.1:<listening port>/code-reception-point", where this server listens
+   *     and expects to get an authorization code returning from a OAuth login flow
+   * @throws IOException wraps any {@link Exception} from Jetty when it failed to start
+   */
+  public String runLocalServer() throws IOException {
+    Preconditions.checkState(server == null, "Don't reuse the receiver.");
```
elharo: checkNotNull

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Waits for a login result to arrive for the given timeout or until interrupted.
+   *
+   * @param timeout in millisecond. Cannot be 0
+   * @return true if the web server consumed the login result
+   */
+  public boolean waitForCode(long timeout) {
+    Preconditions.checkArgument(timeout != 0);
+    Preconditions.checkState(server != null, "Call runLocalServer() first.");
```
elharo: checkNotNull

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+
+    synchronized (server) {
+      try {
+        server.wait(timeout);
+      } catch (InterruptedException e) {}
```
elharo: e --> ex

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+    return resultConsumed;
+  }
+
+  public void shutdown() {
+    Preconditions.checkState(server != null, "Server was never created.");
+
+    try {
+      server.stop();
+      server.join();
+      logger.log(Level.INFO, "Stopped the local web server for login.");
```
elharo: probably don't need to log this

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+    } catch (Exception ex) {
+      logger.log(Level.WARNING, "Failed to stop the local web server for login.", ex);
+    }
+  }
+
+  public String getAuthorizationCode() {
+    Preconditions.checkState(resultConsumed);
```
elharo: I'm not sure a precondition is right here. May want to be more explicit.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +74,60 @@ public void run() {
 
   @Override
   public VerificationCodeHolder obtainVerificationCodeFromExternalUserInteraction(String title) {
-    throw new RuntimeException("Not to be called."); //$NON-NLS-1$
+    AuthorizationCodeReceiver codeReceiver = new AuthorizationCodeReceiver();
+
+    String redirectUrl;
```
elharo: refactor as one big try block with several catch blocks

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +74,60 @@ public void run() {
...
+    } catch (InvocationTargetException ex) {
+      showErrorDialogHelper(Messages.LOGIN_ERROR_DIALOG_TITLE,
+          Messages.LOGIN_ERROR_LOCAL_SERVER_RUN
+              + ex.getLocalizedMessage());
+      return null;
+    } catch (InterruptedException ex) {
+      return null;  // User pushed the cancel button.
+    }
+    finally {
+      codeReceiver.shutdown();
```
elharo: I think you want to do this if the top try fails also

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +74,60 @@ public void run() {
...
-          Messages.LOGIN_ERROR_DIALOG_TITLE, Messages.LOGIN_ERROR_CANNOT_OPEN_BROWSER);
-      return null;
-    }
-
-    // TODO(chanseok): remove InputDialog and set up a web server to receive authorization code.
-    InputDialog dialog = new InputDialog(null,
-        "Enter Authorization Code", "Enter authorization code from the browser.", null, null);
-    dialog.open();
-    return dialog.getValue();
+    throw new RuntimeException("Not to be called."); //$NON-NLS-1$
```
elharo: I wonder if we need this method in the interface at all? Is IntelliJ using it?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Runs a Jetty web server instance that opens up an HTTP listening port to receive OAuth2
+ * authorization code.
+ */
```
patflynn: I'm just taking a look at the IJ implementation and noticing that there are a few parts of the com.google.api.client packages such as VerificationCodeReceiver and Server which aren't being used here.  What's the reason?

chanseokoh: Ah, I see there is this `VerificationCodeReceiver`. I'll add that.

For `Server`, I believe it's really `org.eclipse.jetty.server.Server` (or `org.mortbay.jetty.Server` in the old code), isn't it? Their APIs are identical. I think the reason it has the package name `com.google.api.client.repackaged.org.mortbay.jetty.Server` is that someone repackages this Jetty code and other login-related jars into a single jar called `google-api-java-client-min-repacakged-1.20.0.jar`, right?

patflynn: I'm not actually familiar with the google api client library, but I do
remember that they repackage a few things, I guess as part of providing an
api and support classes for what we're doing here.

@loosebazooka may know more.

patflynn:
Also I think @nbashirbello may be the most familiar with this code.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
loosebazooka: Yeah, I would ask @nbashirbello , I have very little experience with the login code.

chanseokoh:
Regarding your [comment](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/513#issuecomment-238129752), I will go with the simpler approach (passing a `Display`) to focus on more important issues. Feel free to file a bug if you think the more elaborated approach is desired.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
chanseokoh: Not needed for both IntelliJ and Eclipse, since this method is only for entering an authorization code manually. The IDEs won't use it, but I think it makes sense for the method to exist in the ide-login library that covers all cases.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
chanseokoh: Turns out that `waitForCode()` fits well as a polling method in an async progress thread that continuously monitor if the auth code has arrived.

`runLocalServer()` on the other hand is better called in the main thread. If there was an error running a server, we halt. Otherwise, it spins up the async thread that polls the server using `waitForCode()`.

So, I think separating these two methods seems better.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
chanseokoh: Should be checkNull in this case, but there doesn't exist such a method, which makes sense because check methods throws an IllegalArgumentException. Here, I'm checking if the instance is in a proper state, so I think using checkState() makes sense, which throws IllegalStateException.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
chanseokoh: Talked to @nbashirbello and found out that [google-oauth-java-client](https://github.com/google/google-oauth-java-client) already has the [`LocalServerReceiver`](https://github.com/google/google-oauth-java-client/blob/dev/google-oauth-client-jetty/src/main/java/com/google/api/client/extensions/jetty/auth/oauth2/LocalServerReceiver.java) implementation. I spent some time to use it, but eventually I realized it has a critical problem: you are blocked on `waitForCode()` indefinitely until the server receives login result. I will not go into the details, but this must have been one of the reasons that our IntelliJ plugin duplicated the `LocalServerReceiver` code. It makes sense to do the same thing in Eclipse too, so I'll continue with the current state of `AuthorizationCodeReceiver`.

As a side note, `VerificationCodeReceiver` is YAGNI.

chanseokoh: Ultimately, the right way is to make `LocalServerReceiver` implement the features we need (cancallable `waitForCode()` and customizable success/failture landing pages) and make Android Studio, IntelliJ, and Eclipse all use `google-oauth-java-client`.

chanseokoh: Filed

- https://github.com/google/google-oauth-java-client/issues/126
- https://github.com/google/google-oauth-java-client/issues/127
- https://github.com/google/google-oauth-java-client/issues/128
- https://github.com/google/google-oauth-java-client/issues/129

chanseokoh: I found a simple workaround for the cancellation problem. It's not elegant, but I think it's way better to reuse the existing `LocalServerReceiver` and not duplicate code.

chanseokoh:
Work is still in progress. It's currently functional, but I am thinking of trying a different design.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+
+  private volatile boolean resultConsumed;
+
+  // Reads on these shared objects can happen only after server shutdown,
+  // so synchronization is not really necessary.
+  private String error;
+  private String authorizationCode;
+
+  /**
+   * Starts a local server listening through an ephemeral port.
```
elharo: through --> on

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+    Preconditions.checkState(server == null, "Don't reuse the receiver.");
+
+    server = new Server(0);
+    server.setHandler(new RequestHandler());
+    ServerConnector connector = (ServerConnector) server.getConnectors()[0];
+    connector.setHost(LOCAL_HOST);
+    try {
+      server.setStopAtShutdown(true);
+      server.start();
+      logger.log(Level.INFO, "Started a local web server waiting for OAuth2 authorization code.");
```
elharo: I doubt we need this log statement

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
elharo: checkNotNull

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+      try {
+        server.wait(timeout);
+      } catch (InterruptedException e) {}
+    }
+
+    return resultConsumed;
+  }
+
+  public void shutdown() {
+    Preconditions.checkState(server != null, "Server was never created.");
```
elharo: checkNotNull

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+   */
+  public String runLocalServer() throws IOException {
+    Preconditions.checkState(server == null, "Don't reuse the receiver.");
+
+    server = new Server(0);
+    server.setHandler(new RequestHandler());
+    ServerConnector connector = (ServerConnector) server.getConnectors()[0];
+    connector.setHost(LOCAL_HOST);
+    try {
+      server.setStopAtShutdown(true);
```
elharo: How does this interact with server.stop in the shutdown method?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -2,6 +2,8 @@ LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requi
 LOGIN_ERROR_DIALOG_MESSAGE=An error occurred while trying to sign in.
 LOGIN_ERROR_DIALOG_TITLE=Sign in Failed
 LOGIN_ERROR_IN_PROGRESS=Login is already in progress. Please try again later.
+LOGIN_ERROR_LOCAL_SERVER_INIT=Cannot start a local web server required to receive OAuth2 authorization code:\n\n
```
elharo: delete "required"

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -2,6 +2,8 @@ LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requi
...
+LOGIN_ERROR_LOCAL_SERVER_RUN=Error from a local web server while waiting for authorization code:\n\n
```
elharo: a local --> local

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/AuthorizationCodeReceiver.java
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Runs a Jetty web server instance that opens up an HTTP listening port to receive OAuth2
+ * authorization code.
+ */
+public class AuthorizationCodeReceiver {
+
+  private static final Logger logger = Logger.getLogger(AuthorizationCodeReceiver.class.getName());
+
+  // TODO: create landing pages.
```
chanseokoh: #557 created.

chanseokoh:
Fixes #503.

@elharo `AuthorizationCodeReceiver` is deleted and doesn't exist anymore. (You made comments on an outdated diff). We now use `LocalServerReceiver` from the Google OAuth Java Client Library. (The discussions that led to using `LocalServerReceiver` are [here](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/513#discussion_r74070132)).

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUiTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  private Thread createListenerThread(final ServerSocket serverSocket) {
+    return new Thread(new Runnable() {
+      @Override
+      public void run() {
+        try (
+          Socket socket = serverSocket.accept();
+          InputStreamReader reader = new InputStreamReader(socket.getInputStream());
```
elharo: please specify character encoding

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUiTest.java
```diff
@@ -0,0 +1,69 @@
...
+          OutputStreamWriter writer = new OutputStreamWriter(socket.getOutputStream());
```
elharo: please specify character encoding

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -2,8 +2,11 @@ LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requi
 LOGIN_ERROR_DIALOG_MESSAGE=An error occurred while trying to sign in.
 LOGIN_ERROR_DIALOG_TITLE=Sign in Failed
 LOGIN_ERROR_IN_PROGRESS=Login is already in progress. Please try again later.
+LOGIN_ERROR_LOCAL_SERVER_RUN=Error running a local web server listening for authorization code:\n\n
```
elharo: delete "running a local web server"

chanseokoh:
Added unit tests; this PR is now complete and really ready for review.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUiTest.java
```diff
@@ -0,0 +1,117 @@
...
+    return new Thread(new Runnable() {
+      @Override
+      public void run() {
+        try (
+          Socket socket = serverSocket.accept();
+          InputStreamReader reader = new InputStreamReader(socket.getInputStream(), "UTF-8");
+          OutputStreamWriter writer = new OutputStreamWriter(socket.getOutputStream(), "UTF-8");
+        ) {
+          String input = new String();
+          while (true) {
```
elharo: you need to test for end-of stream (-1) here; otherwise this can block indefinitely if you don't get the expected response

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -15,26 +15,51 @@
...
+import java.net.URL;
+import java.util.concurrent.Semaphore;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 public class LoginServiceUi implements UiFacade {
 
+  private static final Logger logger = Logger.getLogger(LoginServiceUi.class.getName());
+
+  private static final String ERROR_MARKER_USER_CANCELLED_LOGIN = "cancelled-by-user"; //$NON-NLS-1$
```
elharo: US-English has one l in canceled: canceled-by-user

(UK English has two)

elharo: CANCELED

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +92,173 @@ public void run() {
...
+          // Fork another sub-job to circumvent the limitation of LocalServerReceiver.
+          // (See the comments of scheduleCodeWaitingJob().)
+          scheduleCodeWaitingJob(
+              new LocalServerReceiverWrapper(codeReceiver), wait, codeHolder, exceptionHolder);
+          wait.acquire();  // Block until signaled.
+        }
+      });
+
+      if (exceptionHolder[0] != null) {
+        throw exceptionHolder[0];
```
elharo: this can't be an InvocationTargetException or  InterruptedException?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +92,173 @@ public void run() {
...
+    } catch (InvocationTargetException | InterruptedException ex) {
+      // Never thrown from the attached task.
+      return null;
+    }
+  }
+
+  /**
+   * Schedule and run a job that calls {@link LocalServerReciever#waitForCode}. The reason for
+   * creating another job inside the job of {@link showProgressDialogAndWaitForCode} is that we
+   * cannot have a 100%-guarantee that {@link LocalServerReciever#waitForCode} will eventually
```
elharo: LocalServerReceiver (i before e, except after c, o when...)

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +92,173 @@ public void run() {
...
+   * return. (If {@link stopCodeWaitingJob} fails to stop {@link LocalServerReciever#waitForCode}
```
elharo: LocalServerReceiver

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -67,22 +92,173 @@ public void run() {
...
+   * gracefully, users will be stuck and the IDE has to be killed forcibly.)
+   *
+   * However, under normal circumstances, {@link stopCodeWaitingJob} will succeed to terminate
+   * this sub-job.
+   *
+   * @return scheduled job. Returning only for unit testing
```
elharo: delete ". Returning only for unit testing" because no need to duplicate @VisibleForTesting

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUiTest.java
```diff
@@ -103,8 +103,13 @@ public void run() {
         ) {
           String input = new String();
           while (true) {
-            input += (char) reader.read();
-            if (input.endsWith("?error=cancelled-by-user")) {
+            int ch = reader.read();
+            if (ch == -1) {
+              break;
```
elharo: You can write this a little more simply as a for loop:

for (int ch = reader.read(); ch != -1; ch = reader.read())

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUiTest.java
```diff
@@ -103,8 +103,13 @@ public void run() {
         ) {
           String input = new String();
```
elharo: Consider a StringBuilder here

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUiTest.java
```diff
@@ -103,8 +103,13 @@ public void run() {
...
-            if (input.endsWith("?error=cancelled-by-user")) {
+            int ch = reader.read();
+            if (ch == -1) {
+              break;
+            }
+
+            input += (char) ch;
+            if (input.endsWith("?error=canceled-by-user")) {
               cancelRequestReceived = true;
               writer.write("HTTP/1.1 200 OK\n\n");
```
elharo: what happens if the remote is still sending data at this point?

chanseokoh: That's not a problem; the both sides of the connection can read and write simultaneously through separate input and output streams as long as they are open by the both sides.

elharo:
LGTM

PR #529, opened 2016-08-09
UI layout for deploy configuration page

akerekes:
WIP! Task 4 in #492 

This is the current layout:
image

What do you think?
The `DeployPropertyPage#createContents()` method will be split up into smaller methods once the editor layout has been finalized.


elharo:
I think there should be colons after the labels, though I'm not 100% sure of that. Otherwise looks good.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,182 @@
...
+    public void widgetSelected(SelectionEvent e) {
+      doSelection(e);
+    }
+
+    @Override
+    public void widgetDefaultSelected(SelectionEvent e) {
+      doSelection(e);
+    }
+
+    private void doSelection(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,182 @@
...
+      this.button = button;
+    }
+
+    @Override
+    public void widgetSelected(SelectionEvent e) {
+      doSelection(e);
+    }
+
+    @Override
+    public void widgetDefaultSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,182 @@
...
+    private Text text;
+    private Button button;
+
+    public RadioSelectionListener(Text text, Button button) {
+      this.text = text;
+      this.button = button;
+    }
+
+    @Override
+    public void widgetSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,182 @@
...
+  private Button customVersionButton;
+  private Button autoPromoteButton;
+  private Button manualPromoteButton;
+  private Button defaultBucketButton;
+  private Button customBucketButton;
+
+  private RadioSelectionListener versionSelectionListener;
+  private RadioSelectionListener bucketSelectionListener;
+
+  public class RadioSelectionListener implements SelectionListener {
```
elharo: This might work a little better as an independent outer class, and could be tested as such.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,182 @@
...
+      text.setEnabled(e.getSource() == button);
+    }
+  }
+
   @Override
   protected Control createContents(Composite parent) {
-    Label label = new Label(parent, SWT.LEFT);
-    label.setText(Messages.getString("preferences.title")); //$NON-NLS-1$
-    return label;
+    Composite container = new Composite(parent, SWT.NONE);
```
elharo: This method is very long. Is it feasible to split into smaller helper methods, purely for clarity?

akerekes:
How about this layout?

image

chanseokoh:
That looks pretty. The staging "bucket" is actually a directory, isn't it?

briandealwis:
My feedback: I'm not sure the groups really helps.  
  - Project ID should indicate that a blank field will prompt on deploy; or perhaps there should be a check _Prompt for Project ID on deploy_, and if not blank then the text field is enabled.  Something like the "Use default location" checkbox on New Project-style wizards
  - I think you could turn the "version" and "promote" radio button sets into a single checkbox with a controlled text field, like the "Use default location"
   - The version could be turned into a single checkbox "_Automatically generate a new version number for each deploy_" with a controlled field: _Version_. Or alternatively _Explicitly specify a version for the deploy_?
   - The promotion could be turned into a single checkbox "_Promote the deployed version to receive all traffic_"

Would it be useful to have a link to open the Google Dev Console?

akerekes:
@chanseokoh no, it's not referring to the local directory we use to create the staged application, but the Google Cloud Storage bucket used by the deploy process to upload the application.

akerekes:
@briandealwis I like the groups here, but I'm down for your suggestion of using checkboxes rather, and then groups can go away.

For project ID I'd go with the checkbox to indicate the prompt during deploy.

For version I'd rather have the checkbox say: "Use custom version" or "Override generated value" and allow the user to specify it (similar to java compiler settings where you can override the general values with project specific settings).

Add a link that leads to the versions page of the console is a good idea. But then we need to know the project ID as well to make it correct, otherwise it will just open the default (probably last visited) project.

Would you get rid of the bucket group+radio and turn it into a checkbox+field?

elharo:
sounds good. I like the checkboxes idea. I'm not sure we need "Project ID should indicate that a blank field will prompt on deploy."

These all work, so let's not overthink this. Let's get something out there that works and then iterate.

akerekes:
@elharo @briandealwis @chanseokoh PTAL

New screenshot (this time from Mac OS):

image

elharo:
Looking good. Some thoughts:

1. I'm not sure we need the advanced section, especially with only one entry.
2. The fonts look differently sized by a point or two. This seems to be a common SWT issue. You may need to call Dialog.applyDialogFont(container) to make them uniform.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+    container.setLayout(new GridLayout(1, false));
+
+    createProjectIdSection(container);
+
+    createProjectVersionSection(container);
+
+    createPromoteSection(container);
+
+    createAdvancedSection(container);
+
```
elharo: call    
here to fix up the font sizing

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+    projectIdComp.setLayout(new GridLayout(3, false));
+    projectIdComp.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+  
+    promptForProjectIdButton = new Button(projectIdComp, SWT.CHECK);
+    promptForProjectIdButton.setText(Messages.getString("deploy.prompt.projectid"));
+    promptForProjectIdButton.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false, 3, 1));
+    promptForProjectIdButton.addSelectionListener(new SelectionListener() {
+      
+      @Override
+      public void widgetSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+        updateProjectIdWidgetsEnablement();
+      }
+      
+      @Override
+      public void widgetDefaultSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+    layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
+    versionLabel.setLayoutData(layoutData);
+  
+    version = new Text(versionComp, SWT.LEFT | SWT.SINGLE | SWT.BORDER);
+    version.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+  
+    overrideDefaultVersionButton.addSelectionListener(new SelectionListener() {
+      
+      @Override
+      public void widgetSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+        updateVersionWidgetsEnablement();
+      }
+      
+      @Override
+      public void widgetDefaultSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+    
+    Link manualPromoteLabel = new Link(promoteComp, SWT.NONE);
+    GridData layoutData = new GridData(SWT.LEFT, SWT.CENTER, true, false);
+    layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
+    manualPromoteLabel.setLayoutData(layoutData);
+    manualPromoteLabel.setText(Messages.getString("deploy.manual.link", APPENGINE_DASHBOARD_URL));
+    manualPromoteLabel.addSelectionListener(new SelectionListener() {
+
+      @Override
+      public void widgetSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+        openAppEngineDashboard();
+      }
+
+      @Override
+      public void widgetDefaultSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+    Composite defaultBucketComp = new Composite(parent, SWT.NONE);
+    defaultBucketComp.setLayout(new GridLayout(1, true));
+    
+    overrideDefaultBucketButton = new Button(defaultBucketComp, SWT.CHECK);
+    overrideDefaultBucketButton.setText(Messages.getString("use.custom.bucket"));
+    overrideDefaultBucketButton.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false));
+    overrideDefaultBucketButton.addSelectionListener(new SelectionListener() {
+      
+      @Override
+      public void widgetSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+        updateCustomBucketWidgetsEnablement();
+      }
+      
+      @Override
+      public void widgetDefaultSelected(SelectionEvent e) {
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,258 @@
...
+
+import com.google.cloud.tools.eclipse.ui.util.FontUtil;
 
 public class DeployPropertyPage extends PropertyPage implements IWorkbenchPropertyPage {
 
+  private static final String APPENGINE_DASHBOARD_URL = "https://console.developers.google.com/appengine";
+  private static final int INDENT_CHECKBOX_ENABLED_WIDGET = 10;
+
+  private Button promptForProjectIdButton;
+  private Label projectIdLabel;
```
elharo: I think the three labels can all be moved to local variables rather than fields

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/RadioSelectionListener.java
```diff
@@ -0,0 +1,31 @@
+package com.google.cloud.tools.eclipse.ui.util;
+
+import org.eclipse.swt.events.SelectionEvent;
+import org.eclipse.swt.events.SelectionListener;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Control;
+
+public class RadioSelectionListener<C extends Control> implements SelectionListener {
```
elharo: I don't see where this class is used

akerekes:
The reason behind the separate advanced section is that there are a few
more options we could expose around jar splitting, env_vars, and jsp
handling.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: If they're fields, it's easier to call `setEnabled()` on them. On Mac
there's not much difference, but on Linux they also greyed out, something
that's consistent with the other preference pages (i.e. disable labels as
well as input widgets).

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/RadioSelectionListener.java
akerekes: RadioSelectionListener is not used anymore, I forgot to remove it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: +1 It's good to mark them as disabled.  I wonder why it's not supported on OS X.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: `#applyDialogFont()` is a bit strange: it only applies the font if the dialog font is different from the 'default' font.  Which makes sense, providing all widgets use the default font.  But `Link` doesn't choose the default font.

briandealwis:
Looks good!

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: There's a long history on [bug 41245](https://bugs.eclipse.org/bugs/show_bug.cgi?id=41245): they experimented having fonts applied by default, but most platforms don't have a single default font (e.g., `Group` uses a slightly smaller font by design), and many fonts are set by OS-level themes.  And changing fonts after creation often throws off layouts.  The recommendation from SWT was to set fonts manually and to not do anything magical.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -1,2 +1,14 @@
...
\ No newline at end of file
+# Global settings
+google.preferences=Google Cloud Platform-related settings
+
+# Deploy settings
+auto.promote=Promote the deployed version to receive all traffic
+bucket.name=Bucket:
+cannot.open.browser=Cannot open browser: {0}
+deploy.prompt.projectid=Prompt for Project ID on deploy
+deploy.manual.link=To manually promote a version use <a href="{0}">Google Developer Console</a>
```
elharo: To manually promote a version use the <a href="{0}">Google Cloud Console</a>.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,261 @@
...
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
 import org.eclipse.ui.dialogs.PropertyPage;
+import org.eclipse.ui.forms.widgets.ExpandableComposite;
+
+import com.google.cloud.tools.eclipse.ui.util.FontUtil;
 
 public class DeployPropertyPage extends PropertyPage implements IWorkbenchPropertyPage {
 
+  private static final String APPENGINE_DASHBOARD_URL = "https://console.developers.google.com/appengine";
```
elharo: is this the same as https://console.cloud.google.com/ ?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -1,19 +1,261 @@
...
 import org.eclipse.ui.IWorkbenchPropertyPage;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
 import org.eclipse.ui.dialogs.PropertyPage;
+import org.eclipse.ui.forms.widgets.ExpandableComposite;
+
+import com.google.cloud.tools.eclipse.ui.util.FontUtil;
 
 public class DeployPropertyPage extends PropertyPage implements IWorkbenchPropertyPage {
```
chanseokoh: Just noticed now: `PropertyPage` already implements `IWorkbenchPropertyPage`, so remove the latter?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: It redirects there. I'll use `console.cloud...`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: Done

elharo:
LGTM

Issue #523, opened 2016-08-09
Logout always puts up a failure dialog


elharo:
As near as I can tell, logout always pops up a failure dialog, even if it succeeds.

chanseokoh:
Invalidated by #530.

Issue #455, opened 2016-07-26
Rename GoogleLoginService methods


akerekes:
Some of the methods do not convey clearly what they are intended to do. E.g. `getCachedActiveCredentials` is basically returns the active user's credential if logged in. Maybe calling it `getActiveLoggedInUserCredential` would be better (although it's slightly more verbose).
Also `getActiveCredential` gets the active logged in user's credential and requests the user to log in case there's no logged in user. Here I don't have a suggestion currently, but I believe we could come up with something more descriptive.

And as a bonus: it would be useful to add a method `isLoggedIn` that returns a boolean by executing `getCachedActiveCredential() != null`. It would help the readability of the code.

chanseokoh:
This issue is no longer relevant, as we don't have the notion of an active account. Closing.

PR #530, opened 2016-08-09
Remove the manual login test that uses "gcloud projects list"

### LINKED ISSUE #519: Failure (see console) does not send output to console
On login with failure:

failure

Note that this bug is not the failure itself but that there is no message in the console

chanseokoh:
This invalidates and closes #519 #523.


plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -22,18 +25,24 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
     Credential credential = loginService.getCachedActiveCredential();
     if (credential == null) {
       credential = loginService.getActiveCredential();
+
+      if (credential != null) {
+        try {
+          new GoogleLoginTemporaryTester().testLogin(credential);
```
elharo: Now that we have deployment, we should be getting rid of this temporary tester class.

chanseokoh: Right, that sounds good.

chanseokoh:
I changed the title of the PR. Now is the time to remove the manual login test that does "gcloud projects login".

(Note: Travis will fail with an transient error.)

With this change, you might think the methods `getCachedActiveCredential()` and `getActiveCredential()` have bad names; we opened #455.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -21,19 +21,14 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
 
     Credential credential = loginService.getCachedActiveCredential();
```
elharo: This code could use some comments explaining how it works. (I.e. pressing the button logs you out if you're already logged in; otherwise it logs you in.) I see it now, but I can imagine it's going to confuse me in 6 months. 

I do wonder if we need a different UI, e.g. swapping out the button; but that can wait until we design multilogin.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
```diff
@@ -21,19 +21,14 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
     if (credential == null) {
-      credential = loginService.getActiveCredential();
+      loginService.getActiveCredential();
```
elharo: I wonder if getActiveCredential might be better named loadCredential or just login. Generally to me a getFoo method suggests that the method merely returns a value, not performs an action.

elharo:
Why is travis failing?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java
chanseokoh: I agree. We have #455 to rename `getActiveCredential()` and other methods.

chanseokoh:
There was a transient problem on the eclipse mars repo last night that made local and Travis builds to fail. It should build fine now.

elharo:
LGTM

Issue #514, opened 2016-08-07
Send Analytics pings asynchronously


chanseokoh:
In the very unlikely event of the Google Analytics backend not listening on the HTTP endpoint, `AnalyticsPingManager().sendPing()` may block for 3 seconds trying to making a connection.

chanseokoh:
Closed by #540.

Issue #359, opened 2016-07-13
Add/fix license headers to source code


chanseokoh:
Many files are missing the license header or have wrong license (Eclipse license rather than Apache).

It'd be also nice to configure the workspace to automate adding the license header.

elharo:
fixed now, I think

PR #540, opened 2016-08-10
Track local run usage / allow sending Analytics pings from non-UI contexts

### LINKED ISSUE #538: Analytics tracking for Local Run and Debug
We probably want separate counters for run and debug launches.

chanseokoh:
Fixes #538.

I realized there should be a way to call `sendPing()` from non-UI contexts. In the case where the caller is not a UI-thread (i.e., `Display.getCurrent()` is null), the ping logic is wrapped inside `Display.asyncExec()`.


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -66,7 +68,9 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
 
     setDefaultSourceLocator(launch, configuration);
 
-    if (ILaunchManager.DEBUG_MODE.equals(mode)) {
+    boolean debugMode = ILaunchManager.DEBUG_MODE.equals(mode);
```
briandealwis: This isn't safe: the mode can be "run", "debug", "profile" — but others can be defined too.  Why not just log the "mode"?

briandealwis: (I'd personally be curious to know what other modes people try. Such as "profile".)

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -133,14 +133,30 @@ void registerOptInStatus(boolean optedIn) {
...
+   * Safe to call from non-UI contexts.
    */
   public void sendPing(String eventName, String metadataKey, String metadataValue) {
     sendPing(eventName, metadataKey, metadataValue, null);
   }
 
-  public void sendPing(String eventName, String metadataKey, String metadataValue,
+  public void sendPing(final String eventName,
+      final String metadataKey, final String metadataValue, final Shell parentShell) {
+    if (Display.getCurrent() != null) {
```
briandealwis: Do you need to get the `Display` when you're passed a `Shell`?  `parentShell.getDisplay()`

And this is definitely a place that warrants `PlatformUI.getWorkbench().getDisplay()`.  `Display#getDefault()` is never good.

chanseokoh: Actually, what I really wanted to know was that if I am in the UI thread or not. (I thought `Display.getCurrent() != null()` served the purpose.) It's about whether I should wrap the main body inside `Display.asyncExec()` or not, to avoid the SWT thread access issue.

I could certainly use `PlatformUI.getWorkbench().getDisplay()`. (Initially, I got the idea of `Display#getDefault()` from [here](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/Activator.java#L68)).

chanseokoh:
Changed the key from "type" to "mode". Records the mode string directly. I see values of "mode=debug" and "mode=run" for now.

`PlatformUI.getWorkbench().getDisplay()` works fine. This is called for local run event pings.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -133,14 +133,30 @@ void registerOptInStatus(boolean optedIn) {
...
+   * Safe to call from non-UI contexts.
    */
   public void sendPing(String eventName, String metadataKey, String metadataValue) {
     sendPing(eventName, metadataKey, metadataValue, null);
   }
 
-  public void sendPing(String eventName, String metadataKey, String metadataValue,
+  public void sendPing(final String eventName,
+      final String metadataKey, final String metadataValue, final Shell parentShell) {
+    if (Display.getCurrent() != null) {  // Check if we are in the UI thread.
```
briandealwis: Proper check then is `display.getThread() == Thread.currentThread()`.

But do we really want to have the pings *always* happen on the UI thread?  I'd have thought we'd want the pings to always happen on a non-UI thread (e.g., via a Job), unless we have to prompt the user.

chanseokoh: What do you think? Currently, `sendPing()` may potentially show the opt-in dialog (if a user has never seen it). That's the only reason we want the UI thread. I had thought about a few alternatives before:

1. For headless jobs (e.g., local run), add something like `sendPingUiFree()` which never pops up dialog.
2. Forget about UI vs. non-UI inside `AnalyticsPingManager`, and have callers responsible for wrapping `sendPing()` with `Display#asyncExec()` if calling from headless jobs.
3. Check if we are in the UI thread, and if not, never pops up the dialog.

The approach in this PR is to liberate any concern from callers about UI vs. non-UI and always be able to show the dialog if necessary. Suggestions?

elharo:
Is it possible to use the UI thread to pop up the dialog in an async job if necessary, and otherwise not use the UI thread? I.e. have a sendPing method that first checks if we've opted in and if that is unknown only then tries to do something in the UI thread?

chanseokoh:
@briandealwis suggested a simple event-ping-queuing framework and actual sent me a sketch of it to take care of the issues from the opt-in dialog, the issue of #514, and some other issues.

chanseokoh:
- Implemented a small ping-queuing framework that @briandealwis suggested. This fixes #514 too.
- Added new unit tests accordingly.
- `OptInDialogCreator`, added previously only for unit testing, is removed.

@briandealwis, FYI, as discussed in the email, this doesn't solve the multiple-concurrent-dialog issue or the issue of pings being ignored until a user makes a choice. (We can ignore the latter issue though.)

chanseokoh:
Ready for review.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -251,6 +252,9 @@ static String getParametersString(Map<String, String> parametersMap) {
     return resultBuilder.toString();
   }
 
+  // To prevent showing multiple opt-in dialogs. See OptInDialog#open() and OptInDialog#close().
```
briandealwis: Another thing to note here is that we treat dialog-close as opt-out rather than unknown.  So we have an answer once we've shown the dialog once and thus don't need to restore the dialogOpenStatus to false.  Maybe `dialogOpenedStatus`?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -259,7 +263,7 @@ public void showOptInDialog(final Shell parentShell) {
       display.syncExec(new Runnable() {
         @Override
         public void run() {
-          new OptInDialog(findShell(parentShell)).open();
+          new OptInDialog(findShell(parentShell), dialogOpenStatus).open();
```
briandealwis: It seems weird to pass this in to the dialog: why is it the dialog's responsibility to manage itself?  Is this for testing? Why not:
```
if (dialogOpenStatus.compareAndSet(false, true)) {
   display.syncExec(…);
}
```
While we're here, the `show` in `showOptInDialog` doesn't convey that it is conditional.  `verifyOptIn` or `verifyConsent` or something?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -62,14 +68,27 @@
 
   // Preference store (should be configuration scoped) from which we get UUID, opt-in status, etc.
   private IEclipsePreferences preferences;
-
-  private OptInDialogCreator optInDialogCreator;
+  private Display display;
+
+  private ConcurrentLinkedQueue<PingEvent> pingEventQueue;
```
akerekes: Is it necessary to declare it `ConcurrentLinkedQueue`? Cannot it be simply `Queue`? Of course it will be instantiated as `ConcurrentLinkedQueue`. If `Queue` is used, then the `@VisibleForTesting` constructor can also use Queue as parameter.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: > why is it the dialog's responsibility to manage itself?

I thought about that too. If you think it's okay to set `dialogOpenStatus` only once and don't care about resetting it back to false when the dialog closes, I'll go with that. It should work.

I'll also chance the method name. Good point. `verifyAndAskConsent`? I'd want to convey that a dialog can be open.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -70,6 +76,10 @@
         }
       });
 
+  @Mock IEclipsePreferences preferences;
+  @Mock private Display display;
+  @Spy private ConcurrentLinkedQueue<PingEvent> pingEventQueue;
```
akerekes: Is it expected that the actual implementation of methods is used in the test, or could this be a @Mock too?

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -70,6 +76,10 @@
         }
       });
 
+  @Mock IEclipsePreferences preferences;
```
akerekes: private

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: I'm afraid I can't. `sendPing()` adds an element to the queue, and the methods can be called concurrently.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Ah, I remembered why I did this weird thing to make the OptInDialog manage the flag. We have this assumption that once a dialog is shown then the choice is made even though the dialog is just closed (e.g., a user just escape it). In this case, we don't need to care to restore the flag. However, once we chance our policy and decide to show the dialog again if a user didn't explicitly made a choice, this won't work.

So, I'll let you make the final decide. Are we good to set the flag only once with the name `dialogOpenedStatus`?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -229,20 +252,28 @@ static String getParametersString(Map<String, String> parametersMap) {
...
   }
 
+  // To prevent showing multiple opt-in dialogs. See OptInDialog#open() and OptInDialog#close().
+  private AtomicBoolean dialogOpenStatus = new AtomicBoolean(false);
+
   /**
    * @param parentShell if null, tries to show the dialog at the workbench level.
    */
-  public void showOptInDialog(Shell parentShell) {
+  public void showOptInDialog(final Shell parentShell) {
```
akerekes: rename to `showOptInDialogIfNeeded`?

chanseokoh: I like it.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
akerekes: But the concurrent calls would be handled by the implementation. What I propose is similar to:
`Queue<E> pingQueue = new ConcurrentLinkedQueue<E>()`

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
akerekes: I'd rather ask the user each time until he decides to opt in or out.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Aha!

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: There was a bit of a discussion around this, and we decided that implicit closure of a dialog is considered opting out. (So, a user will see the opt-in dialog only once.)

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Using an interface is usually nice, but on second thought, I think it's better to explicitly say here that the queue required should be a thread-safe one, to prevent getting a plain `Queue`. I tried to see if there is any other interface such as `ConcurrentQueue`, but there doesn't seem to.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: I guess not restoring the flag is good as long as it's documented. Simplified a bit.

briandealwis: Agreed: just document it

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -1,3 +1,18 @@
+/*******************************************************************************
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * All rights reserved. This program and the accompanying materials are made
```
briandealwis: Umm, hmm.  I think everything in CTE is supposed to be under the Apache license.

chanseokoh: Come to think of it, that sounds about right. I'll forget about the headers for now. (We have #359.)

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -163,6 +187,7 @@ private void sendPostRequest(String parametersString) {
       connection = (HttpURLConnection) url.openConnection();
       connection.setDoOutput(true);
       connection.setRequestMethod("POST");
+      connection.setRequestProperty("User-Agent", "Automated");
```
elharo: named constant; and we're probably going to want something else here

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -266,4 +301,19 @@ private void flushPreferences() {
...
+  @VisibleForTesting
+  static class PingEvent {
+    public PingEvent(String eventName, String metadataKey, String metadataValue, Shell shell) {
+      this.eventName = eventName;
+      this.metadataKey = metadataKey;
+      this.metadataValue = metadataValue;
+      this.shell = shell;
+    }
+
+    private String eventName;
```
elharo: fields usually precede the constructor

elharo:
tests are failing :-(

chanseokoh:
Let me try rerunning the tests.

For the record, the test failure is

```
Tests run: 3, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 59.972 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.integration.appengine.NewNativeAppEngineStandardProjectTest
testWithPackage(com.google.cloud.tools.eclipse.integration.appengine.NewNativeAppEngineStandardProjectTest)  Time elapsed: 11.729 sec  <<< ERROR!
org.eclipse.swtbot.swt.finder.exceptions.WidgetNotFoundException: Could not find widget matching: (of type 'Text' and with label (with mnemonic 'Project name:'))
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:522)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:496)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:484)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntilWidgetAppears(SWTBotFactory.java:466)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.widget(SWTBotFactory.java:419)
	at org.eclipse.swtbot.swt.finder.SWTBot.textWithLabel(SWTBot.java:1478)
	at org.eclipse.swtbot.swt.finder.SWTBot.textWithLabel(SWTBot.java:1466)
	at com.google.cloud.tools.eclipse.integration.appengine.SwtBotAppEngineActions.createNativeWebAppProject(SwtBotAppEngineActions.java:61)
	at com.google.cloud.tools.eclipse.integration.appengine.NewNativeAppEngineStandardProjectTest.createAndCheck(NewNativeAppEngineStandardProjectTest.java:73)
	at com.google.cloud.tools.eclipse.integration.appengine.NewNativeAppEngineStandardProjectTest.testWithPackage(NewNativeAppEngineStandardProjectTest.java:54)
```

chanseokoh:
I'm trying to see if the previous merge that disabled Analytics opt-in dialog will remove the flakiness of the test. Not sure what happened exactly, but I succeeded with local testing after that merge.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
```diff
@@ -26,6 +26,8 @@
   // Don't change the value; this name is used as an originating "application" of usage metrics.
   public static String METRICS_NAME = "gcloud-eclipse-tools";
 
+  public static String ANALYTICS_USER_AGENT = METRICS_NAME + getToolsVersion();
```
elharo: Just USER_AGENT would do. We'll probably want to use this name for non-analytics things too.

elharo: You need a forward slash before the version. Please add a test for this format (I.e. that the constant is what we expect) since the specific format is required by the HTTP spec.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/CloudToolsInfoTest.java
```diff
@@ -24,4 +24,9 @@
   public void testGetToolsVersion() {
     Assert.assertTrue(CloudToolsInfo.getToolsVersion().startsWith("0.1.0."));
   }
+
+  @Test
+  public void testUserAgent() {
+    Assert.assertTrue(CloudToolsInfo.USER_AGENT.startsWith("gcloud-eclipse-tools/0.1.0"));
```
elharo: .equals("gcloud-eclipse-tools/0.1.0")

elharo:
LGTM iff tests pass on travis

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/CloudToolsInfoTest.java
chanseokoh: https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/468#discussion_r72632320

The value is actually something like `gcloud-eclipse-tools/0.1.0.201608160330` (if ran with `mvn verify` on the command-line). For this reason, the other test `testGetToolsVersion()` also uses `startsWith()`.

Issue #365, opened 2016-07-15
Deploy configuration


akerekes:
The options passed into deploy should be configurable by the user.
Current options include:

- ~~bucket~~
- ~~promote~~
- stop-previous-version
- image-url

akerekes:
#610 covers `stop-previous-version`, and `image-url` is used in App Engine Flex, so this issue is obsolete.

PR #554, opened 2016-08-11
Add databinding to deploy properties page

akerekes:
WIP

Preferences are stored on the project level (so different projects in the same workspace can have different deploy settings).

Databinding is setup using `org.eclipse.core.databinding.observable.value.WritableValue` and model class. Validation and error reporting is to be implemented in this PR in future commits.


briandealwis:
Your Model object can be simplified and be bean-like with getters and setters.  You instead use `PojoProperties` (1-way only: UI/target &rarr; model) or `BeanProperties` (requires property-change support in the model) to generate observables and wire them in.  For example:
```
dbc.bindValue(WidgetProperties.selection().observe(promptForProjectIdButton),
    PojoProperties.observe(model, "promptForProjectId"));
```
You can also wire up SWT-observables to other SWT-observables.  So you can join enablement to checkboxes with:
```
dbc.bindValue(WidgetProperties.selection().observe(promptForProjectIdButton), 
   WidgetProperties.enabled().observe(projectId),
   new UpdateValueStrategy().setConverter(BooleanConverter.negate()), // source -> target
   new UpdateValueStrategy().setConverter(BooleanConverter.negate()) // target -> source
);
```
The `UpdateValueStrategy` object encapsulate the validation and conversion logic.  That `BooleanConverter` doesn't exist unfortunately, but it would look something like:
```
public class abstract BooleanConverter extends Converter {
   public static BooleanConverter negate() {
      return new BooleanConverter() {
         public Object convert(Object fromObject) {
            return !(Boolean)fromObject;
         }
      };
   }
   protected BooleanConverter() {
      super(Boolean.class, Boolean.class);
   }
}
```
You would define a string validator to ensure the Project ID is in a valid form.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -140,7 +230,7 @@ private void createPromoteSection(Composite parent) {
     autoPromoteButton = new Button(promoteComp, SWT.CHECK);
     autoPromoteButton.setText(Messages.getString("auto.promote"));
     autoPromoteButton.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false));
-    
+
     Link manualPromoteLabel = new Link(promoteComp, SWT.NONE);
     GridData layoutData = new GridData(SWT.LEFT, SWT.CENTER, true, false);
     layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
```
briandealwis: You'll want to set `manualPromoteLabel`'s font to its parent's font.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -140,7 +230,7 @@ private void createPromoteSection(Composite parent) {
     autoPromoteButton = new Button(promoteComp, SWT.CHECK);
     autoPromoteButton.setText(Messages.getString("auto.promote"));
     autoPromoteButton.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false));
-    
+
     Link manualPromoteLabel = new Link(promoteComp, SWT.NONE);
```
briandealwis: It may be worth extracting the selection listener and turn it into a utility class in .ui.util that opens a browser on the `SelectionEvent`'s `text` field (the value of the `href` attribute).

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -20,30 +29,45 @@
...
 
 public class DeployPropertyPage extends PropertyPage {
 
   private static final String APPENGINE_DASHBOARD_URL = "https://console.cloud.google.com/appengine";
   private static final int INDENT_CHECKBOX_ENABLED_WIDGET = 10;
 
+  private static final String PREFERENCE_STORE_QUALIFIER = "com.google.cloud.tools.eclipse.appengine.deploy";
+  private static final String PREF_PROMPT_FOR_PROJECT_ID = "project.id.prompt"; // boolean
+  private static final String PREF_PROJECT_ID = "project.id";
+  private static final String PREF_OVERRIDE_DEFAULT_VERSIONING = "project.version.default"; // boolean
```
chanseokoh: How about `project.version.use.default`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -20,30 +29,45 @@
...
+  private static final String PREF_CUSTOM_VERSION = "project.version";
+  private static final String PREF_ENABLE_AUTO_PROMOTE = "project.promote"; // boolean
+  private static final String PREF_OVERRIDE_DEFAULT_BUCKET = "project.bucket.default"; // boolean
```
chanseokoh: How about `project.bucket.use.default`?

akerekes:
Still WIP: need to add tests, validation for bucket name and figure out what resizes the window after it's opened the second time.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java

elharo:
The CLA issue is probably just because this switched from @akerekes to @briandealwis

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,35 +33,64 @@
...
 
-import com.google.cloud.tools.eclipse.ui.util.FontUtil;
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class DeployPropertyPage extends PropertyPage {
 
-  private static final String APPENGINE_DASHBOARD_URL = "https://console.cloud.google.com/appengine";
+  public static class VCL implements IValueChangeListener {
```
elharo: VCL is ?
Can we use a full name?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,35 +33,64 @@
...
+    private String name;
+
+    public VCL(String name) {
+      this.name = name;
+    }
+
+    @Override
+    public void handleValueChange(ValueChangeEvent event) {
+      System.out.println(
```
elharo: shouldn't need this. What do we ultimately want to do here?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,35 +33,64 @@
...
+          "VCL[" + name + "]: " + event.diff.getOldValue() + " -> " + event.diff.getNewValue());
+    }
+  }
+
+  private static final String APPENGINE_DASHBOARD_URL =
```
elharo: This URL also appears in gcloud-eclipse-tools/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java and may change in the future. Perhaps move to gcloud-eclipse-tools/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +103,198 @@ protected Control createContents(Composite parent) {
...
+    setupAutoPromoteDataBinding(bindingContext);
+    setupBucketDataBinding(bindingContext);
+
+    PreferencePageSupport.create(this, bindingContext);
+    bindingContext.updateTargets();
+    observables = new ObservablesManager();
+    observables.addObservablesFromContext(bindingContext, true, true);
+  }
+
+  private void setupProjectIdDataBinding(DataBindingContext dbc) {
```
elharo: dbc --> context

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +103,198 @@ protected Control createContents(Composite parent) {
...
+    final ISWTObservableValue overrideObservable =
+        WidgetProperties.selection().observe(overrideProjectIdButton);
+    final ISWTObservableValue projectIdObservable =
+        WidgetProperties.text(SWT.Modify).observe(projectId);
+
+    dbc.addValidationStatusProvider(new MultiValidator(overrideObservable.getRealm()) {
+      @Override
+      protected IStatus validate() {
+        if (overrideObservable.getValue() == Boolean.FALSE) {
```
elharo: I seem to recall that == Boolean.FALSE is an anitpattern for null reasons. Maybe 
!overrideObservable.getValue()

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +103,198 @@ protected Control createContents(Composite parent) {
...
+          System.out.println("VALIDATE: OK");
```
elharo: please remove all printlns

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +103,198 @@ protected Control createContents(Composite parent) {
...
+    dbc.bindValue(WidgetProperties.enabled().observe(versionLabel),
+        model.overrideDefaultVersioning);
+    dbc.bindValue(WidgetProperties.enabled().observe(version), model.overrideDefaultVersioning);
+
+    dbc.addValidationStatusProvider(new MultiValidator(overrideObservable.getRealm()) {
       @Override
-      public void widgetSelected(SelectionEvent event) {
-        updateProjectIdWidgetsEnablement();
+      protected IStatus validate() {
+        if (overrideObservable.getValue() == Boolean.FALSE) {
```
elharo: !overrideObservable.getValue()

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +103,198 @@ protected Control createContents(Composite parent) {
...
+    final ISWTObservableValue overrideObservable =
+        WidgetProperties.selection().observe(overrideDefaultBucketButton);
+    final ISWTObservableValue bucketObservable = WidgetProperties.text(SWT.Modify).observe(bucket);
+
+    dbc.addValidationStatusProvider(new MultiValidator(overrideObservable.getRealm()) {
       @Override
-      public void widgetDefaultSelected(SelectionEvent event) {
-        updateProjectIdWidgetsEnablement();
+      protected IStatus validate() {
+        if (overrideObservable.getValue() == Boolean.FALSE) {
```
elharo: !overrideObservable.getValue()

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -188,73 +340,103 @@ private ExpandableComposite createExpandableComposite(Composite parent) {
...
+    public String getProjectId() {
+      return (String) projectId.getValue();
+    }
+
+    public void setProjectId(String projectId) {
+      this.projectId.setValue(projectId);
+    }
+
+    public boolean isOverrideDefaultVersioning() {
+      return Boolean.TRUE == overrideDefaultVersioning.getValue();
```
elharo: return overrideDefaultVersioning.getValue()

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -188,73 +340,103 @@ private ExpandableComposite createExpandableComposite(Composite parent) {
...
+
+    public void setVersion(String version) {
+      this.version.setValue(version);
+    }
 
-  private void initBucketButtonSelection() {
-    overrideDefaultBucketButton.setSelection(false);
-    updateCustomBucketWidgetsEnablement();
+    public boolean isAutoPromote() {
+      return (boolean) autoPromote.getValue();
```
elharo: autoPromote.getValue().booleanValue() (no cast)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: Oops, this was a debugging helper that should be removed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: `getValue()` returns an `Object`

chanseokoh: I think the issue was that if someone accidentally creates `new Boolean(b)` (which should be avoided), then `==` or `!=` will give different results than `.equals()` will do.

http://findbugs.sourceforge.net/bugDescriptions.html#RC_REF_COMPARISON_BAD_PRACTICE_BOOLEAN

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +103,198 @@ protected Control createContents(Composite parent) {
...
+    setupBucketDataBinding(bindingContext);
+
+    PreferencePageSupport.create(this, bindingContext);
+    bindingContext.updateTargets();
+    observables = new ObservablesManager();
+    observables.addObservablesFromContext(bindingContext, true, true);
+  }
+
+  private void setupProjectIdDataBinding(DataBindingContext dbc) {
+    final ISWTObservableValue overrideObservable =
```
elharo: Can we do ISWTObservableValue<Boolean>?

The superinterface IObservable is generic. I'm not sure how that passes down to the child.

akerekes: The child interface is not generic:

http://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fjface%2Fdatabinding%2Fswt%2FISWTObservableValue.html

akerekes:
@briandealwis @elharo PTAL

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
```diff
@@ -0,0 +1,74 @@
...
+    // It cannot begin with the prefix ah- and the names default and latest are
+    // reserved and cannot be used."
+    if (!(input instanceof String)) {
+      return ValidationStatus.error(Messages.getString("project.version.invalid")); //$NON-NLS-1$
+    }
+    String value = (String) input;
+    if (value.isEmpty()) {
+      return ValidationStatus.warning(Messages.getString("project.version.invalid")); //$NON-NLS-1$
+    } else if (APPENGINE_PROJECT_VERSION_PATTERN.matcher(value).matches() 
+               && !value.endsWith("-")) {
```
chanseokoh: Why not make this check into the pattern itself? The name `APPENGINE_PROJECT_VERSION_PATTERN` is a bit misleading in this sense.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+  
+  private boolean isReservedWord(String value) {
+    for (String reserved : RESERVED_VALUES) {
+      if (reserved.equals(value)) {
+        return true;
+      }
+    }
+    return false;
```
chanseokoh: How about something like
```java
return Arrays.asList(RESERVED_VALUES).contains(value);
```
?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java
```diff
@@ -0,0 +1,34 @@
...
+
+  @Override
+  public IStatus validate(Object input) {
+    if (!(input instanceof String)) {
+      return ValidationStatus.error(Messages.getString("project.id.invalid")); //$NON-NLS-1$
+    }
+    String value = (String) input;
+    if (value.isEmpty()) {
+      return ValidationStatus.warning(Messages.getString("project.id.invalid")); //$NON-NLS-1$
+    } else if (APPENGINE_PROJECTID_PATTERN.matcher((String) value).matches()) {
```
chanseokoh: `(String) value`: casting doesn't look necessary.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BucketNameValidator.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  public IStatus validate(Object input) {
+    if (!(input instanceof String)) {
+      return ValidationStatus.error(Messages.getString("bucket.name.invalid")); //$NON-NLS-1$
+    }
+    String value = (String) input;
+    if (value.isEmpty()) {
+      return ValidationStatus.warning(Messages.getString("bucket.name.invalid")); //$NON-NLS-1$
+    } else if (CLOUD_STORAGE_BUCKET_NAME_PATTERN.matcher((String) value).matches()) {
```
chanseokoh: `(String) value`: casting doesn't look necessary.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BooleanConverter.java
```diff
@@ -0,0 +1,20 @@
+package com.google.cloud.tools.eclipse.ui.util.databinding;
+
+import org.eclipse.core.databinding.conversion.Converter;
+
+public abstract class BooleanConverter extends Converter {
+
+  private static BooleanConverter NEGATE_INSTANCE = new BooleanConverter() {
+    public Object convert(Object fromObject) {
+      return !(Boolean)fromObject;
```
chanseokoh: nit: space after cast

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
akerekes: I tried but couldn't figure out the right regex, I'm open to suggestions.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
akerekes: I made `RESERVED_VALUES` a list instead and inlined this method as it became trivial.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
chanseokoh: I think this can be consistent with `APPENGINE_PROJECTID_PATTERN` in `ProjectIdValidator`:

```java
private static final Pattern APPENGINE_PROJECTID_PATTERN = Pattern.compile("^[a-z][a-z0-9-]{4,28}[a-z0-9]$"); //$NON-NLS-1$
```

So, I think this should work (note `{0,61}`):

```java
private static final Pattern APPENGINE_PROJECT_VERSION_PATTERN = Pattern.compile("^[a-z0-9][a-z0-9-]{0,61}[a-z0-9]$");
```

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BucketNameValidator.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
akerekes: The difference is in the minimal length. project version can be a single letter/digit, and this schema does not allow for it. I just realized that it could be:

akerekes: I'll commit this modification.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BucketNameValidator.java
```diff
@@ -0,0 +1,52 @@
...
+    } else {
+      return ValidationStatus.error(Messages.getString("bucket.name.invalid")); //$NON-NLS-1$
+    }
+  }
+
+  private IStatus allComponentsLengthAreValid(String value) {
+    String[] components = value.split("\\.");
+    for (String component : components) {
+      if (component.length() == 0 || component.length() > 63) {
+        return ValidationStatus.error(Messages.getString("bucket.name.invalid"));
```
elharo: should we include the bad value in the message? and maybe specify that the length is the problem?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.regex.Pattern;
+
+import org.eclipse.core.databinding.validation.IValidator;
+import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.eclipse.core.runtime.IStatus;
+
+import com.google.cloud.tools.eclipse.ui.util.Messages;
+
+public class ProjectIdValidator implements IValidator {
+  // todo: this is different from .newproject.AppEngineProjectIdValidator
```
elharo: we're going to want to get rid of that class after this is in. 
What is the difference between that and this?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java
```diff
@@ -0,0 +1,34 @@
...
+  // But this reflects the reality when creating a project which says:
+  // Project ID must be between 6 and 30 characters.
```
elharo: Please add a URL to the authoritative source here.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.eclipse.core.databinding.validation.IValidator;
+import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+
+import com.google.cloud.tools.eclipse.ui.util.Messages;
+
+/**
+ * Validate a project version.
```
elharo: please add a URL to the authoritative source for this format here

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
```diff
@@ -0,0 +1,65 @@
...
+    // https://cloud.google.com/appengine/docs/java/config/appref
+    // "The version specifier can contain lowercase letters, digits, and hyphens.
+    // It cannot begin with the prefix ah- and the names default and latest are
+    // reserved and cannot be used."
+    if (!(input instanceof String)) {
+      return ValidationStatus.error(Messages.getString("project.version.invalid")); //$NON-NLS-1$
+    }
+    String value = (String) input;
+    if (value.isEmpty()) {
+      return ValidationStatus.warning(Messages.getString("project.version.invalid")); //$NON-NLS-1$
```
elharo: might be useful to include the invalid string in the message

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
```diff
@@ -0,0 +1,65 @@
...
+    } else if (APPENGINE_PROJECT_VERSION_PATTERN.matcher(value).matches()) {
+      if (value.startsWith(RESERVED_PREFIX) || RESERVED_VALUES.contains(value)) {
+        return ValidationStatus.error(Messages.getString("project.version.reserved")); //$NON-NLS-1$
+      }
+      return Status.OK_STATUS;
+    } else {
+      return ValidationStatus.error(Messages.getString("project.version.invalid")); //$NON-NLS-1$
```
elharo: might be useful to include the invalid string in the message

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/messages.properties
```diff
@@ -0,0 +1,10 @@
+bucket.name.invalid=Invalid bucket name
```
elharo: Maybe not in this PR, but in general it's useful to be very specific about failures of this kind. E.g. tell the user that the bucket name is too long, too short, or has invalid characters in three separate messages. And tell them which characters are invalid in the message.

akerekes: https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/issues/605

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java
akerekes: The other validator roughly allows for two strings of letters, digits, hyphen and dot of 1 to 100 length, separated by a colon, and the first string is optional. This is rather different from the format this validator allows.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
akerekes: If the user explicitly selects the override default version checkbox I think it's a good idea to at least give a warning that it's been left empty. (It is not an error, thus not required to fill in).

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectIdValidator.java
briandealwis: I think the other validator is based on internal AppId-related code.  The 6-30 characters comes from error message from the Cloud Console.

akerekes: https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/pull/554#discussion_r76070574

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: Ditto for version.default

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,122 +33,224 @@
...
 
 public class DeployPropertyPage extends PropertyPage {
 
   private static final String APPENGINE_DASHBOARD_URL = "https://console.cloud.google.com/appengine";
+
   private static final int INDENT_CHECKBOX_ENABLED_WIDGET = 10;
 
-  private Button promptForProjectIdButton;
+  private static final String PREFERENCE_STORE_QUALIFIER = "com.google.cloud.tools.eclipse.appengine.deploy";
+  private static final String PREF_PROMPT_FOR_PROJECT_ID = "project.id.prompt"; // boolean
```
briandealwis: `.promptOnDeploy`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,122 +33,224 @@
...
+  private void setupDataBinding() {
+    bindingContext = new DataBindingContext();
+
+    setupProjectIdDataBinding(bindingContext);
+    setupProjectVersionDataBinding(bindingContext);
+    setupAutoPromoteDataBinding(bindingContext);
+    setupBucketDataBinding(bindingContext);
+
+    PreferencePageSupport.create(this, bindingContext);
+    bindingContext.updateTargets();
```
briandealwis: This line is unnecessary (my mistake)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,122 +33,224 @@
...
+    try {
+      if (isValid()) {
+        savePreferences();
+        return true;
+      } else {
+        return false;
+      }
+    } catch (IOException exception) {
+      logger.log(Level.SEVERE, "Could not save deploy preferences", exception);
+      return false;
```
briandealwis: Slightly prettier ?
```
 +   if (isValid()) {
 +     try {
 +        savePreferences();
 +        return true;
 +     } catch (IOException exception) {
 +        logger.log(Level.SEVERE, "Could not save deploy preferences", exception);
 +     } 
 +   }
 +   return false;

File: plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java [hunk 31]
```diff
@@ -188,73 +295,138 @@ private ExpandableComposite createExpandableComposite(Composite parent) {
...
-  private void initVersionButtonSelection() {
-    overrideDefaultVersionButton.setSelection(false);
-    updateVersionWidgetsEnablement();
-  }
+    private WritableValue promptForProjectId = new WritableValue(Boolean.TRUE, Boolean.class);
+    private WritableValue projectId = new WritableValue();
+    private WritableValue overrideDefaultVersioning = new WritableValue(Boolean.FALSE, Boolean.class);
+    private WritableValue version = new WritableValue();
+    private WritableValue autoPromote = new WritableValue(Boolean.TRUE, Boolean.class);
+    private WritableValue overrideDefaultBucket = new WritableValue(Boolean.FALSE, Boolean.class);
```
briandealwis: I think these `Boolean.class` uses should be `Boolean.TYPE` (_mea culpa_)

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BucketNameValidator.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.regex.Pattern;
+
+import org.eclipse.core.databinding.validation.IValidator;
+import org.eclipse.core.databinding.validation.ValidationStatus;
+import org.eclipse.core.runtime.IStatus;
+
+import com.google.cloud.tools.eclipse.ui.util.Messages;
+
+/**
+ * Implements a simplified bucket name validation.
```
briandealwis: "_for Google Cloud Storage"

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BucketNameValidator.java
```diff
@@ -0,0 +1,52 @@
...
+ * <p>
+ * The following rules are verified:
+ * <p>
+ * Use lowercase letters, numbers, hyphens (-), and underscores (_). You can also use a dot (.) to form a valid
+ * top-level domain (e.g., example.com). Format: You must start and end the name with a number or letter. Bucket 
+ * names must contain 3 to 63 characters. Names containing dots can contain up to 222 characters, but each 
+ * dot-separated component can be no longer than 63 characters.
+ * <p>
+ * The actual rules that govern the bucket naming are more complex. See the complete list of bucket name requirements
+ * for more information: https://cloud.google.com/storage/docs/naming
```
briandealwis: This should say whether we are more restrictive or more permissive than the naming standard.  I wonder if there's an approved regex already for this.

briandealwis:
LGTM though it's missing 'stop previous versions' (mentioned in #365).

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: The few examples I found in the Eclipse code used `.class`

akerekes:
Support `stop-previous-versions` - https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/issues/610

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectVersionValidator.java
elharo: Do we need the override default version checkbox? It feels redundant. If the explicit version is empty use the default. If not, use the supplied version. Does that make sense?

akerekes: I like that the user can explicitly make the choice of overriding the versioning. If we omit the checkbox, then we should probably explain the default versioning in a label or a field decoration, which would add another item to the UI.
What we could do is to make the version required if the checkbox is selected. This way the user cannot save the preferences without setting the version (in which case during deploy we'd either give an error message or use default versioning even though the user chose not to). So probably it's better to display an error message right away on the dialog.

akerekes:
@elharo @briandealwis @chanseokoh please take another look, I've added some more stuff since the LGTM. Also changed the behavior of how project ID setting works:

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,39 +33,70 @@
...
-  private Button browseBucketButton;
-  
+
+  private Model model = new Model();
+  private ObservablesManager observables;
+  private DataBindingContext bindingContext;
+
+
+  public DeployPropertyPage() {
+    super();
```
elharo: not needed; this is implicit

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -56,81 +105,161 @@ protected Control createContents(Composite parent) {
...
+  }
+
+  @Override
+  public boolean performOk() {
+    if (isValid()) {
+      try {
+        savePreferences();
+        return true;
+      } catch (IOException exception) {
+        logger.log(Level.SEVERE, "Could not save deploy preferences", exception);
```
elharo: this failure, unlikely as it is, probably calls for more than a log. We should alert the user the save failed, and maybe not close the dialog.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -188,73 +304,172 @@ private ExpandableComposite createExpandableComposite(Composite parent) {
...
 
+  /**
+   * Validates a checkbox and text field as follows:
+   * <ol>
+   * <li>if the checkbox is unselected -> the text field cannot contain the empty string
+   * <li>if the checkbox is selected -> the text field can contain the empty string
+   * <li>if the text field is not empty, the result is determined by {@link ProjectIdValidator} applied to the value
+   * </ol>
+   *
+   * <i>The class does not enforce the {@link ISWTObservableValue}s passed in to the constructor to be of a
```
elharo: I'm having trouble following this sentence

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -5,7 +5,8 @@ google.preferences=Google Cloud Platform-related settings
 auto.promote=Promote the deployed version to receive all traffic
 bucket.name=Bucket:
 cannot.open.browser=Cannot open browser: {0}
-deploy.prompt.projectid=Prompt for Project ID on deploy
+deploy.prompt.projectid=Always prompt for project ID on deploy
+deploy.prompt.projectid.long=During deploy you will have the opportunity to provide another value for project ID that will be used for that execution instead of the project ID set on this page.
```
elharo: delete "for that execution"

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -414,17 +419,30 @@ public void dispose() {
...
+      Preconditions.checkArgument(text.getWidget() instanceof Text,
+                                  "Constructor argument text is an observable for {0}, should be for {1}",
+                                  text.getWidget().getClass().getName(),
+                                  Text.class.getName());
+      Preconditions.checkArgument(selection.getWidget() instanceof Button,
+                                  "Constructor argument selection is an observable for {0}, should be for {1}",
+                                  selection.getWidget().getClass().getName(),
+                                  Button.class.getName());
+      Preconditions.checkArgument((selection.getWidget().getStyle() & SWT.CHECK) != 0,
+                                  "Constructor argument selection must be an observable for a checkbox");
```
elharo: argument selection --> argument validator

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -414,17 +419,30 @@ public void dispose() {
...
 
+    /**
+     * @param selection must be an observable for a checkbox, i.e. a {@link Button} with {@link SWT#CHECK} style
+     * @param text must be an observable for a {@link Text}
+     * @param validator must be a validator for String values, will be applied to <code>text.getValue()</code>
+     */
     public OverrideValidator(ISWTObservableValue selection, ISWTObservableValue text, IValidator validator) {
       super(selection.getRealm());
+      Preconditions.checkArgument(text.getWidget() instanceof Text,
+                                  "Constructor argument text is an observable for {0}, should be for {1}",
```
elharo: I don't think you need "Constrcutor argument" in these messages

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -446,20 +464,29 @@ protected IStatus validate() {
...
 
-    public ProjectIdMultiValidator(ISWTObservableValue selection,
-                                   ISWTObservableValue text) {
+    /**
+     * @param selection must be an observable for a checkbox, i.e. a {@link Button} with {@link SWT#CHECK} style
+     * @param text must be an observable for a {@link Text}
+     */
+    public ProjectIdMultiValidator(ISWTObservableValue selection, ISWTObservableValue text) {
       super(selection.getRealm());
+      Preconditions.checkArgument(text.getWidget() instanceof Text,
```
elharo: I don't think you need "Constructor argument" in these messages

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -5,9 +5,11 @@ google.preferences=Google Cloud Platform-related settings
 auto.promote=Promote the deployed version to receive all traffic
 bucket.name=Bucket:
 cannot.open.browser=Cannot open browser: {0}
-deploy.prompt.projectid=Always prompt for project ID on deploy
-deploy.prompt.projectid.long=During deploy you will have the opportunity to provide another value for project ID that will be used for that execution instead of the project ID set on this page.
 deploy.manual.link=To manually promote a version use the <a href="{0}">Google Cloud Console.</a>
+deploy.preferences.save.error.title=Could not save preferences
+deploy.preferences.save.error.message=There was an error while saving the preferences: {0}
```
elharo: There was an error --> Error

elharo:
LGTM

briandealwis:
Hit a bug:
  - opened _App Engine Deployment_ page
  - hit _Cancel) and got an NPE in `DialogPageSupport.dispose()`: as the `ProjectIdValidator` has no targets.
It's a [databinding bug (312785)](https://bugs.eclipse.org/bugs/show_bug.cgi?id=312785) that seems to have been dropped.  There's a [workaround on the bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=312785#c19).

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,39 +34,70 @@
...
+  private static final String PREFERENCE_STORE_DEFAULTS_QUALIFIER =
+      "com.google.cloud.tools.eclipse.appengine.deploy.defaults";
+  private static final String PREFERENCE_STORE_QUALIFIER = "com.google.cloud.tools.eclipse.appengine.deploy";
+  private static final String PREF_PROMPT_FOR_PROJECT_ID = "project.id.promptOnDeploy"; // boolean
+  private static final String PREF_PROJECT_ID = "project.id";
+  private static final String PREF_OVERRIDE_DEFAULT_VERSIONING = "project.version.overrideDefault"; // boolean
+  private static final String PREF_CUSTOM_VERSION = "project.version";
+  private static final String PREF_ENABLE_AUTO_PROMOTE = "project.promote"; // boolean
+  private static final String PREF_OVERRIDE_DEFAULT_BUCKET = "project.bucket.overrideDefault"; // boolean
+  private static final String PREF_CUSTOM_BUCKET = "project.bucket";
```
briandealwis: You're going to want to move these constants out to something in `.appengine.deploy`, right.  That can wait until your next PR though.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -15,39 +34,70 @@
...
-  
+
+  private Model model = new Model();
+  private ObservablesManager observables;
+  private DataBindingContext bindingContext;
+
+
+  public DeployPropertyPage() {
+    DefaultScope.INSTANCE.getNode(PREFERENCE_STORE_DEFAULTS_QUALIFIER).putBoolean(PREF_PROMPT_FOR_PROJECT_ID, true);
+  }
```
briandealwis: I was just wondering how this interacts with the default scope and whether it makes sense to define a preference initializer.  I would drop this for now and encode the defaults in your `Model`.

briandealwis:
Otherwise LGTM.  I'll put in some more comments in #605

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: I thought this will set the value in the singleton DefaultScope instance. Probably a more proper place would be in the plugin initialization as this constructor is called each time the dialog is opened, so any subsequent calls after the first is redundant.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: I was wondering whether to put these into `.appengine.deploy` or `.appengine.deploy.ui` as preferences are a UI thing in Eclipse (classes are in `org.eclipse.ui` and `org.eclipse.jface`). So moving these to a non-UI bundle may not be the best. What do you think @briandealwis ?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: Sorry, I wasn't clear.  Currently you're configuring defaults in the `Model` initialization, except for this part.  But a user can start a deploy without ever opening this properties page, and so no defaults will have been written out to the project preferences.  So the deploy code won't see any pref values.  If we instead set the defaults using a preference initializer (via the `org.eclipse.equinox.preferences.preferences` extension point), then providing the deploy code uses `IPreferenceService` (either via `Platform.getPreferenceService()` or via the OSGi service) or `ScopedPreferenceStore`, which lookup any unresolved preferences in the default scope, it will see the 'right' defaults.  Then we can set the defaults for `Model` via `loadPreferences()` — and restore the 'restore defaults` button too.

akerekes: I'll do the refactoring for preference initializer in #492 item 7. For now I'll leave the constructor and changes to the default scope to keep the behavior of the dialog .

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: Preferences are a runtime/core thing: they're defined in Equinox (`org.eclipse.equinox.preferences`, incl `IPreferencesService`). `IPreferenceStore` and `ScopedPreferenceStore` are defined in JFace, but `ScopedPreferenceStore` essentially works the same as `IPreferencesService` (consults a set of scopes in order, incl the `DefaultScope`).

PR #565, opened 2016-08-15
This is a new file that was part of the App Engine facet revamp PR.

nbashirbello:
It is used to add the appengine-web.xml to projects that install the facet.

@elharo


elharo:
LGTM

Issue #524, opened 2016-08-09
Unhandled event loop exception when cancelling out of login during deploy


elharo:
I pressed the cancel button in the Enter authorization code dialog box when attempting to deploy:

eclipse.buildId=4.6.0.I20160606-1100
java.version=1.8.0_91
java.vendor=Oracle Corporation
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product
Command-line arguments:  -os macosx -ws cocoa -arch x86_64 -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -product org.eclipse.epp.package.java.product -data file:/Users/demouser/workspace/ -product org.eclipse.epp.package.java.product

org.eclipse.ui
Error
Tue Aug 09 14:21:42 EDT 2016
Unhandled event loop exception

org.eclipse.e4.core.di.InjectionException: org.eclipse.core.commands.ExecutionException: Deploy failed.
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:64)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invokeUsingClass(InjectorImpl.java:282)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invoke(InjectorImpl.java:264)
	at org.eclipse.e4.core.contexts.ContextInjectionFactory.invoke(ContextInjectionFactory.java:132)
	at org.eclipse.e4.core.commands.internal.HandlerServiceHandler.execute(HandlerServiceHandler.java:152)
	at org.eclipse.core.commands.Command.executeWithChecks(Command.java:494)
	at org.eclipse.core.commands.ParameterizedCommand.executeWithChecks(ParameterizedCommand.java:488)
	at org.eclipse.e4.core.commands.internal.HandlerServiceImpl.executeHandler(HandlerServiceImpl.java:210)
	at org.eclipse.e4.ui.workbench.renderers.swt.HandledContributionItem.executeItem(HandledContributionItem.java:433)
[...]

PR #568, opened 2016-08-16
Inform to sign in for deploy if not signed in

### LINKED ISSUE #472: Let user know that he/she needs to log in for deploy

chanseokoh:
Fixes #472 #524.

prompt-login-for-deploy

Moved out `login()` from `launchDeployJob()`, since `launchDeployJob()` sounds like it will launch a job no matter what.

Did some tests: cancel login, deny authorization, deploy while logged in, etc.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/messages.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/messages.properties
@@ -10,5 +10,6 @@ deploy.standard.runnable.name=Deploying to App Engine Standard
 deploy.job.stagingdir.create.failed=Cannot create staging directory.
 deploy.job.sourcedir.missing=Staging source directory does not exist.
 deploy.failed.error.message=Deploy failed.
-login.failed=Login failed.
 save.credential.failed=Saving credential to file failed.
+prompt.login.dialog.title=Sign-in Required for Deploying
+prompt.login.dialog.message=Do you want to sign in to authorize deploying to the target project?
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/messages.properties
```diff
@@ -10,5 +10,6 @@ deploy.standard.runnable.name=Deploying to App Engine Standard
 deploy.job.stagingdir.create.failed=Cannot create staging directory.
 deploy.job.sourcedir.missing=Staging source directory does not exist.
 deploy.failed.error.message=Deploy failed.
-login.failed=Login failed.
-save.credential.failed=Saving credential to file failed.
\ No newline at end of file
+save.credential.failed=Saving credential to file failed.
+prompt.login.dialog.title=Sign-in Required for Deploying
+prompt.login.dialog.message=Do you want to sign in to authorize deploying to the target project?
```
elharo: Do you want to sign in to deploy the target project?

elharo:
per #472 "We should tell the user the reason for the login page."

I don't think this means we should ask them if they want to sign in. Assume they do. Ideally the browser page that pops up should include the message. However if that's difficult or infeasible a simple message dialog (Please sign in to deploy. OK) will do.

chanseokoh:
Customizing the title of the login page doesn't seem possible: https://developers.google.com/identity/protocols/OAuth2InstalledApp

I'll change the question dialog with a simple message dialog just with a OK button. (Alternatively, it is also possible to change the title of the login progress monitor dialog. However, since the external login browser will pop up and get the focus, a simple message dialog before launching the browser seems a good idea.)

chanseokoh:
The dialog looks like this now:

sign-in-to-deploy

patflynn:
This wording doesn't sound ideal to me.  Looking at it I'd wonder 'They
want me to open an external browser and login...how?  Go to google.com? How
does that authenticate the IDE?"

chanseokoh:
@patflynn Right... I'm too accustomed to our login flow.

elharo:
They don't have to press the button to get the browser, right? It just opens automatically. So here we just want a statement, not a call to action. E.g. 

"Browser opened to authorize deployment." 
OK

chanseokoh:
So, the last commit just customizes the message of the default login progress dialog. (No extra dialog; a browser will open automatically for deployment if users are not logged in.)

login-progress-dialog

elharo:
LGTM iff the tests pass on travis

PR #569, opened 2016-08-16
Disable Analytics in dev mode

chanseokoh:
Disables Analytics altogether if in the dev mode or the Analytics tracking ID is not defined.


elharo:
LGTM

PR #450, opened 2016-07-26
GAE Facet revamp

nbashirbello:
The App Engine Standard facet is used to expose the App Engine functionality (run, debug and deploy). To use the App Engine runtime, the project must have the App Engine facet. This means that adding the App Engine runtime without first adding the facet, would cause the facet to be added to the project.
- Installing the App Engine facet does the following
-- Adds the App Engine Runtime and SDK jars to the project's classpath
-- Adds the appengine-web.xml if it doesn't exist
- Uninstalling the App Engine facet does the following
-- Removes the App Engine runtimes from the project
-- Removes the App Engine Runtime and SDK jars from the project's classpath

Creating new App Engine projects would also include adding the runtimes but it doesn't need to happen when users are adding the facet to an already existing project. The runtimes would be available to projects with the App Engine facet

PR #570, opened 2016-08-16
fix integration test for new maven project

elharo:
@chanseokoh @briandealwis @nbashirbello

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/NewM2EclipseGeneratedProjectTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/NewM2EclipseGeneratedProjectTest.java
@@ -17,9 +17,10 @@
 package com.google.cloud.tools.eclipse.integration.appengine;
 
 import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
 import com.google.cloud.tools.eclipse.swtbot.SwtBotProjectActions;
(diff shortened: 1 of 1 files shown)


elharo:
so as best I can tell, the problem seems to be that this test has different behavior locally and on travis:

testNewMavenAppEngineProject(com.google.cloud.tools.eclipse.integration.appengine.NewM2EclipseGeneratedProjectTest)  Time elapsed: 3.703 sec  <<< FAILURE!
java.lang.AssertionError: Project should be faceted
	at org.junit.Assert.assertNotNull(Assert.java:712)
	at 

I.e. the project is faceted locally and not faceted on travis. weird

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/NewM2EclipseGeneratedProjectTest.java
```diff
@@ -47,14 +48,11 @@ public void testNewMavenAppEngineProject() throws CoreException {
...
 
-    /*** Commented as it's not true at the moment. ***/
-    // IFacetedProject facetedProject = new FacetedProjectHelper().getFacetedProject(project);
-    // assertNotNull("Project should be faceted", facetedProject);
-    // assertTrue( new FacetedProjectHelper().projectHasFacet(facetedProject,
-    // AppEngineStandardFacet.ID));
-
     IFacetedProject facetedProject = new FacetedProjectHelper().getFacetedProject(project);
-    assertNull("Project should not be faceted", facetedProject);
+    assertNotNull("Project should be faceted", facetedProject);
```
chanseokoh: I know we will remove this test entirely, but it's really strange that the Travis build failed exactly at this point and that previously the opposite test was succeeding. I wonder if the test below isn't working somehow.

briandealwis:
Hmm… I wonder if this is a consequence of the App Engine facet installation jobs from #450.  I suspect we'd see consistent behaviour with a `SwtBotWorkbenchActions.waitForIdle()` before the is-faceted test.

PR #573, opened 2016-08-16
delete flaky test

elharo:
@nbashirbello @chanseokoh @briandealwis We can add this back in when we figure out how to get the same behavior on travis and locally.


chanseokoh:
LGTM

PR #581, opened 2016-08-17
Use freemarker templates to generate files for App Engine.

nbashirbello:
Use freemarker templates to generate App Engine files


briandealwis:
> com.google.cloud.tools.eclipse.ui.util.templates.appengine

I haven't looked at the code, but I wouldn't expect this in `.ui`.

nbashirbello:
Still not ready for review :) Will add a note when it is ready for review

nbashirbello:
This PR is ready for review.

The tests are failing in travis in AppEngineTemplateUtility#createConfiguration (line 58). Runs successfully in eclipse. Need suggestions on how to get path to /templates/appengine that passes in travis.  

Thanks

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
```diff
@@ -111,16 +112,10 @@ private static void createConfigFiles(IProject project, IProgressMonitor monitor
...
-      IStatus status = StatusUtil.error(FacetInstallDelegate.class,
-          "Could not load template for " + APPENGINE_WEB_XML);
-      throw new CoreException(status);
-    }
-
-    IFile configFile = configDir.getFile(APPENGINE_WEB_XML);
-    if (!configFile.exists()) {
-      configFile.create(in, true, monitor);
-    }
+    appEngineWebXml.create(new ByteArrayInputStream(new byte[0]), true, monitor);
```
elharo: this is strange. Why do we need an empty stream?

plugins/com.google.cloud.tools.eclipse.util.test/testData/templates/appengine/web.txt
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns:web="http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
+         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd" 
+         version="2.5">
+  <servlet>
+    <servlet-name>HelloAppEngine</servlet-name>
+    <servlet-class>com.test.HelloAppEngine</servlet-class>
```
elharo: we should use com.example instead of com.test because exmaple.com domain name is reserved for exactly this sort of thing

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
nbashirbello: I want to create an empty file.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/templates/appengine/AppEngineTemplateUtility.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  private AppEngineTemplateUtility() {
+  }
+
+  private static Configuration createConfiguration() throws IOException, URISyntaxException{
+    Configuration cfg = new Configuration(Configuration.VERSION_2_3_25);
+    Bundle bundle = FrameworkUtil.getBundle(AppEngineTemplateUtility.class);
+    URL templatesUrl = bundle.getEntry("/templates/appengine");
+    File appengineTemplatesDir = new File(FileLocator.resolve(templatesUrl).toURI());
```
briandealwis: Use `FileLocator.toFileURL()`: that will extract the contents as required.

plugins/com.google.cloud.tools.eclipse.util/build.properties
```diff
@@ -3,6 +3,7 @@ output.. = target/classes/
 bin.includes = META-INF/,\
                .,\
                plugin.xml,\
```
briandealwis: I think you're missing the template directory here too, which might account for why the templates aren't found.

plugins/com.google.cloud.tools.eclipse.util.test/testData/templates/appengine/helloAppEngine.txt
```diff
@@ -0,0 +1,18 @@
+package com.example
```
chanseokoh: Missing `;`?

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/templates/appengine/AppEngineTemplateUtilityTest.java
```diff
@@ -0,0 +1,134 @@
...
+    InputStream expectedFileStream = getDataFile("appengineWebXmlWithoutProjectId.txt");
+    compareFileContent(expectedFileStream, testFileStream);
+  }
+
+  @Test
+  public void testCreateFileContent_helloAppEngine()
+      throws CoreException, IOException {
+    String fileLocation = testFile.getLocation().toString();
+    Map<String, String> dataMap = new HashMap<String, String>();
+    dataMap.put("package", "package com.example");
```
chanseokoh: `package com.example;` (missing `;`)?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/templates/appengine/AppEngineTemplateUtility.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public static void createFileContent(String outputFileLocation, String templateName, Map<String, String> dataMap)
+      throws CoreException {
+    Preconditions.checkNotNull(outputFileLocation, "output file is null");
+    Preconditions.checkNotNull(templateName, "template name is null");
+    Preconditions.checkNotNull(dataMap, "data map is null");
+
+    try {
+      if (configuration == null) {
+        configuration = createConfiguration();
```
chanseokoh: This doesn't look like it's thread-safe, and I wonder if it needs to.

chanseokoh: Maybe this is called while a workspace is locked, so no issue with the thread-safety?

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/templates/appengine/AppEngineTemplateUtilityTest.java
```diff
@@ -79,7 +79,7 @@ public void testCreateFileContent_helloAppEngine()
       throws CoreException, IOException {
     String fileLocation = testFile.getLocation().toString();
     Map<String, String> dataMap = new HashMap<String, String>();
-    dataMap.put("package", "package com.example");
+    dataMap.put("package", "com.example");
```
briandealwis: The previous code (`package X`) was necessary to handle when the user selects the default package.  It's unfortunately a valid package to select and we need to support it.  Can freemarker support optional template pieces, like in sh:
```
${package:+package ${package};}
```
?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
briandealwis: What's the rationale of the empty file?  Would be useful to see in a comment.  Is it to ensure Eclipse sees the file?  If so, you should execute `appEngineWebXml.refreshLocal(IResource.DEPTH_ZERO, monitor)`.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xmlns:web="http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
+         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd" 
+         version="2.5">
+  <servlet>
+    <servlet-name>HelloAppEngine</servlet-name>
+    <servlet-class>${package}HelloAppEngine</servlet-class>
```
briandealwis: I think you'll need a trailing period here

briandealwis:
I've never played with Freemarker, but from the little I've seen it looks like a great addition.  But this implementation is tied too much to the appengine templates.  We should be able to use this functionality to create GCE- or GKE-specific examples without having any AppEngine bundles available.  You could move the functionality into `.appengine.newproject`, which would be acceptable — that's where `CodeTemplates` lives — and we'll just copy code for subsequent uses.  Or you could refactor this to be more general purpose:
  - rename `AppEngineTemplateUtility` &rarr; `TemplateUtility`
  - move the templates to `.appengine.newproject`, not in the general purpose util bundle.  Maybe callers should create an instance of `TemplateUtility` and provide the bundle or class and a path to the templates?
  - move the named constants into `.appengine.newproject`

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/HelloAppEngine.java.ftl
```diff
@@ -1,4 +1,4 @@
-${package}
+<#if package != "">package ${package};</#if>
```
briandealwis: nice

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/templates/appengine/AppEngineTemplateUtility.java
```diff
@@ -0,0 +1,59 @@
...
+
+public class AppEngineTemplateUtility {
+  public static final String APPENGINE_WEB_XML_TEMPLATE = "appengine-web.xml.ftl";
+  public static final String HELLO_APPENGINE_TEMPLATE = "HelloAppEngine.java.ftl";
+  public static final String INDEX_HTML_TEMPLATE = "index.html.ftl";
+  public static final String WEB_XML_TEMPLATE = "web.xml.ftl";
+
+  private static Configuration configuration;
+
+  public static void createFileContent(String outputFileLocation, String templateName, Map<String, String> dataMap)
```
briandealwis: Need Javadoc to describe how to handle template exceptions?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
```diff
@@ -111,16 +112,10 @@ private static void createConfigFiles(IProject project, IProgressMonitor monitor
...
+    String configFileLocation = appEngineWebXml.getLocation().toString();
+    AppEngineTemplateUtility.createFileContent(
+        configFileLocation, AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE, new HashMap<String, String>());
```
akerekes: instead of a new instance you could use `Collections.EMPTY_MAP`

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
nbashirbello: The templates can only write to a file that already exists.

nbashirbello:
Templates are used in com.google.cloud.tools.eclipse.appengine.newproject and also in com.google.cloud.tools.eclipse.appengine.facets. The templates package is com.google.cloud.tools.eclipse.util.template and the appengine specific templates are in com.google.cloud.tools.eclipse.util.template.appengine. In the future when we have other templates, e.g. for xyz feature, they could be in com.google.cloud.tools.eclipse.util.template.xyz

briandealwis:
I disagree with that decision: IMHO the templates should live alongside the code that uses them.  We shouldn't be modifying a utility bundle to add templates.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/HelloAppEngine.java.ftl
```diff
@@ -0,0 +1,18 @@
...
```
elharo: following blank line should also be part of if block

elharo:
LGTM

PR #592, opened 2016-08-19
Persist user credential using Java Preferences

### LINKED ISSUE #397: Enable persisting and reusing credentials for Google login
Created credentials should be managed in a persistent credential datastore for reuse.

For #342.

chanseokoh:
Fixes #397.

[We put the Java Preferences-based DataStore into the ide-login library](https://github.com/GoogleCloudPlatform/ide-login/pull/18). We are just using it.

I had some difficulty initializing the login menu item and the icon tooltip based on the login status at Eclipse startup. (For example, when a user is logged in, the menu item should change to reflect that.) When I was tinkering with `<visibleWhen>`, I noticed that using a `PropertyTester` and forcing the activation of the bundle (`forcePluginActivation="true"`) triggers [`GoogleLoginCommandHandler.updateElement()`](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginCommandHandler.java#L36), which updates the menu and the icon properly. However, I feel this a bit hacky. I wonder what could be the right way.


briandealwis:
Take a look at the `org.eclipse.ui.startup` extension point.  GPE didn't have this problem as its login widget was a `ToolControl`, so it actually ran code to render itself.

chanseokoh:
Using `org.eclipse.ui.startup` works like a charm. :+1:

The `IStartup` interface Javadoc gives an example of using `PlatformUI` and `Display.asyncExec()`, so I'm following that.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -19,11 +19,11 @@
 import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeRequestUrl;
 import com.google.cloud.tools.eclipse.appengine.login.ui.LoginServiceUi;
 import com.google.cloud.tools.ide.login.GoogleLoginState;
+import com.google.cloud.tools.ide.login.JavaPreferenceOAuthDataStore;
 import com.google.cloud.tools.ide.login.LoggerFacade;
 import com.google.cloud.tools.ide.login.OAuthDataStore;
 import com.google.common.annotations.VisibleForTesting;
 
-import org.eclipse.e4.core.contexts.IEclipseContext;
```
briandealwis: You'll also be able to remove the `Import-Package: org.eclipse.e4.core.contexts` from `MANIFEST.MF`

chanseokoh: Good catch!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -43,6 +43,9 @@
  */
 public class GoogleLoginService implements IGoogleLoginService {
 
+  private static final String OAUTH_DATA_STORE_PREFERENCE_PATH =
```
briandealwis: This is the limitation I mentioned yesterday: we're persisting a single account.  How do you envision this working with multiple accounts?  (And I'm totally fine with punting on this for now.)

Hmm, I suppose `ide-login` could use:
```
Preferences prefs = Preferences.userRoot().node(preferencePath).node(credential.getStoredEmail());
```
as the node, and then have a single key "active" -> email to note the active account.  (_active_ is ok since it's not a valid email address.)

briandealwis:
Otherwise LGTM.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: Yeah, @nbashirbello was also talking a very similar thing (although what she was talking about was to having a single node, no children, and have things like `prefs.put("KEY_EMAIL" + credential.getStoredEmail(), ...)`.

The reason I hesitated a bit now is that, in IntelliJ, we sometimes observe an email is null. I need to look into that to figure out why it's sometimes null.

PR #600, opened 2016-08-22
Send Analytics login events

### LINKED ISSUE #537: Analytics Tracking for Login action
Be very careful to only track that the action occurred, not who logged in

chanseokoh:
Fixes #537.

Did actual testing; working fine.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
@@ -20,6 +20,8 @@ import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeReque
 import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
 import com.google.cloud.tools.eclipse.appengine.login.GoogleLoginService;
 import com.google.cloud.tools.eclipse.appengine.login.Messages;
+import com.google.cloud.tools.eclipse.usagetracker.AnalyticsEvents;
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
```diff
@@ -126,16 +131,23 @@ private String showProgressDialogAndWaitForCode(final String message,
...
-      new ProgressMonitorDialog(shellProvider.getShell()) {
+      final Shell parentShell = shellProvider.getShell();
+      new ProgressMonitorDialog(parentShell) {
         @Override
         protected void configureShell(Shell shell) {
           super.configureShell(shell);
           shell.setText(Messages.LOGIN_PROGRESS_DIALOG_TITLE);
+
+          AnalyticsPingManager.getInstance().sendPing(
+              AnalyticsEvents.LOGIN_START, null, null, shell);
```
briandealwis: A minor quibble, but a constructor doesn't seem the right place for this event.  Why not either within the run() or just above?

briandealwis:
LGTM.

I wonder if `AnalyticsPingManager#sendPingEvent()` could take an `IShellProvider`, since you're having to 'dereference' the shell here repeatedly even though it may not be called for.  But I wasn't sure if there might be hidden consequences to calling `IShellProvider#getShell` outside of the UI thread.

chanseokoh:
Thanks. There is actually only one instance of calling `IShellProvider.getShell()` for `sendPing()` in this PR. (Although `getParentShell()` could be replaced with it, I like calling `getParentShell()` that seems conceptually more fitting). If I find many instances of calling `IShellProvider.getShell()` coming along, I'll add a new method.

PR #602, opened 2016-08-23
Release lock first when a user cancels login

chanseokoh:
When a user cancels login, we release a lock which terminates the dialog task and closes the dialog.

I think it's a better idea to release the lock first, since we shouldn't assume `stopCodeWaitingJob()` will throw an unchecked exception.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/LoginServiceUi.java
@@ -137,8 +137,8 @@ public class LoginServiceUi implements UiFacade {
         }
         @Override
         protected void cancelPressed() {
-          stopCodeWaitingJob(redirectUrl);
           wait.release();  // Allow termination of the attached task.
+          stopCodeWaitingJob(redirectUrl);
 
           AnalyticsPingManager.getInstance().sendPing(
               AnalyticsEvents.LOGIN_CANCELED, null, null, getParentShell());


briandealwis:
If I were to change this code, I think I'd drop the semaphore entirely — we're essentially creating a dummy thread that just blocks on the semaphore.  I think you can change the runnable to use `fork=false` and instead loop like:
```
      }.run(false /* fork */, true /* cancelable */, new IRunnableWithProgress() {
        @Override
          monitor.beginTask(message != null ? message : Messages.LOGIN_PROGRESS_DIALOG_MESSAGE,
              IProgressMonitor.UNKNOWN);
          // Fork another sub-job to circumvent the limitation of LocalServerReceiver.
          // (See the comments of scheduleCodeWaitingJob().)
          Job job = scheduleCodeWaitingJob(
              new LocalServerReceiverWrapper(codeReceiver), wait, codeHolder, exceptionHolder);
          // when the job completes, awaken the UI thread
          job.addJobListener(new JobChangeAdapter() {
            public void done(IJobChangeEvent event) { display.asyncExec(new Runnable() {})); }
           });
          // check if cancelled, or if the job completed
          while(!monitor.isCanceled() && job.getResult() == null) {
             while(display.readAndDispatch()) { /*spin*/ }
             display.sleep();
          }
        }
      });
```

Or alternatively use a `Future` instead of checking the `Job`'s result.

chanseokoh:
Ah, that's how one can consume UI events.

Do you think it is necessary to add an event listener that wakes the UI thread? It's execution after the UI thread's `display.sleep()` is not guaranteed (e.g., the sub-job terminates immediately due to an error and the UI thread reaches `display.sleep()`), so I see no point of doing so here. I think pressing the cancel button wakes the UI thread anyway; in my test, it works even if the job is not terminating.

briandealwis:
Good catch:  there is a race condition between checking Job.getResult() and the sleep.

The asyncExec is needed when the UI thread is in sleep() (since actions are happening in the browser) and the login completes.  Without the asyncExec, the UI won't wake up (until some other action occurs of course).

briandealwis:
…which points out that maybe it is easier to have an extra thread around :-)

Anyways, your reasoning seems sound to me.  So I'm LGTM.

PR #607, opened 2016-08-24
Wait uninterruptibly to ignore spurious wake-ups 

chanseokoh:
Sadly, I missed it.


briandealwis:
LGTM.  What were the causes of the spurious wakes?

chanseokoh:
I have never encountered spurious wakes. I think it can happen when a process gets a signal:

http://stackoverflow.com/questions/1050592/do-spurious-wakeups-actually-happen

PR #616, opened 2016-08-26
Move UI related deploy code from c.g.c.t.e.appengine.deploy to c.g.c.t.e.appengine.deploy.ui

akerekes:
Additionally the job config parameters have been extracted into a value class.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
```diff
@@ -2,6 +2,67 @@
...
+      </enabledWhen>
+    </handler>
+  </extension>
+
+  <extension
+    point="org.eclipse.ui.console.consolePatternMatchListeners">
+    <consolePatternMatchListener
+      class="com.google.cloud.tools.eclipse.ui.util.console.PatternToHyperlinkConverter"
+      id="com.google.cloud.tools.eclipse.appengine.deploy.consolePatternMatchListener"
+      regex="https://.+\.appspot\.com">
```
briandealwis: I missed this before, but we should constrain this regex as `.+` is unnecessarily broad.  `[a-zA-Z][a-zA-Z0-9-.]+[a-zA-Z0-9]`?  (I'm not sure about the `.`: I know AE is moving from _version.service.app_.appspot.com to the _version_-_service_-_app_.appspot.com model, but does that show here?)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/META-INF/MANIFEST.MF
```diff
@@ -37,5 +46,7 @@ Import-Package: com.google.cloud.tools.eclipse.ui.util,
  org.eclipse.swt.graphics,
```
elharo: do we still need any of these swt imports since the ui code is in a different package?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
```diff
@@ -1,2 +1,6 @@
 pluginName=Google App Engine Standard Deployment Support UI
```
elharo: delete "Support"

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -0,0 +1,141 @@
+/*******************************************************************************
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * All rights reserved. This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License v1.0 which
```
elharo: apache license please

plugins/com.google.cloud.tools.eclipse.appengine.deploy/plugin.xml
```diff
@@ -1,70 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
```
elharo: plugin.xml is supposed to have at least one extension point. I believe Eclipse warns about this.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: I agree, that it's too broad, however I'd only restrict it to only non-whitespace. I don't want to add a more stricter logic as then it needs to be maintained should the valid format change. I think we can rely on the fact, that if something starts with `https://` and ends with `appspot.com` and appears in the output of the deploy process, it is most likely a valid URL.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/META-INF/MANIFEST.MF
akerekes: This the UI bundle's manifest file.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
akerekes: @elharo This comes from the name of the non-UI bundle by appending `UI`, but I can remove it. Or should we remove `Support` from that plugin's name as well?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
briandealwis: Domain names have a very clamped down character space.  There were some performance issues in the Console view with very long lines that were traced to an [overly broad regex](https://bugs.eclipse.org/bugs/show_bug.cgi?id=175888#c29) to search for Java exception stacktraces.

elharo:
LGTM

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
```diff
@@ -43,7 +43,7 @@
     <consolePatternMatchListener
       class="com.google.cloud.tools.eclipse.ui.util.console.PatternToHyperlinkConverter"
       id="com.google.cloud.tools.eclipse.appengine.deploy.consolePatternMatchListener"
-      regex="https://\S+\.appspot\.com">
+      regex="https://[a-zA-Z][a-zA-Z0-9-.]*[a-zA-Z0-9]?\.appspot\.com">
```
briandealwis: That still matches `abc-`.  Since a Project ID ≥ 6 characters you can leave off the trailing `?`

Issue #614, opened 2016-08-25
_Run_ launches must register a debug target or process with the ILaunch


briandealwis:
If we use _Run_ to start the server, a launch configuration appears in the _Debug_ view that cannot be terminated as there are no associated processes or debug target resources.
<img width="867" alt="will not disappear" src="https://cloud.githubusercontent.com/assets/202851/17985458/e6603a70-6ae3-11e6-9427-263b3f1f2718.PNG">

Basically `LocalAppEngineServerLaunchConfigurationDelegate#launch()` needs to associate an `IProcess` (perhaps the CloudSdk's process?) or an `IDebugTarget` with 
the `Launch` instance.  The normal JavaLaunchDelegate (_Run As Java Application_) uses `DebugPlugin.newProcess()` to add a wrapper around a `java.lang.Process`.

Thence the launch can be terminated and removed.

PR #618, opened 2016-08-26
Cause server launch to listen and react to server state changes

### LINKED ISSUE #516: LocalAppEngineServerLaunchConfigurationDelegate must listen to LocalAppEngineServer state changes
When launching the local server from within Eclipse, the `dev_appserver` may fail to launch (e.g., cannot listen on configured port).  Our ServerBehaviour has a [process exit listener](https://github.com/GoogleCloudPlatform/gcloud-eclipse-tools/blob/30532621dde075f09e6e1f2437732c0ec9aa6972/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java#L199) that sets the server state to STOPPED.  But our LaunchConfigurationDelegate doesn't listen to the state change and assumes the server is still running.  And because there is no process or debug target attached to the launch, the launch configuration cannot be _terminated_ and lives forever.

briandealwis:
Cause server launch to listen and react to server state changes such that stopping the server terminates the launch.

_Note:_ this only clears up the _Debug_ situation; #614 will fix the _Run_ situation

Also cleaned up server stopping to set state to STOPPED once the server process is reported as dead.  This was to fix strange race condition where the server could be reported as STOPPED but continued to write to the console.

Fixes #516

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
@@ -25,15 +25,21 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


elharo:
LGTM once the tests pass

PR #619, opened 2016-08-26
Use deployment preferences during deploy.

akerekes:
Also cleanup some manifests.

Implements task 7 of #492


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -189,41 +169,13 @@ public boolean performOk() {
...
-    preferenceStore.setValue(PREF_ENABLE_AUTO_PROMOTE, model.isAutoPromote());
-    preferenceStore.setValue(PREF_OVERRIDE_DEFAULT_BUCKET, model.isOverrideDefaultBucket());
-    preferenceStore.setValue(PREF_CUSTOM_BUCKET, model.getBucket());
-
-    ScopedPreferenceStore scopedPreferenceStore = (ScopedPreferenceStore) preferenceStore;
-    if (scopedPreferenceStore.needsSaving()) {
-      scopedPreferenceStore.save();
-    }
+  private void savePreferences() throws BackingStoreException {
+    model.savePreferences();
```
briandealwis: Would it make sense to move the exception logging into `savePreferences()` and instead have `savePreferences()` return true/false on success?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployPreferences.java
```diff
@@ -0,0 +1,93 @@
...
+  private static final String PREF_OVERRIDE_DEFAULT_BUCKET = "project.bucket.overrideDefault"; // boolean
+  private static final String PREF_CUSTOM_BUCKET = "project.bucket";
+
+  private IEclipsePreferences preferenceStore;
+
+  public StandardDeployPreferences(IProject project) {
+    preferenceStore = new ProjectScope(project).getNode(StandardDeployPreferences.PREFERENCE_STORE_QUALIFIER);
+  }
+
+  public void save() throws BackingStoreException {
```
briandealwis: Does this method actually do anything?  Are the values being saved being obtained from the same preferenceStore?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployPreferencesConverter.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class StandardDeployPreferencesConverter {
+
+  private StandardDeployPreferences preferences;
+
+  public StandardDeployPreferencesConverter(StandardDeployPreferences preferences) {
+    this.preferences = preferences;
+  }
+
+  public DefaultDeployConfiguration toDeployConfiguration() {
```
briandealwis: Would it make sense to call this an `Applicator` and have it `applyTo()` a deploy configuration?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployer.java
```diff
@@ -25,49 +20,19 @@
...
-    deployInfo = new AppEngineDeployInfo();
-  }
-
-  @VisibleForTesting
-  AppEngineProjectDeployer(AppEngineDeployInfo deployInfo) {
-    this.deployInfo = deployInfo;
-  }
-
-  public void deploy(IPath stagingDirectory, CloudSdk cloudSdk, IProgressMonitor monitor) throws CoreException {
+  public void deploy(IPath stagingDirectory, CloudSdk cloudSdk,
```
briandealwis: Does `stagingDirectory` need to be an `IPath`?  It's not really in the workspace, right?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineDeployInfo.java
```diff
@@ -12,17 +18,14 @@
...
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
+import com.google.cloud.tools.eclipse.util.status.StatusUtil;
 
 /**
  * Parses and returns project ID and version from appengine-web.xml
+ *
+ * @deprecated Deployment should not rely on the project Id and version set in the appengine-web.xml
  */
+@Deprecated
```
briandealwis: Hmm, I think you want `@Deprecated` on those specific elements — this will cause warnings for any reference to `AppEngineDeployInfo`.

akerekes: My intention was to get rid of this class as we should not parse this information form the XML for our current functionality. Maybe in the future it will be useful again to do so. The reason I could not remove this class is that an integration test depends on it and I did not want to change that in this PR.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployer.java
akerekes: Currently it is in the workspace, see `com.google.cloud.tools.eclipse.appengine.deploy.ui.standard.StandardDeployCommandHandler.getTempDir()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployPreferences.java
akerekes: Thanks for catching this, the only meaningful code here is the `flush()` call. I've deleted the rest.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployPreferencesConverter.java
akerekes: I don't know. I don't feel it as a significant improvement, but I agree that this class and method could be better named.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
```diff
@@ -43,7 +43,7 @@
     <consolePatternMatchListener
       class="com.google.cloud.tools.eclipse.ui.util.console.PatternToHyperlinkConverter"
       id="com.google.cloud.tools.eclipse.appengine.deploy.consolePatternMatchListener"
-      regex="https://[a-zA-Z][a-zA-Z0-9-.]*[a-zA-Z0-9]?\.appspot\.com">
+      regex="https://[a-zA-Z0-9][a-zA-Z0-9-.]*[a-zA-Z0-9]\.appspot\.com">
```
elharo: It might be worth pulling out a small fix like this into a separate PR so it can be quickly reviewed and committed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModel.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public void setPromptForProjectId(boolean promptForProjectId) {
+    this.promptForProjectId = promptForProjectId;
+  }
+
+  public String getProjectId() {
+    return projectId;
+  }
+
+  public void setProjectId(String projectId) {
```
elharo: null is allowed?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -132,98 +114,97 @@ private void setupDataBinding() {
...
-    preferenceStore.setValue(PREF_OVERRIDE_DEFAULT_VERSIONING, model.isOverrideDefaultVersioning());
-    preferenceStore.setValue(PREF_CUSTOM_VERSION, model.getVersion());
-    preferenceStore.setValue(PREF_ENABLE_AUTO_PROMOTE, model.isAutoPromote());
-    preferenceStore.setValue(PREF_OVERRIDE_DEFAULT_BUCKET, model.isOverrideDefaultBucket());
-    preferenceStore.setValue(PREF_CUSTOM_BUCKET, model.getBucket());
-
-    ScopedPreferenceStore scopedPreferenceStore = (ScopedPreferenceStore) preferenceStore;
-    if (scopedPreferenceStore.needsSaving()) {
-      scopedPreferenceStore.save();
+  private boolean savePreferences() {
```
elharo: Error return codes are unusual in java. Consider throwing an exception on failure to save instead and handling it up the call stack.

elharo:
Test failures look flaky. LGTM iff the tests pass.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: Eventually the Eclipse method calling this one will return a boolean status. I accepted @briandealwis 's suggestion to move the error handling down into this method, to make the two method more balanced (wrt lines of code).
But I agree, it's better to use exceptions for error handling in general.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModel.java
akerekes: I'd keep any validations separate from this class. `null` wouldn't cause a problem in the code (as in no NPE being thrown).

PR #624, opened 2016-08-28
remove redundant groupId and version elements from pom.xmls

akerekes:


elharo:
@briandealwis What do you think of this? I've looked into this in the past, and it seems many people believe the warnings Eclipse throws up here are spurious, and perhaps even ill-advised. I wasn't able to come to any definite conclusion myself.

briandealwis:
Certainly the `<groupId>` is redundant since the groupId is inherited from the `<parent>` element.  The `<version>` field feels a bit more problematic, as we could conceivably vary it, and these version fields have to correspond to the `MANIFEST.MF` or `feature.xml` versions.

To be honest, I think it might be better exploit Tycho 0.25's _pomless_ support and drop the `pom.xml`s where possible.  This pommels support builds off of Maven's polyglot support to retrieve information from the `MANIFEST.MF`, `build.properties` and `feature.xml`.  I started experimenting with this in the [pomless](https://github.com/GoogleCloudPlatform/cloud-tools-for-eclipse/tree/pomless) but Tycho 0.25 requires Java 8, and I wasn't sure about setting up the `toolchains.xml` for Travis to enable testing against Java 7 and Java 8.  Now that have the `quay.io` travis images, that could be done.

akerekes:
+1 for pomless build. However I'd rather not let the versions differ (even
if there are no changes from version to version) to simplify dependency
management and the release process (to be able to use the ability to set a
new version in one step to all bundle using the tycho-version-plugin).

briandealwis:
LGTM

PR #629, opened 2016-08-29
Enable 'Restore Defaults' button for deploy preferences page

akerekes:
Item 8 in #492


plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployPreferencesTest.java
```diff
@@ -0,0 +1,45 @@
+package com.google.cloud.tools.eclipse.appengine.deploy.standard;
+
+import static org.junit.Assert.*;
```
elharo: google style is not to use wildcard imports: https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

elharo:
LGTM assuming tests pass

PR #631, opened 2016-08-30
don't set project ID from new project wizard

elharo:
@akerekes I'm not sure if we should take it out from the wizard or not. Are we going to use the wizard to set the project default preference?


akerekes:
The deploy preference defaults are set by `com.google.cloud.tools.eclipse.appengine.deploy.standard.DeployPreferenceInitializer`. We don't set any preferences per-project until the preferences page is opened and saved for the first time. As long as any code that needs the project ID uses `com.google.cloud.tools.eclipse.appengine.deploy.standard.StandardDeployPreferences` (which in turn falls back to the defaults if none set) we should be OK.

akerekes:
LGTM

PR #639, opened 2016-08-30
Fixes #630 - Locked up Eclipse on "Deploy to App Engine"

### LINKED ISSUE #630: Locked up Eclipse on "Deploy to App Engine"
OSX.  Tried the new plugin.  Froze up when deploying to GAE after the uploaded files step.  A Cancel request didn't cancel.  Eventually had to force-quit Eclipse.

akerekes:


chanseokoh:
I tested this PR, but unfortunately, it doesn't look like this really fixes #630. Pressing the cancel button still does nothing if the control is stuck inside `deployer.deploy()`.

However, I can verify that at least this fix allows users to quit Eclipse gracefully while the deploy job is still "running".

akerekes:
How did you test it? How do you simulate that it's stuck in deploy?

chanseokoh:
Sorry. I'll take back what I said in the above comment. I simulated the situation where the job is blocked indefinitely, but on second thought, that's not realistic. In fact, I realized that `deployment.deploy()` becomes unblocked if the gcloud process is forcibly killed, so this looks good.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -50,7 +51,9 @@
   private AppEngineProjectDeployer deployer;
   
   //temporary way of error handling, after #439 is fixed, it'll be cleaner
-  protected boolean cloudSdkProcessError;
+  protected IStatus cloudSdkProcessStatus = Status.OK_STATUS;
```
elharo: should be final or private

elharo:
LGTM

PR #646, opened 2016-08-31
fill in project ID from wizard

elharo:
@akerekes

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
@@ -4,10 +4,12 @@ import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IProjectDescription;
 import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.ProjectScope;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -57,6 +59,17 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
     AppEngineStandardFacet.installAllAppEngineRuntimes(facetedProject, true /* force */, monitor);
+    
+    setProjectIdPreference(newProject);
+  }
+
+  void setProjectIdPreference(IProject project) {
+    String projectId = config.getAppEngineProjectId();
+    if (projectId != null && !projectId.isEmpty()) {
+      IEclipsePreferences preferences = new ProjectScope(project)
+          .getNode("com.google.cloud.tools.eclipse.appengine.deploy");
```
akerekes: Is this to avoid depending on the `c.g.c.t.e.appengine.deploy` bundle? Then I'd rather move the preference node IDs into the `c.g.c.t.e.util` bundle and refer to it from everywhere else.

elharo: Yes, though now I'm thinking we may just want to eliminate this completely and just ask for it at deploy time

elharo:
PTAL. After reflection I think we should move aherad with this, regardless of whether we set the preferneces on project properties or at deploy time. Either way we'll be storing this value here.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
akerekes: #660 filed to tackle this when a new feature will actually need the project ID.

akerekes:
LGTM

PR #657, opened 2016-09-01
Fix bug: Set max module instance = 1 in *debug* mode, not in *run* mode

briandealwis:
In c0e409b1946d92bc5975cc28eb9e6127c225ecca, I mistakenly restored the call to `DefaultRunConfiguration.setMaxModuleInstances(1)` to the _run_ server setup instead of the _debug_ server setup.


elharo:
Any not too terribly onerous way we could test this to avoid regressions like this?

briandealwis:
Not that I can think of.  Perhaps if `DefaultRunConfiguration` had a way to to produce the command-line?

elharo:
LGTM

PR #658, opened 2016-09-01
Improve SERVER_STARTED detection

### LINKED ISSUE #556: Improve server-start detection to mark server a STATE_STARTED
`LocalAppEngineServerBehaviour` uses an `appengine-plugins-core` `ProcessStartedListener`  to set the server's state to `STATE_STARTED` (in `LocalAppEngineStartListener`). This listener is fired once `appengine-plugins-core` has launched the process, but does not necessarily indicate that the server was actually started (e.g., could be delayed, could have errored out).  We should not set `STATE_STARTED` until we're able to probe the deserver's API port. An alternative is to push that responsibility should be to `appengine-plugins-core`.

briandealwis:
Scan the `dev_appserver` output for special strings to identify the server start or failure state as per #556.

Fixes #556


elharo:
LGTM for now though I really wish there were a more authoritative way we could do this. Those strings are going to change out from under us one of these days. Any ideas?

briandealwis:
The strings are used in appengine-plugins-core too.  It would be better to push this down.  Filed https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/201.

PR #659, opened 2016-09-02
release scripts for kokoro

elharo:
@akerekes @jmicco

### DIFF
--- a/kokoro/ubuntu/release.sh
+++ b/kokoro/ubuntu/release.sh
@@ -8,7 +8,7 @@ set -x
 # create a timestamp file
 date +%s > timestamp.txt
 
-cd git/kokoro-codelab-elharo
+cd git/google-cloud-eclipse
 ./build.sh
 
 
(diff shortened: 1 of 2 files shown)


elharo:
PTAL @jmicco

akerekes:
LGTM (however I have no kokoro knowledge yet)

kokoro/ubuntu/release.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# Fail on any error.
+set -e
+# Display commands being run.
+set -x
+
+# create a timestamp file
+date +%s > timestamp.txt
+
+cd git/kokoro-codelab-elharo
```
briandealwis: It looks odd to embed a particular user id in a script instead of `$USER`.  But that could be because I know nothing about kokoro!

jmicco: This was copped directly from the codelab and this directory likely wont exist in your actual repo.

elharo: Yes, I'm trying to bootstrap this and make sure it works, whihc is tricky since the configs are split across two repos; but I've sent a cl to create a non-codelab configuration on the internal side. Meanwhile I'll update this.

PR #663, opened 2016-09-02
Fixes #491 - Add a Cancel/Terminate button to the deploy console

### LINKED ISSUE #491: Add a Cancel/Terminate button to the deploy console
The deploy job can be cancelled by clicking on the icon on status bar that indicates that background jobs are running and then clicking on the corresponding terminate button. However it'd be easier for a user to click on a terminate button on the console showing the output of deploy like when the user stops the local dev app server.

akerekes:
Creates a new console for each deployment.
Also adds a close button to the console that is enabled once the deploy job is done.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
```diff
@@ -99,4 +89,14 @@
       </adapt>
     </definition>
   </extension>
+
+  <extension point="org.eclipse.ui.console.consolePageParticipants">
+    <consolePageParticipant
+        class="com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployConsolePageParticipant"
+        id="com.google.cloud.tools.eclipse.appengine.deploy.consolePageParticipant">
+      <enablement>
+        <instanceof value="com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployConsole"/>
```
briandealwis: This can't use the same `org.eclipse.ui.console.consoleTypeTest` property test as above?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+    toolbarManager.appendToGroup(IConsoleConstants.LAUNCH_GROUP, terminateAction);
+
+    closeAction = createCloseAction();
+    toolbarManager.appendToGroup(IConsoleConstants.LAUNCH_GROUP, closeAction);
+  }
+
+  private void addJobchangeListener() {
+    console.getJob().addJobChangeListener(new JobChangeAdapter() {
+      public void done(IJobChangeEvent event) {
+        System.out.println("done update");
```
briandealwis: This should be a log message

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsole.java
```diff
@@ -0,0 +1,33 @@
...
+    super(name, null);
+    setType(TYPE);
+  }
+
+  public StandardDeployJob getJob() {
+    return job;
+  }
+
+  public void setJob(StandardDeployJob deployJob) {
+    this.job = deployJob;
```
briandealwis: Is it possible that the job could be set *after* the participants are configured?  Perhaps fire a property change event here, and hook it in your termination participant? Just in case the order of operations varies.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+  private void configureToolBar(IToolBarManager toolbarManager) {
+    terminateAction = createTerminateAction();
+    toolbarManager.appendToGroup(IConsoleConstants.LAUNCH_GROUP, terminateAction);
+
+    closeAction = createCloseAction();
+    toolbarManager.appendToGroup(IConsoleConstants.LAUNCH_GROUP, closeAction);
+  }
+
+  private void addJobchangeListener() {
+    console.getJob().addJobChangeListener(new JobChangeAdapter() {
```
briandealwis: If order of `setJob` and _add-participants_ can vary you'll need to guard against null here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+      public void done(IJobChangeEvent event) {
+        System.out.println("done update");
+        update();
+      }
+    });
+  }
+
+  private void update() {
+    StandardDeployJob tag = console.getJob();
```
briandealwis: ditto re: nullness.  `tag` is an interesting choice of name.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+    close.setHoverImageDescriptor(getSharedImage(ISharedImages.IMG_ELCL_REMOVE));
+    close.setDisabledImageDescriptor(getSharedImage(ISharedImages.IMG_ELCL_REMOVE_DISABLED));
+    return close;
+  }
+
+  private Action createTerminateAction() {
+    Action terminate = new Action(Messages.getString("action.stop")) {
+      @Override
+      public void run() {
+        console.getJob().cancel();
```
briandealwis: nullness check

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+    if (terminateAction != null) {
+      terminateAction.setEnabled(tag.getState() != Job.NONE);
+    }
+
+    if (closeAction != null) {
+      closeAction.setEnabled(tag.getState() == Job.NONE);
+    }
+  }
+
+  private Action createCloseAction() {
```
briandealwis: This is interesting…  Do you really think developers will want this?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: Fixed

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
akerekes: Deleted

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+    terminate.setDisabledImageDescriptor(getSharedImage(ISharedImages.IMG_ELCL_STOP_DISABLED));
+    return terminate;
+  }
+
+  private ImageDescriptor getSharedImage(String image) {
+    return PlatformUI.getWorkbench().getSharedImages().getImageDescriptor(image);
+  }
+
+  @Override
+  public void activated() {
```
briandealwis: I think you need to call `update()` to update the enablement of the terminate action.  It could have terminated in that time.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+    StandardDeployJob tag = console.getJob();
+
+    if (terminateAction != null) {
+      terminateAction.setEnabled(tag.getState() != Job.NONE);
+    }
+
+    if (closeAction != null) {
+      closeAction.setEnabled(tag.getState() == Job.NONE);
+    }
+  }
```
briandealwis: It would also be useful to update the console description with the job state.  For example, a terminated launch shows "`<terminated> gcloud [Eclipse Application] …`" when dead.  (We should do the same with the server launch console too)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsole.java
```diff
@@ -0,0 +1,33 @@
+package com.google.cloud.tools.eclipse.appengine.deploy.ui;
+
+import org.eclipse.ui.console.MessageConsole;
+
+import com.google.cloud.tools.eclipse.appengine.deploy.standard.StandardDeployJob;
+import com.google.cloud.tools.eclipse.ui.util.MessageConsoleUtilities.ConsoleFactory;
+
+public class DeployConsole extends MessageConsole {
```
briandealwis: It would also be useful to update the console description with the job state. For example, a terminated launch shows "<terminated> gcloud [Eclipse Application] …" when dead. (We should do the same with the server launch console too)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -21,3 +21,7 @@ settings.advanced=Advanced
 stop.previous.version=Stop previous version
 use.custom.bucket=Override default Cloud Storage bucket for staging
 use.custom.versioning=Override generated version
+
+# Deploy console
+action.stop=Stop
+action.close=Close
```
briandealwis: JDT/Debug uses `Remove Launch`.  `Remove Deploy Log`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsole.java
akerekes: In the current implementation it is not possible, as the console is created on the UI thread as part of the command handler, and the console page creation also happens on the UI thread via `asyncExec`.
Nevertheless, it's definitely a fragile construction, so I'll look into it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
chanseokoh: null-check for `console` needed here.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/MessageConsoleUtilities.java
```diff
@@ -61,4 +61,31 @@ public static MessageConsole getMessageConsole(String consoleName,
     }
     return messageConsole;
   }
+  
+  public static <C extends MessageConsole> C findOrCreateConsole(String name, ConsoleFactory<C> factory) {
```
briandealwis: I wonder if this should be based on names — by using a timestamp for the deploy case, you'll never find a match.  Are we cleaning up old consoles?

An alternative mechanism would be to provide the console class, and / or a predicate to test.  `DeployConsole` could take the `IProject` and use that for comparison purposes.  That would allow having multiple deploys in parallel, but removing old/stale deploy logs.  (Or are older consoles automatically cleaned up?)

(This could be so much nicer with Java 8 with streams.)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,129 @@
...
+
+  private Action createCloseAction() {
+    Action close = new Action(Messages.getString("action.close")) {
+      @Override
+      public void run() {
+        ConsolePlugin.getDefault().getConsoleManager().removeConsoles(new IConsole[] { console });
```
chanseokoh: `console` may be null, so what happens it's null?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -170,4 +171,57 @@ public void onExit(int exitCode) {
     }
   }
 
+  public IProject getProject() {
+    return config.getProject();
+  }
+
+  public void setConfig(StandardDeployJobConfig config) {
```
chanseokoh: Is this used?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
akerekes: Yeah, the idea was to provide similar functionality.
If I'm correct and `org.eclipse.debug.internal.ui.views.console.ProcessConsolePageParticipant` is the class used there, then it's something more complex then what we need here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
akerekes: In my tests this is called when the user clicks into the console. If the job is terminated then job change listener should have handled it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsole.java
akerekes: +1

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
akerekes: `Remove`

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/MessageConsoleUtilities.java
akerekes: This method is actually not used by deploy, only the for local App Engine runs. I did not want to change that to use console type in this PR, just renamed the method to describe better what it does.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
akerekes: I changed to code (to be pushed soon) to not allow for a case when `console` is null.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
akerekes: Removed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
briandealwis: I believe it's also clicked when the user switches console pages.  So they could have switched out to something else and then back again.

akerekes: How would that interfere with the enablement of the actions? Not that I'm fighting against adding `update()` to `activate()` just curious, because I'd rather not add code that is redundant.

briandealwis: The job could have completed by the time the user switches back.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsole.java
```diff
@@ -20,7 +26,17 @@ public StandardDeployJob getJob() {
   }
 
   public void setJob(StandardDeployJob deployJob) {
-    this.job = deployJob;
+    firePropertyChange(this, PROPERTY_JOB, job, this.job = deployJob);
+    job.addJobChangeListener(new JobChangeAdapter() {
+      @Override
+      public void done(IJobChangeEvent event) {
+        PlatformUI.getWorkbench().getDisplay().asyncExec(new Runnable() {
```
briandealwis: Yuck, unfortunately I don't see a better way to get it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -53,34 +58,37 @@ private void configureToolBar(IToolBarManager toolbarManager) {
   }
 
   private void addJobchangeListener() {
```
briandealwis: Job_C_hange

briandealwis:
LGTM (barring the addJobChange name)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
```diff
@@ -0,0 +1,144 @@
...
+
+    closeAction = createCloseAction();
+    toolbarManager.appendToGroup(IConsoleConstants.LAUNCH_GROUP, closeAction);
+  }
+
+  private void addJobChangeListener() {
+    StandardDeployJob job = console.getJob();
+    if (job != null) {
+      job.addJobChangeListener(new JobChangeAdapter() {
+        public void done(IJobChangeEvent event) {
```
elharo: @Override I suspect

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -119,19 +124,26 @@ public void done(IJobChangeEvent event) {
     }
   }
 
+  private String getConsoleName(String project) {
+    Date now = new Date();
```
elharo: Google style strongly recommends using JodaTime instead of Date class. If all you're doing is showing the current time to the the user it probably doesn't matter that much. For anything else, use JodaTime.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -119,19 +124,26 @@ public void done(IJobChangeEvent event) {
...
+    String dateStr = DateFormat.getDateInstance(DateFormat.MEDIUM, Locale.getDefault()).format(now);
```
elharo: you could have a single string with getDateTimeInstance if you like

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployConsolePageParticipant.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
akerekes: Yes, I also prefer JodaTime. However it is not a dependency of our plugin yet, so I'm not going to add it just to get the current time (which as you noted doesn't really matter if we do it via `java.util.Date`)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
akerekes: Thanks, I missed that. Fixed!

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -119,19 +124,26 @@ public void done(IJobChangeEvent event) {
     }
   }
 
+  private String getConsoleName(String project) {
+    Date now = new Date();
+    String nowStr =
```
elharo: nowString

elharo:
LGTM

Issue #620, opened 2016-08-26
Should terminating a launch also stop the server?


briandealwis:
Terminating the launch doesn't stop the server instance.  Should it?

Minimal code I can find to do this, though I suspect there must be something else available.
```
    getLaunchManager().addLaunchListener(new ILaunchesListener2() {

      @Override
      public void launchesRemoved(ILaunch[] launches) {}

      @Override
      public void launchesChanged(ILaunch[] launches) {}

      @Override
      public void launchesAdded(ILaunch[] launches) {}

      @Override
      public void launchesTerminated(ILaunch[] launches) {
        for (ILaunch terminated : launches) {
          if (terminated == launch) {
            server.stop(true);
          }
        }
      }
    });
```

elharo:
I suspect it should, but what exactly do you mean by terminating the launch?

briandealwis:
If you terminate the _launch_ object in the _Debug_ view:

![screen shot 2016-08-29 at 9 18 18 am](https://cloud.githubusercontent.com/assets/202851/18052687/c8543a5c-6dc9-11e6-88f8-d5bbcd888d29.PNG)

the server remains in the _started_ state:

![screen shot 2016-08-29 at 9 19 07 am](https://cloud.githubusercontent.com/assets/202851/18052696/d1d2de1c-6dc9-11e6-83ac-8dfdc0f40c02.PNG)

I can't think of a reason why this would be useful.  I took a quick look at GPE and didn't see any obvious hooks in its WTP support, but I haven't yet tested.

elharo:
Seems reasonable.

chanseokoh:
@briandealwis I think #664 closes this?

briandealwis:
I knew there was another issue open, but couldn't find it!

PR #664, opened 2016-09-02
Cause server listen and react to to launch termination

### LINKED ISSUE #597: [Windows] Eclipse STOP button does not stop server process
Hello,

it seems that the STOP button in Eclipse Server view
image

Does not truly kill the server process.

Here you can see that I used the stop button
image

But the inside the Debug the process is still running
image

and even inside the Task Manager the Java process is running
image

Also... 
If I use the Terminate button on the running process
image

The Java process still is not killed
image

I need to explicitly kill it from the Task Manager

briandealwis:
#618 fixes caused the launch to be terminated when the server is stopped.
This fix handles the opposite case to ensure the server is stopped when the launch is terminated.

I also pulled the listeners into a single class to simplify listener cleanup.

Fixes #597

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
@@ -32,6 +32,7 @@ import org.eclipse.debug.core.DebugException;
 import org.eclipse.debug.core.ILaunch;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchManager;
+import org.eclipse.debug.core.ILaunchesListener2;
 import org.eclipse.jdt.launching.AbstractJavaLaunchConfigurationDelegate;
 import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 import org.eclipse.jdt.launching.IVMConnector;
@@ -61,6 +62,22 @@ import java.util.logging.Logger;
 
 public class LocalAppEngineServerLaunchConfigurationDelegate
     extends AbstractJavaLaunchConfigurationDelegate {
+
+  /**
+   * Encapsulates monitoring state for the server and its corresponding launch configuration.
+   */
+  public class LaunchMonitor {
+
+    /**
+     * @param configuration
+     * @param server
+     * @param launch
+     */
+    public LaunchMonitor(ILaunchConfiguration configuration, IServer server, ILaunch launch) {
+      // TODO Auto-generated constructor stub
+    }
+
+  }
 
   private final static Logger logger =
       Logger.getLogger(LocalAppEngineServerLaunchConfigurationDelegate.class.getName());
@@ -97,13 +114,28 @@ public class LocalAppEngineServerLaunchConfigurationDelegate
     console.clearConsole();
     console.activate();
 
-    if (shouldOpenStartPage()) {
-      String pageLocation = determinePageLocation(server, configuration);
-      if (pageLocation != null) {
-        // odd: addServerListener(..., IServer.SERVER_STARTED) doesn't work
-        server.addServerListener(new OpenBrowserListener(pageLocation));
-      }
+    stopServerOnLaunchTerminate(launch, server);
+    terminateLaunchOnServerStop(server, launch);
+    addPageOpenListener(configuration, server);
+
+    if (ILaunchManager.DEBUG_MODE.equals(mode)) {
+      int debugPort = getDebugPort();
+      setupDebugTarget(launch, configuration, debugPort, monitor);
+      serverBehaviour.startDebugDevServer(runnables, console.newMessageStream(), debugPort);
+    } else {
+      serverBehaviour.startDevServer(runnables, console.newMessageStream());
     }
+  }
+
+  /**
+   * Hook launch-termination so that it stops the server.
+   */
+  private void stopServerOnLaunchTerminate(ILaunch launch, IServer server) {
+    getLaunchManager().addLaunchListener(new StopServerOnLaunchTermination(launch, server));
+  }
+
+  /** Listen for server STOPPED state and terminate the launch */
+  private void terminateLaunchOnServerStop(IServer server, final ILaunch launch) {
     // If the server is stopped, then terminate any ancillary processes
     server.addServerListener(new IServerListener() {
       @Override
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -127,6 +109,46 @@ public void serverChanged(ServerEvent event) {
...
 
+  /**
+   * Listen for the server to enter STARTED and open a web browser on the server's main page
+   */
+  protected void openBrowserPage(final ILaunchConfiguration configuration, final IServer server) {
+    if (!shouldOpenStartPage()) {
+      return;
+    }
+    final String pageLocation = determinePageLocation(server, configuration);
+    if (pageLocation != null) {
```
akerekes: should not this use `==`?

chanseokoh:
I have not been able to reproduce #597 on Linux. I thought #597 happens only on Windows. I wonder why #597 doesn't happen on Linux even without this fix.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -127,6 +109,46 @@ public void serverChanged(ServerEvent event) {
...
+          int style = IWorkbenchBrowserSupport.LOCATION_BAR
+              | IWorkbenchBrowserSupport.NAVIGATION_BAR | IWorkbenchBrowserSupport.STATUS;
+          browserSupport.createBrowser(style, server.getId(), server.getName(), server.getName())
+              .openURL(url);
+        } catch (PartInitException ex) {
+          // Unable to use the normal browser support, so punt to the OS
+          logger.log(Level.WARNING, "Cannot launch a browser", ex);
+          Program.launch(pageLocation);
+        } catch (MalformedURLException ex) {
+          logger.log(Level.WARNING, "Unable to determine dev_appserver URL", ex);
```
akerekes: Shouldn't this be `Level.Error`? Also the message could be `Invalid dev_appserver URL`

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -175,50 +197,81 @@ private String determinePageLocation(IServer server, ILaunchConfiguration config
...
     }
 
     @Override
     public void serverChanged(ServerEvent event) {
-      if (event.getState() != IServer.STATE_STARTED) {
-        return;
-      }
-      final IServer server = event.getServer();
-      final IWorkbench workbench = PlatformUI.getWorkbench();
+      assert server == event.getServer();
```
akerekes: We shouldn't use assert. Maybe `Preconditions`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
briandealwis: Oops.  I had had a browser page up during testing and hadn't noticed.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
briandealwis: Ok.  Is there a rationale?  I suppose it's because nobody runs with `-enableassertions` and so they're never caught…

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
briandealwis: True.  It's a generated URL and so shouldn't happen.  I changed the error text too as it's an invalid URL.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
akerekes: Yeah, that's what I'd use as rationale.

briandealwis:
> I have not been able to reproduce #597 on Linux. I thought #597 happens only on Windows. I wonder why #597 doesn't happen on Linux even without this fix.

I think it only happens if you start up the datastore.  My test app has a little warmup servlet (below).  When I terminate the launch, the server waits a bit before emitting a message like:
```
INFO     2016-09-02 21:29:12,060 api_server.py:648] Applying all pending transactions and saving the datastore
INFO     2016-09-02 21:29:12,068 api_server.py:651] Saving search indexes
```

Here's the servlet:
```
public class WarmupServlet extends HttpServlet {
	private static final Logger logger = Logger.getLogger(WarmupServlet.class.getName());

	public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
		logger.info("Brr it's chilly in here! Time to warm up!");
		
		DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
		
		Entity employee = new Entity("Employee", "asalieri");
		employee.setProperty("firstName", "Antonio");
		employee.setProperty("lastName", "Salieri");
		employee.setProperty("hireDate", new Date());
		employee.setProperty("attendedHrTraining", true);
		datastore.put(employee);

		response.setStatus(HttpServletResponse.SC_OK);
		response.addHeader("FOO", "BAR");
		response.flushBuffer();
	}
}
```

akerekes:
LGTM (provided the `assert` is replaced with something more explicit)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
elharo: yes, google forbids asserts in its Java code for this reason.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -175,50 +197,81 @@ private String determinePageLocation(IServer server, ILaunchConfiguration config
...
+          } catch (DebugException ex) {
+            logger.log(Level.WARNING, "Unable to terminate launch", ex);
           }
-          return Status.OK_STATUS;
+      }
+    }
+
+    @Override
+    public void launchesTerminated(ILaunch[] launches) {
+      for (ILaunch l : launches) {
```
elharo: please don't use l as local variable name; depending on font it may not be readable; maybe launch and this.launch?

briandealwis:
Fixes #620

PR #673, opened 2016-09-08
Create an IDebugTarget for the Dev App Server

### LINKED ISSUE #614: _Run_ launches must register a debug target or process with the ILaunch

briandealwis:
Required for _Run_ launches to be able to be terminated.  Not required for _Debug_ launches as the the remote JVMs are added to the launch as they attach to the debugger.

Fixes #614 

Hard to test: will enhance the ITs test to test the various combinations separately.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDebugTarget.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDebugTarget.java
@@ -19,11 +19,11 @@ package com.google.cloud.tools.eclipse.appengine.localserver.server;
 import org.eclipse.core.resources.IMarkerDelta;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
-import org.eclipse.debug.core.DebugEvent;
 import org.eclipse.debug.core.DebugException;
 import org.eclipse.debug.core.DebugPlugin;
 import org.eclipse.debug.core.ILaunch;
 import org.eclipse.debug.core.ILaunchesListener;
+import org.eclipse.debug.core.model.DebugElement;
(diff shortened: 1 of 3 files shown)


chanseokoh:
I absolutely have no idea, but it sounds like you said in #614 we might be able to use `DebugPlugin.newProcess()` if we can get a `java.lang.Process` of `CloudSdk`? (I believe getting a process is trivial with `addStartListener`, which is being done for deployment.) By that way, I wonder if we can avoid implementing `IDebugTarget` ourselves, if that's simpler. But I really have no idea about what's going on here; just throwing out a random thought.

briandealwis:
I didn't explore the `IProcess` approach.  My intuition was that the _Stop/Terminate_ button would actually call `Process#destroy()` on that instance, and we wouldn't necessarily get any proper server cleanup.  The `IDebugTarget` seemed conceptually simple.

The other implementations of `IDebugTarget` inherit from a `DebugElement` abstract class, but it seemed more of a hindrance than a help :-/

briandealwis:
> My intuition was that the Stop/Terminate button would actually call Process#destroy() on that instance

And [confirmed](https://github.com/eclipse/eclipse.platform.debug/blob/master/org.eclipse.debug.core/core/org/eclipse/debug/core/model/RuntimeProcess.java#L204).

chanseokoh:
Well thought out!

briandealwis:
>  a DebugElement abstract class, but it seemed more of a hindrance than a help :-/

But inheriting from `DebugElement` does bring some other advantages, like the adapter support.  I'll revisit this.

elharo:
LGTM assuming tests pass

PR #674, opened 2016-09-08
Fixes #642 - Allow specify all deploy preferences on deploy

### LINKED ISSUE #642: Allow specify all deploy preferences on deploy
Convert the "Prompt for Project ID" checkbox into a general "Prompt for preferences" checkbox which is when set will cause the prompt dialog come up each time the user deploys.
Change the prompt dialog to prompt for all deploy preferences not only for project ID.

akerekes:
Since we display the dialog on each deploy, the setting to prompt for project ID on deploy has been removed.

image


patflynn:
stop previous version only applies to flex projects where 'promote' is true.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -147,55 +141,15 @@ private StandardDeployJobConfig getDeployJobConfig(IProject project,
...
-                      // The input dialog does not provide decoration for the error message,
-                      // make errors more recognizable with modified validation messages
-                      // FIXME https://github.com/GoogleCloudPlatform/cloud-tools-for-eclipse/issues/635
-                      public String isValid(String newText) {
-                        String error = super.isValid(newText);
-                        if (error != null) {
-                          if (error.isEmpty()) {
-                            return Messages.getString("error.projectId.invalid");
-                          } else {
-                            return Messages.getString("error.projectId.prefix", error);
```
chanseokoh: I see no usage of `error.projectId.prefix elsewhere`. Removing from `message.properties`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -147,55 +141,15 @@ private StandardDeployJobConfig getDeployJobConfig(IProject project,
...
     if (deployPreferences.getProjectId() == null || deployPreferences.getProjectId().isEmpty()) {
       throw new ExecutionException(Messages.getString("error.projectId.missing"));
     }
     return new StandardDeployPreferencesConverter(deployPreferences).toDeployConfiguration();
   }
 
-  private String promptForProjectId(Shell shell, String initialValue) throws DeployCancelledException {
-    InputDialog dialog = new InputDialog(shell,
-                    Messages.getString("dialog.prompt.projectId.title"),
-                    Messages.getString("dialog.prompt.projectId.message"),
```
chanseokoh: Looks like `dialog.prompt.projectId.*` are no longer used. Remove form `messages.properties`?

akerekes:
@patflynn `stop-previous-version` works with Standard as well if the scaling is set to `basic` or `manual`. I just did a quick test with basic scaling to verify it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
akerekes: done

patflynn:
Interesting. @etanshall, something to think about.

Andras, can you add a screenshot of what this UI looks like?

akerekes:
Screenshot added to description

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,114 @@
...
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.ISharedImages;
+import org.eclipse.ui.PlatformUI;
+
+import com.google.common.base.Preconditions;
+
+public class DeployPreferencesDialog extends Dialog {
```
briandealwis: Typically new dialogs inherit from `TrayAreaDialog`.  The tray is a little area that can swipe out on the right, which can be used for hosting in-place help.  It also provides a message area (see #setErrorMessage and #setMessage).  That should cut down on the code required here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,114 @@
...
+    Composite dialogArea = (Composite) super.createDialogArea(parent);
+    ScrolledComposite scrolledComposite = new ScrolledComposite(dialogArea, SWT.V_SCROLL | SWT.H_SCROLL);
+    scrolledComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+    scrolledComposite.setShowFocusedControl(true);
+    scrolledComposite.setExpandHorizontal(true);
+    scrolledComposite.setExpandVertical(true);
+    scrolledComposite.addControlListener(new ControlListener() {
+      @Override
+      public void controlResized(ControlEvent e) {
+        Point computeSize = DeployPreferencesDialog.this.getShell().computeSize(SWT.DEFAULT, SWT.DEFAULT);
```
briandealwis: I don't think using the *Shell*'s size is correct.  You should be using the ScrolledComposite's content.

patflynn:
I really like the wording here. I think that you've really improved clarity of what these flags do.  

One thing we did is made the 'stop previous version' checkbox a child of 'promote' and only enabled if promote is set.  To have 'promote' unchecked and 'stop previous version' checked apparently does nothing.

@etanshaul maybe we can borrow some of this wording for our UI.

etanshaul:
:+1: yes I like the wording too. "promote" alone never really meant much to me. I'll incorporate this into IntelliJ as well.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,114 @@
...
+  
+  private void updateValidationStatus(String errorMessage) {
+    Preconditions.checkNotNull(errorMessage, "errorMessage is null");
+    
+    if (!validationMessage.isDisposed()) {
+      getButton(IDialogConstants.OK_ID).setEnabled(errorMessage.isEmpty());
+      validationMessage.setText(errorMessage);
+      validationMessage.setImage(errorMessage.isEmpty() ? null : ERROR_ICON);
+      validationMessage.setVisible(!errorMessage.isEmpty());
+      ((GridData)validationMessage.getLayoutData()).exclude = errorMessage.isEmpty();
```
elharo: nit: space after cast 
https://google.github.io/styleguide/javaguide.html#s4.6-whitespace

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,411 @@
...
+    projectIdLabel = new Label(projectIdComp, SWT.LEFT);
+    projectIdLabel.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false));
+    projectIdLabel.setText(Messages.getString("project.id"));
+
+    projectId = new Text(projectIdComp, SWT.LEFT | SWT.SINGLE | SWT.BORDER);
+    projectId.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+  }
+
+  private void createProjectVersionSection() {
+    Composite versionComp = new Composite(this, SWT.NONE);
```
elharo: Google style is no abbreviations; i.e. versionComposite

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,411 @@
...
+    GridData layoutData = new GridData(SWT.LEFT, SWT.CENTER, false, false);
+    layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
+    versionLabel.setLayoutData(layoutData);
+
+    version = new Text(versionComp, SWT.LEFT | SWT.SINGLE | SWT.BORDER);
+    version.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
+  }
+
+  private void createPromoteSection() {
+    Composite promoteComp = new Composite(this, SWT.NONE);
```
elharo: promoteComposite

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,411 @@
...
+  }
+
+  /**
+   * Update the ancestral ScrolledComposite that there's been a change.
+   */
+  protected void updateLayout(ExpandableComposite expandableComposite) {
+    Control parent = expandableComposite.getParent();
+    while (parent != null) {
+      if (parent instanceof ScrolledComposite) {
+        ScrolledComposite sc = (ScrolledComposite) parent;
```
elharo: sc --> scrolledComposite

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,411 @@
...
+        new ExpandableComposite(this, SWT.NONE, ExpandableComposite.TWISTIE | ExpandableComposite.CLIENT_INDENT);
+    expandableComposite.setText(Messages.getString("settings.advanced"));
+    expandableComposite.setExpanded(false);
+    expandableComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+    FontUtil.convertFontToBold(expandableComposite);
+    return expandableComposite;
+  }
+
+  private Composite createBucketSection(Composite parent) {
+    Composite defaultBucketComp = new Composite(parent, SWT.NONE);
```
elharo: defaultBucketComposite or defaultBucketSection if you prefer

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,411 @@
...
+    promoteComp.setLayout(new GridLayout(1, false));
+    autoPromoteButton = new Button(promoteComp, SWT.CHECK);
+    autoPromoteButton.setText(Messages.getString("auto.promote"));
+    autoPromoteButton.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, true, false));
+
+    Link manualPromoteLink = new Link(promoteComp, SWT.NONE);
+    GridData layoutData = new GridData(SWT.LEFT, SWT.CENTER, true, false);
+    layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
+    manualPromoteLink.setLayoutData(layoutData);
+    manualPromoteLink.setText(Messages.getString("deploy.manual.link", APPENGINE_DASHBOARD_URL));
```
etanshaul: This doesn't look like the right link for manual promoting.

etanshaul: What about using this link: `https://pantheon.corp.google.com/appengine/versions` ?

This is what I currently have in intellij. But we could go further though and add:

`https://pantheon.corp.google.com/appengine/versions?project={project_name}`

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,114 @@
...
+    validationMessage.setLayoutData(layoutData);
+    validationMessage.setVisible(false);
+
+    Composite dialogArea = (Composite) super.createDialogArea(parent);
+    ScrolledComposite scrolledComposite = new ScrolledComposite(dialogArea, SWT.V_SCROLL | SWT.H_SCROLL);
+    scrolledComposite.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+    scrolledComposite.setShowFocusedControl(true);
+    scrolledComposite.setExpandHorizontal(true);
+    scrolledComposite.setExpandVertical(true);
+    scrolledComposite.addControlListener(new ControlListener() {
```
chanseokoh: Using `ControlAdapter` instead of `ControlListener` will make the code simpler which allows to remove the no-op `controlMoved()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,114 @@
...
+  
+  
+  private void updateValidationStatus(String errorMessage) {
+    Preconditions.checkNotNull(errorMessage, "errorMessage is null");
+    
+    if (!validationMessage.isDisposed()) {
+      getButton(IDialogConstants.OK_ID).setEnabled(errorMessage.isEmpty());
+      validationMessage.setText(errorMessage);
+      validationMessage.setImage(errorMessage.isEmpty() ? null : ERROR_ICON);
+      validationMessage.setVisible(!errorMessage.isEmpty());
```
chanseokoh: If we are hiding/showing this control, maybe we could remove the call `setImage()` right above?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -84,6 +81,7 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
       if (project != null) {
         Credential credential = loginIfNeeded(event);
         if (credential != null) {
+          new DeployPreferencesDialog(HandlerUtil.getActiveShell(event), project).open();
```
chanseokoh: I think you need to check the output of the dialog to see if a user pressed cancel. Currently, clicking the cancel button is like clicking the OK button, and it gives the following exception:

`org.eclipse.core.commands.ExecutionException: Project ID must be set for deploy.`

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -1,114 +1,88 @@
...
-        if (status.isOK()) {
-          updateValidationStatus("");
-        } else {
-          updateValidationStatus(status.getMessage());
-        }
+      public void handleExpansionEvent(ExpansionEvent e) {
+        Shell shell = getShell();
+        shell.setMinimumSize( shell.getSize().x, 0 );
+        shell.pack();
+        ((ExpandableComposite)e.getSource()).getParent().layout();
```
elharo: space after cast

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -408,4 +381,12 @@ public void dispose() {
     }
     super.dispose();
   }
+
+  private void handleExpansionStateChanged(final ExpansionEvent e) {
+    expansionHandler.handleExpansionEvent(e);
+  }
+
+  public static interface AdvancedSectionExpansionHandler {
+    void handleExpansionEvent(ExpansionEvent e);
```
elharo: e --> event

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -408,4 +381,12 @@ public void dispose() {
     }
     super.dispose();
   }
+
+  private void handleExpansionStateChanged(final ExpansionEvent e) {
```
elharo: e --> event
I don't think the argument needs to be final

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -31,11 +35,29 @@
...
+  private AdvancedSectionExpansionHandler getExpansionHandler() {
+    return new AdvancedSectionExpansionHandler() {
+
+      @Override
+      public void handleExpansionEvent(ExpansionEvent e) {
+        Composite parent = ((Control)e.getSource()).getParent();
+        parent.layout();
+        while (parent != null) {
+          if (parent instanceof ScrolledComposite) {
+            ScrolledComposite sc = (ScrolledComposite) parent;
```
elharo: sc --> scrolledComposite

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,392 @@
...
+  private void createPromoteSection() {
+    Composite promoteComposite = new Composite(this, SWT.NONE);
+    autoPromoteButton = new Button(promoteComposite, SWT.CHECK);
+    autoPromoteButton.setText(Messages.getString("auto.promote"));
+
+    Link manualPromoteLink = new Link(promoteComposite, SWT.NONE);
+    GridData layoutData = GridDataFactory.swtDefaults().create();
+    layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
+    manualPromoteLink.setLayoutData(layoutData);
+    manualPromoteLink.setText(Messages.getString("deploy.manual.link", APPENGINE_DASHBOARD_URL));
```
elharo: Is this the right link, or should it be the one Etan suggested?

chanseokoh:
I can press the OK button while the project ID validation fails.

deployment preferences _008

Also, my project doesn't set any default project ID (i.e., created via the wizard while leaving the project ID field blank. I noticed that when the preference page first pops up, the project ID validation isn't executed.

deployment preferences _007

Lastly, dialog margins don't seem to be set properly.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
akerekes: I remember now why I decided to send the user to the App Engine dashboard: if they have worked with a different project previously, the web console will bring that up by default, not the current project. But I agree it's closer to where the user should be, also if we have the project ID we can add it to the URL as you suggested.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -45,12 +45,15 @@ private AdvancedSectionExpansionHandler getExpansionHandler() {
 
       @Override
       public void handleExpansionEvent(ExpansionEvent e) {
-        Composite parent = ((Control)e.getSource()).getParent();
-        parent.layout();
+        Composite expandableComposite = (Composite)e.getSource();
+        DeployPreferencesPanel panel = (DeployPreferencesPanel) expandableComposite.getParent();
```
briandealwis: Since the `DeployPropertyPage` creates the panel, why don't you stash it in a field.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -45,12 +45,15 @@ private AdvancedSectionExpansionHandler getExpansionHandler() {
...
+        Composite parent = panel.getParent();
         while (parent != null) {
```
briandealwis: Reference [Eclipse bug 265237](https://bugs.eclipse.org/bugs/show_bug.cgi?id=265237)?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,88 @@
...
+import org.eclipse.ui.forms.widgets.ExpandableComposite;
+
+import com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployPreferencesPanel.AdvancedSectionExpansionHandler;
+import com.google.cloud.tools.eclipse.appengine.ui.AppEngineImages;
+
+public class DeployPreferencesDialog extends TitleAreaDialog {
+
+  // if the image is smaller (e.g. 32x32, it will break the layout of the TitleAreaDialog)
+  // seems like an Eclipse/JFace bug
+  private Image titleImage = AppEngineImages.appEngine(64).createImage();
```
briandealwis: Images should be created in `createContents()` to ensure it's created on the SWT thread.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -0,0 +1,392 @@
...
+      observables.dispose();
+    }
+    super.dispose();
+  }
+
+  private void handleExpansionStateChanged(final ExpansionEvent e) {
+    expansionHandler.handleExpansionEvent(e);
+  }
+
+  public static interface AdvancedSectionExpansionHandler {
```
briandealwis: This name seems to expose too specific and too detail focussed.  For example, we add another section.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: It's already in a field (`content`), thanks for pointing this out.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: It's 7 years old and nothing happened to it since then... anyways, I'll add it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
akerekes: <joke>considering how much effort was to make resizing work properly (?), I don't think we should add more collapsible sections</joke>
I'll rename it to `SectionExpansionHandler`.

akerekes:
Latest screenshot:

image

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: It's so Future Us remember that this code was added to work around a problem

briandealwis:
The dialog should be retitled as it is opened on the actual "Deploy" operation.  _Deployment Configuration_?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,116 @@
...
+        return type;
+      }
+    });
+    return dialogArea;
+  }
+
+  private void setMargin(Composite dialogArea) {
+    Layout layout = dialogArea.getLayout();
+    if (layout instanceof GridLayout) {
+      ((GridLayout)layout).marginWidth = 5;
```
elharo: space after cast

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/event/OpenUriSelectionListener.java
```diff
@@ -0,0 +1,76 @@
...
+      errorHandler.handle(ex);
+    }
+  }
+
+  private URI appendProjectId(URI uri, String projectId) throws URISyntaxException {
+    if (projectId != null && !projectId.isEmpty() && !projectId.trim().isEmpty()) {
+      projectId = projectId.trim();
+      String query = uri.getQuery();
+      if (query == null) {
+        query = URI_PARAM_PROJECT + "=" + projectId;
```
elharo: There's a potential problem here if the project ID contains a colon, which it may (google.com:something). That needs to be URL encoded.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/event/OpenUriSelectionListener.java
```diff
@@ -0,0 +1,76 @@
...
+      } else {
+        query += "&" + URI_PARAM_PROJECT + "=" + projectId;
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -247,9 +249,14 @@ private void createPromoteSection() {
     GridData layoutData = GridDataFactory.swtDefaults().create();
     layoutData.horizontalIndent = INDENT_CHECKBOX_ENABLED_WIDGET;
     manualPromoteLink.setLayoutData(layoutData);
-    manualPromoteLink.setText(Messages.getString("deploy.manual.link", APPENGINE_DASHBOARD_URL));
+    manualPromoteLink.setText(Messages.getString("deploy.manual.link", APPENGINE_VERSIONS_URL));
```
briandealwis: Do we need a note here to handle cases where the Project ID doesn't exist?  We could have a combo box, etc.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,116 @@
...
+        shell.pack();
+        ((ExpandableComposite) event.getSource()).getParent().layout();
+        shell.setMinimumSize(shell.getSize());
+      }
+    };
+  }
+
+  @Override
+  protected void okPressed() {
+    content.savePreferences();
```
briandealwis: An aside, it feels a bit strange that a the *panel* knows how to persist the model. It would seem the responsibility of the dialog or the dialog's creator.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,116 @@
...
+  protected Control createDialogArea(final Composite parent) {
+    Composite dialogArea = (Composite) super.createDialogArea(parent);
+    setMargin(dialogArea);
+    content = new DeployPreferencesPanel(dialogArea, project, getExpansionHandler());
+    GridDataFactory.fillDefaults().grab(true, false).applyTo(content);
+    TitleAreaDialogSupport.create(this, content.getDataBindingContext()).setValidationMessageProvider(new ValidationMessageProvider() {
+      @Override
+      public int getMessageType(ValidationStatusProvider statusProvider) {
+        int type = super.getMessageType(statusProvider);
+        Button okButton = getButton(IDialogConstants.OK_ID);
```
briandealwis: The actual button-enablement part seems better extracted to a `setValid()` method

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -0,0 +1,116 @@
...
+    // TitleAreaDialogSupport does not validate initially, let's trigger validation this way
+    content.getDataBindingContext().updateTargets();
+
+    return contents;
+  }
+  
+  @Override
+  protected Control createDialogArea(final Composite parent) {
+    Composite dialogArea = (Composite) super.createDialogArea(parent);
+    setMargin(dialogArea);
```
briandealwis: So `setMargin()` doesn't produce the right results as you can see in the following image: note the separator is now indented.
screen shot 2016-09-12 at 9 51 25 pm
It turns out `TitleAreaDialog` overrides `Dialog#createDialogArea()` and creates the content area `Composite` with 0 margins so that the separator (shown above) goes to the edge.  What it should have done is create and return a child composite with the same margins.

I'll push a change that DTRT here, following practices in JDT to create a child composite and applying the same margins and spacing as was applied in `Dialog#createDialogArea()`.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/event/OpenUriSelectionListener.java
akerekes: It seems that `:` is not required to be encoded: http://stackoverflow.com/questions/13713671/do-colons-require-encoding-in-uri-query-parameters

From [rfc3986](https://tools.ietf.org/html/rfc3986):
https://tools.ietf.org/html/rfc3986#section-3.4
```
query       = *( pchar / "/" / "?" )
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
```
so `:` is explicitly allowed.
The multiparameter constructor of `URI` used here does escaping as needed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
akerekes: Thanks!

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
akerekes: done

akerekes:
@elharo @chanseokoh @briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/event/OpenUriSelectionListener.java
```diff
@@ -42,35 +45,37 @@ public void widgetDefaultSelected(SelectionEvent event) {
 
   private void openAppEngineDashboard(String uriString) {
```
briandealwis: This seems a funny name for a method in the .ui.util?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/event/OpenUriSelectionListener.java
```diff
@@ -42,35 +45,37 @@ public void widgetDefaultSelected(SelectionEvent event) {
...
+    StringBuilder query = new StringBuilder(queryString);
+    for (Entry<String, String> parameter : queryParameterProvider.getParameters().entrySet()) {
+      if (query.length() > 0) {
+        query.append('&');
       }
-      return new URI(uri.getScheme(), uri.getUserInfo(), uri.getHost(),
-                     uri.getPort(), uri.getPath(), query, uri.getFragment());
-    } else {
-      return uri;
+      query.append(parameter.getKey())
```
briandealwis: Don't we need to ensure the parameter name and values are URL encoded?  (e.g., in case the key contains a `=`, or the value contains a `&`)

akerekes: The multiparameter URI constructor does escaping.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/event/OpenUriSelectionListener.java
akerekes: indeed, will fix it.

briandealwis:
LGTM

PR #677, opened 2016-09-08
remove import to let the Messages class in the same package to be used

akerekes:
After experimenting with DeployConsolePageParticipant the import for the Messages class was not cleaned up and thus the tooltips for the action buttons were missing.


chanseokoh:
LGTM

elharo:
Wow, this is a bug class I haven't run into before. I wonder if there's a findbugs check for this? Anyway, LGTM.

akerekes:
We could put the Messages classes into their own separate package that should never be exported from the bundle.

chanseokoh:
That being said, our codebase currently has two different types of Messages class: one auto-generated by Eclipse with `Externalize Strings...` that uses `org.eclipse.osgi.util.NLS`, and another one that uses the standard java `ResourceBundle`.

akerekes:
I think both classes are generated by Eclipse. I'd prefer to use the one that uses `NLS` I just could not find a way to generate that, so all the `Messages` classes my IDE generated uses `ResourceBundle`

elharo:
Let's not overthink this.  For now, let'd merge this fix. We can improve it later.

PR #683, opened 2016-09-12
don't install extra jars by default

elharo:


briandealwis:
Wouldn't it be easier to not install the App Engine facet by default when adding the runtime?

elharo:
Presumably we need the facet to ID projects as app engine projects, even if it's not adding jars to the classpath. The change here is that we do not automatically add these jars to all app engine apps.

briandealwis:
I had been thinking that we could check if the project specified an _App Engine Standard_ runtime. But I'll retract that: having a facet means we could have a project targeted to the generic HTTP runtime.

briandealwis:
So shouldn't we remove this classpath initializer and the corresponding `org.eclipse.jdt.core.classpathContainerInitializer` extension in `plugin.xml` too?

Seems we should also add support for a facet detector (`org.eclipse.wst.common.project.facet.core.detectors`) to automagically detect and add the facet on import.

akerekes:
If we don't install anything, why don't we just get rid of the whole class?

briandealwis:
Or better yet, move it to a new facet.

elharo:
PTAL. I was trying for the minimal necessary change, but now I've removed the installation of the jars when the facet is installed.

akerekes:
LGTM (assuming Travis passes)

PR #684, opened 2016-09-12
gcloud to 125.0.0

elharo:
@nbashirbello @briandealwis @akerekes Travis seems to be timing out when it tries to upgrade 124 to 125


briandealwis:
LGTM

PR #699, opened 2016-09-14
Create App Engine Flex Facet

nbashirbello:
Allows users to install and uninstall the flex facet to any project that has the Java 1.8 facet installed.


nbashirbello:
@elharo

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/flex/AppEngineFlexFacetTest.java
```diff
@@ -0,0 +1,12 @@
+package com.google.cloud.tools.eclipse.appengine.facets.flex;
```
elharo: Is there a reason for an extra package? I would think one package for the different facets is enough.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.properties
```diff
@@ -2,5 +2,8 @@ google = Google
 Bundle-Vendor = Google Inc.
 Bundle-Name = Facets for Google Cloud Platform Apps
 
-facetName = Google App Engine Standard Java
-facetDescription = Supply App Engine Functionality
+flexFacetName = Google App Engine Flex Java
+flexFacetDescription = Supply App Engine Flex Functionality
```
chanseokoh: Are we good to use the abbreviated `Flex`?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -4,8 +4,8 @@
    
    <extension point="org.eclipse.wst.common.project.facet.core.facets">        
     <project-facet id="com.google.cloud.tools.eclipse.appengine.facet">
```
chanseokoh: Add `.standard` to align with the flex id?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -4,8 +4,8 @@
...
-       <label>%facetName</label>
-       <description>%facetDescription</description>
+       <label>%standardFacetName</label>
+       <description>%standardFacetDescription</description>
     </project-facet>
     <project-facet-version facet="com.google.cloud.tools.eclipse.appengine.facet" version="1">
```
chanseokoh: Ditto: `.standard`

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -22,13 +22,13 @@
       id="com.google.cloud.tools.eclipse.appengine.facet.install.action"
       facet="com.google.cloud.tools.eclipse.appengine.facet"
```
chanseokoh: Ditto: `.standard`?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
briandealwis: My question is somewhat related: what are the trade-offs of having separate facets for _flex_ vs _standard_ as compared to having a single facet with two different versions?  The facet versions can be strings, like _Flex_ and _Standard_, and we can provide a comparator too.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.properties
patflynn: I don't think so. 

This naming seems off to me. 

II would remove the 'google' and call these facets 'App Engine Standard' and 'App Engine Flexible Environment'.

@arouzrokh should review this.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
patflynn: FYI, I don't know what the right thing for Eclipse is in this case, but just figured I'd chime in on the differences of the environments:
There are pretty fundamental differences between Standard and Flexible.  The APIs you can use are different, the build steps required are different, and the parameters to deploy are different.

nbashirbello: You can put "conflicts" requirement on the flex facet that it conflicts with  the standard facet. I actually put it in and took it out for 2 reasons: (i) standard and flex depend on different versions of java (7 and 8 respectively), so that prevents users from adding both facets to a project, (ii) I was thinking about compat, I am not quite sure of how it works but theoretically, a user can deploy a standard project to a flex environment

nbashirbello:
I am planning on renaming the App Engine Standard id. The standard id is currently "com.google.cloud.tools.eclipse.appengine.facet", which was named after the package name. I want to change it to "com.google.cloud.tools.eclipse.appengine.facet.standard" and have the flex id "com.google.cloud.tools.eclipse.appengine.facet.flex". One thing I want to note is that after this change, instances of the App Engine standard facets in the old release would not be automatically updated and users would have to reinstall the new standard facet. Since we are still in fishfood versions, I think its ok. But should we hold off before the soft launch?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineRuntimeChangeListenerTest.java
```diff
@@ -27,9 +27,11 @@
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardRuntimeChangeListener;
```
chanseokoh: This is probably not needed because we are at the same package now.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacetTest.java
```diff
@@ -17,19 +17,27 @@
...
 
+import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
 import org.eclipse.wst.server.core.IRuntimeType;
 import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
```
chanseokoh: Ditto.

plugins/com.google.cloud.tools.eclipse.appengine.facets/META-INF/MANIFEST.MF
```diff
@@ -30,4 +30,5 @@ Import-Package: com.google.cloud.tools.eclipse.util,
  org.eclipse.wst.common.project.facet.core.events,
  org.eclipse.wst.common.project.facet.core.runtime,
  org.eclipse.wst.server.core
-Export-Package: com.google.cloud.tools.eclipse.appengine.facets
+Export-Package: com.google.cloud.tools.eclipse.appengine.facets,
+ com.google.cloud.tools.eclipse.appengine.facets.standard
```
chanseokoh: Ditto.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/META-INF/MANIFEST.MF
```diff
@@ -21,6 +21,7 @@ Require-Bundle: com.google.cloud.tools.eclipse.appengine.newproject;bundle-versi
  org.eclipse.m2e.maven.runtime
 Bundle-ActivationPolicy: lazy
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
+ com.google.cloud.tools.eclipse.appengine.facets.standard,
```
chanseokoh: Ditto.

nbashirbello:
Hey, I addressed all the comments. Any more changes needed?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
```diff
@@ -0,0 +1,14 @@
...
+import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineFlexFacet;
+
+public class AppEngineFlexFacetTest {
+  @Test
+  public void testFlexFacetExists() {
+    Assert.assertTrue(ProjectFacetsManager.isProjectFacetDefined(AppEngineFlexFacet.ID));
```
elharo: If you use the string literal here instead of the constant, then you also test the value of the constant.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacetTest.java
```diff
@@ -30,6 +31,11 @@
   @Mock private IRuntimeType runtimeType;
 
   @Test
+  public void testStandardFacetExists() {
+    Assert.assertTrue(ProjectFacetsManager.isProjectFacetDefined(AppEngineStandardFacet.ID));
```
elharo: If you use the string literal here instead of the constant, then you also test the value of the constant.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
```diff
@@ -0,0 +1,17 @@
...
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.wst.common.project.facet.core.IDelegate;
+import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
+
+public class FlexFacetInstallDelegate implements IDelegate {
+
+  @Override
+  public void execute(IProject project, IProjectFacetVersion fv, Object config, IProgressMonitor monitor)
```
elharo: no abbreviations please; fv --> facetVersion

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
briandealwis: Meaning that you pick up accidental changes.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacet.java
```diff
@@ -0,0 +1,5 @@
+package com.google.cloud.tools.eclipse.appengine.facets;
+
+public class AppEngineFlexFacet {
```
briandealwis: This would be a good place for a `isFlexProject` (or `isFlexible` :-))

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
```diff
@@ -0,0 +1,17 @@
...
+      throws CoreException {
+
```
briandealwis: Todo to create minimal `app.yaml` if not present?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetUninstallDelegate.java
```diff
@@ -0,0 +1,17 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.wst.common.project.facet.core.IDelegate;
+import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
+
+public class FlexFacetUninstallDelegate implements IDelegate {
+
+  @Override
+  public void execute(IProject project, IProjectFacetVersion fv, Object config, IProgressMonitor monitor)
+      throws CoreException {
+
```
briandealwis: Do we remove the generated `app.yaml` if untouched?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
nbashirbello: We decided not to do that since users can do deploy with a project that doesn't have the app.yaml because gcloud would create one on its own.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetUninstallDelegate.java
nbashirbello: We decided not to do that since users can do deploy with a project that doesn't have the app.yaml because gcloud would create one on its own.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
briandealwis: Huh, that's handy.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacet.java
nbashirbello: Expanding this class in a subsequent pr

elharo:
LGTM

PR #707, opened 2016-09-16
Fixes #129 - Generated native project should include a unit test for HelloAppEngine servlet (called HelloAppEngineTest)

### LINKED ISSUE #129: Generated native project should include a unit test for HelloAppEngine servlet (called HelloAppEngineTest)
for the HelloAppEngine servlet, autogenerated by the wizard

akerekes:


plugins/com.google.cloud.tools.eclipse.util/templates/appengine/MockHttpServletResponse.java.ftl
```diff
@@ -0,0 +1,174 @@
...
+import java.util.Locale;
+
+import javax.servlet.ServletOutputStream;
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * This class is created only for the purpose of demonstrating unit testing of the
+ * {@link HelloAppEngine} class and the only those methods have meaningful implementation
+ * which are use in the unit test.
```
chanseokoh: I guess what you tried to say is

"Only those methods that have meaningful implementation are used in the unit test."

or

"Only those methods that are used in the unit test have meaningful implementation."

I think it's also good to cut this long statement into two, e.g., `... HelloAppEngine} class. (period) Only those ...`

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/MockHttpServletResponse.java.ftl
```diff
@@ -0,0 +1,174 @@
...
+ * 
+ * Feel free to change this class or replace it using other ways for testing HttpServlets,
```
chanseokoh: `{@link HttpServlet}`?

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/MockHttpServletResponse.java.ftl
akerekes: Splitting the sentence makes sense, also a `d` is missing from `are use in` (should be `are used in`). But otherwise I don't see the improvement over my version (which can be due to me being non-native speaker).

akerekes:
@chanseokoh @briandealwis @nbashirbello PTAL

chanseokoh:
Just for curiosity: what does the new settings in `pom.xml` do and why are they needed for what? I remember they weren't there in the initial PR.

akerekes:
The new test `CreateAppEngineStandardWtpProjectTest#testUnitTestCreated` actually executes the `CreateAppEngineStandardWtpProject#execute` method which will create the project and setup the classpath. But to do that we need the workbench running, and that's what the additional config for the `tycho-surefire-plugin` does.

chanseokoh:
Thanks for the explanation!

PR #711, opened 2016-09-18
Update README.md

branflake2267:
Change the github repo path. The current repo path works fine, so there isn't a need to change it other than clarifying it.


briandealwis:
Will have to remember this for the next repo rename.

branflake2267:
Oh oops funny. Good catch back.

PR #712, opened 2016-09-19
Fixes #703 - Add an ellipsis after the "Deploy to App Engine Standard" menu item because it brings up a dialog.

### LINKED ISSUE #703: Deploy to App Engine Standard...
Add an ellipsis after the "Deploy to App Engine Standard" menu item because it brings up a dialog.

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
```diff
@@ -3,4 +3,4 @@ providerName=Google, Inc.
 
 deployStandardCommandDescription=Deploy the project to App Engine Standard
 deployStandardCommandName=Deploy to App Engine Standard
```
elharo: Looking at this now, why do we have a command name and a menu label? Are we using this in two different UI items?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
```diff
@@ -3,4 +3,4 @@ providerName=Google, Inc.
 
 deployStandardCommandDescription=Deploy the project to App Engine Standard
```
elharo: This is a tooltip, right? Ideally it should use different language than the menu item it describes. E.g. Upload the project to App Engine. If someone doesn't understand "Deploy to App Engine Standard" they're probably not going to understand "Deploy the project to App Engine Standard"

I.e. tooltips should not parrot the thing they're describing.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
akerekes: For future reference (thanks @briandealwis ):
 - command name is used when a menu label is not provided or the handler does not implement the `org.eclipse.ui.commands.IElementUpdater` interface, it also appears in the `Quick Access (Ctrl+3)` view
 - command description is usually a tooltip, but also appears in the `Quick Access (Ctrl+3)` view
 - menu extension label (the value of `deployStandardMenuLabel` property in this case) is shown in the menu this command is added to.

I'll update the command description.

PR #722, opened 2016-09-20
Updated GAE standard facet for deploy and localserver run.

nbashirbello:
@akerekes

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/ProjectFromSelectionHelper.java
+++ b/plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/ProjectFromSelectionHelper.java
@@ -9,6 +9,7 @@ import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.ui.handlers.HandlerUtil;
 import org.eclipse.wst.common.project.facet.core.IFacetedProject;
 
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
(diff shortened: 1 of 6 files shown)


elharo:
That we didn't notice this in the last PR suggests that we're missing some test coverage...

nbashirbello:
Good point. Would add tests.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/ProjectFromSelectionHelper.java
```diff
@@ -32,7 +32,7 @@ public IProject getProject(ExecutionEvent event) throws CoreException, Execution
 
         IFacetedProject facetedProject = facetedProjectHelper.getFacetedProject(project);
         // TODO replace with constant from AppEngineFacet (after possibly relocating that class)
-        if (facetedProjectHelper.projectHasFacet(facetedProject, "com.google.cloud.tools.eclipse.appengine.facet")) {
+        if (facetedProjectHelper.projectHasFacet(facetedProject, "com.google.cloud.tools.eclipse.appengine.facets.standard")) {
```
akerekes: Also we should use `com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet.ID` instead of the string literal (as the TODO says)

akerekes:
There's one more line to change: https://github.com/GoogleCloudPlatform/cloud-tools-for-eclipse/blob/d65fa14c32f7dd625eb4d142e8e59bd1c55f19d8/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml#L152

elharo:
LGTM

PR #727, opened 2016-09-22
Fixes #725 - FontUtil.convertToBoldFont should consider using the bold JFace font type for default font

### LINKED ISSUE #725: FontUtil.convertToBoldFont should consider using the bold JFace font type for default font
If the font passed into `FontUtil.convertToBoldFont` is a JFace dialog font (`JFaceResources.DIALOG_FONT`), then the returned font should be it's bold version.
The current implementation creates a new font from the provided one by converting it to bold, however this causes problems if the user resizes the dialog font in Eclipse/Preferences/Appearance. This resize mechanism will resize the font of all controllers that has the control's default font. If we create a new bold font with the current implementation, that won't be recognized as the default, and won't be resized. E.g.: as a result, all the controls will have fonts of the new size in `DeployPreferencesPanel` except for the `Advanced` section`.

akerekes:
The solution fixes the problem, but in a different way than the issue's title suggests. The `FontUtil.convertToBoldFont` method does the same conversion (but also disposes the font once it's not needed). The way the font is set for the `ExpandableComposite` widget on the `DeployPreferencesPanel` has changed.


briandealwis:
So I'm ok with this change, though it feels like garbage-collection which doesn't usually work so well with SWT/OS resources.  The one downside in that the bolded font will hang around until the control is disposed.  This is a hypothetical problem though: it's only going to happen if our `DeployPreferencesPanel#setFont()` is called multiple times, which is pretty unlikely.

An alternative is to have the disposal on `DeployPrefernecesPanel`:
  - `FontUtil#boldify(Font)` would return the bolded font or `null` if no bold font found; no control or dispose listener
  - `DeployPreferencesPanel#setFont()` would attempt to `boldify()` and stash the bold font aside to be disposed in its `dispose()`.  If a boldFont is already set, then `setFont()` will dispose of it (after generating and installing the new boldFont). This pushes the responsibility onto the consumer of the font and ensures we only have a single bold font hanging around.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/FontUtil.java
```diff
@@ -2,20 +2,40 @@
...
+        return;
+      }
+    }
     FontDescriptor boldDescriptor = FontDescriptor.createFrom(control.getFont()).setStyle(SWT.BOLD);
-    Font boldFont = boldDescriptor.createFont(control.getDisplay());
+    final Font boldFont = boldDescriptor.createFont(control.getDisplay());
     control.setFont(boldFont);
+    control.addDisposeListener(new DisposeListener() {
+      @Override
+      public void widgetDisposed(DisposeEvent e) {
```
elharo: e --> event

akerekes:
@briandealwis I agree that this is like garbage collecting. I like that it's easy to use on the callers end, just one line. Should we come upon a case where we need to call more often this method, we should add the version where the `Font` is changed to bold, and the caller is responsible for disposing it.

PR #734, opened 2016-09-23
Fixes #733 - New bundle for API management

### LINKED ISSUE #733: New bundle for API management
Create a new bundle to define classes and an extension points needed to support the API management configuration.

akerekes:


plugins/com.google.cloud.tools.eclipse.apimanagement.test/.classpath
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
branflake2267: Out of curiosity, are the .classpath, .project and .settings files needed if the project is imported with maven?

akerekes:
So far as I can recall we don't do any customization of these files, so
they're not required to be committed. If we added them to `.gitignore` they
wouldn't show up as changes, but we don't do that currently.

elharo:
I'd prefer not to use the word "API" in package/bundle names here. We may want that for One Platform service APIs, and even if we don't it's likely to be confusing. Maybe "library" or "jar"? I'm not sure.

Does the design doc answer this? I.e. does this potentially include One Platform APIs? Random jars such as Guava? Can we name this such that the purpose of this package is unambiguous?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/.project
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <projectDescription>
-	<name>com.google.cloud.tools.eclipse.appengine.deploy.test</name>
+	<name>com.google.cloud.tools.eclipse.appengine.libraries.test</name>
 	<comment></comment>
```
elharo: do we actually need these empty elements?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/DummyTest.java
```diff
@@ -0,0 +1,11 @@
+package com.google.cloud.tools.eclipse.appengine.libraries;
+
+import org.junit.Test;
+
+public class DummyTest {
```
elharo: You could have a test for the plugin.xml file to shut surefire up

akerekes: That assumes a `plugin.xml` which I have not added as it would have been empty.

elharo:
LGTM once tests pass

akerekes:
I cannot reproduce the test failure locally and it does not seem to connected to the changes. Might be a flaky test. I've pushed another small change to trigger a new build.

PR #736, opened 2016-09-26
don't include application element in appengine-web.xml

elharo:


elharo:
This does not work yet, and will not until he fix for b/31177580 is pushed. :-(

briandealwis:
Need to fix up the `${app.id}` assertions in thehttps://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/NewMavenBasedAppEngineProjectWizardTest.java

Issue #740, opened 2016-09-26
Show signed-in accounts panel at bottom right


chanseokoh:
However, I think it's better to show the panel at the center when it's triggered by the menu item.

chanseokoh:
#778 invalidates this. Currently, the panel is shown at the center of the workbench, so we are good.

chanseokoh:
For the record:

> I think you need to override getInitialLocation() and return a Point that's near the mouse cursor. Use Display#getCursorLocation(). See Window#getInitialLocation() for how it uses the Monitor to constrain the location so the contents aren't shown off-screen.

PR #742, opened 2016-09-26
Implement multiuser login

chanseokoh:
However, functionality-wise, everything is working, including adding and switching new accounts. The existing deploy flow works. If you're curious, you can check out this branch and run Eclipse to test them.

UI is not pretty (which is my best):

java - nothing-src-main-java-something-nothing-helloappengine java - eclipse platform _014

The accounts panel is shown only when there is at least one account signed in. The menu item and the icon tooltip will also change based on the login status.

Things to do:
 - Add as many tests as possible, in this PR.
 - Move the accounts panel to the bottom-right corner, in another PR. (#740)


plugins/com.google.cloud.tools.eclipse.appengine.login/plugin.properties
```diff
@@ -1,2 +1,3 @@
 pluginName=Google Cloud Platform Login Support
-providerName=Google, Inc.
\ No newline at end of file
+providerName=Google, Inc.
+loginIconTooltip=Sign in with your Google Account to integrate with Google services
```
elharo: integrate with --> access

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -96,42 +111,108 @@ public Shell getShell() {
...
+  public boolean switchActiveAccount(String email) {
+    Preconditions.checkNotNull(email);
+
     synchronized (loginState) {
-      if (loginState.isLoggedIn()) {
-        return loginState.getCredential();
+      Account account = findAccount(accounts, email);
+      if (account != null) {
+        activeAccount = account;
+        persisteActiveAccount();
```
elharo: persistActiveAccount

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -96,42 +111,108 @@ public Shell getShell() {
...
+      return false;
     }
   }
 
   @Override
-  public void clearCredential() {
+  public Set<Account> listAccounts() {
     synchronized (loginState) {
-      loginState.logOut(false /* Don't prompt for logout. */);
+      return accounts;
```
elharo: does this expose mutable internal state?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,91 @@
...
    *
    * Must be called from a UI context.
    *
-   * @param dialogMessage custom login dialog message. Can be null
+   * @param dialogMessage custom login dialog message. Can be {@code null}
+   * @return signed-in {@link Account} for successful login; {@code null} otherwise,
+   *     including failed and canceled login
    */
-  public Credential getActiveCredential(String dialogMessage);
+  public Account logIn(String dialogMessage);
```
elharo: we don't use public on interface methods since it's implied

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,91 @@
...
+   *  }
+   *  return logIn(dialogMessage);
+   * }
+   * </pre>
+   *
+   * Must be called from a UI context.
+   *
+   * @see #logIn
+   */
+  public Account getActiveAccountWithAutoLogin(String dialogMessage);
```
elharo: we don't use public on interface methods since it's implied

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,91 @@
...
+
+  /**
+   * Clears all accounts. ("Logging out" from users' perspective.)
    *
    * Safe to call from non-UI contexts.
    */
-  public void clearCredential();
+  public void logOutAll();
```
elharo: we don't use public on interface methods since it's implied

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,91 @@
...
+  /**
+   * If there exists an account that matches the given {@code email}, makes it an active account.
+   *
+   * Safe to call from non-UI contexts.
+   *
+   * @param email cannot be {@null}
+   * @return true if there existed an account matching the {@code email} and it became active;
+   *     false otherwise
+   */
+  public boolean switchActiveAccount(String email);
```
elharo: we don't use public on interface methods since it's implied

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,91 @@
...
+
+  /**
+   * Returns a list of currently logged-in accounts.
+   *
+   * Safe to call from non-UI contexts.
+   *
+   * @return never {@code null}
+   */
+  public Set<Account> listAccounts();
```
elharo: we don't use public on interface methods since it's implied

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -1,12 +1,14 @@
...
-LOGIN_MENU_LOGGED_IN=Sign out from Google
+LOGIN_MENU_LOGGED_IN=Sign in to/Sign out from Google...
 LOGIN_MENU_LOGGED_OUT=Sign in to Google...
 LOGIN_PROGRESS_DIALOG_MESSAGE=Please complete signing in using the external browser.
 LOGIN_PROGRESS_DIALOG_TITLE=Sign in to Google
-LOGIN_TOOLTIP_LOGGED_IN=Sign out from Google
-LOGIN_TOOLTIP_LOGGED_OUT=Authorize Eclipse to operate on Google Cloud Platform resources
 LOGOUT_CONFIRM_DIALOG_MESSAGE=Do you want to sign out from Google?
 LOGOUT_CONFIRM_DIALOG_TITILE=Sign out from Google
+MESSAGE_LABEL_ACTIVE_ACCOUNT=Default Account
```
elharo: Why MESSAGE_?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,163 @@
...
+    if (accounts.size() > 1) {
+      new Label(container, SWT.NONE).setText(Messages.MESSAGE_LABEL_OTHER_ACCOUNTS);
+
+      Composite accountsIndenter = new Composite(container, SWT.NONE);
+      GridLayoutFactory.swtDefaults().margins(EMAIL_LEFT_MARGIN, 0).applyTo(accountsIndenter);
+
+      for (Account account : accounts) {
+        if (!account.getEmail().equals(activeAccount.getEmail())) {
+          Link link = new Link(accountsIndenter, SWT.NO_FOCUS);
+          link.setText("<a href=\"" + account.getEmail() + "\">" + account.getEmail() + "</a>"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
elharo: strange use of a element; maybe it makes sense in context of Eclipse

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,163 @@
...
+    Composite buttonArea = new Composite(container, SWT.NONE);
+    GridDataFactory.defaultsFor(buttonArea).align(SWT.END, SWT.BEGINNING).applyTo(buttonArea);
+    GridLayoutFactory.swtDefaults().numColumns(2).margins(0, 0).applyTo(buttonArea);
+
+    Button addAccountButton = new Button(buttonArea, SWT.PUSH);
+    addAccountButton.setText(Messages.BUTTON_ACCOUNTS_PANEL_ADD_ACCOUNT);
+    addAccountButton.addSelectionListener(new LogInOnClick());
+    GridDataFactory.defaultsFor(addAccountButton).applyTo(addAccountButton);
+
+    boolean loggedIn = loginService.getActiveAccount() != null;
```
elharo: I've seen this several times now. Maybe worth adding an isLoggedIn method to loginService?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: Yhea, the `Link` widget uses the `<A>` element to mark a link: http://www.java2s.com/Code/Java/SWT-JFace-Eclipse/Createalinkwidget.htm

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -152,7 +155,7 @@ public Account logIn(String dialogMessage) {
       if (account != null) {
         activeAccount = account;
         accounts = loginState.listAccounts();
-        persisteActiveAccount();
+        persistActiveAccount();
```
elharo: did you mean this to be recursive? Could it be an infinite loop?

akerekes:
I can see that `Sign Out` will affect all accounts. We may want to make this more fine grained to offer sign out per account.
Also instead of `Other Accounts` label I'd like to suggest `Switch to...` label in the dialog so it may convey better the effect of clicking on of the links.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -60,14 +60,19 @@
...
   /**
    * Clears all accounts. ("Logging out" from users' perspective.)
    *
    * Safe to call from non-UI contexts.
    */
-  public void logOutAll();
+  void logOutAll();
+
+  /**
+   * @return true iff {@link #getActiveAccount} does not return {@code null}.
```
elharo: would prefer a less implementation dependent description of what this means

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+
+  private IGoogleLoginService loginService;
+
+  public AccountsPanel(Shell parent, IGoogleLoginService loginService) {
+    super(parent, SWT.MODELESS,
+        true /* takeFocusOnOpen */,
+        false /* persistSize */,
+        false /* persistLocation */,
+        false /* showDialogMenu */,
+        true /* showPersistActions */,
```
akerekes: From the Javadoc:
`This parameter has no effect if <code>showDialogMenu</code> is <code>false</code>.`

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+        null /* no title area */, null /* no info text area */);
+    this.loginService = loginService;
+  }
+
+  @Override
+  protected Color getBackground() {
+    return getShell().getDisplay().getSystemColor(SWT.COLOR_WIDGET_BACKGROUND);
```
akerekes: Does this change when the theme changes (e.g. to dark)?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+
+  @Override
+  protected Color getForeground() {
+    return getShell().getDisplay().getSystemColor(SWT.COLOR_WIDGET_FOREGROUND);
+  }
+
+  @Override
+  protected Control createDialogArea(Composite parent) {
+    Composite container = (Composite) super.createDialogArea(parent);
+    GridLayoutFactory.swtDefaults().margins(DEFAULT_MARGIN, DEFAULT_MARGIN).applyTo(container);
```
akerekes: `GridLayoutFactory.swtDefaults()` uses 5,5 for margin by default.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  @Override
+  protected Color getBackground() {
+    return getShell().getDisplay().getSystemColor(SWT.COLOR_WIDGET_BACKGROUND);
+  }
+
+  @Override
+  protected Color getForeground() {
+    return getShell().getDisplay().getSystemColor(SWT.COLOR_WIDGET_FOREGROUND);
```
akerekes: Does this change when the theme changes (e.g. to dark)?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+          link.setText("<a href=\"" + account.getEmail() + "\">" + account.getEmail() + "</a>"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+          link.addSelectionListener(new SwitchAccountOnClick());
+        }
+      }
+    }
+  }
+
+  private void createButtons(Composite container) {
+    Composite buttonArea = new Composite(container, SWT.NONE);
+    GridDataFactory.defaultsFor(buttonArea).align(SWT.END, SWT.BEGINNING).applyTo(buttonArea);
```
akerekes: `GridDataFactory.fillDefaults().applyTo(buttonArea)` might achieve the same effect.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+    GridLayoutFactory.swtDefaults().numColumns(2).margins(0, 0).applyTo(buttonArea);
```
akerekes: `GridLayoutFactory.fillDefaults()` uses 0 margins by default.

chanseokoh:
Enabling removing (logging out) a single account sounds like a good idea: https://github.com/GoogleCloudPlatform/ide-login/issues/31

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+
+    Button addAccountButton = new Button(buttonArea, SWT.PUSH);
+    addAccountButton.setText(Messages.BUTTON_ACCOUNTS_PANEL_ADD_ACCOUNT);
+    addAccountButton.addSelectionListener(new LogInOnClick());
+    GridDataFactory.defaultsFor(addAccountButton).applyTo(addAccountButton);
```
akerekes: Unless you want to change anything on the defaults, GridLayoutFactory.applyTo(parentContainer) usually takes care of the child widgets without the need to set GridData objects explicitly.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+
+    if (loginService.isLoggedIn()) {
+      Button logOutButton = new Button(buttonArea, SWT.PUSH);
+      logOutButton.setText(Messages.BUTTON_ACCOUNTS_PANEL_LOGOUT);
+      logOutButton.addSelectionListener(new LogOutOnClick());
+      GridDataFactory.defaultsFor(logOutButton).applyTo(logOutButton);
```
akerekes: Unless you want to change anything on the defaults, GridLayoutFactory.applyTo(parentContainer) usually takes care of the child widgets without the need to set GridData objects explicitly.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  private class SwitchAccountOnClick extends SelectionAdapter {
+    @Override
+    public void widgetSelected(SelectionEvent event) {
+      String email = event.text;
+      loginService.switchActiveAccount(email);
+
+      close();
+      new AccountsPanel(getParentShell(), loginService).open();
```
akerekes: Wouldn't it be better to update the dialog instead of recreating it?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
briandealwis: Rather than overriding `getForeground()` and `getBackground()`, call `setForeground()` and `setBackground()` in the constructor.  But I think these overrides are the default colours anyways?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
briandealwis: I think you're thinking of `generateLayout()`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: In fact, the `PopUp` dialog has black background. I'll set these in the constructor.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
akerekes: yes, thanks for the correction!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: (`PopupDialog` doesn't have `setForeground()` and `setBackground()`)

Turns out that `PopupDialog` does `applyColors()` which happens after `createDialogArea()`, so any attempt to override the default colors will need overriding a method anyway. It feels right to override `getBackground()` and `getForeground()`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: Cool!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: :+1:

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: Good catch.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: This is no recursion!

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: I don't see a particular problem in closing the dialog and opening a new one. I think it's the simplest approach that gets the job done. However, I would seriously considering updating the dialog if you could think of an implementation that is as simple as these two lines (`close()` and `open()`).

chanseokoh:
Addressed all the review comments (if I hadn't missed any) and added unit tests for `AccountsPanel`.

I am still working on adding unit tests for `GoogleLoginService`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -1,12 +1,14 @@
+BUTTON_ACCOUNTS_PANEL_ADD_ACCOUNT=Add Account
+BUTTON_ACCOUNTS_PANEL_LOGOUT=Sign Out
 LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requires an open browser.
 LOGIN_ERROR_DIALOG_MESSAGE=An error occurred while trying to sign in.
 LOGIN_ERROR_DIALOG_TITLE=Sign in Failed
 LOGIN_ERROR_LOCAL_SERVER_RUN=Error listening for authorization code:\n\n
-LOGIN_MENU_LOGGED_IN=Sign out from Google
+LOGIN_MENU_LOGGED_IN=Sign in to/Sign out from Google...
```
elharo: I don't understand this message. It feels like it should be one or the other. If the user can do both, this should be two separate menu items.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
akerekes: My concern was around resource consumption and the slightly increased GC-ing. Probably not too much for this rather simple dialog.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
chanseokoh: Okay, I tried to implement dynamically switching between one menu item ("sign in...") and two menu items ("add account..." and "sign out...") depending on the login status, but it got a bit ugly. @elharo Do you think having two menu items is a way to go?

elharo:
Dynamically switching menu items is a UI antipattern because it confuses muscle memory. A non-functional menu item should be disabled, not removed.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,96 @@
...
+   *  return logIn(dialogMessage);
+   * }
+   * </pre>
    *
    * Must be called from a UI context.
    *
-   * @param dialogMessage custom login dialog message. Can be null
+   * @see #logIn
+   */
+  Account getActiveAccountWithAutoLogin(String dialogMessage);
```
briandealwis: I'm not a big fan of this name as _auto-login_ doesn't really correspond to the behaviour.  It would be good to use a different prefix for methods that may block.  Maybe `obtain`?  So `getActiveAccount()` will return null if there's no active account, but `obtainActiveAccount()`?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,96 @@
...
  *******************************************************************************/
 
 package com.google.cloud.tools.eclipse.appengine.login;
 
-import com.google.api.client.auth.oauth2.Credential;
+import com.google.cloud.tools.ide.login.Account;
+
+import java.util.Set;
 
 public interface IGoogleLoginService {
```
briandealwis: Should have interface javadoc describing the invariants.  E.g., 
  - when there is a logged-in account, one will always be selected as the active account

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,96 @@
...
+   *
+   * Safe to call from non-UI contexts.
+   */
+  void logOutAll();
+
+  /**
+   * @return true if there is at least one signed-in account; false otherwise
    */
-  public Credential getActiveCredential(String dialogMessage);
+  boolean isLoggedIn();
```
briandealwis: This name doesn't seem quite right either: is it that we have an active account, or we have at least one account?  `hasLogins()`?  `hasAccounts()`?  `hasActiveAccount()`?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,96 @@
...
   /**
-   * Clears all credentials. ("logging out" from user perspective.)
+   * Returns a list of currently logged-in accounts.
    *
    * Safe to call from non-UI contexts.
+   *
+   * @return never {@code null}
    */
-  public void clearCredential();
+  Set<Account> listAccounts();
```
briandealwis: `getAccounts()`?

elharo:
As I was discussing in person with @chanseokoh earlier, I'm not sure I like the idea of an active account. It doesn't seem that different from a single account that logs in and out to switch accounts. 

Is it conceivable that we could drop the requirement that the user choose an active account? E.g. deployment works if any of the logged in accounts have the right permissions, not just the active account. 

We can discuss further at our team meeting.

patflynn:
I tend to agree although I'm concerned by the statement: " E.g. deployment
works if any of the logged in accounts have the right permissions, not just
the active account."

That sounds wrong to me.  Two users can have access to the same project. A
user needs to be able to clearly indicate as which user they are performing
an action.

chanseokoh:
For the record (and a heads-up for @briandealwis), let me summarize the direction after our discussions:
- Remove the activeness notion
- Deploy dialog (at the time of deploy) has a user selector (to be done in this PR)

That's really it. Profit.

elharo:
Just capturing the output of our in-person meeting:

1. We are going to remove the notion of an active account (at least insofar as it's visible to the end user.
2. The deploy dialog will contain a user selector drop-down that enables the end user to pick from the logged in accounts.
3. The deploy dialog will remember which account was most recently selected and present that as the selected account in the drop down.

In future PRs we may:

1. Add some sort of "Login to new account" button to the deploy dialog, especially if the user has not logged into any accounts. 
2. Use different user selectors that have different currently selected accounts for operations other than deploy.

chanseokoh:
- Added an `AccountSelector`. Deploy works as intended. Users' credential choice is not being validated and saved yet. (#759)

deployment configuration _017

properties for nativeproj _016

- No active account notion:

selection_018

chanseokoh:
Will spend some more time on cleaning up the code, but the work is basically done.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -9,6 +9,8 @@ deploy.login.dialog.message=Browser opened to authorize deployment.
 deploy.manual.link=To manually promote a version use the <a href="{0}">Google Cloud Console.</a>
 deploy.preferences.dialog.title=Deployment configuration
 deploy.preferences.dialog.title.withProject=Deployment configuration for {0}
+deploy.preferences.dialog.label.selectCredential=Credential to use:
```
elharo: probably, Credential --> Account
Credential is not a term users think of in this context

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -9,6 +9,8 @@ deploy.login.dialog.message=Browser opened to authorize deployment.
...
+deploy.preferences.dialog.accountSelector.login=<Add a new account...>
```
elharo: why the <>?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -59,7 +59,7 @@
 /**
  * Command handler to deploy an App Engine web application project to App Engine Standard.
```
elharo: "an App Engine web application" --> a web application

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -33,64 +40,171 @@
...
+  }
+
+  @Test
+  public void testLogIn_successfulLogin() {
+    GoogleLoginService loginService = newLoginServiceWithMockLoginState(true /* set up logins */);
+    Account account = loginService.logIn(null /* no dialog message */);
+
+    assertEquals(account1, account);
+    assertTrue(loginService.hasAccounts());
+    // Comparison between Account's is conveniently based only on email. (See 'Account.equals().')
```
elharo: accounts

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorTest.java
```diff
@@ -0,0 +1,184 @@
...
+    assertEquals(-1, selector.combo.getSelectionIndex());
+    assertNull(selector.getSelectedCredential());
+    String emailAtIndex1 = selector.combo.getItem(1);
+
+    selector.combo.select(1);
+    selector.logInOnSelect.widgetSelected(mock(SelectionEvent.class));
+
+    assertEquals(1, selector.combo.getSelectionIndex());
+    assertEquals(emailAtIndex1, selector.combo.getItem(1));
+    switch (emailAtIndex1) {
```
elharo: There shouldn't be a switch like this in a test. It means some of your asserts are not being activated. The test should be deterministic and hit all cases.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -64,6 +64,9 @@ public static String getGoogleLoginUrl(String redirectUrl) {
         GoogleLoginService.OAUTH_SCOPES).toString();
   }
 
+  // We expose the reference 'accounts' to callers as-is. 'GoogleLoginService' must not modify
```
elharo: This invariant does not seem to be maintained. The logOut method violates it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/IGoogleLoginService.java
```diff
@@ -1,50 +1,61 @@
...
+  void logOutAll();
+
+  /**
+   * @return true if there is at least one signed-in account; false otherwise
+   */
+  boolean hasAccounts();
 
   /**
-   * Clears all credentials. ("logging out" from user perspective.)
+   * Returns a list of currently logged-in accounts.
```
elharo: delete "a list of"

patflynn:
+1 to not using the credential term. A simple 'Account:' label would do IMO

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
chanseokoh: I occasionally saw the angle brackets for a special "value" in a combo box in order to distinguish it from other actual values. However, I'll remove them as I don't have any objection doing so.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: > This invariant does not seem to be maintained. The logOut method violates it.

I'll update the comment to eliminate any possible confusion.

chanseokoh:
Ready for review. PTAL.

chanseokoh:
Builds are failing with Tycho where they shouldn't: `Caused by: java.util.zip.ZipException: archive is not a ZIP archive`

Should be a transient error, but it's consistently happening at the moment.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -9,6 +9,8 @@ deploy.login.dialog.message=Browser opened to authorize deployment.
 deploy.manual.link=To manually promote a version use the <a href="{0}">Google Cloud Console.</a>
 deploy.preferences.dialog.title=Deployment configuration
 deploy.preferences.dialog.title.withProject=Deployment configuration for {0}
+deploy.preferences.dialog.label.selectAccount=Account to use:
```
elharo: Maybe just "Account" because "to use" is implied

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -57,9 +58,9 @@
 import com.google.common.annotations.VisibleForTesting;
 
 /**
- * Command handler to deploy an App Engine web application project to App Engine Standard.
+ * Command handler to deploy an web application project to App Engine Standard.
```
elharo: a web

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -82,10 +83,19 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
+        // TODO(chanseok): remove the condition (dialog.getCredential() != null) after fixing
+        // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/759. The fix will
+        // make the condition redundant. Also remove the stopgap message dialog.
+        if (dialog.open() == Window.OK) {
+          if (dialog.getCredential() != null) {
+            launchDeployJob(project, dialog.getCredential(), event);
+          } else {
+            // Stopgap (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/759)
+            MessageDialog.openInformation(HandlerUtil.getActiveShell(event),
+                "Account Required for Deploy", "Please select an account to deploy.");
```
elharo: move these strings to a message bundle

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -1,12 +1,12 @@
+BUTTON_ACCOUNTS_PANEL_ADD_ACCOUNT=Add Account
+BUTTON_ACCOUNTS_PANEL_LOGOUT=Sign Out
 LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requires an open browser.
 LOGIN_ERROR_DIALOG_MESSAGE=An error occurred while trying to sign in.
 LOGIN_ERROR_DIALOG_TITLE=Sign in Failed
 LOGIN_ERROR_LOCAL_SERVER_RUN=Error listening for authorization code:\n\n
-LOGIN_MENU_LOGGED_IN=Sign out from Google
+LOGIN_MENU_LOGGED_IN=Manage Google Accounts...
 LOGIN_MENU_LOGGED_OUT=Sign in to Google...
 LOGIN_PROGRESS_DIALOG_MESSAGE=Please complete signing in using the external browser.
```
elharo: delete "Please" http://ux.stackexchange.com/questions/25500/why-avoid-the-word-please-in-error-messages

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -24,4 +24,8 @@ use.custom.versioning=Override generated version
 # Deploy console
 action.stop=Stop
 action.remove=Remove
-job.terminated.template=<terminated> {0}
\ No newline at end of file
+job.terminated.template=<terminated> {0}
```
elharo: why <>?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -24,4 +24,8 @@ use.custom.versioning=Override generated version
...
+
+# TODO(chanseok): remove after fixing https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/759.)
+deploy.account.missing.dialog.title=Account Required for Deploy
+deploy.account.missing.dialog.message=Please try again after selecting an account.
```
elharo: No "Please"

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
chanseokoh: I didn't touch the message. @akerekes may have put the angle brackets, and I think they are suitable for console messages.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -33,64 +40,171 @@
...
 
-    GoogleLoginService loginService = new GoogleLoginService(dataStore, uiFacade, loggerFacade);
-    Assert.assertNull(loginService.getCachedActiveCredential());
+    new GoogleLoginService(dataStore, uiFacade, loggerFacade);
+    verify(dataStore, times(1)).removeOAuthData("my-email@example.com");
   }
 
   @Test
   public void testGoogleLoginService_restoreSavedCredential() {
     // Persisted credential in the data store is valid.
```
elharo: Nit: "persisted" is redundant with "in the data store". Suggest just "Credential in the data store is valid"

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
```diff
@@ -33,64 +40,171 @@
...
+    when(savedOAuthData.getStoredScopes()).thenReturn(OAUTH_SCOPES);
+    when(savedOAuthData.getRefreshToken()).thenReturn(null);
+
+    new GoogleLoginService(dataStore, uiFacade, loggerFacade);
+    verify(dataStore, times(1)).removeOAuthData("my-email@example.com");
+  }
+
+  @Test
+  public void testGoogleLoginService_removeSavedCredentialIfScopesChanged() {
     // Persisted credential in the data store has an out-dated scopes.
```
elharo: Nit: "Persisted" is redundant with "in the data store". Suggest just "Credential in the data store"

chanseokoh:
I have retried many many times, but build always fails on `orackejdk8`. Needs some investigation.

```
$ mvn -Ptravis --fail-at-end verify ${ECLIPSE_TARGET:+-Declipse.target=${ECLIPSE_TARGET}}
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=192m; support was removed in 8.0
[INFO] Scanning for projects...
[WARNING] Error initializing: org.eclipse.tycho.p2.resolver.P2DependencyResolver@4287d447
java.lang.RuntimeException: org.apache.maven.MavenExecutionException: Failed to unpack Tycho's OSGi runtime: Error while expanding /home/travis/.m2/repository/org/eclipse/tycho/tycho-bundles-external/0.24.0/tycho-bundles-external-0.24.0.zip
```

```
Caused by: org.apache.maven.MavenExecutionException: Failed to unpack Tycho's OSGi runtime: Error while expanding /home/travis/.m2/repository/org/eclipse/tycho/tycho-bundles-external/0.24.0/tycho-bundles-external-0.24.0.zip
```

```
Caused by: org.codehaus.plexus.archiver.ArchiverException: Error while expanding /home/travis/.m2/repository/org/eclipse/tycho/tycho-bundles-external/0.24.0/tycho-bundles-external-0.24.0.zip
```

```
Caused by: java.util.zip.ZipException: archive is not a ZIP archive
```

chanseokoh:
I cleared the Travis cache for this PR, and the problem seems to be gone.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A panel listing all currently logged-in accounts. The panel allows adding new accounts and
+ * logging out all accounts.
+ *
+ * TODO(chanseok): move the panel to the bottom-right corner: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/740
+ */
+public class AccountsPanel extends PopupDialog {
```
briandealwis: I think you need to override getInitialLocation() and return a `Point` that's near the mouse cursor.  Use `Display#getCursorLocation()`.  See `Window#getInitialLocation()` for how it uses the `Monitor` to constrain the location so the contents aren't shown off-screen.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,137 @@
...
+  public AccountsPanel(Shell parent, IGoogleLoginService loginService) {
+    super(parent, SWT.MODELESS,
+        true /* takeFocusOnOpen */,
+        false /* persistSize */,
+        false /* persistLocation */,
+        false /* showDialogMenu */,
+        false /* showPersistActions */,
+        null /* no title area */, null /* no info text area */);
+    this.loginService = loginService;
+  }
```
briandealwis: I think this dialog would benefit from:

```
    setTitleText("Google Accounts");
    setInfoText(MessageFormat.format("Click outside or hit {0} to dismiss", SWTKeyLookup.ESC_NAME));
```
(or something along that lines)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -22,4 +24,8 @@ use.custom.versioning=Override generated version
 # Deploy console
 action.stop=Stop
 action.remove=Remove
-job.terminated.template=<terminated> {0}
\ No newline at end of file
+job.terminated.template=<terminated> {0}
+
+# TODO(chanseok): remove after fixing https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/759.)
+deploy.account.missing.dialog.title=Account Required for Deploy
```
chanseokoh: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/759

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
akerekes: I wouldn't add that info text. In Eclipse many popups work this way, so I'd think users expect it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
briandealwis: +1 to _Manage Google Accounts…_

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
briandealwis: Ok, we can always add it if we get reports of confusion.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
```diff
@@ -1,12 +1,12 @@
+BUTTON_ACCOUNTS_PANEL_ADD_ACCOUNT=Add Account
+BUTTON_ACCOUNTS_PANEL_LOGOUT=Sign Out
```
briandealwis: You might want to try using `SWT.FLAT` in your buttons.  They're smaller and look a bit better in the dialog IMHO.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
```diff
@@ -0,0 +1,137 @@
...
+import org.eclipse.swt.widgets.Shell;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A panel listing all currently logged-in accounts. The panel allows adding new accounts and
+ * logging out all accounts.
+ *
+ * TODO(chanseok): move the panel to the bottom-right corner: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/740
```
elharo: Is this still relevant? Looks like that issue is closed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -9,6 +9,8 @@ deploy.login.dialog.message=Browser opened to authorize deployment.
 deploy.manual.link=To manually promote a version use the <a href="{0}">Google Cloud Console.</a>
 deploy.preferences.dialog.title=Deployment configuration
 deploy.preferences.dialog.title.withProject=Deployment configuration for {0}
+deploy.preferences.dialog.label.selectAccount=Account:
+deploy.preferences.dialog.accountSelector.login=Add a new account...
```
chanseokoh: This is an item in the combo box, and I recall I have seen many occasions where ankle brackets are used to distinguish this kind of instructive item from actual values in the combo list. I think this is not unusual as @briandealwis also suggested it.

akerekes:
I used the `<>` because I wanted to follow the format Eclipse uses in the
console when a process has terminated.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
elharo: so this message shouldn't be needed, right?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
```diff
@@ -97,41 +102,56 @@ public Shell getShell() {
   public GoogleLoginService() {}
 
   @VisibleForTesting
-  GoogleLoginService(
+  GoogleLoginService(OAuthDataStore dataStore, LoginServiceUi uiFacade, LoggerFacade loggerFacade) {
+    this(new GoogleLoginState(Constants.getOAuthClientId(), Constants.getOAuthClientSecret(),
```
elharo: I have no idea what is meant by an emoji here. is this addressed?

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
elharo: ping

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginService.java
chanseokoh: The formatting in my opinion is correct.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/GoogleLoginServiceTest.java
chanseokoh: This had been fixed.

elharo:
please merge

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountsPanel.java
chanseokoh: Closed #740, so no longer relevant.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/messages.properties
chanseokoh: Good idea. Used `SWT.FLAT`.

Issue #732, opened 2016-09-23
Implement API attacher for new native Eclipse projects


akerekes:
Part of #682 

This issue covers adding App Engine APIs to a new native project.

Tasks:

 - update the new project wizard UI
 - upon creating the project, add the selected APIs to the build path

YnR:
Is this related to what we saw in the UX study with libraries?

akerekes:
It will fix some of the problems we have seen there, but it originates from the intention of downloading APIs for App Engine, Endpoints and Objectify instead of pulling them from CloudSDK. However this will not make us independent of CloudSDK and the path issues we have seen will need to be addressed separately.

PR #744, opened 2016-09-27
Adds UI for managing libraries

akerekes:
Implements part of #732 
The selected libraries are added as classpath containers to the project, but they are not yet resolved.
They are visible in the `Configure Build Path` dialog, but do not cause build errors, etc.

UI (proper display name for libraries will be added in subsequent PR)
image

Build Path dialog
image


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
```diff
@@ -0,0 +1,30 @@
+package com.google.cloud.tools.eclipse.appengine.libraries;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+
+/**
+ * Represents a library that can be added to App Engine projects. E.g. AppEngine Endpoints library.
```
elharo: Theoretically that could include something like Guava. Can this be more specific?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
```diff
@@ -0,0 +1,30 @@
...
+
+  public Library(String id) {
+    this.id = id;
+  }
+
+  public String getId() {
+    return id;
+  }
+
+  public void setId(String id) {
```
elharo: We need a setter method?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
```diff
@@ -0,0 +1,30 @@
...
+ * Represents a library that can be added to App Engine projects. E.g. AppEngine Endpoints library.
+ *
+ */
+public class Library {
+  private static final String CONTAINER_PATH_PREFIX = "com.google.cloud.tools.eclipse.appengine.libraries";
+  
+  private String id;
+
+  public Library(String id) {
+    this.id = id;
```
elharo: should this be nullable?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProjectTest.java
```diff
@@ -72,14 +75,37 @@ public void testUnitTestCreated() throws InvocationTargetException, CoreExceptio
...
+    assertAppEngineContainerOnClasspath(library);
+  }
+
+  private void assertAppEngineContainerOnClasspath(Library library) throws CoreException {
+    assertTrue(project.hasNature(JavaCore.NATURE_ID));
+    IJavaProject javaProject = JavaCore.create(project);
+    boolean found = false;
+    for (IClasspathEntry iClasspathEntry : javaProject.getRawClasspath()) {
+      if (iClasspathEntry.getPath().equals(library.getContainerPath())) {
+        found = true;
```
elharo: Option: you could simply return when you find it and fail at the end if you don't.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProjectTest.java
```diff
@@ -72,14 +75,37 @@ public void testUnitTestCreated() throws InvocationTargetException, CoreExceptio
...
+      }
+    }
+    assertTrue(found);
```
elharo: Can you include the name or path of the library that wasn't found in the failure message?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
```diff
@@ -60,4 +65,12 @@ public void setEclipseProjectLocationUri(URI uri) {
     this.eclipseProjectLocationUri = uri;
   }
 
+  public List<Library> getAppEngineLibraries() {
+    return appEngineLibraries;
+  }
+
+  public void setAppEngineLibraries(List<Library> libraries) {
+    this.appEngineLibraries = libraries;
```
elharo: danger: mutable private state. Consider copying instead.

patflynn:
What do you think about adding links to the API's home pages?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -67,9 +80,31 @@ public void createControl(Composite parent) {
...
 
+    // Manage APIs
+    addManageLibrariesWidgets(container);
+
     Dialog.applyDialogFont(container);
   }
 
+  private void addManageLibrariesWidgets(Composite container) {
+    apiGroup = new Group(container, SWT.NONE);
+    apiGroup.setText("Manage Libraries");
```
elharo: 1. "Libraries to add to classpath"
2. Move this into the message bundle

akerekes:
@patflynn you mean add links to the dialog (not only show their names), right? Good idea, I'll add it along with the display name in the corresponding PR (to be submitted soon)

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
akerekes: I don't see why we should allow that.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
akerekes: removed

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProjectTest.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProjectTest.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
akerekes: I'd rather just say "Add libraries", what do you think?

Message bundles are not yet used: #494

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -25,6 +36,8 @@
 
   private Text javaPackageField;
   private Text projectIdField;
+  private Group apiGroup;
+  private Map<Button, Library> libraries = new HashMap<>();
```
briandealwis: Just thought I'd point out that you can also use the `Button`'s widget data (viz `setData()`) to directly associate the information.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -67,9 +80,31 @@ public void createControl(Composite parent) {
...
+    for (Library library : getLibraries()) {
+      Button libraryButton = new Button(apiGroup, SWT.CHECK);
+      libraryButton.setText(library.getId());
+      libraries.put(libraryButton, library);
+    }
+
+    GridLayoutFactory.fillDefaults().applyTo(apiGroup);
+  }
+
+  // mock method until the libraries are defined via an extension point
```
briandealwis: I think being able to configure them independently of an extension point is a good thing though.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -73,9 +76,29 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
+  private void addAppEngineLibrariesToBuildPath(IProject newProject, List<Library> libraries, IProgressMonitor monitor) throws CoreException {
+    SubMonitor subMonitor = SubMonitor.convert(monitor, libraries.size());
+    subMonitor.setTaskName("Adding AppEngine libraries");
+    IJavaProject javaProject = JavaCore.create(newProject);
+    IClasspathEntry[] rawClasspath = javaProject.getRawClasspath();
+    IClasspathEntry[] newRawClasspath = Arrays.copyOf(rawClasspath, rawClasspath.length + libraries.size());
+    for (int i = 0; i < libraries.size(); i++) {
+      Library library = libraries.get(i);
+      IClasspathEntry libraryContainer = JavaCore.newContainerEntry(library.getContainerPath(),
+                                                                    new IAccessRule[0],
```
briandealwis: Maybe a TODO about adding the access rules and javadoc references

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
elharo: then you should check for null here

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
elharo: I could be wrong, but it feels to me like a noun is needed in this label, not a verb because clicking the checkbox does not, itself, take an action.

elharo: We shoudl add message bundle in this PR then. You don't need to convert the existing messages in this package here, but let's not add to the problem.

briandealwis:
Uh ignore the previous entry — I thought my comments hadn't gone through. :blush:

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
akerekes: done

akerekes:
@elharo @briandealwis PTAL

For now I'd keep the libraries on the same wizard page.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
```diff
@@ -7,3 +7,6 @@ Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-ActivationPolicy: lazy
+Require-Bundle: org.eclipse.equinox.common
```
elharo: Unless it causes problems, I think you should use Import-Package for this instead: http://stackoverflow.com/questions/1865819/when-should-i-use-import-package-and-when-should-i-use-require-bundle

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
```diff
@@ -73,9 +76,29 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
     addJunit4ToClasspath(monitor, newProject);
   }
 
+  private void addAppEngineLibrariesToBuildPath(IProject newProject, List<Library> libraries, IProgressMonitor monitor) throws CoreException {
+    SubMonitor subMonitor = SubMonitor.convert(monitor, libraries.size());
+    subMonitor.setTaskName("Adding AppEngine libraries");
+    IJavaProject javaProject = JavaCore.create(newProject);
+    IClasspathEntry[] rawClasspath = javaProject.getRawClasspath();
+    IClasspathEntry[] newRawClasspath = Arrays.copyOf(rawClasspath, rawClasspath.length + libraries.size());
+    for (int i = 0; i < libraries.size(); i++) {
```
elharo: I think you can use an enhanced for loop over libraries here since you aren't using the loop variable for anything except retrieving an element

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
akerekes: It causes problems in this case: https://bugs.eclipse.org/bugs/show_bug.cgi?id=398886

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineStandardWtpProject.java
akerekes: the index is also used to set the entry in teh `newRawClasspath` array

PR #748, opened 2016-09-27
Add schema to describe extension point for appengine libraries

akerekes:
The schema is not final (docs, example, etc missing), but I wanted to put it out there to get some feedback. Also an example of how the App Engine SDK can be defined via the extension point is included.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Schema file written by PDE -->
+<schema targetNamespace="com.google.cloud.tools.eclipse.appengine.libraries" xmlns="http://www.w3.org/2001/XMLSchema">
```
elharo: This would normally be an absolure URL, unless Eclipse does something very wonky?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+<annotation>
+      <appinfo>
+         <meta.schema plugin="com.google.cloud.tools.eclipse.appengine.libraries" id="com.google.cloud.tools.eclipse.appengine.libraries" name="App Engine Libraries"/>
+      </appinfo>
+      <documentation>
+         [Enter description of this extension point.]
```
elharo: fill in

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+         </appinfo>
+      </annotation>
+      <complexType>
+         <choice minOccurs="1" maxOccurs="unbounded">
+            <element ref="library"/>
+         </choice>
+         <attribute name="point" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="id" type="string">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+   <element name="library">
+      <complexType>
+         <choice minOccurs="1" maxOccurs="unbounded">
+            <element ref="jarFile"/>
+         </choice>
+         <attribute name="id" type="string">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="name" type="string">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+                  <element ref="exclusionFilter"/>
+                  <element ref="inclusionFilter"/>
+               </choice>
+            </sequence>
+            <element ref="mavenCoordinates"/>
+         </sequence>
+         <attribute name="javadocUri" type="string">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="sourceUri" type="string">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="siteUri" type="string">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            </annotation>
+         </attribute>
+         <attribute name="sourceUri" type="string">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="siteUri" type="string">
```
elharo: perhaps type="anyURI"

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="exclusionFilter">
+      <complexType>
+         <attribute name="pattern" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="inclusionFilter">
+      <complexType>
+         <attribute name="pattern" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="mavenCoordinates">
+      <complexType>
+         <attribute name="repositoryUri" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="groupId" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="artifactId" type="string" use="required">
+            <annotation>
+               <documentation>
+
```
elharo: fill in or delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="since"/>
+      </appinfo>
+      <documentation>
+         [Enter the first release in which this extension point appears.]
```
elharo: fill in

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="examples"/>
+      </appinfo>
+      <documentation>
+         [Enter extension point usage example here.]
```
elharo: fill in

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="apiinfo"/>
+      </appinfo>
+      <documentation>
+         [Enter API information here.]
```
elharo: fill in

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="implementation"/>
+      </appinfo>
+      <documentation>
+         [Enter information about supplied implementation of this extension point.]
```
elharo: fill in

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -1,4 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
+  <extension-point id="com.google.cloud.tools.eclipse.appengine.libraries" name="App Engine Libraries" schema="schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd"/>
+  
+  <extension point="com.google.cloud.tools.eclipse.appengine.libraries">
+    <library 
+          id="com.google.cloud.tools.eclipse.appengine.libraries/appengine-api"
```
elharo: ids don't usually contain slashes

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="mavenCoordinates">
```
briandealwis: Seems to be missing `version`, `packaging`, and `classifier`.  I likely won't need to use `packaging` or `classifier`, but it's trivial to include.  And who knows, maybe we'll need to pull down an `eclipse-bundle`.  `version` will default to `LATEST`, `packaging` to `jar`, and empty classifier.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
briandealwis: Especially if this is considered an internal extension point.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
briandealwis: If this is to match a `org.eclipse.jdt.core.classpathContainerInitializer`, then you can make this an `identifier` which supports an XPath-like spec for referencing attribute values from other definitions.  Something like `org.eclipse.jdt.core.classpathContainerInitializer/classpathContainerInitializer/@id`, I believe.

(And I haven't yet looked at your previous PR — will do next — but does it really make sense to have multiple classpath containers, instead of one container whose contents are populated by this extension point?  In which case this library id could just go away.)

akerekes: Changing it to a reference to the `id` in `classpathContainerInitializer` sounds good.

My idea was to provide multiple containers, one for each feature. So if the user looks at the project on the navigator, he'll see one for App Engine SDK, one for Endpoints, so it'll be easy to check what is added, what's missing. I could imagine breaking the App Engine SDK further down into datastore, task queues, etc.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -1,4 +1,21 @@
...
+          name="App Engine API">
+      <jarFile>
+        <mavenCoordinates
+              artifactId="appengine-api-1.0-sdk"
+              groupId="com.google.appengine"
+              repositoryUri="http://repo1.maven.org/maven2/">
```
akerekes: My concern here is that we want to define the repo to be used, not relying on what is defined in the users' settings.xml where they can override what is central, or define a mirror for it. But to allow for user customization, probably we should allow them to override it in the CT4E preferences.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -1,4 +1,21 @@
...
+        </mavenCoordinates>
+        <exclusionFilter pattern="com.google.appengine.repackaged.*" />
```
akerekes: Yeah, I wasn't sure where to find these patterns, will update the filter.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
briandealwis: 👍

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,194 @@
...
+               </documentation>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="jarFile">
+      <complexType>
+         <sequence>
+            <sequence minOccurs="0" maxOccurs="unbounded">
```
akerekes: That way the filter elements are optional.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
akerekes: Looking at the Eclipse JDT schemas the `targetNamespace` seems to be the bundle's symbolic name, [example](https://github.com/eclipse/eclipse.jdt.core/blob/947633cd3e536904a3c62f9fa98185b52f85fb30/org.eclipse.jdt.core/schema/classpathContainerInitializer.exsd#L3)

elharo: Looking into this further, plugin.xml never seems to use anything except the default namespace. Is that correct? 

If so, using any targetNamespace here is simply incorrect and if the other document does it, I'm 70% sure it's wrong (not 90% because the schema spec is ridiculously complex). If something in Eclipse expects this to happen, then that something is in violation of the schema spec.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -18,177 +18,226 @@
...
+'com.google.cloud.tools.eclipse.appengine.libraries'. The second section of the container path (the ID after the '/') 
+will determine how to resolve the entries associated with the container.
                </documentation>
             </annotation>
          </attribute>
          <attribute name="name" type="string">
             <annotation>
                <documentation>
-                  
+                  A user-friendly name for the library that can be displayed in the UI, e.g.: &quot;AppEngine Endpoints&quot;.
```
elharo: You don't actually need to use &quot; here. Literal " works just fine and i a tad more legible.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -18,177 +18,226 @@
...
                </documentation>
             </annotation>
          </attribute>
       </complexType>
    </element>
 
    <element name="exclusionFilter">
+      <annotation>
+         <documentation>
+            A pattern that can be used to exclude classes and packages to be visible. Uses the same format as the JDT&apos;s build path inclusion/exclusion filters: http://ant.apache.org/manual/dirtasks.html#patterns
```
elharo: &apos; to '

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -18,177 +18,226 @@
...
-            </annotation>
-         </attribute>
+         <attribute name="pattern" type="string" use="required" />
       </complexType>
    </element>
 
    <element name="inclusionFilter">
+      <annotation>
+         <documentation>
+            A pattern that can be used to include classes and packages to be visible. Uses the same format as the JDT&apos;s build path inclusion/exclusion filters: http://ant.apache.org/manual/dirtasks.html#patterns
```
elharo: apos --> '

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -18,177 +18,226 @@
...
                </documentation>
             </annotation>
          </attribute>
-         <attribute name="groupId" type="string" use="required">
+         <attribute name="groupId" type="string" use="required" />
+         <attribute name="artifactId" type="string" use="required" />
+         <attribute name="scope">
+            <simpleType>
+               <restriction base="string">
+                  <enumeration value="compile">
```
elharo: might want to use empty-element tags here

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
akerekes: these were converted by the Eclipse editor, I'll fix them

akerekes:
@elharo @briandealwis  I've spent some time trying to figure out the namespace issue, but could not find any docs related to this, and digging in the code yielded no definite answer either.
The documentation recommends using the provided schema editor instead of authoring the file manually because of the complexity of the schema definition. And neither the editor, nor the wizard offers an option to set the `targetNamespace`, but they set it automatically. We have followed the same convention in the preference area schema, so while it may not follow the specification, I don't see any downsides to it.

akerekes:
@briandealwis PTAL

elharo:
We need to resolve the targetNamespace issue. I'll dig into the schema spec and try to figure out if maybe I'm misinterpreting how targetNamespace is supposed to work. However if I'm write we may need to file bugs against Eclipse. I'm not comfortable accepting this as is.

elharo:
There's another technical point to confuse matters. elementFormDefault="unqualified" (by default) so the targetNamespace does not apply to locally defined elements.

elharo:
Have you tried validating the document against the schema? If that doesn't work, then something is wrong about this.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
elharo: How does Eclipse use this schema? What happens if it is not present? If nothing changes, then we can simply remove it.

elharo:
Per https://wiki.eclipse.org/FAQ_What_is_an_extension_point_schema%3F we don't need the extension point schema at all, so unless I'm wrong and the targetNamespace attribute is correct, we should save ourselves the hassle and remove this file.

akerekes:
PDE sets the targetNamespace to the plugin ID explicitly:

https://github.com/eclipse/eclipse.pde.ui/blob/3c1c04392b0d8943874debe1bec9abe1f89f133b/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/schema/Schema.java#L1008

akerekes:
The extension point contribution in plugin.xml is validated against the schema. If I change it to something that is not permitted by the schema, Eclipse will put an error marker on it. I find it useful.

briandealwis:
The schema is used by the PDE manifest editors to provide the UI for creating extension definitions.  Very useful and should be kept.

elharo:
I've consulted with some experts and they agree with my interpretation of the schema and the targetNamespace. This schema does not correctly describe or validate plugin.xml. PDE is simply wrong. We can't check it in. :-( 👎

briandealwis:
I'm not disagreeing that it's wrong, but it's a file generated by PDE and only consumed by PDE.  We can fix PDE (patches welcome :smile:) but that should not prevent us from committing this.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd
```diff
@@ -0,0 +1,243 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!-- Schema file written by PDE -->
+<schema targetNamespace="com.google.cloud.tools.eclipse.appengine.libraries" xmlns="http://www.w3.org/2001/XMLSchema">
+<annotation>
+      <appinfo>
+         <meta.schema plugin="com.google.cloud.tools.eclipse.appengine.libraries" id="com.google.cloud.tools.eclipse.appengine.libraries" name="App Engine Libraries"/>
+      </appinfo>
+      <documentation>
+         Provides an extension point to define libraries that will be resolved using the Maven coordinates. These libraries then can be added to App Engine projects.
```
briandealwis: Probably should add that this is intended for internal use only and may change without notice.

elharo:
Sorry, I'm not convinced. It may be PDE's fault, but that's not enough reason to commit this, especially since we can live without it. The equivalent is "This Java file does not compile, but it's only read by this special processor and not javac so we can allow it."

akerekes:
Just as an experiment I removed the schema reference from the plugin.xml. Eclipse immediately shows an error for unknown extension point. We can get over this by changing the severity of the error in the IDE settings. Then I ran the code and tried to get access to the extension declared in plugin.xml (the App Engine API library) by using:
```
IExtensionRegistry extensionRegistry = Platform.getExtensionRegistry();
    IConfigurationElement[] configurationElementsFor = extensionRegistry.getConfigurationElementsFor("com.google.cloud.tools.eclipse.appengine.libraries");
```
I got back an empty array.
When the schema was referenced in the plugin.xml, this returned an `org.eclipse.core.internal.registry.ConfigurationElementHandle` that contained the extension defined in plugin.xml.

We can try to check if changing the targetNamespace to something that adheres to the spec breaks Eclipse or we keep ourselves from leveraging the extension framework capabilities of Eclipse just in the name of enforcing the spec.

briandealwis:
@elharo: We've already committed broken PDE schema files elsewhere.  Consider it an internal XML-like format to PDE.  We don't have to include the schema in our actual plugins (don't include it in the `build.properties`) — the Eclipse SDK only includes the scheme files in the source bundles.

@akerekes: theres something else going on as the scheme isn't actually used by the extension registry.  The schema is only used by the PDE editors.  Are you sure the .libraries bundle is included in your launch, and is being resolved?

elharo:
To change the schema to correctly describe the plugin.xml you should remove the targetNamespace attribute.
We should also seriously consider adding a unit test that validates plugin.xml against the schema.

akerekes:
@briandealwis 
I've double-checked, but just changing whether the line
```
<extension-point id="com.google.cloud.tools.eclipse.appengine.libraries" name="App Engine Libraries" schema="schema/com.google.cloud.tools.eclipse.appengine.libraries.exsd"/>
```
is present in the plugin.xml or not makes the difference. No other differences between the two execution.

akerekes:
Removing just the `schema` attribute is enough to fix the missing extension in my test.

briandealwis:
I was digging in the Eclipse bugs database and there are [several][79122] [bugs][213255] about `.exsd` being non-conformant.  But the PDE committers [note](https://bugs.eclipse.org/bugs/show_bug.cgi?id=213255#c2):

> We'll take a look at it for M5. However, note, we have never claimed exsd's are xsd compatible. Ever. :)

[and](https://bugs.eclipse.org/bugs/show_bug.cgi?id=213255#c5):

> As Chris pointed out, the schemas don't claim any compatibility.  

They aren't `.xsd` but `.exsd` and the fact that they look like XML is merely coincidence.

[79122]: https://bugs.eclipse.org/bugs/show_bug.cgi?id=79122 "psd generated by PDE does not comform to W3 specs"
[213255]: https://bugs.eclipse.org/bugs/show_bug.cgi?id=213255 "Extension point schemas aren't valid W3C Schemas"

elharo:
Hardly a coincidence since they use the namespace
 xmlns="http://www.w3.org/2001/XMLSchema"
and not
 xmlns="http://www.ecipse.org/XMLSchema"

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PluginXmlValidationTest.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.xml.validation.Schema;
+import javax.xml.validation.SchemaFactory;
+
+import org.junit.Test;
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+public class PluginXmlValidationTest {
+
+  private Document doc;
```
elharo: unused field

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+         <meta.schema plugin="com.google.cloud.tools.eclipse.appengine.libraries" id="com.google.cloud.tools.eclipse.appengine.libraries" name="App Engine Libraries"/>
+      </xs:appinfo>
+      <xs:documentation>
+         Provides an extension point to define libraries that will be resolved using the Maven coordinates. These libraries then can be added to App Engine projects.
+      </xs:documentation>
+   </xs:annotation>
+
+   <xs:element name="plugin">
+      <xs:complexType>
+         <xs:sequence>
```
elharo: try changing xs:sequence to

<xs:choice minOccurs="1" maxOccurs="unbounded">

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+
+   <xs:element name="libraryFile">
+      <xs:annotation>
+         <xs:documentation>
+            Describes a single jar file located in a Maven repository.
+         </xs:documentation>
+      </xs:annotation>
+      <xs:complexType>
+         <xs:sequence>
+            <xs:sequence minOccurs="0" maxOccurs="unbounded">
```
elharo: you can simplify this by removing the sequence and pushing the minOccurs/maxOccurs attributes onto the choice

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+               <xs:choice>
+                  <xs:element ref="exclusionFilter" />
+                  <xs:element ref="inclusionFilter" />
+               </xs:choice>
+            </xs:sequence>
+            <xs:element ref="mavenCoordinates"/>
+         </xs:sequence>
+         <xs:attribute name="javadocUri" type="xs:string">
```
elharo: xs:string --> xs:anyURI

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+            <xs:annotation>
+               <xs:documentation>
+                  URI that can be used by the IDE to lookup javadoc for the classes contained in this jar file.
+               </xs:documentation>
+            </xs:annotation>
+         </xs:attribute>
+         <xs:attribute name="sourceUri" type="xs:string">
```
elharo: xs:string --> xs:anyURI

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+               </xs:documentation>
+            </xs:annotation>
+         </xs:attribute>
+      </xs:complexType>
+   </xs:element>
+
+   <xs:element name="exclusionFilter">
+      <xs:annotation>
+         <xs:documentation>
+            A pattern that can be used to exclude classes and packages to be visible. Uses the same format as the JDT&apos;s build path inclusion/exclusion filters: http://ant.apache.org/manual/dirtasks.html#patterns
```
elharo: &apos; --> '

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+      </xs:annotation>
+      <xs:complexType>
+         <xs:attribute name="pattern" type="xs:string" use="required" />
+      </xs:complexType>
+   </xs:element>
+
+   <xs:element name="inclusionFilter">
+      <xs:annotation>
+         <xs:documentation>
+            A pattern that can be used to include classes and packages to be visible. Uses the same format as the JDT&apos;s build path inclusion/exclusion filters: http://ant.apache.org/manual/dirtasks.html#patterns
```
elharo: &apos; --> '

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+If the value starts with [protocol]:// then it is treated as a URI, otherwise an ID of a repository that is expected to be defined.
+               </xs:documentation>
+            </xs:annotation>
+         </xs:attribute>
+         <xs:attribute name="groupId" type="xs:string" use="required" />
+         <xs:attribute name="artifactId" type="xs:string" use="required" />
+         <xs:attribute name="scope">
+            <xs:simpleType>
+               <xs:restriction base="xs:string">
+                  <xs:enumeration value="compile">
```
elharo: might want to convert these to empty-element tags

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+      </xs:complexType>
+   </xs:element>
+
+
+   <xs:annotation>
+      <xs:appinfo>
+         <xs:meta.section type="xs:examples"/>
+      </xs:appinfo>
+      <xs:documentation>
+         &lt;extension point=&quot;com.google.cloud.tools.eclipse.appengine.libraries&quot;&gt;
```
elharo: don't need to escape " here

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+        &lt;exclusionFilter pattern=&quot;com.google.appengine.repackaged.*&quot; /&gt;
+      &lt;/jarFile&gt;
+    &lt;/library&gt;
+  &lt;/extension&gt;
+      </xs:documentation>
+   </xs:annotation>
+
+   <xs:annotation>
+      <xs:appinfo>
+         <xs:meta.section type="xs:apiinfo"/>
```
elharo: xs:apiinfo --> apiinfo

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+         <xs:meta.section type="xs:implementation"/>
+      </xs:appinfo>
+      <xs:documentation>
+         See the plugin.xml of the com.google.cloud.tools.eclipse.appengine.libraries for an example of how to contribute to the extension point.
+      </xs:documentation>
+   </xs:annotation>
+   
+   <xs:annotation>
+      <xs:appinfo>
+         <xs:meta.section type="xs:copyright"/>
```
elharo: remove xs: from xs:meta

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -0,0 +1,263 @@
...
+      </xs:appinfo>
+      <xs:documentation>
+         Copyright 2016 Google Inc. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);
```
elharo: don't escape &quot;

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
akerekes: to?

PR #752, opened 2016-09-29
Libraries exsd

akerekes:
For future reference, should we decide to add the eXSD to the plugin.


akerekes:
Blocked on the resolution of https://bugs.eclipse.org/bugs/show_bug.cgi?id=502693

PR #757, opened 2016-09-30
Ignore plugins/com.google.cloud.tools.eclipse.util.test/testData, generated from running tests

briandealwis:
This directory and its contents are generated when running the tests.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.util.test/.gitignore
+++ b/plugins/com.google.cloud.tools.eclipse.util.test/.gitignore
@@ -1,0 +1,1 @@
+testData


chanseokoh:
Shouldn't we remove the directory from our repo too then?

elharo:
We have to figure out how to fix the flakiness in our tests. It seems to consistently be a problem  on neon, oraclejdk8, and push. 

But LGTM on this once the tests pass.

briandealwis:
Abandoning this change: at @akerekes's prompting, I checked this again and I must have had some stray changes hanging around from earlier as I can't reproduce, and these files are present in the tree.

PR #760, opened 2016-09-30
Kokoro

elharo:
@chanseokoh use separate signing script

### DIFF
--- a/kokoro/ubuntu/release.cfg
+++ b/kokoro/ubuntu/release.cfg
@@ -2,3 +2,4 @@
 
 # Location of the release bash script in git.
 build_file: "google-cloud-eclipse/kokoro/ubuntu/release.sh"
+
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -6,8 +6,5 @@ set -e
 set -x
 
 if [ "$1" == "release" ]; then
-  # TODO load keys from internal storage
-  # mvn clean install
-else
-  # mvn clean install
+  mvn clean install
```
chanseokoh: Previously the intention was to do `mvn clean install` regardless of `$1` being `"release"`. Now building only when `"release"`?

elharo: previously there was a bash syntax error. Apparently bash doesn't like empty if blocks, so I put the mvn commands back and moved kokoro to using a separate script for now. we're not actually using release.sh for the time being. This is all WIP until kokoro supports maven.

PR #763, opened 2016-10-02
remove deployment preferences

### LINKED ISSUE #761: Allow saving null Project ID (i.e., make it optional) in project preferences
For a project created without a project ID given, when the Deploy preference page is open, a user is forced to enter a project ID before the user can move to another preference page or dismiss the dialog. If the user tries to select another preference node, it shows the following error:

selection_019

This could be annoying particularly for a user who doesn't know what project ID means or what to enter.

elharo:
fix #761 @akerekes @chanseokoh 

With this PR deployment preferences are adjusted and stored only when actually deploying the project. reasons to do this:
1. See #761 for different validation requirements in this panel depending on when we bring it up.
2. We always bring this dialog up at deploy. Having an extra preference panel does not add any functionality and adds complexity to the UI.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/XmlTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/XmlTest.java
@@ -39,9 +39,9 @@ public class XmlTest {
   @Test
   public void testLimitedVisibility() {
     NodeList pages = doc.getElementsByTagName("page");
-    Assert.assertEquals(2, pages.getLength());
+    Assert.assertEquals(1, pages.getLength());
     NodeList enabledWhen = doc.getElementsByTagName("enabledWhen");
-    Assert.assertEquals(3, enabledWhen.getLength());
+    Assert.assertEquals(2, enabledWhen.getLength());
     NodeList tests = doc.getElementsByTagName("test");
     Assert.assertEquals(3, tests.getLength());
     NodeList adapts = doc.getElementsByTagName("adapt");
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
@@ -66,18 +66,6 @@
        </enabledWhen>
     </page>
   </extension>
-  
-  <extension point="org.eclipse.ui.propertyPages">
-    <page
-        id="com.google.cloud.tools.eclipse.appengine.deploy"
-        name="App Engine Deployment"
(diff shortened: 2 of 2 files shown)


elharo:
@nbashirbello looks like this might affect what you're working on too

akerekes:
I'd still prefer to have a place to set deploy preferences without executing an actual deploy, and this panel was like that. As for the different requirements for project ID, we could just make it optional on the property page (instead of checking whether it was empty initially) as the dialog will be shown anyways, and we can require the project ID to filled in there.

elharo:
What's the use case for a place to set deploy preferences without executing an actual deploy? We always bring up the panel deploy, no matter what is set here. I'd rather not do the extra work to make this a little different than the one in the deploy dialog absent a compelling use case.

Further, once we add flex deploy there are going to be two separate panels. It's going to get more confused.

akerekes:
You can create/change a run configuration  without executing it. A use case may be "make a change to the deploy settings, but I don't want to make a deploy now, only the change, otherwise I'm going to forget I wanted to change it by the next time I want to do a deploy". Possibly a niche use case. Maybe just adding an `Apply` button to the deploy dialog will achieve the same and then no need for the property page.

elharo:
On Tue, Oct 4, 2016 at 11:18 AM, Andras Kerekes <notifications@github.com>
wrote:

> You can create/change a run configuration without executing it. A use case
> may be "make a change to the deploy settings, but I don't want to make a
> deploy now, only the change, otherwise I'm going to forget I wanted to
> change it by the next time I want to do a deploy". Possibly a niche use
> case. Maybe just adding an Apply button to the deploy dialog will achieve
> the same and then no need for the property page.
>

An Apply button sounds good.


-- 
Elliotte Rusty Harold
elharo@macfaq.com

nbashirbello:
I don't agree with removing the deploy property page. In Eclipse, there are different project actions that are accessible via the context menu and also in the Properties Page e.g "Configure Build Path..." In Eclipse if you want to update the attributes of an action you do it in the Property Dialog, its pretty standard.

Also the Flex property page is not going to complicate the UI. Right now, if the standard facet is installed, the current deploy page is available. If the flex facet is installed, the flex deployment page will be available.

elharo:
On Tue, Oct 4, 2016 at 1:53 PM, nbashirbello <notifications@github.com>
wrote:

> I don't agree with removing the deploy property page. In Eclipse, there
> are different project actions that are accessible via the context menu and
> also in the Properties Page e.g "Configure Build Path..." In Eclipse if you
> want to update the attributes of an action you do it in the Property
> Dialog, its pretty standard.


Yes, but Build does not bring up the Build Path with each build. For
deploy, our plan is to bring up the deploy settings with each deploy.


-- 
Elliotte Rusty Harold
elharo@macfaq.com

briandealwis:
Perhaps we can have a checkbox button within the deploy dialog to _remember these values_, rather than apply?

chanseokoh:
@elharo close this since #761 is closed?

elharo:
Not yet

PR #765, opened 2016-10-02
fix font in API group

elharo:
@briandealwis @akerekes Not sure why, but merely calling 

does not seem to be enough to set the correct font size on the group label, at least not on the Mac.


briandealwis:
`Dialog.applyDialogFont(container)` is misnamed given its current behaviour.  The method only performs a font change if the configured dialog font is different from the configured general font.  If configured dialog and configured general fonts are the same (the usual situation) then the method does nothing, even if the dialog font is different from the container's font.  This behaviour came in later, and it seems puzzling.

The title font on a Group is set deliberately smaller by OS X.  You can see an example in _System Preferences_ > _Mission Control_.

PR #772, opened 2016-10-05
Sign the jars

elharo:
@nbashirbello @chanseokoh


chanseokoh:
Just to be sure: `${KOKORO_GFILE_DIR}` is always freshly empty, right? Otherwise LGTM.

elharo:
Yes, it actually points to /tmpfs/gfile and is created anew on each build.

Issue #768, opened 2016-10-02
Provide actionable development urls 


branflake2267:
Would it be possible to get the launcher urls or have them provided in some way where the user could easily access and click on them to launch them? Or maybe provide a App engine view. Not sure what you want to do. Below I'll show you what I did for the GWT users. 

Here is what I did in the GPE-Fork, I retrieved the urls that the user could launch. I always found it annoying that I had to look up the admin url because I'd forget. 
<img width="1027" alt="screen shot 2016-10-02 at 12 48 54 pm" src="https://cloud.githubusercontent.com/assets/1326504/19023252/bd1797f2-889e-11e6-8fcd-5305c471e7e3.png">

Example of starting the App Engine server and the urls show up in the the DevMode View. 
https://youtu.be/SUueCocqf_U?t=76

elharo:
I'm not sure what you mean by "launcher URLs". Are you referring to the URL of the local server? and if so, which ones? There can be many.

briandealwis:
We have an open-browser-on-launch option that will open the main page.

@branflake2267 is referring to the all the URLs that are output by the `dev_appserver`: the main launch page, the admin page, etc.
```
INFO     2016-08-19 19:49:38,095 devappserver2.py:769] Skipping SDK update check.
INFO     2016-08-19 19:49:38,161 api_server.py:205] Starting API server at: http://localhost:60320
INFO     2016-08-19 19:49:38,166 dispatcher.py:197] Starting module "default" running at: http://localhost:8080
INFO     2016-08-19 19:49:38,172 admin_server.py:116] Starting admin server at: http://localhost:8000
Aug 19, 2016 7:49:39 PM com.google.appengine.tools.development.AbstractContainerService configure
WARNING: Null value for containerConfigProperties.get(devappserver.portMappingProvider)
Aug 19, 2016 7:49:39 PM com.google.apphosting.utils.jetty.JettyLogger info
INFO: Logging to JettyLogger(null) via com.google.apphosting.utils.jetty.JettyLogger
Aug 19, 2016 7:49:39 PM com.google.apphosting.utils.jetty.JettyLogger info
INFO: jetty-6.1.x
Aug 19, 2016 7:49:40 PM org.apache.jasper.EmbeddedServletOptions <init>
WARNING: Warning: Invalid value for the initParam keepgenerated. Will use the default value of "false"
[...]

PR #780, opened 2016-10-06
Highlight urls in launch console

briandealwis:
Adds a URL link converter to the launch console to convert URLs into clickable links.

Goes part way to fixing #768.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
@@ -145,6 +145,7 @@
       </consolePageParticipant>
    </extension>
    <extension
+         id="urlLinker"
          point="org.eclipse.ui.console.consolePatternMatchListeners">
     <!-- pretty liberal URL pattern matcher -->
     <consolePatternMatchListener
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
@@ -145,6 +145,18 @@
       </consolePageParticipant>
    </extension>
    <extension
+         point="org.eclipse.ui.console.consolePatternMatchListeners">
+    <!-- pretty liberal URL pattern matcher -->
+    <consolePatternMatchListener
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -145,6 +145,18 @@
       </consolePageParticipant>
    </extension>
    <extension
+         point="org.eclipse.ui.console.consolePatternMatchListeners">
+    <!-- pretty liberal URL pattern matcher -->
+    <consolePatternMatchListener
+      class="com.google.cloud.tools.eclipse.ui.util.console.PatternToHyperlinkConverter"
+      id="com.google.cloud.tools.eclipse.appengine.deploy.consolePatternMatchListener"
+      regex="https?://([\w-]+\.?)+(:\w+)?(/[^\s]*)?">
```
elharo: I'm imagining a test that would extract the regex from the XML and compare it to strings.

branflake2267:
Does the local dev server run with port https by default?

briandealwis:
No, I don't think it supports https at all.  I kept it there just in case something logs a message with https.

branflake2267:
Ah, gotcha. I was wondering, b/c we would have a mixed content issue if that were the case.

briandealwis:
Hmm, I wonder if we could pull out the module/service information from the server logs to provide with [`org.eclipse.wst.server.core.model.IURLProvider`](https://github.com/gwt-plugins/gwt-eclipse-plugin/blob/master/plugins/com.gwtplugins.gwt.eclipse.wtp/src/com/google/gwt/eclipse/wtp/facet/GwtLaunchableAdapterDelegate.java#L34):
```
INFO     2016-07-28 01:33:21,421 dispatcher.py:197] Starting module "default" running at: http://localhost:8080
INFO     2016-07-28 01:33:21,424 admin_server.py:116] Starting admin server at: http://localhost:8000
```
[…]
```
Jul 28, 2016 1:33:23 AM com.google.appengine.tools.development.AbstractModule startup
INFO: Module instance default is running at http://localhost:54220/
Jul 28, 2016 1:33:23 AM com.google.appengine.tools.development.AbstractModule startup
INFO: The admin console is running at http://localhost:54220/_ah/admin
```
Unfortunately the log formats are different, and the `default` module port is different too.

PR #785, opened 2016-10-07
Fix mismatching server type ID

chanseokoh:
Server type ID defined in plugin.xml is `com.google.cloud.tools.eclipse.appengine.standard.server`.

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml#L29

With the correct ID, our App Engine Local Server launch config tab can load server instances properly.


akerekes:
Assuming tests pass.

PR #786, opened 2016-10-07
Change deploy property settings title to include flex/standard environment

nbashirbello:


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -111,8 +111,10 @@ private DeployPreferencesPanel getPreferencesPanel(Composite container) throws C
     IFacetedProject facetedProject = ProjectFacetsManager.create(project);
 
     if (AppEngineStandardFacet.hasAppEngineFacet(facetedProject)) {
+      setTitle(Messages.getString("standard.page.title"));
```
elharo: realizing I don't like the name of this method. Should probably be just hasFacet; but that's for another PR

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -111,8 +111,10 @@ private DeployPreferencesPanel getPreferencesPanel(Composite container) throws C
...
       return new StandardDeployPreferencesPanel(container, project, getLayoutChangedHandler());
     } else if (AppEngineFlexFacet.hasAppEngineFacet(facetedProject)) {
+      setTitle(Messages.getString("flex.page.title"));
       return new FlexDeployPreferencesPanel(container);
     } else {
```
elharo: This shouldn't happen, but if it does is this really where we want to throw an exception? It feels safer to just set the title in this case.

chanseokoh:
The SWT bot test fails only on Neon, and it seems relevant to this PR.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
nbashirbello: At the stage, the default title "App Engine Deployment" will be used.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -115,12 +116,15 @@ private DeployPreferencesPanel getPreferencesPanel(Composite container) throws C
...
-          getLayoutChangedHandler(), false /* requireValues */);
+      setTitle(Messages.getString("standard.page.title"));
+      return new StandardDeployPreferencesPanel(
+          container, project, loginService, getLayoutChangedHandler(), false /* requireValues */);
     } else if (AppEngineFlexFacet.hasAppEngineFacet(facetedProject)) {
+      setTitle(Messages.getString("flex.page.title"));
       return new FlexDeployPreferencesPanel(container);
     } else {
-      throw new IllegalStateException(project.getName() + " does not have an App Engine facet.");
+      // App Engine Deployment property page is only visible if project contains an App Engine facet
```
briandealwis: Do you think this is ok?  Or should we at least log an error?

nbashirbello:
@briandealwis

PR #787, opened 2016-10-07
Enable setting arbitrary host for local runs

### LINKED ISSUE #775: Allow the specification of the --host address as part of local run
dev_appserver.py has the following argument:

  --host HOST           host name to which application modules should bind
                        (default: localhost)

We should allow the user to change this. A common use-case is to set the host to 0.0.0.0, which then binds to localhost but allows the host to be hit from another machine.

chanseokoh:
Fixes #775.

Of course, attempting to use non-local hosts and IP addresses except for 0.0.0.0 will fail, which is expected.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
@@ -34,7 +34,7 @@
             runtime="true"
             runtimeTypeId="com.google.cloud.tools.eclipse.appengine.standard.runtime"
             startBeforePublish="false"
-            supportsRemoteHosts="false">
+            supportsRemoteHosts="true">
       </serverType>
    </extension>
    <extension
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
@@ -130,14 +130,14 @@ public class LocalAppEngineServerBehaviour extends ServerBehaviourDelegate {
    * Starts the development server.
    *
    * @param runnables the path to directories that contain configuration files like appengine-web.xml
-   * @param stream the stream to send development server process output to
-   * @param host
+   * @param console the stream (Eclipse console) to send development server process output to
+   * @param host the host name to which application modules should bind
    */
-  void startDevServer(List<File> runnables, MessageConsoleStream stream, String host) {
+  void startDevServer(List<File> runnables, MessageConsoleStream console, String host) {
     setServerState(IServer.STATE_STARTING);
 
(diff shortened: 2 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -131,8 +131,9 @@ public final void setModulePublishState2(IModule[] module, int state) {
    *
    * @param runnables the path to directories that contain configuration files like appengine-web.xml
    * @param stream the stream to send development server process output to
+   * @param host
```
elharo: document this.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -162,7 +164,8 @@ void startDevServer(List<File> runnables, MessageConsoleStream stream) {
    * @param stream the stream to send development server process output to
    * @param debugPort the port to attach a debugger to if launch is in debug mode
    */
-  void startDebugDevServer(List<File> runnables, MessageConsoleStream stream, int debugPort) {
+  void startDebugDevServer(List<File> runnables, MessageConsoleStream stream,
```
elharo: suggest stream --> console since this is not either of the things Java devs usually think of as a "stream"

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -34,7 +34,7 @@
             runtime="true"
             runtimeTypeId="com.google.cloud.tools.eclipse.appengine.standard.runtime"
             startBeforePublish="false"
-            supportsRemoteHosts="false">
+            supportsRemoteHosts="true">
```
elharo: why this change if non-local hosts fail?

elharo:
Why will IP addresses other than 0.0.0.0 fail if they are assigned to the local machine?

chanseokoh:
> Why will IP addresses other than 0.0.0.0 fail if they are assigned to the local machine?

Let me say it precisely. An address assigned to the local machine will not fail. On the other hand, non-local IP addresses will fail. Actually, the IP address is simply passed to `dev_appserver.py` through `--host`, so if an address fails, then it is because `dev_appserver.py` failed bind it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
chanseokoh: This is purely a UI constraint. If it's `false`, the run dialog complains with the below error message and the _Finish_ button is disabled when attempting enter non-local hosts:

run_on_server

It has no function effect other than the UI constraint. The host value is simply passed on to the dev app server.

PR #793, opened 2016-10-11
Explicitly examine problem markers rather than probing the problems view

briandealwis:
Since we're really using `SwtBotProjectActions#getErrorsInProblem()` to determine if there are any build errors, and expanding tree items is painful, just check for error markers directly.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotProjectActions.java
+++ b/plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotProjectActions.java
@@ -36,7 +36,7 @@ import org.eclipse.swtbot.swt.finder.widgets.SWTBotShell;
 import org.eclipse.swtbot.swt.finder.widgets.SWTBotTree;
 import org.eclipse.swtbot.swt.finder.widgets.SWTBotTreeItem;
 
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotProjectActions.java
```diff
@@ -197,27 +200,36 @@ public static SWTBotTree getProjectRootTree(SWTWorkbenchBot bot) {
...
-    List<String> errors = new ArrayList<>();
-    
-    for (SWTBotTreeItem item : tree.getAllItems()) {
-      String text = item.getText();
-      if (text != null && text.startsWith("Errors")) {
-        errors.add(text);
+  /** Return the list of all problems in the workspace. */
+  public static List<String> getAllBuildErrors(SWTWorkbenchBot bot) {
+    IWorkspaceRoot root = getWorkspaceRoot();
+    List<String> foundProblems = new LinkedList<>();
```
elharo: ArrayList

plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotProjectActions.java
```diff
@@ -197,27 +200,36 @@ public static SWTBotTree getProjectRootTree(SWTWorkbenchBot bot) {
...
+      try {
+        IMarker[] problems = project.findMarkers(IMarker.PROBLEM, true /* includeSubtypes */,
+            IResource.DEPTH_INFINITE);
+        for (IMarker problem : problems) {
+          int severity = problem.getAttribute(IMarker.SEVERITY, IMarker.SEVERITY_INFO);
+          if (severity >= IMarker.SEVERITY_ERROR) {
+            foundProblems.add(formatProblem(problem));
+          }
+        }
+      } catch (CoreException ex) {
```
elharo: Any particular reason not to just bubble this up as is? Why wrap?

briandealwis: It's a checked exception, so it would require percolating that up.  I can't think of a reason why or how we would create erroring projects, so it didn't seem worth exposing it.

PR #797, opened 2016-10-11
Save and validate account selection

### LINKED ISSUE #759: Wire up credential choice into project-scoped deploy preferences and add validation
- Save (as an email address) the user's choice of credential for deploy in deploy preferences.
- Add validation to the deploy dialog so that the OK button is enabled only when a credential is selected. However, do NOT do validation in the preference page.
- Next time the dialog or the page is opened, load the saved email and auto-select the corresponding credential, if it exists.
  - In the case of the preference dialog when the corresponding credential doesn't exist: if a user closes the page while a user didn't explicitly select a credential (recall that we don't do validation), don't overwrite the previously saved preference.

chanseokoh:
@akerekes @briandealwis

Fixes #759.

- Saves and loads an account selection choice (email address) into deploy preferences .
  - `AccountSelectorObservableValue` defined for databinding.
- Validates the deploy dialog. (The _Deploy_ button is disabled if no account is selected.) Subsequently, the stopgap error popup dialog is removed.
- However, this is an optional preference setting (like project ID), so no validation in the preference page.

Misc:
- `OverrideValidator` constructor: I removed `super(Realm)`, as calling the default constructor (which passes the default realm) seems working fine. @akerekes, is there a reason that we should pass a specific realm?
- `allowedEmptyProjctId` --> `!requireValues`

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
@@ -66,7 +66,7 @@ public class DeployPreferencesDialog extends TitleAreaDialog {
 
     Composite container = new Composite(dialogArea, SWT.NONE);
     content = new StandardDeployPreferencesPanel(container, project, loginService,
-        getLayoutChangedHandler(), false /* allowEmptyProjectId */);
+        getLayoutChangedHandler(), true /* requireProjectId */, true /* requireAccount */);
     GridDataFactory.fillDefaults().grab(true, false).applyTo(content);
 
     // we pull in Dialog's content margins which are zeroed out by TitleAreaDialog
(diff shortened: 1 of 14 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -66,7 +66,7 @@ protected Control createDialogArea(final Composite parent) {
 
     Composite container = new Composite(dialogArea, SWT.NONE);
     content = new StandardDeployPreferencesPanel(container, project, loginService,
-        getLayoutChangedHandler(), false /* allowEmptyProjectId */);
+        getLayoutChangedHandler(), true /* requireProjectId */, true /* requireAccount */);
```
elharo: realizing now that this could really be one flag: requireValues. I don't think the preference dialog requires anything to be set. What do you think?

chanseokoh: I like one less parameter. Will go for it.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorObservableValue.java
```diff
@@ -0,0 +1,61 @@
...
+ * An object that represents and binds to the email value of an {@link AccountSelector}, whose
+ * changes (i.e., account selection changes in the combo box) can be tracked by value change
+ * listeners.
+ */
+public class AccountSelectorObservableValue extends AbstractObservableValue {
+
+  private String oldValue;
+  private AccountSelector accountSelector;
+
+  public AccountSelectorObservableValue(final AccountSelector accountSelector) {
```
briandealwis: Because this is an SWT-specific value, you need to set the realm.  Something like the following should work:
```
super(DisplayRealm.forDisplay(accountSelector.getDisplay()));
```
(The default `Realm` may well be a `DisplayRealm`, but better to be safe than sorry.)

PR #805, opened 2016-10-12
getGroupId() --> groupIdField.getText()

### LINKED ISSUE #801: StringIndex out of bounds when entering Maven group ID
To reproduce:

1. Open maven new project wizard
2. Type a letter into the group ID field.
3. Type a space into the group ID field.
3. Type another letter.

java.lang.StringIndexOutOfBoundsException: String index out of range: 2
	at java.lang.String.substring(String.java:1963)
	at com.google.cloud.tools.eclipse.appengine.newproject.maven.MavenAppEngineStandardWizardPage$AutoPackageNameSetterOnGroupIdChange.verifyText(MavenAppEngineStandardWizardPage.java:365)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:273)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4230)

chanseokoh:
Fixes #801.

`getGroupId()` calls `trim()` and doesn't return an actual value typed in the text field.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
@@ -91,11 +91,15 @@ public class MavenArchetypeProjectWizardTest {
   @Test
   public void testSuggestPackageName() {
     assertEquals("aa.bb", MavenAppEngineStandardWizardPage.suggestPackageName("aa.bb"));
+
     assertEquals("aA.Bb", MavenAppEngineStandardWizardPage.suggestPackageName("aA.Bb"));
+
     assertEquals("aa.bb",
         MavenAppEngineStandardWizardPage.suggestPackageName(" a  a\t . b\r b \n"));
+
     assertEquals("aa.bb",
         MavenAppEngineStandardWizardPage.suggestPackageName("....aa....bb..."));
+
     assertEquals("aa._01234bb", MavenAppEngineStandardWizardPage.suggestPackageName(
         "aa`~!@#$%^&*()-+=[]{}<>\\|:;'\",?/._01234bb"));
   }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
@@ -120,10 +120,16 @@ public class MavenArchetypeProjectWizardTest {
     page.groupIdField.setText(" a ");
     assertEquals("a", page.javaPackageField.getText());
 
-    page.groupIdField.setText(" a a");
-    assertEquals("aa", page.javaPackageField.getText());
+    page.groupIdField.setText(" a b");
+    assertEquals("a", page.javaPackageField.getText());
 
-    page.groupIdField.setText(" a a ");
-    assertEquals("aa", page.javaPackageField.getText());
+    page.groupIdField.setText(" a ");
+    assertEquals("a", page.javaPackageField.getText());
+
+    page.groupIdField.setText(" a");
+    assertEquals("a", page.javaPackageField.getText());
+
+    page.groupIdField.setText(" ac");
+    assertEquals("ac", page.javaPackageField.getText());
   }
 }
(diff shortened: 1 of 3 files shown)


elharo:
This should have a test that exposes the bug. Otherwise we won't be sure it's fixed, and it may reappear.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
```diff
@@ -79,19 +90,40 @@ public void testArchetypeDefaultSelection() {
...
+    page.groupIdField.setText(" ");  // setText() triggers VerifyEvent.
+    assertEquals("", page.javaPackageField.getText());
+
+    page.groupIdField.setText(" a");
+    assertEquals("a", page.javaPackageField.getText());
+
+    page.groupIdField.setText(" a ");
+    assertEquals("a", page.javaPackageField.getText());
+
+    page.groupIdField.setText(" a a");
```
elharo: This does not seem right. The package shouldn't autofill form an incorrect groupId.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
```diff
@@ -16,25 +16,36 @@
...
 public class MavenArchetypeProjectWizardTest {
 
   private MavenArchetypeProjectWizard wizard;
+  private Shell shell;
 
   @Before
   public void setUp() {
     assertNotNull(Display.getDefault());
-
+    shell = new Shell(Display.getDefault());
```
akerekes: You could create a [JUnit rule](https://github.com/junit-team/junit4/wiki/rules) out of this, similar to `com.google.cloud.tools.eclipse.appengine.libraries.TestProject`

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
chanseokoh: The current behavior is to try its best to suggest a valid package name by removing, e.g., invalid characters. Since users may go back and forth between a valid group ID state and an invalid state while typing, I think this behavior is acceptable. What else do you think it should do if the group ID is incorrect?

elharo: Off the top of my head, I suspect it should do nothing if the group ID is incorrect, just leave the package name alone.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
chanseokoh: Thanks for the reference. It's good to learn new things. Since the setup and tear-down processes for this test is very small, I'll go with simplicity here.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
chanseokoh: Okay, let me see what I can do.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
akerekes: There are several places we create and dispose a shell in tests, so it could be reused and would be reused in future tests.

chanseokoh: Ah, if for only shell, I think it's a good idea. Coincidentally, I have another PR ready that would refactor unit tests in other packages, so I'll use the rule there.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
chanseokoh: Done. PTAL.

Issue #367, opened 2016-07-15
Missing app-engine-java components


elharo:
We do not recognize or handle the case where gcloud is installed but the app-engine-java components are not:

If not previously installed, install the app-engine-java components into the gcloud SDK: $ gcloud components install app-engine-java 

They are not installed by default with the gcloud SDK, and need to be installed as an extra step.

akerekes:
Should not installing components  be part of the common-lib?

elharo:
Possibly. If they aren't, though, we at least need to give a better error message on our end.

elharo:
Yes

YnR:
Should also check that the right SDK folder was passed. I gave it the bin, and it failed silently.

briandealwis:
@YnR how did you "give it the bin"?  Via the `GOOGLE_CLOUD_SDK_HOME` env var, or via the _Google Cloud Platform_ preference page?  If the latter, there should have been a warning if the directory didn't validate.

elharo:
Initial check is committed. We should still add a better message telling the user what to do when they hit this. Possibly with a hidden label saying to run

gcloud components install app-engine-java

at command line

branflake2267:
The link to installing the sdk in the preferences dialog is opaque because it doesn't mention App Engine.

elharo:
We only validate the app engine java components when the cloud SDK location is explicitly changed. We don't do this when a project is created.

rdayal:
status on this one?

elharo:
I'm verifying it now. There are a couple more pieces we want to do, but I think we've got enough of this done to remove the beta blocker label if not quite close it out.

elharo:
This isn't quite done--I still want to include better instructions for installing the app-engine-java components on the preferences page--but I think we've got enough of it done that we can take it off the beta blockers list for now. If we find users still have problems because appengine-java-components is not available, we can reprioritize additional notifications.

elharo:
Next steps in this dialog:

1. Consider moving the appengine-java-components missing error message below everything else, rather than replacing the "Google Cloud Tools" title at the top.
2. Add instructions for installing the component to the error message.
3. Make sure the error message fires when the dialog starts up.
4. Make sure the  the error message fires when the location is edited. Right now it fires on Apply, OK, or changing the Usage Statistics preference but *not* immediately when the location is changed. 


[...]

PR #808, opened 2016-10-13
Checks for missing App Engine Component

nbashirbello:
Checks that the App Engine Component is installed during the facet installation. If the App Engine component is not installed, it throws an error (with instructions on how to install the component) and uninstalls the facet. So the only time a project can be flagged to have App Engine functionality is when the App Engine component is installed. 

I originally went the route of creating a builder that puts an error marker on App Engine projects when the App Engine component is missing, but it was seemed confusing because (i) the project themselves did not have any problems, the fix was outside the project (ii) even though it was an incremental builder, the times when there was a change in the project could not be correlated with when the app engine component was installed, so we were basically checking at random times if the app engine component had been installed.

Fix for #367

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
@@ -32,10 +32,10 @@ public abstract class AppEngineFacetInstallDelegate implements IDelegate   {
       IProjectFacetVersion version,
       Object config,
       IProgressMonitor monitor) throws CoreException {
-    validateAppEngineComponent();
+    validateAppEngineJavaComponents();
   }
 
-  private void validateAppEngineComponent() throws CoreException {
+  private void validateAppEngineJavaComponents() throws CoreException {
     CloudSdk cloudSdk = new CloudSdk.Builder().build();
     try {
       cloudSdk.validateAppEngineJavaComponents();
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
@@ -26,7 +26,7 @@ import com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInsta
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.eclipse.util.status.StatusUtil;
 
-public abstract class FacetInstallDelegate implements IDelegate   {
+public abstract class AppEngineFacetInstallDelegate implements IDelegate   {
   @Override
   public void execute(IProject project,
       IProjectFacetVersion version,
@@ -42,7 +42,7 @@ public abstract class FacetInstallDelegate implements IDelegate   {
     } catch (AppEngineJavaComponentsNotInstalledException ex) {
       // to properly display error in message box
       throw new CoreException(StatusUtil.error(getClass(),
-          Messages.getString("appengine.component.missing"), ex));
+          Messages.getString("appengine.java.component.missing"), ex));
     }
   }
 }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
@@ -22,12 +22,27 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.wst.common.project.facet.core.IDelegate;
 import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
 
-public class FlexFacetUninstallDelegate implements IDelegate {
+import com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInstalledException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
+import com.google.cloud.tools.eclipse.util.status.StatusUtil;
 
+public abstract class FacetInstallDelegate implements IDelegate   {
   @Override
-  public void execute(IProject project, IProjectFacetVersion facetVersion, Object config, IProgressMonitor monitor)
-      throws CoreException {
-    
+  public void execute(IProject project,
+      IProjectFacetVersion version,
+      Object config,
+      IProgressMonitor monitor) throws CoreException {
+    verifyGcloudAppEngineComponent();
   }
 
+  private void verifyGcloudAppEngineComponent() throws CoreException {
+    CloudSdk cloudSdk = new CloudSdk.Builder().build();
+    try {
+    cloudSdk.validateAppEngineJavaComponents();
+    } catch (AppEngineJavaComponentsNotInstalledException ex) {
+      // to properly display error in message box
+      throw new CoreException(StatusUtil.error(getClass(),
+          Messages.getString("appengine.component.missing"), ex));
+    }
+  }
 }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
@@ -32,13 +32,13 @@ public abstract class FacetInstallDelegate implements IDelegate   {
       IProjectFacetVersion version,
       Object config,
       IProgressMonitor monitor) throws CoreException {
-    verifyGcloudAppEngineComponent();
+    validateAppEngineComponent();
   }
 
-  private void verifyGcloudAppEngineComponent() throws CoreException {
+  private void validateAppEngineComponent() throws CoreException {
     CloudSdk cloudSdk = new CloudSdk.Builder().build();
     try {
-    cloudSdk.validateAppEngineJavaComponents();
+      cloudSdk.validateAppEngineJavaComponents();
     } catch (AppEngineJavaComponentsNotInstalledException ex) {
       // to properly display error in message box
       throw new CoreException(StatusUtil.error(getClass(),
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
@@ -21,7 +21,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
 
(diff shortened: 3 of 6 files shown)


nbashirbello:
screenshot from 2016-10-13 08 47 02

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
```diff
@@ -0,0 +1,48 @@
...
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.wst.common.project.facet.core.IDelegate;
+import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
+
+import com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInstalledException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
+import com.google.cloud.tools.eclipse.util.status.StatusUtil;
+
+public abstract class FacetInstallDelegate implements IDelegate   {
```
elharo: This should probably have a more specific name. We already have several different kinds of FacetInstallDelegate. We want to distinguish them. Maybe AppEngineFacetInstallDelegate?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
```diff
@@ -0,0 +1,48 @@
...
+    validateAppEngineComponent();
+  }
+
+  private void validateAppEngineComponent() throws CoreException {
+    CloudSdk cloudSdk = new CloudSdk.Builder().build();
+    try {
+      cloudSdk.validateAppEngineJavaComponents();
+    } catch (AppEngineJavaComponentsNotInstalledException ex) {
+      // to properly display error in message box
+      throw new CoreException(StatusUtil.error(getClass(),
```
elharo: @briandealwis Is this the right way to warn of a problem like this? I thought this was mostly for unanticipated problems. Can we put up a more specific dialog box here?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
```diff
@@ -1,17 +1,32 @@
...
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.wst.common.project.facet.core.IDelegate;
 import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
 
-public class FlexFacetInstallDelegate implements IDelegate {
+public class FlexFacetInstallDelegate extends FacetInstallDelegate {
 
   @Override
   public void execute(IProject project, IProjectFacetVersion facetVersion, Object config, IProgressMonitor monitor)
```
elharo: I think you can just inherit this method instead.

nbashirbello:
@elharo

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/messages.properties
```diff
@@ -0,0 +1 @@
+appengine.component.missing=Missing Java App Engine component
```
elharo: appengine.java.component.missing perhaps

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
nbashirbello: The facet installation framework already has a way for displaying messages during installation failure. Once you throw a CoreException, it handles the message display. However just throwing the AppEngineJavaComponentsNotInstalledException caused the "Reason:" to be the same as the detailed message (look at dialog image)

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FlexFacetInstallDelegate.java
elharo: You don't need to override it at all. Just delete it.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetInstallDelegate.java
elharo: Yes, I'm suggesting  that the dialog it pops up is not very user friendly and we might well do better with a custom dialog.

briandealwis: I don't think we should pop up our own dialog here since this isn't a UI component and we don't know what we're going to interrupt.

But we could make our status message ("_Missing App Engine Java component_"), which is shown in the dialog text itself, more helpful. The user may have made several facet changes and so it may not be clear that the error was caused by this App Engine facet.  It looks like we could include the `AppEngineJavaComponentsNotInstalledException` message, or maybe "_Google Cloud SDK installation is missing App Engine Java component_: please run 'gcloud components …etc…'"?

nbashirbello:
screenshot from 2016-10-13 13 17 55

Dialog before clicking "Details>>"

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/messages.properties
```diff
@@ -0,0 +1 @@
+appengine.java.component.missing=Missing Java App Engine component
```
elharo: Cloud SDK App Engine Java components not installed

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
```diff
@@ -0,0 +1,48 @@
...
+public abstract class AppEngineFacetInstallDelegate implements IDelegate   {
+  @Override
+  public void execute(IProject project,
+      IProjectFacetVersion version,
+      Object config,
+      IProgressMonitor monitor) throws CoreException {
+    validateAppEngineComponent();
+  }
+
+  private void validateAppEngineComponent() throws CoreException {
```
elharo: validateAppEngineJavaComponents

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
```diff
@@ -0,0 +1,48 @@
...
+    CloudSdk cloudSdk = new CloudSdk.Builder().build();
+    try {
+      cloudSdk.validateAppEngineJavaComponents();
+    } catch (AppEngineJavaComponentsNotInstalledException ex) {
+      // to properly display error in message box
+      throw new CoreException(StatusUtil.error(getClass(),
+          Messages.getString("appengine.java.component.missing"), ex));
```
elharo: Where is the message in the details section coming from?

briandealwis: The details message is the `AppEngineJavaComponentsNotInstalledException` message.

elharo: I wonder if we should replace that message with one of our own choosing? Or maybe this is OK as long as we update the main message to be clearer like we discussed yesterday so users don't need to read the details.

This approach still feels wonky to me. It's as if facets were designed under the assumption that a failure to attach one was a very unusual condition and not a normal part of operation.

elharo:
I don't think the test failures are related, but I kicked off another build to find out.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFacetInstallDelegate.java
elharo: I don't think we have time to explore that path right now, maybe down the road. For now let's just try to get the best error message we can, and then improve it.

chanseokoh: I vote for a custom dialog for reasons I will talk about later when there is a chance. For now, let's just try to get the best error message here.

briandealwis: I think it's risky to open our own dialog here: we don't know what state WTP will be in, and it may lead to problems, even possibly data corruption.

We know the dialog here is pretty terrible, so let's avoid getting to this situation in the first place.

elharo: The problem is this is the first time we may realize we're in this situation. We don't have an earlier point to avoid getting into it. :-(

elharo: I think we need at least one of two changes here:

A) Provide the command to install the components above the fold (i.e. without expanding the details section) 
B) Replace the AppEngineJavaComponentsNotInstalledException message with a message from our own messages.properties

nbashirbello:
@elharo What message do you want under "Reason" and under "Details"?

Currently, the messages are:
"Reason": The Cloud SDK App Engine Java component is not installed
"Details": Validation Error: Java App Engine components not installed. Fix by running 'gcloud components install app-engine-java' on command-line.

briandealwis:
Having just hit this on Windows, having the `Fix by running 'gcloud components install app-engine-java' on command-line.` would be handy.

elharo:
Up top how about:

Project setup failed because the Cloud SDK App Engine Java component is not installed.
Fix by running 'gcloud components install app-engine-java' on command-line.

If we get all that, reasonably wrapped, into the top part then I don't care what's in details. Otherwise, details should maybe be "Fix by running 'gcloud components install app-engine-java' on command-line."

I wonder, does Eclipse let us make any of this a hyperlink?

briandealwis:
I don't think there's hyperlink support.

nbashirbello:
screenshot from 2016-10-17 09 00 26

elharo:
Screenshot doesn't seem to reflect the new message?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/messages.properties
```diff
@@ -0,0 +1 @@
+appengine.java.component.missing=Project setup failed because the Cloud SDK App Engine Java component is not installed. Fix by running 'gcloud components install app-engine-java' on command-line.
```
elharo: on the command-line

nbashirbello:
screenshot from 2016-10-18 09 58 09

briandealwis:
FWIW LGTM

PR #815, opened 2016-10-13
define artifacts when signing

elharo:
@chanseokoh


chanseokoh:
My mistake. Removed the reference.

PR #826, opened 2016-10-14
use library name for display and fall back to library id if it's null or empty

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -128,6 +129,14 @@ private void addManageLibrariesWidgets(Composite container) {
     GridLayoutFactory.fillDefaults().applyTo(apiGroup);
   }
 
+  private String getLibraryName(Library library) {
```
elharo: static

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -118,7 +119,7 @@ private void addManageLibrariesWidgets(Composite container) {
     List<Library> libraries = getLibraries();
     for (Library library : libraries) {
       Button libraryButton = new Button(apiGroup, SWT.CHECK);
-      libraryButton.setText(library.getId());
+      libraryButton.setText(getLibraryName(library));
```
elharo: I'm thinking about how to test this method, but probably not right now.

PR #827, opened 2016-10-14
Fixes #825 - Enhance API attacher extension point

### LINKED ISSUE #825: Enhance API attacher extension point
Add the possibility to define dependencies among libraries and the recommendation level (?) of a library, which can be:

 - required: the library is presented with a disabled, selected checkbox
 - suggested: the library is presented with an enabled, selected checkbox
 - optional: the library is presented with an enabled, unselected checkbox

in the API attacher UI widget.

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
```diff
@@ -102,4 +104,14 @@ public void setLibraryDependencies(List<String> libraryDependencies) {
       this.libraryDependencies = new ArrayList<>(libraryDependencies);
     }
   }
+
+  public void setRecommendation(LibraryRecommendation recommendation) {
+    if (recommendation != null) {
+      this.recommendation = recommendation;
```
elharo: should we have a response to a null value rather than simply ignoring it?

akerekes: The response I can imagine would be to throw an exception. My feeling is that the intention of clients setting this to null is the same as setting it to an empty collection -> clearing it.

elharo: This doesn't clear it though. Maybe we should throw the exception?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
```diff
@@ -66,6 +69,11 @@ public Library create(IConfigurationElement configurationElement) throws Library
         if (exportString != null) {
           library.setExport(Boolean.parseBoolean(exportString));
         }
+        String recommendationString = configurationElement.getAttribute(ATTR_RECOMMENDATION);
+        if (recommendationString != null) {
+          library.setRecommendation(LibraryRecommendation.valueOf(recommendationString));
```
elharo: This can throw an IllegalArgumentException if the string isn't recognized.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Library.java
akerekes: True, I totally missed that. Ok, I'll check for null (and in other setter methods as well)

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
akerekes: I'll catch it in the surrounding try/catch.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
```diff
@@ -151,6 +159,15 @@ private MavenCoordinates getMavenCoordinates(IConfigurationElement[] children) {
     return filters;
   }
 
+  private List<String> getLibraryDependencies(IConfigurationElement[] children) {
+    List<String> libraryDependencies = new ArrayList<>();
+    for (IConfigurationElement libraryDependencyElement : children) {
+      if (libraryDependencyElement.getName().equals(ELMT_LIBRARY_DEPENDENCY)) {
+      }
```
chanseokoh: Why is this empty? Need to add a TODO?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
chanseokoh: It seems like you also have to do `recommendationString.toUpperCase()`. Otherwise, `valueOf()` fails in my small test.

akerekes:
@chanseokoh @elharo PTAL

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
```diff
@@ -151,6 +159,19 @@ private MavenCoordinates getMavenCoordinates(IConfigurationElement[] children) {
     return filters;
   }
 
+  private List<String> getLibraryDependencies(IConfigurationElement[] children) {
+    List<String> libraryDependencies = new ArrayList<>();
+    for (IConfigurationElement libraryDependencyElement : children) {
+      if (libraryDependencyElement.getName().equals(ELMT_LIBRARY_DEPENDENCY)) {
```
chanseokoh: Do we need to be this much conservative to check the name again, because on line 76, we get these `children` by `getChildren(ELMT_LIBRARY_DEPENDENCY)`, i.e., `children` are already all `ELMT_LIBRARY_DEPENDENCY`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
```diff
@@ -71,7 +71,7 @@ public Library create(IConfigurationElement configurationElement) throws Library
         }
         String recommendationString = configurationElement.getAttribute(ATTR_RECOMMENDATION);
         if (recommendationString != null) {
-          library.setRecommendation(LibraryRecommendation.valueOf(recommendationString));
+          library.setRecommendation(LibraryRecommendation.valueOf(recommendationString.toUpperCase()));
```
elharo: Needs a Locale

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/config/LibraryFactory.java
elharo: If you do, reverse the equals clause to protect against null pointers

akerekes:
@chanseokoh @elharo PTAL

PR #835, opened 2016-10-17
JDK 1.7_79

elharo:
@briandealwis


briandealwis:
Looking on the [JavaSE Archives page](http://www.oracle.com/technetwork/java/javase/downloads/java-archive-downloads-javase7-521261.html), the last release was 7u80.

PR #838, opened 2016-10-17
Force dialog re-validation for account email

### LINKED ISSUE #831: Deploy dialog may complain to select an account even if it is selected
This requires some steps to reproduce:

1) Select and save an account as project preferences.
2) Log out.
3)  Do 'Deploy to App Engine Standard...'. (The deploy dialog pops up, no account is selected, and the only item in the account combo box is `<Add a new account...>`.)
4) Select `<Add a new account...>`, and log in with the account previously saved.
5) Observe that the account is selected in the dialog, but the dialog complains to select an account.

I suspect there is a databinding sync issue with the control and the data model. Triggering a sync may be necessary.

chanseokoh:
Fixes #831.

Currently, we have 3 objects being sync'ed through databinding for saving and loading deploy preferences:

Standard Java Preferences <--> Model Object <--> Account Selector

This PR employs a trick that invalidates the email value in the model (by setting a special value) right after sync'ing it with the Account Selector. As a result, databinding will always trigger dialog re-validation whenever the value in the Account Selector changes.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModelTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModelTest.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+
+package com.google.cloud.tools.eclipse.appengine.deploy.ui;
+
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import com.google.cloud.tools.eclipse.appengine.deploy.standard.StandardDeployPreferences;
+
+import org.junit.Test;
+import org.osgi.service.prefs.BackingStoreException;
+
+public class DeployPreferencesModelTest {
+
+  @Test
+  public void testSavePreferences_ignoreRevalidationTrickEmail() throws BackingStoreException {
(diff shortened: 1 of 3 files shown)


briandealwis:
@chanseokoh, I'm sorry I missed this earlier.  The way to do this is to provide an `UpdateValueStrategy`.  So in `setupAccountEmailDataBinding()`, you'd do something like:
```
    context.bindValue(new AccountSelectorObservableValue(accountSelector), accountEmailModel,
        null, // target to model update
        new UpdateValueStrategy().setAfterGetValidator(new ValidateAccountEmail(loginService)));
```
That `ValidateAccountEmail` would verify that the retrieved account email is a valid account and fail if not.

chanseokoh:
I spent a lot of time trying that idea, but I failed. I see `UpdateStrategy` can play a part in validation, but ultimately, I wasn't able to make it work to fix #831 in our situation. In #831, we need to clear an existing validation error. Making `UpdateStrategy()` return OK doesn't clear it. I think it can be used to raise validation errors early while syncing values, but it doesn't have the authority to override/clear the overall dialog error declared by the last validation result.

briandealwis:
Did you try a converter, to at least convert the invalid value to null? (because that's what we want to do, right?)

chanseokoh:
We are not trying to set an invalid email here. The situation is like this: the model holds a valid email, but the value in the account selector is am empty string due to the account not being logged in. The dialog will first rightly say "you have to select an account." Then, when a user logs in with the email in the model, the account selector will have the email as its value. Now the user has selected the account, the dialog should clear the error. However, the databinding framework sees that the email in the model and the email in the account are same, and there has been no change. So, even if the account notifies the framework by firing the value-change event to notify the framework, the framework sees no actual change, so it skips the dialog validation, leaving the dialog error untouced.

briandealwis:
A few things:

> However, the databinding framework sees that the email in the model and the email in the account are same, it thinks that there has been no change to this field.

Databinding does not swallow changes.  If I place a breakpoint on `DeployPreferencesModel#setAccountEmail()` and repeat the steps in #831, I see the `setAccountEmail()` method called.

> the model holds a valid email, but the value in the account selector is am empty string due to the account not being logged in.

It seems to me then that the model does *not* hold a valid email, since it doesn't correspond to a known account.

——
All the above said, I've rewritten `StandardDeployPreferencesPanel#setupAccountEmailDataBinding()` in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/alt831 that uses an model-to-target `UpdateValueStrategy` that:

  - provides an converter transforms any saved accounts that aren't found into a null
  - adds a set validator that raises an error when `requireValues == true` and there is no valid account being set

That seems to fix the issue. PTAL.

chanseokoh:
Aha! Now I see how the Converter can work here: you give the selector a null value if the selector cannot set it, instead of giving the actual email. That did the trick. Thanks a lot for giving a try on this! I'll do some refactoring based on your input shortly.

> Databinding does not swallow changes. If I place a breakpoint on DeployPreferencesModel#setAccountEmail() and repeat the steps in #831, I see the setAccountEmail() method called.

Certainly, it calls `setAccountEmail()`. I was talking about skipping validation deep inside databinding. Below is the code of `SimplePropertyObservableValue.notifyIfChanged()`. As you can see, `fireValueChange()`, which eventually calls our `invalidate()`, is not called if `newValue` and `oldValue` are same.

selection_001

PR #839, opened 2016-10-18
Use databinding converter and update strategy to avoid validation issues on deploy dialog

### LINKED ISSUE #831: Deploy dialog may complain to select an account even if it is selected

briandealwis:
Alternative fix #831

(Or if you prefer to refactor, we can close this.)

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
@@ -102,7 +102,6 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
 
   private Runnable layoutChangedHandler;
   private boolean requireValues = true;
-  private IGoogleLoginService loginService;
 
   public StandardDeployPreferencesPanel(Composite parent, IProject project,
       IGoogleLoginService loginService, Runnable layoutChangedHandler, boolean requireValues) {
@@ -110,7 +109,6 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
 
     this.layoutChangedHandler = layoutChangedHandler;
     this.requireValues = requireValues;
-    this.loginService = loginService;
 
     createCredentialSection(loginService);
 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
@@ -152,11 +152,11 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
         return ValidationStatus.ok();
       }
     };
-    UpdateValueStrategy targetToModel =
-        new UpdateValueStrategy().setBeforeSetValidator(accountSelectedChecker);
-    UpdateValueStrategy modelToTarget =
-        new UpdateValueStrategy().setBeforeSetValidator(accountSelectedChecker);
-    modelToTarget.setConverter(new Converter(String.class, String.class) {
+    UpdateValueStrategy targetToModel = new UpdateValueStrategy()
+        .setBeforeSetValidator(accountSelectedChecker);
+    UpdateValueStrategy modelToTarget = new UpdateValueStrategy()
+        .setBeforeSetValidator(accountSelectedChecker)
+        .setConverter(new Converter(String.class, String.class) {
       @Override
       public Object convert(Object fromObject /* email */) {
         return accountSelector.isEmailAvailable((String) fromObject) ? fromObject : null;
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
@@ -143,7 +143,7 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
   }
 
   private void setupAccountEmailDataBinding(DataBindingContext context) {
-    IValidator checker = new IValidator() {
+    IValidator accountSelectedChecker = new IValidator() {
       @Override
       public IStatus validate(Object value /* email */) {
         if (requireValues && Strings.isNullOrEmpty((String) value)) {
@@ -152,8 +152,10 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
         return ValidationStatus.ok();
       }
     };
-    UpdateValueStrategy targetToModel = new UpdateValueStrategy().setBeforeSetValidator(checker);
-    UpdateValueStrategy modelToTarget = new UpdateValueStrategy().setBeforeSetValidator(checker);
+    UpdateValueStrategy targetToModel =
+        new UpdateValueStrategy().setBeforeSetValidator(accountSelectedChecker);
+    UpdateValueStrategy modelToTarget =
+        new UpdateValueStrategy().setBeforeSetValidator(accountSelectedChecker);
     modelToTarget.setConverter(new Converter(String.class, String.class) {
       @Override
       public Object convert(Object fromObject /* email */) {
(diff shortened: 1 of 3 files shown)


elharo:
If you'e going to take this, please assign the issue to yourself first. Thanks.

chanseokoh:
There is another issue. I think we need a targetToModel update strategy too, so that when an account is de-selected (for whatever reason) in the selector, it can raise a validation error. Steps to reproduce:

1) Select and save an account. Close any dialog.
2) Open the deploy dialog. Observe that the saved account is auto-selected.
3) Select `<Add a new account>`. Don't complete login.
4) Come back to IDE and cancel login. Observe that the selector clears the previous account selection. (Intended behavior.)
5) Observe that the dialog doesn't say "you need to select an account."

briandealwis:
@chanseokoh I disagree with this statement:

> The email values in the combo are the source of truth.

The login service is the source of truth, the combo is a view on that truth.

Adding a similar target-to-model UpdateValueStrategy makes sense to me.

chanseokoh:
@briandealwis Okay, if you think the combo is just a view, then I insist you re-initialize the combo whenever you call `getAccounts()` to do something with the combo, since the emails in the combo may be out of sync with the account list returned from `getAccounts()`.

Would you continue to work on this?

chanseokoh:
Had an offline discussion with @briandealwis. I will take and complete this PR.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -141,23 +143,27 @@ private void setupDataBinding() {
...
-        @Override
-        protected IStatus validate() {
-          String email = (String) accountEmailModel.getValue();
-          // It's possible that no account is selected while a valid email has been saved in the
-          // model (if the corresponding account is signed out), so check the actual selection too.
-          if (email.isEmpty() || accountSelector.getSelectedEmail().isEmpty()) {
-            return ValidationStatus.error(Messages.getString("error.account.missing"));
-          }
-          return ValidationStatus.ok();
+    IValidator checker = new IValidator() {
```
briandealwis: Can we give this a more suggestive name?  `accountSelectedChecker`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -141,23 +143,27 @@ private void setupDataBinding() {
...
-    }
+        return ValidationStatus.ok();
+      }
+    };
+    UpdateValueStrategy targetToModel = new UpdateValueStrategy().setBeforeSetValidator(checker);
+    UpdateValueStrategy modelToTarget = new UpdateValueStrategy().setBeforeSetValidator(checker);
+    modelToTarget.setConverter(new Converter(String.class, String.class) {
+      @Override
+      public Object convert(Object fromObject /* email */) {
+        return accountSelector.isEmailAvailable((String) fromObject) ? fromObject : null;
```
briandealwis: Seems a bit unusual for `isEmailAvailable()` to support null or empty string. WDYT?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -73,7 +82,7 @@ public String getSelectedEmail() {
   }
 
   public int selectAccount(String email) {
-    int index = combo.indexOf(email);
+    int index = Strings.isNullOrEmpty(email) ? -1 : combo.indexOf(email);
     if (index != -1) {
```
briandealwis: Shouldn't this have an `else { selectedCredential = null; }`?

chanseokoh: That's intended not to change the previously selected account if the method fails to find a new selection.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
chanseokoh: Makes two lines wrapped :cry:. (Just kidding. Done.)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
chanseokoh: On second thought, I think it's convenient to support null and the empty string for callers. Also, I think it's conceptually not wrong: if the given email — whether it's null (cannot ever match) or the empty string (may match a blank item if it exists) — matches a value in the combo, it will return true (or selected for `selectAccount`). I'll add a javadoc.

briandealwis: Should we be using the email as an account key then?  If the model stores "", and there are two accounts with blank email, which one should be selected?  Perhaps we should be using our own UID.

chanseokoh: Yes, we can conveniently use the email as a unique account key. No logged-in account can have a null or empty email. The ide-login library guarantees that. So, `""` in the model implies no selection.

chanseokoh: > whether it's null (cannot ever match) or the empty string (may match a blank item if it exists)

After some thought, I decided to treat null and the empty string equally — they never match any account. It sounds weird to say "an empty email address matches an empty string item in the combo."

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
chanseokoh: Done.

chanseokoh:
It may be useful for users to be able to unset an account. I'd though about that before, and I did expect that someone will bring up this. I'll create a new issue for this.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -141,23 +143,29 @@ private void setupDataBinding() {
...
         }
-      });
-    }
+        return ValidationStatus.ok();
+      }
+    };
+    UpdateValueStrategy targetToModel =
+        new UpdateValueStrategy().setBeforeSetValidator(accountSelectedChecker);
+    UpdateValueStrategy modelToTarget =
+        new UpdateValueStrategy().setBeforeSetValidator(accountSelectedChecker);
```
briandealwis: You may want to change the formatting here to cascade the `.setConverter()` call to be consistent with the other calls.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -58,6 +59,14 @@ public AccountSelector(Composite parent, IGoogleLoginService loginService,
   }
 
   /**
+   * @return true if this selector has (non-null and non-empty) {@code email} as an item;
```
briandealwis: That reads to me that `email` is expected to be non-null and non-empty.  Maybe:
```
@return true if this selector lists an account with
  {@code email}. For convenience of callers,
  {@code email} may be {@code null} or empty,
  which always returns false.
```

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -72,8 +81,18 @@ public String getSelectedEmail() {
     return combo.getText();
   }
 
+  /**
+   * Selects an account corresponding to the given {@code email} and returns its index of the
+   * combo item. If there doesn't exist an account corresponding to the {@code email}, returns -1
```
briandealwis: doesn't exist an -> is no

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -72,8 +81,18 @@ public String getSelectedEmail() {
...
+   * while retaining current selection (if any).
+   *
+   * @param email to use to select an account. If {@code null} or the empty string, or
```
briandealwis: IMHO `to use` doesn't read right here

elharo:
fix #831

PR #845, opened 2016-10-18
Checks that the App Engine component exists in the Create New App Engine project wizard

nbashirbello:
(non-maven only)
When the user opens the Create New App Engine Standard wizard, a check for the App Engine component is done. If the App Engine component does not exist, a message is shown and users would not be able to create the project.
Fix for #367


nbashirbello:
@elharo

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -36,45 +61,51 @@ public void addPages() {
...
-    AnalyticsPingManager.getInstance().sendPing(
-        AnalyticsEvents.APP_ENGINE_NEW_PROJECT_WIZARD_COMPLETE,
-        AnalyticsEvents.APP_ENGINE_NEW_PROJECT_WIZARD_TYPE,
-        AnalyticsEvents.APP_ENGINE_NEW_PROJECT_WIZARD_TYPE_NATIVE);
-
-    if (config.getCloudSdkLocation() == null) {
-      File location = CloudSdkPrompter.getCloudSdkLocation(getShell());
-      if (location == null) {
-        return false;
+    if (hasAppEngineComponent) {
```
briandealwis: Could be argued that we shouldn't allow finish if they don't have the components installed?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -20,12 +39,18 @@
...
-  private AppEngineStandardWizardPage page;
+  private WizardPage page;
   private AppEngineStandardProjectConfig config = new AppEngineStandardProjectConfig();
+  private boolean hasAppEngineComponent;
 
   public StandardProjectWizard() {
     this.setWindowTitle("New App Engine Standard Project");
-    page = new AppEngineStandardWizardPage();
+    hasAppEngineComponent = doesAppEngineJavaComponentExist();
+    if (hasAppEngineComponent) {
```
briandealwis: Would it make sense to:
  - prepend `AppEngineComponentPage` to be the first page, such that `AppEngineStandardWizardPage` will be the second page
  - have `AppEngineComponentPage` have a _Check_ button that would run the `doesAppEngineJavaComponentExist()` check
  - and `AppEngineComponentPage#canFlipToNextPage()` check that `doesAppEngineJavaComponentExist()` is true

Otherwise the user needs to cancel and restart the new wizard.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
nbashirbello: Discussed this scenario earlier with @elharo. Decided on this flow.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineComponentPage.java
```diff
@@ -51,12 +53,24 @@ public void createControl(Composite parent) {
...
+
+    int startIndex = message.indexOf("\'");
+    int endIndex = message.indexOf("\'", startIndex + 1);
+    if ((-1 < startIndex) && (startIndex < endIndex)) {
+      StyleRange styleRange = new StyleRange();
+      styleRange.start = startIndex + 1;
+      styleRange.length = endIndex - startIndex;
+      styleRange.fontStyle = SWT.BOLD;
+      styledText.setStyleRange(styleRange);
+    }
```
briandealwis: Nice idea.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/messages.properties
```diff
@@ -1 +1,2 @@
-AppEngineStandardWizardPage_librariesGroupLabel=Libraries to add to build path
+appengine.standard.wizard.page.libraries.group.label=Libraries to add to build path
+appengine.java.component.missing=An App Engine project cannot be created because the Cloud SDK App Engine Java component is not installed. Fix by running 'gcloud components install app-engine-java' on the command-line
```
chanseokoh: Period at the end of the message?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineComponentPage.java
```diff
@@ -0,0 +1,80 @@
...
+    setDescription("The Cloud SDK App Engine Java component is not installed"); 
+  }
+
+  @Override
+  public void createControl(Composite parent) {
+    // TODO: should we create a ping for missing app engine component
+    AnalyticsPingManager.getInstance().sendPing(
+        AnalyticsEvents.APP_ENGINE_NEW_PROJECT_WIZARD,
+        AnalyticsEvents.APP_ENGINE_NEW_PROJECT_WIZARD_TYPE,
+        AnalyticsEvents.APP_ENGINE_NEW_PROJECT_WIZARD_TYPE_NATIVE, parent.getShell());
```
chanseokoh: I think we should send a ping only when a user actually creates a project through wizard. Poking around the wizard while not actually using it isn't really using this feature.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineComponentPage.java
```diff
@@ -0,0 +1,80 @@
...
+
+    Composite container = new Composite(parent, SWT.NONE);
+    GridLayoutFactory.swtDefaults().numColumns(1).applyTo(container);
+
+    GridData gridData = new GridData(GridData.FILL_HORIZONTAL);
+    gridData.widthHint = parent.getSize().x;
+
+    String message = Messages.getString("appengine.java.component.missing");
+    StyledText styledText = new StyledText(container, SWT.MULTI | SWT.READ_ONLY | SWT.WRAP);
```
chanseokoh: I noticed that it shows a blinking caret when clicked on. Below will hide it completely.

```java
styledText.setCaret(null /* hide caret */);
```

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
elharo: We thought it would be simpler to have the user cancel out of the wizard and restart, rather than adding a button whose only purpose was to check for the component.

If there were a way the wizard could notice the component was installed and enable the next button, it would certainly be better to check in a single page.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
elharo: Yes, that does make sense except we couldn't figure out an easy way to change state once the component was installed.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/messages.properties
elharo: "An App Engine project cannot be created" --> "Cannot create an App Engine Eclipse project"

elharo: Can we make Cloud SDK App Engine Java component a hyperlink like we do in the preference page?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
chanseokoh: This extra check isn't really necessary because `AppEngineStandardWizardPage` disables the finish button. WDYT?

In any case, this method returns true if `hasAppEngineComponent` is false, which doesn't sound quite right. Since this is a huge if-stmt block, I suggest the following for better readability:

```java
  if (hasAppEngineComponent) {
    return false;
  }
  ...
```

chanseokoh: Oops. Reverted the if condition above.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -36,45 +61,51 @@ public void addPages() {
...
-    }
-    // todo is this the right time/place to grab these?
-    config.setAppEngineProjectId(page.getAppEngineProjectId());
-    config.setPackageName(page.getPackageName());
 
-    config.setProject(page.getProjectHandle());
-    if (!page.useDefaults()) {
-      config.setEclipseProjectLocationUri(page.getLocationURI());
-    }
+      if ((page != null) && (page instanceof AppEngineStandardWizardPage)) {
```
chanseokoh: `(page != null)` is unnecessary, since `null instanceof Object` is safely false.

elharo:
This is what you get if the cloud SDK is not installed (not just the app engine java component):

image

If you open the details you get a bit more info but still no fix instructions:

Google Cloud SDK path was not provided and could not be found in any known install locations.

I'd much rather have a dialog like this one that tells the user how to install the Cloud SDK and doesn't include an exception stack trace, or require them to open the details section. Again, probably in another PR unless that somehow changes what we do here.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
briandealwis: I think it would be more helpful if this page did the following:
  - check if there is a Cloud SDK: if none (or not found), then show a `Button` or `Link` that would open the Cloud SDK pref page to configure the location.
  - if missing java components: provide a `Button` or `Link` that would offer to invoke `gcloud components install app-engine-java` with a little transcript to show the results of the command.  If successful, mark the page as valid so the user can move on with _Next_.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
briandealwis: If `AppEngineComponentPage` is never valid, then the _Finish_ button will never be enabled and `performFinish()` can never be called.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -20,12 +39,18 @@
 
 public class StandardProjectWizard extends Wizard implements INewWizard {
 
-  private AppEngineStandardWizardPage page;
+  private WizardPage page;
```
briandealwis: So maybe we should hold onto both the `AppEngineStandardWizardPage` and the `AppEngineComponentPage`.  That would avoid the `instanceof` checking elsewhere.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
elharo: Those are very good ideas. We definitely want to make this as easy on and obvious to the user as possible. That's going to take quite a bit more work, and probably several PRs so I'd like to move ahead with this much and iterate.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineComponentPage.java
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
elharo: Delete "All Rights Reserved."

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -1,18 +1,19 @@
-/*******************************************************************************
- * Copyright 2016 Google Inc.
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
elharo: Delete "All Rights Reserved."

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/Messages.java
```diff
@@ -1,15 +1,43 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
elharo: Delete "All Rights Reserved."

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
```diff
@@ -1,5 +1,23 @@
+/*
+ * Copyright 2016 Google Inc. All Rights Reserved.
```
elharo: Delete "All Rights Reserved."

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/Messages.java
```diff
@@ -16,28 +16,18 @@
...
 
-import java.util.MissingResourceException;
-import java.util.ResourceBundle;
-
 import org.eclipse.osgi.util.NLS;
 
 public class Messages extends NLS {
   private static final String BUNDLE_NAME = "com.google.cloud.tools.eclipse.appengine.newproject.messages"; //$NON-NLS-1$
-  private static final ResourceBundle RESOURCE_BUNDLE = ResourceBundle.getBundle(BUNDLE_NAME);
+  public static String AppEngineStandardWizardPage_librariesGroupLabel;
```
elharo: FWIW, I prefer the version that doesn't rely on mutable public fields.

chanseokoh:
Sorry, I was wrong about removing the Analytics ping here. I forgot that there were two different Analytics events. One for just bringing up the wizard, and the other for actually creating a project. Talked to @nbashirbello offline, and she will restore the ping she removed.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/StandardProjectWizard.java
chanseokoh: FYI: doing this will (at the moment) affect Analytics pings, as each page will send a ping when it is created.

PR #871, opened 2016-10-21
Enable configuring service port for local run

chanseokoh:
The server port field (spinner) is placed below the runtime environment combo. That's practically the only place we can add this field. This spinner is dynamically shown/hide as you change the server type to/from our App Engine Server.

new server _009

One minor problem: the new spinner makes the server selection box even smaller on my machine, to the point that it looks like a single-line text box. I can still clearly see our "App Engine Standard" server though.

Lastly, I removed `<useUIThread>false</useUIThread>` (which will set it to true by default) in `pom.xml` to make tests work. Tests can't access `Display` or `Shell` when they are run in a background thread.


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -52,23 +63,14 @@
...
-import java.util.logging.Logger;
-
 public class LocalAppEngineServerLaunchConfigurationDelegate
     extends AbstractJavaLaunchConfigurationDelegate {
 
   private final static Logger logger =
       Logger.getLogger(LocalAppEngineServerLaunchConfigurationDelegate.class.getName());
 
   private static final String DEBUGGER_HOST = "localhost";
+  private static final String LOCAL_HOST = "localhost";
```
briandealwis: seems funny to have an underscore in the name?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -199,9 +200,11 @@ private boolean shouldOpenStartPage() {
...
   }
 
-  private String determinePageLocation(IServer server, ILaunchConfiguration config) {
-    // todo[issue #259]: pull this from the server or launch configuration
-    return "http://" + DEBUGGER_HOST + ":8080";
+  @VisibleForTesting
+  static String determinePageLocation(IServer server) {
+    int port = server.getAttribute(ServerPortExtension.SERVER_ATTRIBUTE_PORT,
+                                   ServerPortExtension.DEFAULT_SERVICE_PORT);
+    return "http://" + LOCAL_HOST + ":" + port;
```
briandealwis: Should we not be using `server.getHost()`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -52,23 +63,14 @@
...
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
 public class LocalAppEngineServerLaunchConfigurationDelegate
     extends AbstractJavaLaunchConfigurationDelegate {
 
   private final static Logger logger =
       Logger.getLogger(LocalAppEngineServerLaunchConfigurationDelegate.class.getName());
 
   private static final String DEBUGGER_HOST = "localhost";
```
briandealwis: Is `DEBUGGER_HOST` still needed then?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,77 @@
...
+  @VisibleForTesting Spinner portSpinner;
+
+  @Override
+  public void createControl(UI_POSITION position, Composite parent) {
+    if (position == UI_POSITION.BOTTOM) {
+      portLabel = new Label(parent, SWT.NONE);
+      portLabel.setVisible(false);
+      portLabel.setText(Messages.NEW_SERVER_DIALOG_PORT);
+
+      portSpinner = new Spinner(parent, SWT.BORDER);
```
briandealwis: Rather than a `Spinner` perhaps this should be a `Text` with a `VerifyListener` that ensures it's a number?  After all, is it likely for someone to really use up/down?

chanseokoh: Others use `Spinner`s too. The only UI difference from a `Text` is the spinner added at the right. We don't have to implement code to validate input ourselves, so I think it's a huge advantage.

Basic HTTP Server:
new server _010

Jetty:
selection_012

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
chanseokoh: I intended it, but on second thought, I'll just use `server.getHost()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
chanseokoh: I'll remove `LOCAL_HOST`. `DEBBUGER_HOST` is still used somewhere.

elharo:
I agree with @briandealwis Text makes more sense than spinner here.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtentionTest.java
```diff
@@ -0,0 +1,125 @@
...
+import java.beans.PropertyChangeEvent;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.wst.server.core.IServerType;
+import org.eclipse.wst.server.ui.wizard.ServerCreationWizardPageExtension.UI_POSITION;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ServerPortExtentionTest {
```
elharo: ExtensionTest

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
elharo: Isn't there a numeric text field somewhere? Spinners are really for *quantities* and this isn't that.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,77 @@
...
+      portSpinner.setVisible(false);
+      portSpinner.setMinimum(1);
+      portSpinner.setMaximum(65535);
+      portSpinner.setSelection(DEFAULT_SERVICE_PORT);
+      portSpinner.addModifyListener(new ModifyListener() {
+        @Override
+        public void modifyText(ModifyEvent event) {
+          serverWc.setAttribute(SERVER_ATTRIBUTE_PORT, portSpinner.getSelection());
```
elharo: what's Wc?

chanseokoh: Server Working Copy. It's a variable defined in the base class (Eclipse code), so we can't rename it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
briandealwis: I'm pretty sure `DEBUGGER_HOST` is a todo.  It would be good to use `getServer().getHost()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegateTest.java
```diff
@@ -30,8 +30,9 @@
   @Test
   public void testDeterminePageLocation() {
     IServer server = mock(IServer.class);
+    when(server.getHost()).thenReturn("0.0.0.0");
     when(server.getAttribute(eq("appEngineDevServerPort"), anyInt())).thenReturn(8085);
     String url = LocalAppEngineServerLaunchConfigurationDelegate.determinePageLocation(server);
-    assertEquals("http://localhost:8085", url);
+    assertEquals("http://0.0.0.0:8085", url);
```
briandealwis: Hmm, `http://0.0.0.0:8085` isn't a working URL (0.0.0.0 isn't a valid host)

chanseokoh:
All review comments address, including replacing the `Spinner` with a `Text`.

chanseokoh:
#259 (#259 is in a broader context than just a port.)

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,202 @@
...
+    assertEquals("1", portExtension.portText.getText());
+  }
+
+  @Test
+  public void testPortText_limitMaximumPortOnFocusLost() {
+    portExtension.portText.setText("123456");
+    assertEquals("123456", portExtension.portText.getText());
+
+    portExtension.portText.notifyListeners(SWT.FocusOut, null /* event */);
+    assertEquals("65535", portExtension.portText.getText());
```
elharo: I'm not sure this is the right behavior here. If the use sets 123456 it seems more likely they meant 12345 or 12456 or some such rather than 65535. It's a minor point but can we simply stop them from entering a value like this? Or somehow limit the field to a maximum of five digits?

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,202 @@
...
+  @Test
+  public void testPortText_allowEmpty() {
+    portExtension.portText.setText("");
+    assertTrue(portExtension.portText.getText().isEmpty());
+  }
+
+  @Test
+  public void testPortText_preventNonNumericCharacters() {
+    portExtension.portText.setText("80A80");
+    assertEquals("8080", portExtension.portText.getText());
```
elharo: I don't think we should just strip characters here. This calls for an error of some kind.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,202 @@
...
+    verify(server).setAttribute("appEngineDevServerPort", "54321");
+  }
+
+  @Test
+  public void testPortText_limitMinimumPortOnFocusLost() {
+    portExtension.portText.setText("0");
+    assertEquals("0", portExtension.portText.getText());
+
+    portExtension.portText.notifyListeners(SWT.FocusOut, null /* event */);
+    assertEquals("1", portExtension.portText.getText());
```
elharo: This one definitely needs to change because in TCP port 0 means "pick an available port", not use port 1 (which isn't necessarily available as non-root on Unix). This case we should just treat it the same as an empty port.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,102 @@
...
+    public void focusLost(FocusEvent event) {
+      try {
+        Integer port = Integer.valueOf(portText.getText());
+        port = Math.max(1, Math.min(port, 65535));
+        portText.setText(port.toString());
+      } catch (NumberFormatException ex) {
+        portText.setText(DEFAULT_SERVICE_PORT.toString());
+      }
+
+      serverWc.setAttribute(SERVER_ATTRIBUTE_PORT, portText.getText());
```
elharo: what is serverWc? can we use a non-abbreviated name?

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: I tried to see if there is a way to show an arbitrary error message and hook into the validation framework in this first wizard page. Unfortunately, there does not seem a way to do so from our extension point.

The current `Text` implementation in this PR is to blocking typing non-numeric characters, which is the same behavior of `Spinner`. I'm sure people have put much thought into the workings of `Spinner`, so I think there isn't really a problem in this behavior. Also, I get no unusual sensation when I actually used the `Spinner`. Nothing feels wrong.

Considering all these, especially that there is no way to hook into the wizard's validation framework, I think the current behavior is what we can do best.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: I first tried a very similar idea to block users from entering invalid values, such as values exceeding 65535, on the fly. I immediately got a sense that this UX was very wrong. For example, let's assume that a user first entered 10000 and later wanted to change it to 12000. The user could first change it to 120000 by inserting 2, and then delete one '0' to make it 12000. However, this cannot work if we prevent entering 120000.

So, I backed off and consulted the behavior of the `Spinner`. I learned that a user should be given complete freedom on what she can enter as a numeric value — whether she is typing, doing cut-and-paste, or selecting multiple characters. It gives smooth UX; it feels right. If we immediately blocks usual actions such as a key stroke, it doesn't feel right. Lastly, the `Spinner` forces a minimum and a maximum when losing a focus. I followed the same UX.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
chanseokoh: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/871#discussion_r84555552

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: Done. Now it automatically picks an available port when it is set to 0.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
elharo: We may need to talk about this then. These autofixes strike me as near-dangerous (and only not actively dangerous because it's just for a test server.) We should not change input to something a user neither entered nor wanted. Much better to put up an error message.

chanseokoh: Yeah, thinking again, I agree that these autofixes are dangerous.

chanseokoh:
We had a discussion about these autofixing. Although (I believe) we can't hook into the bulit-in error message and validation framework of the wizard, we can at least, e.g., insert our own label widget on top of the port text for displaying an error message (or a balloon message) explaining what is wrong.

Also, I'll look into again if there is a way to disable the finish button. Even if we can't, I think showing an error message basically addresses this issue. The worst thing with an invalid port value is a failure of starting a server.

briandealwis:
> Even if we can't, I think showing an error message basically addresses this issue.

I suggested to @chanseokoh that he could show a popup/hover to say _Ports must be 0-65535_.  JFace provides some easy support for that using `org.eclipse.jface.fieldassist.ControlDecoration` (`showHoverText()`).

chanseokoh:
The last commit added `ControlDecoration` to show info/error messages.

selection_014

selection_015

It doesn't seem there is a way to disable the finish button. So, users can start a server with a port greater than 65535, which will result in an immediate failure.

chanseokoh:
@elharo PTAL

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -163,11 +178,11 @@ void startDevServer(List<File> runnables, MessageConsoleStream console, String h
    *
    * @param runnables the path to directories that contain configuration files like appengine-web.xml
    * @param console the stream (Eclipse console) to send development server process output to
-   * @param host the host name to which application modules should bind
+   * @param port the app service port
```
briandealwis: Why are we determining the port in the launch configuration instead of here in the server implementation? Especially since we're using a server attribute to fetch the port?

(We should be able to start the server without the launch configuration.)

We should have some methods to return the actual ports used. That would simplify launch code where you're passing around the activePort.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -102,28 +103,32 @@ public void launch(ILaunchConfiguration configuration, String mode, final ILaunc
     console.clearConsole();
     console.activate();
 
-    new ServerLaunchMonitor(configuration, launch, server).engage();
+    int port = server.getAttribute(ServerPortExtension.SERVER_ATTRIBUTE_PORT,
```
briandealwis: `ServerPortExtension` is a UI class; `SERVER_ATTRIBUTE_PORT` on the other hand is a core concept and should be available elsewhere.  I think it belongs on `LocalAppEngineServerBehaviour`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -102,28 +103,32 @@ public void launch(ILaunchConfiguration configuration, String mode, final ILaunc
...
-      serverBehaviour.startDevServer(runnables, console.newMessageStream(), server.getHost());
+      serverBehaviour.startDevServer(runnables, console.newMessageStream(), port);
     }
   }
 
   /**
    * Listen for the server to enter STARTED and open a web browser on the server's main page
    */
-  protected void openBrowserPage(final ILaunchConfiguration configuration, final IServer server) {
+  protected void openBrowserPage(final IServer server, int actualPort) {
```
briandealwis: actualPort should be obtained from the server instance.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,128 @@
...
+import org.eclipse.swt.widgets.Text;
+import org.eclipse.wst.server.core.IServerType;
+import org.eclipse.wst.server.ui.wizard.ServerCreationWizardPageExtension;
+
+/**
+ * An extension that adds a server port field to the WTP server creation wizard page.
+ */
+public class ServerPortExtension extends ServerCreationWizardPageExtension {
+
+  public static final String SERVER_ATTRIBUTE_PORT = "appEngineDevServerPort"; //$NON-NLS-1$
```
briandealwis: I think these public statics belong on `LocalAppEngineServerBehaviour`.  After all, we should be able to access these values without the UI component even being loaded.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegateTest.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.mockito.Mockito.when;
+
+import org.eclipse.wst.server.core.IServer;
+import org.junit.Test;
+
+public class LocalAppEngineServerLaunchConfigurationDelegateTest {
+
+  @Test
+  public void testDeterminePageLocation() {
+    IServer server = mock(IServer.class);
```
akerekes: You could extract this to a field, annotate with `@Mock` and let the `MockitoJUnitRunner` initialize it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -131,9 +145,9 @@ public final void setModulePublishState2(IModule[] module, int state) {
...
```
akerekes: I think this is not the proper description for the port. Maybe 'server listener port'? @elharo

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+    portExtension.portText.setText("80A80");
+    assertEquals("8080", portExtension.portText.getText());
+
+    portExtension.portText.setText("80 clipboard pasted 80");
+    assertEquals("8080", portExtension.portText.getText());
+  }
+
+  @Test
+  public void testPortText_numericCharacters() {
+    portExtension.portText.setText("808");
```
akerekes: I can see little value in testing for these two specific values, not sure what this test demonstrates. To test that numerical value is accepted, one of them is enough. How long would a test run that would loop over 0..65535 so the test can verify that all possible port numbers are valid?

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+    assertEquals("808", portExtension.portText.getText());
+
+    portExtension.portText.setText("8108");
+    assertEquals("8108", portExtension.portText.getText());
+  }
+
+  @Test
+  public void testPortText_allowArbitrarilyLargeNumbers() {
+    portExtension.portText.setText("80808");
```
akerekes: Again I don't see the value in these specific values. How about Integer.MAX_VALUE, Long.MAX_VALUE, and a string of 20-30 numbers?
You could also add a test for negative numbers, and non-integer values.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+    portExtension.portText.setText("480808");
+    assertEquals("480808", portExtension.portText.getText());
+
+    portExtension.portText.setText("9480808");
+    assertEquals("9480808", portExtension.portText.getText());
+  }
+
+  @Test
+  public void testAttributeSetOnValueChange() {
+    portExtension.portText.setText("12345");
```
akerekes: I believe one (or at most two) case(s) should be sufficient to verify that the attribute is set.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+  @Test
+  public void testAttributeSetToZeroOnEmptyField() {
+    portExtension.portText.setText("");
+    assertTrue(portExtension.portText.getText().isEmpty());
+    verify(server).setAttribute("appEngineDevServerPort", 0);
+  }
+
+  @Test
+  public void testNoPopUpDecorationForValidPorts() {
+    portExtension.portText.setVisible(true);
```
akerekes: Again, I'd prefer either looping over 0..65535 or use 1 instead of 1024 and 8080 (to test for extremal values).

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+
+  @Test
+  public void testShowErrorDecorationOnInvalidPorts() {
+    portExtension.portText.setVisible(true);
+
+    portExtension.portText.setText("65536");
+    assertTrue(portExtension.portDecoration.isVisible());
+    assertEquals(portExtension.errorImage, portExtension.portDecoration.getImage());
+
+    portExtension.portText.setText("65535");
```
akerekes: If the point here to verify that the decorator disappears if the invalid value is changed to a valid, then I'd move this case into a separate test method.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+    portExtension.portText.setText("8080");
+    assertFalse(portExtension.portDecoration.isVisible());
+
+    portExtension.portText.setText("65535");
+    assertFalse(portExtension.portDecoration.isVisible());
+  }
+
+  @Test
+  public void testShowErrorDecorationOnInvalidPorts() {
+    portExtension.portText.setVisible(true);
```
akerekes: if the previous test would cover all valid ports, I think 65536 should be sufficient as a test case here.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+    assertEquals(portExtension.informationImage, portExtension.portDecoration.getImage());
+
+    portExtension.handlePropertyChanged(newNonAppEngineServerTypeEvent());
+    assertFalse(portExtension.portDecoration.isVisible());
+
+    portExtension.handlePropertyChanged(newAppEngineServerTypeEvent());
+    assertTrue(portExtension.portDecoration.isVisible());
+    assertEquals(portExtension.informationImage, portExtension.portDecoration.getImage());
+
+    portExtension.portText.setText("65536");
```
akerekes: This step is rather testing whether using an invalid value will trigger an error decorator, I think it'd be better in it's own test case (unless it's already covered).

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,128 @@
...
+  public static final int DEFAULT_SERVICE_PORT = 8080;
```
akerekes: DEFAULT_SERVER_PORT?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,128 @@
...
+      portText.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
+
+      portDecoration = new ControlDecoration(portText, SWT.LEFT | SWT.TOP);
+      portDecoration.hide();
+
+      FieldDecorationRegistry registry = FieldDecorationRegistry.getDefault();
+      informationImage =
+          registry.getFieldDecoration(FieldDecorationRegistry.DEC_INFORMATION).getImage();
+      errorImage = registry.getFieldDecoration(FieldDecorationRegistry.DEC_ERROR).getImage();
+    }
```
akerekes: What if `position` has a different value? Error/warning in log? If it's a normal case, then maybe a comment explaining it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
briandealwis: I don't think we need to test 0..65535: just the boundary cases are necessary.

elharo: Please do test something in the middle too. If not all 65,535. It's surprising how often the edges work and the middle fails.

Also, -1, 0,1, 1023, 1024, 1025, 65534, 65535, 65536 are edge cases for ports.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/messages.properties
```diff
@@ -0,0 +1,3 @@
+NEW_SERVER_DIALOG_EMPTY_PORT_FIELD=Leaving the port empty will pick a free port.
```
elharo: I'm not sure this is needed.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,128 @@
...
+      String newText = portText.getText().substring(0, event.start)
+          + event.text + portText.getText().substring(event.end);
+      event.doit = updatePortAndTriggerDecoration(newText);
+    }
+  };
+
+  private boolean updatePortAndTriggerDecoration(String newPortString) {
+    if (newPortString.isEmpty()) {
+      showPortDecoration(informationImage, Messages.NEW_SERVER_DIALOG_EMPTY_PORT_FIELD);
+      serverWc.setAttribute(SERVER_ATTRIBUTE_PORT, 0);
```
elharo: I still don't understand what serverWc is. Please unabbreviate.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -0,0 +1,284 @@
...
+    assertTrue(portExtension.portText.getText().isEmpty());
+  }
+
+  @Test
+  public void testPortText_preventNonNumericCharacters() {
+    portExtension.portText.setText("80A80");
+    assertEquals("8080", portExtension.portText.getText());
+
+    portExtension.portText.setText("80 clipboard pasted 80");
+    assertEquals("8080", portExtension.portText.getText());
```
elharo: I thought we were getting rid of this autofixing?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
chanseokoh: There is some complication involved here. If we determine the port in `Behavior`, there should be a way to pass the actual port used to `LaunchConfigurationDelegate` so that `openBrowserPage` can know which port to use. What could be the way? FYI, you can't get it from `IServer`, since there is no way you can update server attributes inside `Behavior`. (`IServer` is basically immutable.)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
chanseokoh: Please see https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/871#discussion_r84920391.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
briandealwis: Have the ports on the `LocalAppEngineServerBehaviour` instance.  It represents the actual running instance, right?  Certainly we're holding on to other state (the dev process, the `AppEngineDevServer`)

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegateTest.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: We removed auto-limiting the max and min port vlaues.

I think we still want to block entering non-numbers, since we cannot disable the finish button. It's not clear what the server should do if it's given a non-numeric string as a port. I think no one will complain if she cannot enter non-numeric character in the port field.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/messages.properties
chanseokoh: I'll remove this message.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
chanseokoh: @elharo About `serverWc`, please see https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/871#discussion_r84555552.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: Non-integer values are tested in another test. For negative numbers, users cannot even enter them (although I just found an edge case where users can enter a negative number.)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
chanseokoh: I removed the parameters, so no longer relevant.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
chanseokoh: Added a comment we add controls only at the `BOTTOM`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
elharo: 1024 is an extremal value here due to Unix's rules about root ports.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -27,36 +27,40 @@
...
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
 /**
  * A {@link ServerBehaviourDelegate} for App Engine Server executed via the Java App Management
  * Client Library.
  */
 public class LocalAppEngineServerBehaviour extends ServerBehaviourDelegate {
+
+  public static final String SERVER_ATTRIBUTE_PORT = "appEngineDevServerPort";
```
elharo: What is a "SERVER_ATTRIBUTE_PORT"? Maybe SERVER_PORT_ATTRIBUTE_NAME?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
```diff
@@ -0,0 +1,126 @@
...
+      String newText = portText.getText().substring(0, event.start)
+          + event.text + portText.getText().substring(event.end);
+      event.doit = updatePortAndTriggerDecoration(newText);
+    }
+  };
+
+  private boolean updatePortAndTriggerDecoration(String newPortString) {
+    if (newPortString.isEmpty()) {
+      showPortDecoration(informationImage, Messages.NEW_SERVER_DIALOG_EMPTY_PORT_FIELD);
+      serverWc.setAttribute(LocalAppEngineServerBehaviour.SERVER_ATTRIBUTE_PORT, 0);
```
elharo: Please, please change this variable name!

chanseokoh: Sorry, I think I failed to communicate clearly. This variable is in the super class, which is Eclipse code. @briandealwis agreed that there is nothing we can do about it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: I'll go with checking middle values and edges cases in the tests as @elharo suggested.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: Changed to test `Integer.MAX_VALUE`. (Soon to be pushed.) Values large than that will be blocked from entering for technical reasons (e.g., CloudSdk accepts `Integer` as a port), and I think it's acceptable.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: I think the logic to set the attribute based on the value in the text field isn't trivial, so it's good to test middle values and edge cases.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
chanseokoh: Done. The step is already covered in another test, so I removed it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtension.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -49,6 +49,7 @@
  */
 public class LocalAppEngineServerBehaviour extends ServerBehaviourDelegate {
   public static final String SERVER_ATTRIBUTE_PORT = "appEngineDevServerPort";
```
chanseokoh: Done.

chanseokoh:
I think I've address all the review comments. If you think I missed something, do let me know. I did some manual testing, and things are working properly.

PTAL.

elharo:
Got it on non-numeric characters. That sounds fine.

I'm still confused on what we're doing with out of bounds ports such as 100000. Where/when/how does this break?

chanseokoh:
> I'm still confused on what we're doing with out of bounds ports such as 100000. Where/when/how does this break?

Users will see an error dialog almost instantly after clicking the finish button. (The dialog is a general error WTP dialog which doesn't help at all.)  In more details, a server instance will be created, but the server will not start since it can't use an invalid port. Note that since we don't yet provide means to change ports of existing server instances, this instance cannot be re-used.

I wish we really had a way to disable the finish button. We can do it if we set the port in a new wizard screen, but unfortunately, there doesn't seem a way to disable it in the first wizard screen.

chanseokoh:
@briandealwis The error dialog we talked about is from `Sever.java`:3532:

```java
		
		if (!monitor.isCanceled() && getServerState() == IServer.STATE_STOPPED)
			return new Status(IStatus.ERROR, ServerPlugin.PLUGIN_ID, 0, NLS.bind(Messages.errorStartFailed, getName()), null);
```

Can there be a way to pass up our own error message to the UI when a sever fails to start?

patflynn:
I haven't read this whole thread but FYI the dev appserver defaults to 8080
when no port is specified.  Admin port defaults to 8000, and API port
defaults to '0' according to the help docs.  I'm gonna guess that 0 turns
into random available.

chanseokoh:
In any case, as soon as a user enters a port greater than 65535, the user will see an error message in the wizard screen, so even if the WTP error dialog doesn't tell why the server failed, I think it's okay.

chanseokoh:
Okay, problem solved. We could throw `CoreException` which we can use to change the message in the WTP dialog. Will update the PR soon.

PR #875, opened 2016-10-23
add properties for hello world

elharo:
@lesv @akerekes This is a quick and dirty fix so the plugin (and integration tests) continue to work with the latest version of the Hello world archetype. Suggestions for improvement appreciated, though we may need to get this in sooner rather than later, since otherwise I think our integration tests are about to break. We'll definitely need to do something better when we add library configuration to the maven project wizard.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateMavenBasedAppEngineStandardProject.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateMavenBasedAppEngineStandardProject.java
@@ -70,6 +70,11 @@ public class CreateMavenBasedAppEngineStandardProject extends WorkspaceModifyOpe
     properties.put("appengine-version", appengineArtifactVersion);
     properties.put("gcloud-version", gcloudArtifactVersion);
     properties.put("application-id", appId);
+    properties.put("useJstl", "true");
+    properties.put("useObjectify", "false");
+    properties.put("useEndpoints1", "false");
+    properties.put("useEndpoints2", "false");
+    properties.put("useAppEngineApi", "false");
 
     ProjectImportConfiguration importConfiguration = new ProjectImportConfiguration();
     String packageName = this.packageName == null || this.packageName.isEmpty() 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateMavenBasedAppEngineStandardProject.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/CreateMavenBasedAppEngineStandardProject.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Google Inc. All Rights Reserved.
+ * Copyright 2016 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


chanseokoh:
The sooner we merge this, the better. Our plugin is currently broken.

briandealwis:
It seems we'll break with any new additional option.  I wonder if `appengine-standard-archetype` should instead use a single property (`options`?) with comma-separated list of optional dependencies.

lesv:
LGTM from me - though I suspect these aren't required as they have defaults.

elharo:
I've merged this to hopefully unbreak us.

@lesv I thought these wouldn't be needed either, but that's not what my testing found. Might perhaps be a maven bug? :-(

@briandealwis An options property does sound like a good idea, though I'm not immediately sure how to implement the splitting of the options inside the archetype pom.xml. I suppose Velocity could do this somehow. Talk to @lesv and if he agrees with your approach feel free to take a crack at it inside the archetypes project.

Clearly we will need to refactor/replace the code checked in here as we add support for specifying options in the UI. I think @nbashirbello is working on that? This is just a quick fix to unblock us.

lesv:
@briandealwis I like the idea in principle, though the manual UX becomes a bit more complex.  (It would be nice to repurpose this Archetype for pure maven users as well.

Based on Elliott's comment, I'm going to retest things manually as it may just be the automated tests that work.  (Which will make me grumpier)

I don't know how you'll do a split in http://velocity.apache.org/ but you are welcome to look.

lesv:
Well, the SO link pointed to the exact thing.  Cool.  Let me test and I'll comment to the thread.

briandealwis:
[DefaultFilesetArchetypeGenerator#isArchetypeConfigured()](https://github.com/apache/maven-archetype/blob/master/archetype-common/src/main/java/org/apache/maven/archetype/generator/DefaultFilesetArchetypeGenerator.java#L454) doesn't check if the properties have a default value.

It looks like we could obtain the properties from the archetype artifact and thus pre-populate the properties.  Something like the following:
```
ArchetypeManager am = MavenPluginActivator.getDefault().getArchetypeManager();
ArchetypeArtifactManager amm = manager.getArchetypeArtifactManager();
File archetypeFile = amm.getArchetypeFile(groupId, artifactId, version, …);
ArchetypeDescriptor descriptor = null;
if(amm.isFileSetArchetypeDescriptor(file)) {
   descriptor = amm.getFileSetArchetypeDescriptor(file);
} else if(amm.isOldArchetypeDescriptor(file)) {
   descriptor = amm.getOldArchetypeDescriptor(file);
} else { throw something; }

Properties properties = new Properties();
for(RequiredProperty required : descriptor.getRequiredProperties()) {
   if(required.getDefaultValue() != null) {
     properties.put(required.getKey(), required.getDefaultValue());
   }
}
return properties;
```

lesv:
So, I ran through the process manually and things worked, so I'm inclined to recommend keeping things the way they are and getting you to figure out what your doing wrong to Maven.  I managed to use Defaults.

```
Choose a number or apply filter (format: [groupId:]artifactId, case sensitive contains): : 3
Choose com.google.appengine.archetypes:appengine-standard-archetype version:
1: 0.1.0
2: 0.1.1
3: 0.2.0
4: 0.3.0
Choose a number: 4: 4
Define value for property 'groupId': : mygroup
Define value for property 'artifactId': : myArtifact
Define value for property 'version':  1.0-SNAPSHOT: :
Define value for property 'package':  mygroup: : org.example.appengine
[INFO] Using property: appengine-version = 1.9.44
[INFO] Using property: useAppEngineApi = false
[INFO] Using property: useEndpoints1 = false
[INFO] Using property: useEndpoints2 = false
[INFO] Using property: useJstl = true
[INFO] Using property: useObjectify = false
Confirm properties configuration:
groupId: mygroup
artifactId: myArtifact
version: 1.0-SNAPSHOT
package: org.example.appengine
appengine-version: 1.9.44
useAppEngineApi: false
useEndpoints1: false
useEndpoints2: false
useJstl: true
useObjectify: false
```

PR #877, opened 2016-10-24
move every plugin to appengine-plugins-core 0.1.13

elharo:
@briandealwis


elharo:
@briandealwis is right of course. 0.1.3 --> 0.1.13

Issue #783, opened 2016-10-07
Changing GAE Facets in the Property Dialog is not updating the deployment settings page


nbashirbello:
What happens when the user unselects the facet and goes back to the GAE deployment page?

elharo:
Can you elaborate on this? I'm not totally clear on the issue. Thanks.

nbashirbello:
If you open the Properties dialog of an App Engine Standard project it will contain both the App Engine Deployment page and the Project Facets page. If you uninstall the App Engine Standard facet or change it to the App Engine Flex facet, the App Engine Deployment tab will not be updated to either hide the App Engine Deployment for Standard (as in the first case) or show the App Engine Deployment for Flex.

akerekes:
Fixed by #1567

PR #880, opened 2016-10-25
Update deployment properties page on facet change

nbashirbello:
Show an error message on the deployment properties page when it shows standard deployment information but the standard facet has been uninstalled. Same for flex. 
Fix for #783


nbashirbello:
@briandealwis

chanseokoh:
I tried to do a manual testing, and looks like it breaks validation in the preferences page. For example, if you check `Override generated version`, it doesn't show the error message on top. When a user tried to move to another preference page, it's blocked with the following error:

selection_002

nbashirbello:
@elharo 

When you saw the "Could not accept changes" message, had you set a value for "Bucket". It looks like if "Override default Cloud Storage bucket for staging" is checked, you have to have an input for "Bucket". (The "Apply" button also gets disabled when this happens).

elharo:
I think @chanseokoh saw that message, not me.

chanseokoh:
Explained the error to @nbashirbello offline.

nbashirbello:
Not ready for review!

nbashirbello:
PR ready for review :)

PR #881, opened 2016-10-25
Relative path for lib jars

akerekes:
Uses M2Eclipse's local repository as base directory for location of library jar files, persists relative paths based on that.
Serialization of libraries is delegated more along the Serializable* classes.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -133,4 +143,15 @@ public IPath getContainerStateFile(IJavaProject javaProject, IPath containerPath
       return containerFile.getLocation();
     }
   }
+
+  private static class M2LocalRepositoryLocationProvider implements ArtifactBaseLocationProvider {
+
+    /* (non-Javadoc)
```
elharo: why non-Javadoc? MIght as well make it Javadoc is only to avoid the extra comment :-)

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/SerializableClasspathEntry.java
```diff
@@ -33,53 +33,79 @@
...
-                                    getAccessRules(accessRules),
-                                    getAttributes(attributes),
+                                    getAccessRules(),
+                                    getAttributes(),
                                     true);
   }
 
-  private IClasspathAttribute[] getAttributes(SerializableAttribute[] attributes) {
+  // if path is relative, it's appended to baseDirectory, otherwise unchanged
+  private static IPath makePathAbsolute(String path, IPath baseDirectory) {
```
elharo: These next two methods might benefit from @VisibleForTesting and direct tests.

briandealwis:
I'm fine with this.  I was expecting to see the use of a variable classpath entry (`JavaCore.newVariableEntry`) and you could serialize it out as `$M2_REPO/xxx`.  It also prettifies the presentation within the Explorers, since the library isn't shown with an absolute path name.  But I can't think of any issues with your current approach.

akerekes:
@briandealwis it basically uses the value of M2_REPO by obtaining the local repository from M2Eclipse, however it does not exposes it by using a classpath variable (thus using longer, less readable paths). Should we change where do we store the jars, we can do so without changing the persisted paths. Probably the best would be to use our own classpath variable that defaults to M2_REPO.

Issue #803, opened 2016-10-12
Implement API attacher for new Maven projects


akerekes:
When a new Maven GCP project is generated using our wizards, users should be able to attach libraries similar to how they are/will be able to do it when generating a new GCP native Eclipse project via #732 

This is part of #682

elharo:
I think this is likely to require some support on the archetypes. I.e. we'll want conditional sections that add the relevant dependencies iff certain variable are set.

akerekes:
This is a good idea: extend the library attachment to the archetypes thus it would work outside of Eclipse as well. Only reason I see why we could implement it by adding the libraries "manually" when creating the new project is to have a solution faster in.

elharo:
We'll still need a section in the dialog for the user to check off libraries, like the one you have for non-maven projects. I think each checkbox will set/unset a variable that determines whether a particular dependency is included in the pom.xml.

I suspect we'll want to extract the library panel you wrote for AppEngineStandardWizardPage into a shared class that can be used by both maven and non-maven projects.

elharo:
m2e completely handles adding dependencies to a pom.xml and downloading the jars and installing them in the build path  after project creation. All we need to do is offer the checkboxes when the project is created.

akerekes:
The reason for a dedicated dialog/action to add libraries to existing maven
 projects is to make it easier for the user to find/discover GCP libraries.

briandealwis:
I have a proof-of-concept working using the m2e `PomEdits` class.  Now needs to integrate with the `.libraries` support.

rdayal:
status?

elharo:
@nbashirbello is working on extracting the libraries panel into a shareable plugin we can use for both standard and maven projects. (Currently it's only in standard projects.)

At least for beta I don't want to add a dedicated dialog for Maven projects. There isn't likely time, and the utility seems less than other features we can spend our time on.

rdayal:
Ok, then I'm confused about what's being implemented here. 

I thought we agreed that as a blocker for beta, we'd ensure that users had a way to attach the relevant app engine dependencies to their project when they create it. Is that what this issue is tracking? Or is it tracking something else?

elharo:
Yes, we need to add the same panel for selecting libraries that we have not in standard to the maven wizard as well.

rdayal:
Ok - so are you suggesting we drop this particular issue as a beta blocker? Or are you suggesting something else?

[...]

PR #882, opened 2016-10-25
Pull out the APIs Library panel from the the standard wizard into a separate plugin

nbashirbello:
so that it can be shared by both native and maven wizards.
Part of the fix for #803


nbashirbello:
@elharo @akerekes

akerekes:
LGTM

nbashirbello:
@akerekes I need the approval thingy :)

PR #885, opened 2016-10-26
less public dependency injection

elharo:
slight refactoring suggested by an attendee at EclipseCon


briandealwis:
Seems like decent change?

akerekes:
The approach looks good, but in this case the `ServletClasspathProvider` has changed and do not depend on CloudSdk anymore.

elharo:
It was made irrelevant by a recent commit from Andras.

--
Elliotte Rusty Harold

PR #888, opened 2016-10-26
Don't let user choose a directory for project they can't write to

### LINKED ISSUE #834: Maven Wizard doesn't check if directory is writable
When creating a project in a user specified directory outside the workspace, we don't check that the directory exists. It doesn't fail until a couple of screens later when we try to create the project:

Unable to create project from archetype [com.google.appengine.archetypes:appengine-standard-archetype:LATEST]
org.apache.maven.archetype.exception.ArchetypeGenerationFailure: Error merging velocity templates: /usr/tmp/foo/pom.xml (No such file or directory)

elharo:
fix #834


elharo:
@briandealwis

briandealwis:
Hmm, most of the Eclipse wizards will create the path (e.g., the _Google App Engine Standard Java Project_ will create the directories in the location).  I'll check what the m2e wizards do.

briandealwis:
Verified that our wizard also creates the directories.  I think we should close this.

elharo:
It creates the directories in the workspace. Does it do so anywhere else? Maybe it's a permissions issue, or a chain of directories it won't create; but I can definitely get this to fail as described in the bug.

briandealwis:
I was verifying under `c:\temp` for my tests (`c:\temp\foo\bar\baz`).  The directories were created, and the project structure looked fine.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -207,17 +208,24 @@ protected void checkFlipToNext() {
...
     // order here should match order of the UI fields
 
     String location = locationField.getText().trim();
-    if (!useDefaults() && location.isEmpty()) {
-      setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
-      return false;
-    }
+    if (!useDefaults()) {
+      if (location.isEmpty()) {
+	    setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
```
akerekes: Could be this a warning?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -207,17 +208,24 @@ protected void checkFlipToNext() {
...
+	    return false;
+      } else if (!Paths.get(location).toFile().exists()) {
+  	    String message = MessageFormat.format(
+  	    		Messages.getString("MISSING_LOCATION"), location); //$NON-NLS-1$
+		setMessage(message, INFORMATION);
```
akerekes: Could be this a warning/error?

chanseokoh:
Will you do the same thing in the native wizard so that the behavior is perfectly same? (I know the native wizard has a single screen, so it's not that critical to do so there.)

elharo:
PTAL @briandealwis 

Not a full solution yet, but it does catch some common cases. Not sure if we want to check this in as an intermediate step or continue on to a full solution that walks up the path until it finds a directory it either can or cannot write to.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/messages.properties
```diff
@@ -9,6 +9,8 @@ LOCATION_GROUP_TEXT=Location
 JAVA_PACKAGE_LABEL=Java package:
 GENERATED_PROJECT_LOCATION=Select the directory to contain the generated project
 PROVIDE_LOCATION=Provide a location
+MISSING_LOCATION=Directory {0} does not exist.
```
chanseokoh: This isn't being used.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
+	    setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
+	    return false;
+      } else {
+          java.nio.file.Path path = Paths.get(location);
+    	  if (Files.exists(path) && !Files.isDirectory(path)) {
+            String message = MessageFormat.format(Messages.getString("FILE_LOCATION"), location); //$NON-NLS-1$
+    	  	setMessage(message, INFORMATION);
+            return false;    		  
+          } else if (Files.exists(path) && !Files.isWritable(path)) {
+            String message = MessageFormat.format(Messages.getString("NONWRITABLE"), location); //$NON-NLS-1$
```
chanseokoh: `NONWRITABLE` isn't defined.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
+      	  	setMessage(message, INFORMATION);
+            return false;
```
chanseokoh: Indentation is off across the entire block here. There could be tab characters too.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
+    	  if (Files.exists(path) && !Files.isDirectory(path)) {
+            String message = MessageFormat.format(Messages.getString("FILE_LOCATION"), location); //$NON-NLS-1$
+    	  	setMessage(message, INFORMATION);
+            return false;    		  
+          } else if (Files.exists(path) && !Files.isWritable(path)) {
+            String message = MessageFormat.format(Messages.getString("NONWRITABLE"), location); //$NON-NLS-1$
+      	  	setMessage(message, INFORMATION);
+            return false;         	  
+          }
+    	  // TODO how to check if a directory that doesn't exist could be created?
```
chanseokoh: Maybe checking if the parent directory `isWritable()`?

briandealwis: @chanseokoh's suggestion is what I was going to suggest.

elharo: It's a little more complex than that. We have to walk up the tree to find out whether the first directory that exists is writable or not.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
+    if (!useDefaults()) {
+      String location = locationField.getText().trim();
+      if (location.isEmpty()) {
+        setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
+        return false;
+      } else {
+        java.nio.file.Path path = Paths.get(location);
+        if (Files.exists(path) && !Files.isDirectory(path)) {
+          String message = MessageFormat.format(Messages.getString("FILE_LOCATION"), location); //$NON-NLS-1$
+              setMessage(message, INFORMATION);
```
chanseokoh: Indentation is off for this one.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
+          return false;              
+        } else if (Files.exists(path) && !Files.isWritable(path)) {
+          String message = MessageFormat.format(Messages.getString("NONWRITABLE"), location); //$NON-NLS-1$
+              setMessage(message, INFORMATION);
```
chanseokoh: Indentation is off for this one too.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
-      return false;
-    }
+    if (!useDefaults()) {
+      String location = locationField.getText().trim();
+      if (location.isEmpty()) {
+        setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
+        return false;
+      } else {
+        java.nio.file.Path path = Paths.get(location);
+        if (Files.exists(path) && !Files.isDirectory(path)) {
```
briandealwis: I think you could quickly turn this into the directory-doesn't-exist case with:
```
// must eventually get to the root, which exists
while(path != null && !Files.exists(path)) {
   path = path.getParent();
}
if (path == null) {
  return false;
} 
// remaining ifs …
```

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -226,17 +228,32 @@ protected void checkFlipToNext() {
...
-      setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
-      return false;
-    }
+    if (!useDefaults()) {
+      String location = locationField.getText().trim();
+      if (location.isEmpty()) {
+        setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
+        return false;
+      } else {
+        java.nio.file.Path path = Paths.get(location);
```
briandealwis: You should handle the `InvalidPathException` that's possible from `Paths#get()`

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -268,6 +253,38 @@ private boolean validatePage() {
...
+            } else if (!Files.isWritable(path)) {
+              String message = MessageFormat.format(Messages.getString("NONWRITABLE"), location); //$NON-NLS-1$
+              setMessage(message, WARNING);
+              return false;
+            } else { // writable directory
+              return true;
+            }
+          }
+          path = path.getParent();
+        } while (path != null);
```
briandealwis: If `path == null` then this will return true.  I'm not exactly sure what the circumstances are where this could happen.

elharo: I don't think that can happen because:

1. Paths.get doesn't return null
2. If Paths.get does return null, a NullPointerException is thrown earlier

briandealwis:
Path.getParent() is documented as being able to return null.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
chanseokoh: From my testing, providing a relative path such as `abc` or `a/b/c` can eventually make `path == null`, escaping the loop.

I think we should block entering a relative path. It creates directories relative to the work directory of Eclipse. I filed an issue for this: #929

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -251,6 +253,38 @@ public boolean validatePage() {
...
+    if (location.isEmpty()) {
+      setMessage(Messages.getString("PROVIDE_LOCATION"), INFORMATION); //$NON-NLS-1$
+      return false;
+    } else {
+      try {
+        java.nio.file.Path path = Paths.get(location);
+        do {
+          if (Files.exists(path)) {
+            if (!Files.isDirectory(path)) {
+              String message = MessageFormat.format(Messages.getString("FILE_LOCATION"), path.toString()); //$NON-NLS-1$
```
chanseokoh: Nit: `location` instead of `path.toString()` to be consistent with others?

elharo: No, that;'s quite deliberate. In this example it it path (which can be a parent of location) that is problematic. E.g. in /foo/bar/baz we don't want to say that /foo/bar/baz is a file if in fact it's /foo/bar that's a file and baz does not even exist.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
elharo: abc and a/b/c should not be able to be entered in the location text field. Are you seeing something different?

However if path.getParent() == null we want to escape the loop That is WAI.

akerekes: Would it make sense to extract this logic into a separate class and cover it with junit tests?

briandealwis: Seems NTFS supports [separate permissions](https://msdn.microsoft.com/en-us/library/bb727008.aspx) for _Create Files_ vs _Create Folders_.

PR #893, opened 2016-10-26
Fixes #855 - AppEngineLibraryContainerInitializer should re-resolve artifacts if saved paths are invalid

### LINKED ISSUE #855: AppEngineLibraryContainerInitializer should re-resolve artifacts if saved paths are invalid
When `AppEngineLibraryContainerInitializer` loads the container state from file, the path stored for the given jar files might be invalid. In this case it should request the ILibraryRepositoryService to resolve the jars again to obtain a valid local path.

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/AppEngineLibraryContainerInitializer.java
```diff
@@ -100,6 +117,16 @@ public void initialize(IPath containerPath, IJavaProject project) throws CoreExc
     }
   }
 
+  private void validateJarPaths(LibraryClasspathContainer container) throws LibraryRepositoryServiceException {
+    IClasspathEntry[] classpathEntries = container.getClasspathEntries();
+    for (int i = 0; i < classpathEntries.length; i++) {
+      IClasspathEntry classpathEntry = classpathEntries[i];
+      if (!classpathEntry.getPath().toFile().exists()) {
+        classpathEntries[i] = repositoryService.getLibraryClasspathEntry(classpathEntry);
```
briandealwis: `getLibraryClasspathEntry` seems a bit odd.  `refresh` or `rebuild` would seem to make more sense?

It seems a bit odd to have so much knowledge in this class.  Would it make sense to push this method into `LibraryClasspathContainer` itself:
```
if(container.revalidate(repositoryService)) {
   JavaCore.setClasspathContainer(…);
```

PR #896, opened 2016-10-26
update package when group ID is deleted

### LINKED ISSUE #889: Package suggestion bug in maven project wizard
Steps to reproduce:

1. Type 'a' in the Group ID.
2. Observe that the Java package field has the same value: 'a'.
3. Delete 'a' with a backspace in the Group ID.

Observe that 'a' still remains in the Java package field.

elharo:
@chanseokoh fix #889

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
@@ -367,7 +367,7 @@ public class MavenAppEngineStandardWizardPage extends WizardPage {
         String suggestion = suggestPackageName(newGroupId);
         javaPackageField.setText(suggestion);
       } else if (newGroupId.trim().isEmpty()) {
-    	javaPackageField.setText("");
+        javaPackageField.setText("");
       }
     }
 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
@@ -412,6 +412,14 @@ public class MavenAppEngineStandardWizardPage extends WizardPage {
 	  if (event.character == '\t' || event.character == '\r' || event.character == '\n') {
 		  return;
 	  }
+	  if (event.keyCode == SWT.ARROW_UP || event.keyCode == SWT.ARROW_DOWN
+	      || event.keyCode == SWT.ARROW_LEFT || event.keyCode == SWT.ARROW_RIGHT) {
+        return;
+	  }
+	  if (event.keyCode == SWT.ESC) {
+	    return;
+	  }
+	  
 	  userChosePackageName = true;
 	}
   }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
@@ -348,13 +348,10 @@ public class MavenAppEngineStandardWizardPage extends WizardPage {
   }
 
   /**
-   * Auto-fills {@link #javaPackageField} as Group ID when
-   * 1) {@link #javaPackageField} is empty; or
-   * 2) the field matches previous auto-fill before ID modification.
+   * Auto-fills {@link #javaPackageField} as Group ID if the user has not explicitly
+   * chosen a package name.
    */
   private final class AutoPackageNameSetterOnGroupIdChange implements VerifyListener {
-
-    private String previousSuggestion = "";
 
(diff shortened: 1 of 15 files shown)


elharo:
FYI, this one picked up some extra junk causing the tests to fail since I neglected to branch of master. That should take care of itself once the previous PR is merged in.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -364,14 +364,16 @@ public void verifyText(VerifyEvent event) {
...
+      } else if (newGroupId.trim().isEmpty()) {
+        updatePackageField("");
       }
     }
 
     private void updatePackageField(String newSuggestion) {
       if (getPackageName().isEmpty() || getPackageName().equals(previousSuggestion)) {
         javaPackageField.setText(newSuggestion);
       }
+      previousSuggestion = newSuggestion;
```
briandealwis: This will override the package field when explicitly set. Steps to repeat:
  1. Open the new maven wizard
  2. Put "foo" in the Java package field
  3. Start writing "footer" in the group field.  See the Java package field change.

chanseokoh: Just saw this. This was an intended behavior. If the suggested package name based on the group ID becomes to match what is entered in the package field, we go back into the "tracking" mode, in which we continue to auto-fill the package field based on group ID changes. If at any point a user manually modifies the package field and the value becomes different from suggested package based on the group ID, we go into the "detached" mode, and we stop updating the package field. Then, if a user changes her mind to modify the package to match the group ID, then we go back to the "tracking" mode.

So, this is the current behavior. What else could you suggest for updating the package?

elharo:
Good catch. I'd prefer that if the user ever types anything into the package field we don't autoupdate it, no matter what is typed into the group ID field. Maybe we could even detach the listener in this case. @chanseokoh what do you think?

chanseokoh:
@elharo That's also a good option, which will also simplify implementation significantly.

elharo:
PTAL. This now also changes the auto-package setting so it no longer happens after the user types a package name.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -362,17 +364,13 @@ public void verifyText(VerifyEvent event) {
 
       // getGroupId() trims whitespace, so we do the same to sync with the dialog validation error.
       if (MavenCoordinatesValidator.validateGroupId(newGroupId.trim())) {
-        String newSuggestion = suggestPackageName(newGroupId);
-        updatePackageField(newSuggestion);
-        previousSuggestion = newSuggestion;
+        String suggestion = suggestPackageName(newGroupId);
+        javaPackageField.setText(suggestion);
+      } else if (newGroupId.trim().isEmpty()) {
+    	javaPackageField.setText("");
```
chanseokoh: Indentation is slightly off. (There is a tab character).

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/UserChose.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public void keyReleased(KeyEvent event) {
+    // filter out navigation keys
+    if (event.character == '\t' || event.character == '\r' || event.character == '\n') {
+	  return;
+    }
+    if (event.keyCode == SWT.ESC || event.keyCode == SWT.ARROW_UP 
+        || event.keyCode == SWT.ARROW_DOWN || event.keyCode == SWT.ARROW_LEFT 
+        || event.keyCode == SWT.ARROW_RIGHT) {
```
chanseokoh: We should add `SWT.HOME` and `SWT.END`. But there are other problems, unfortunately. For example, I tested that if you press Ctrl+Left arrow (which is same as HOME), `userChosePackageName` is set to true. There may be other unknowns too.

I can think of using a `ModifyListener` (with some care) instead of `KeyListener`. For example,

```java
    manualChangeMonitor = new ModifyListener() {
      @Override
      public void modifyText(ModifyEvent event) {
        groupIdField.removeVerifyListener(autoPackageNameSetter);
      }
    };
    javaPackageField.addModifyListener(manualChangeMonitor);
```

and you detach and re-attach the listener in the following way:

```java
  private final class AutoPackageNameSetterOnGroupIdChange implements VerifyListener {
    @Override
    public void verifyText(VerifyEvent event) {
      javaPackageField.removeModifyListener(manualChangeMonitor);
      ... method body ...
      javaPackageField.addModifyListener(manualChangeMonitor);
    }
```

This seems to work.

elharo: I tried ModifyListener first. The problem was that ModifyLIstener fired on a call to setText, not just actual user typing.

chanseokoh: Yeah, so what I suggested was to detach it before setText() (at the beginning of verifyText()) and re-attach before exiting.

However, I'm not against using the KeyListener. It's more or less working.

chanseokoh:
#889 fixed by #907. Closing.

PR #899, opened 2016-10-26
Build with Oxygen

briandealwis:
Revises the version ranges on our `com.google.cloud.tools.eclipse.jdt.launching` bundle to allow installing in anything from Mars on.  Change our launch configuration to first try using the Eclipse-provided _Socket Listener_ VM connector, providing it supports the _connectionLimit_ argument that was accepted for 4.7M3.  Otherwise uses our _connectionLimit_-supporting VM connector.

We could put the version ranges back into `com.google.cloud.tools.eclipse.jdt.launching`, but we'd need to ship an additional feature.  But as our VM connector won't interfere with future versions, we may as well keep it around until we move our support level to Oxygen.

### DIFF
--- a/eclipse/oxygen/gcp-eclipse-oxygen.tpd
+++ b/eclipse/oxygen/gcp-eclipse-oxygen.tpd
@@ -1,0 +1,47 @@
+/* 
+ * Target Platform Definition created using Mikael Barbero's TPD editor 
+ * <https://github.com/mbarbero/fr.obeo.releng.targetplatform/>
+ * 
+ * If you make changes to this file, either:
+ * 
+ *    * Right-click in the editor and choose 'Create Target Definition File'
+ *      to update the corresponding .target file.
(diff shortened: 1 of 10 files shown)


briandealwis:
As we don't have a way to implicitly pull in bundles/fragments for specific versions of the platform, this is good enough.

eclipse/oxygen/gcp-eclipse-oxygen.tpd
```diff
@@ -0,0 +1,47 @@
...
+	org.eclipse.datatools.sdk.feature.feature.group
+	org.eclipse.swtbot.eclipse.feature.group
+	
+	org.eclipse.jetty.http
+	org.eclipse.jetty.servlet
+	org.eclipse.jetty.server
+	org.eclipse.jetty.util
+}
+
+// /webtools/repository/oxygen currently has some issues
```
akerekes: I think we could revert the location to the original, and the same for neon TP

Issue #906, opened 2016-10-27
Use ModifyListener instead of VerifyListener for AutoPackageNameSetterOnGroupIdChange


chanseokoh:
This is about `AutoPackageNameSetterOnGroupIdChange` in the class `MavenAppEngineStandardWizardPage`.

We no longer need to get a previous group ID before text change, so we don't need `VerifyListener`. `ModifyListener` will make the code simpler.

chanseokoh:
Closed by #907.

PR #907, opened 2016-10-27
Alternative fix for #889

### LINKED ISSUE #889: Package suggestion bug in maven project wizard

briandealwis:
Alternative fix outlined in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/896#pullrequestreview-6093155

Fixes #889

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
@@ -17,7 +17,9 @@
 package com.google.cloud.tools.eclipse.appengine.newproject.maven;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 
 import org.eclipse.core.databinding.observable.Realm;
@@ -142,6 +144,37 @@ public class MavenArchetypeProjectWizardTest {
         assertEquals("ac", page.javaPackageField.getText());
       }
     });
-
   }
+
+  @Test
+  public void testAutoPackageNameSetterOnGroupIdChange_disbledOnUserChange() {
+    Realm.runWithDefault(DisplayRealm.getRealm(Display.getDefault()), new Runnable() {
+      @Override
+      public void run() {
+        wizard.setContainer(mock(IWizardContainer.class));
+        wizard.createPageControls(shell);
+        MavenAppEngineStandardWizardPage page =
+            (MavenAppEngineStandardWizardPage) wizard.getPage("basicNewProjectPage");
+        assertTrue(page.autoGeneratePackageName);
+
+        page.groupIdField.setText("abc");
+        assertEquals("abc", page.javaPackageField.getText());
+        assertTrue(page.autoGeneratePackageName);
+
+        page.javaPackageField.setText("def");
+        assertFalse(page.autoGeneratePackageName);
+
+        // javaPackageField should no longer auto-gen
+        page.groupIdField.setText("xyz");
+        assertEquals("def", page.javaPackageField.getText());
+
+        // we shouldn't auto-gen even if the user clears the contents
+        page.javaPackageField.setText("");
+        assertFalse(page.autoGeneratePackageName);
+        page.groupIdField.setText("abc");
+        assertEquals("", page.javaPackageField.getText());
+      }
+    });
+  }
+
 }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
@@ -70,6 +70,8 @@ public class MavenAppEngineStandardWizardPage extends WizardPage {
 
   private boolean shouldUpdatePackageName = true;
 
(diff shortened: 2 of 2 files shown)


chanseokoh:
Doesn't work on my machine. :(

briandealwis:
Grrr.  SWT-GTK's `verifyText()` attempts to retrieve the current event and if it's a key-pressed event, it sets the key details, so we can't implicitly differentiate between programmatic updates and user-triggered updates.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -350,8 +366,6 @@ public void modifyText(ModifyEvent event) {
    */
   private final class AutoPackageNameSetterOnGroupIdChange implements VerifyListener {
```
chanseokoh: If you want, you can change it to `ModifyListener` and close #906. That way, you don't have to do the complex computation to get `newGroupId`.

briandealwis:
Fixes #906 
PTAL @chanseokoh

chanseokoh:
Generally looks good to me. What about https://github.com/GoogleCloudPlatform/google-cloud-eclipse/commit/39a186e386eb48ee7271fee3577bb93dee8cc7de#commitcomment-19597981?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -184,6 +186,20 @@ private void createAppEngineProjectDetailsArea(Composite container) {
...
+      @Override
+      public void verifyText(VerifyEvent event) {
+        // check if set programmatically
+        if (javaPackageProgrammaticUpdate) {
+          return;
+        }
+        String current = ((Text) event.widget).getText();
+        String newText =
+            current.substring(0, event.start) + event.text + current.substring(event.end);
+        shouldUpdatePackageName = newText.isEmpty();
```
elharo: Why check if it's empty? If the user explicitly set the package name, including setting it to empty by deleting the auto-generated value, we should not overwrite that.

briandealwis: I missed your [previous comment](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/896#issuecomment-256464103):
> I'd prefer that if the user ever types anything into the package field we don't autoupdate it, no matter what is typed into the group ID field.

Easily fixed.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -348,30 +364,27 @@ public void modifyText(ModifyEvent event) {
    * 1) {@link #javaPackageField} is empty; or
    * 2) the field matches previous auto-fill before ID modification.
```
chanseokoh: Need to change the comment. See @elharo's PR for that.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java
```diff
@@ -142,6 +144,37 @@ public void run() {
         assertEquals("ac", page.javaPackageField.getText());
       }
     });
+  }
 
+  @Test
+  public void testAutoPackageNameSetterOnGroupIdChange_disbledOnUserChange() {
+    Realm.runWithDefault(DisplayRealm.getRealm(Display.getDefault()), new Runnable() {
```
chanseokoh: Interesting. Perhaps this will make the test work even if `<useUIThread>false<useUIThread>`?

chanseokoh: Seems not. Mind explaining when we need to do `Realm.runWithDefault`, and why and by what in our test below we need it?

chanseokoh:
@elharo Want to take a look? It's blocked by your change request.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -344,34 +359,30 @@ public void modifyText(ModifyEvent event) {
...
       }
     }
+  }
 
-    private void updatePackageField(String newSuggestion) {
-      if (getPackageName().isEmpty() || getPackageName().equals(previousSuggestion)) {
-        javaPackageField.setText(newSuggestion);
-      }
+  private void updatePackageField(String newSuggestion) {
+    if (autoGeneratePackageName) {
```
akerekes: Why don't we check this even before we generate the suggestion in `modifyText()`?

briandealwis: In case there are other paths that call in here?

PR #909, opened 2016-10-28
Fixes #903 - Objectify and Endpoints do not fully lock the App Engine API dependency

### LINKED ISSUE #903: Objectify and Endpoints do not fully lock the App Engine API dependency
1. Create a new project with wizard.
2. Check all three libraries.
3. Uncheck Objectify

What happens:
Objectify and App Engine API are unchecked

What should happen:
Only Objectify is unchecked. App Engine API (needed for endpoints) should remain checked.

Related bug:
1. Manually check App Engine API. 
2. Check Endpoints.
3. Uncheck endpoints

What happens:
Endpoints and App Engine API are unchecked

What should happen:
Only Endpoints is unchecked. App Engine API (needed for endpoints) should remain checked.

I suggest we simply require that App Engine API not be automatically unchecked without user action. There may also be an issue that as long as one of the two other dependencies is checked, App Engine API  should not be able to be unchecked.

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+  private Shell shell;
+  private AppEngineLibrariesSelectorGroup librariesSelector;
+  private SWTBotCheckBox appengineButton;
+  private SWTBotCheckBox endpointsButton;
+  private SWTBotCheckBox objectifyButton;
+
+  @Before
+  public void setUp() throws Exception {
+    Display.getDefault().syncExec(new Runnable() {
```
chanseokoh: Ah, this is the way to make it work even with `<useUIThread>false<useUIThread>`.

plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+      @Override
+      public void run() {
+        shell = new Shell(Display.getDefault());
```
chanseokoh: Can you add the same TODO as in [here](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven.test/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizardTest.java#L40)?

plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
```diff
@@ -0,0 +1,256 @@
...
+    });
+  }
+
+  @After
+  public void tearDown() {
+    Display.getDefault().syncExec(new Runnable() {
+
+      @Override
+      public void run() {
+        if (shell != null) {
```
chanseokoh: I don't think we need this. If `shell` were `null`, other tests will fail earlier. Even if isn't the case, I think there is no problem throwing a `NullPointerException`. WDYT?

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,205 @@ private static String getLibraryName(Library library) {
...
+    public void widgetSelected(SelectionEvent e) {
+      setManualSelection(e);
+    }
+
+    @Override
+    public void widgetDefaultSelected(SelectionEvent e) {
+      setManualSelection(e);
+    }
+
+    private void setManualSelection(SelectionEvent e) {
```
chanseokoh: Nit: `e` --> `event` here (and other places).

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,205 @@ private static String getLibraryName(Library library) {
...
+      Button source = (Button)e.getSource();
```
chanseokoh: Nit: space after the cast.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,205 @@ private static String getLibraryName(Library library) {
...
+  private static final class DependentLibrarySelected extends ComputedValue {
+    private String libraryId;
+    private IObservableList libraries;
+    private Getter<Boolean> condition;
+    private boolean selectedResult;
+
+    /**
+     * @param libraries the list of libraries to search
+     * @param libraryId the id of the library to be searched for
+     * @param foundResult return value if the library is found
```
chanseokoh: I am confused by this comment initially. I suggest to change it to `the value calculate() will return if the library is found` or something along the line.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,205 @@ private static String getLibraryName(Library library) {
...
+        public Boolean get() {
+          return !foundResult;
+        }});
+    }
+
+    /**
+     * @param libraries the list of libraries to search
+     * @param libraryId the id of the library to be searched for
+     * @param foundResult return value if the library is selected
+     * @param condition if the library is not found in the list, return the result of <code>condition.get()</code>
```
chanseokoh: I suggest the comment to `..., calculate() returns the result of condition.get()`.

I am also confused by the name `condition`. `resultIfNotFound`? Then, how about `foundResult` --> `resultIfFound`?

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,205 @@ private static String getLibraryName(Library library) {
...
+      Preconditions.checkNotNull(condition);
+      this.selectedResult = foundResult;
+      this.libraries = libraries;
+      this.libraryId = libraryId;
+      this.condition = condition;
+    }
+
+    @Override
+    protected Object calculate() {
+      for (int i = 0; i < libraries.size(); i++) {
```
chanseokoh: Enhanced `for`. `for (Object object : libraries)` will work.

I wish we were at Java 8.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,205 @@ private static String getLibraryName(Library library) {
...
+    /**
+     * @param libraries the list of libraries to search
+     * @param libraryId the id of the library to be searched for
+     * @param foundResult return value if the library is found
+     */
+    private DependentLibrarySelected(Realm realm,
+                                     IObservableList libraries,
+                                     String libraryId,
+                                     final boolean foundResult) {
+      this(realm, libraries, libraryId, foundResult, new Getter<Boolean>(){
```
chanseokoh: (Not sure if this is strict) space before the opening bracket?

plugins/com.google.cloud.tools.eclipse.appengine.ui.test/META-INF/MANIFEST.MF
```diff
@@ -9,3 +9,5 @@ Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.ui
 Require-Bundle: org.junit,
  org.eclipse.equinox.registry
+Import-Package: org.eclipse.swtbot.swt.finder.widgets,
+ org.apache.log4j
```
akerekes: Indeed it is not anymore.

plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
akerekes: I prefer to avoid NPEs as I cannot always tell easily whether it's intended and OK or programming error. Also, we already throw so many ignored exceptions during tests that I'd rather not add to the noise.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
akerekes: We're using an `IObservable` (i.e. `libraries`) to compute the value, and the databinding API requires to use the annotated getters (javadoc of `calculate()`):

> Subclasses must override this method to provide the object's value. Any dependencies used to calculate the value must be IObservable, and implementers must use one of the interface methods tagged TrackedGetter for ComputedValue to recognize it as a dependency.

I could not find a doc describing how the enhanced for loop is translated and whether that's part of the specification or implementation detail. Replacing the loop with enhanced for loop still makes the tests pass. But until I see an authoritative source that says this is intentional and not just happens because this given implementation uses one of those getters, I'd prefer to leave it as it is.

chanseokoh: IObservavble extends List, and you are simply getting each element through List.get(). This is just the iteration on the list throught the List interface. AFAIK, that is how the enhanced loop works.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+    final Library library = (Library) libraryButton.getData();
+    ISWTObservableValue libraryButtonSelection = WidgetProperties.selection().observe(libraryButton);
+    ISWTObservableValue libraryButtonEnablement = WidgetProperties.enabled().observe(libraryButton);
+    // library selection UI -> model
+    bindingContext.bindValue(libraryButtonSelection,
+                             new NullComputedValue(getDisplayRealm()),
+                             new UpdateValueStrategy().setConverter(new HandleLibrarySelectionConverter(selectedLibraries,
+                                                                                                        library)),
+                             new UpdateValueStrategy(UpdateValueStrategy.POLICY_NEVER));
+    // library selection model -> UI
```
briandealwis: It would match the code better if you flipped this around
// UI <- library selection model

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+    bindingContext.bindValue(libraryButtonSelection, 
+                             new DependentLibrarySelected(getDisplayRealm(),
+                                                          selectedLibraries,
+                                                          library.getId(),
+                                                          true /* resultIfFound */,
+                                                          new ButtonManuallySelected(libraryButton)),
+                             new UpdateValueStrategy(UpdateValueStrategy.POLICY_NEVER),
+                             new UpdateValueStrategy());
+    // library checkbox enablement model -> UI
```
briandealwis: Hmm you don't have an explicit library checkbox enablement model:
`// UI enablement <- library is a dependency`

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+                             new UpdateValueStrategy().setConverter(new HandleLibrarySelectionConverter(selectedLibraries,
+                                                                                                        library)),
+                             new UpdateValueStrategy(UpdateValueStrategy.POLICY_NEVER));
+    // library selection model -> UI
+    bindingContext.bindValue(libraryButtonSelection, 
+                             new DependentLibrarySelected(getDisplayRealm(),
+                                                          selectedLibraries,
+                                                          library.getId(),
+                                                          true /* resultIfFound */,
+                                                          new ButtonManuallySelected(libraryButton)),
```
briandealwis: This `ButtonManuallySelected` class complicates the code.  It took me a good 10 minutes to wrap my head around it.  I think it would be easier to understand if you just subclassed `DepedentLibrarySelected` and override a method.

Is `DependentLibrarySelected` right?  Maybe `IsSelectedOrRequiredLibrary`?

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
-        bindingContext.bindValue(dependencyButtonSelection, intermediate);
-        bindingContext.bindValue(dependencyButtonEnablement, intermediate,
-                                 new UpdateValueStrategy().setConverter(BooleanConverter.negate()),
-                                 new UpdateValueStrategy().setConverter(BooleanConverter.negate()));
-      }
+  private void addDatabindingForButton(final Button libraryButton) {
+    final Library library = (Library) libraryButton.getData();
+    ISWTObservableValue libraryButtonSelection = WidgetProperties.selection().observe(libraryButton);
+    ISWTObservableValue libraryButtonEnablement = WidgetProperties.enabled().observe(libraryButton);
+    // library selection UI -> model
```
briandealwis: I think you need a comment here to say something along the lines of:

We have three bindings for each library:
  1. The first is a one-way binding of the checkbox selection state to add or remove the corresponding library from our selected-libraries list.
  2. The second is the opposite of the first, a one-way binding to set the checkbox selection state when the corresponding library has been selected or if it is a dependency of another selected library
  3. The last is a one-way binding to set the checkbox enablement when the corresponding library is a dependency of a selected library

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+    bindingContext.bindValue(libraryButtonSelection,
+                             new NullComputedValue(getDisplayRealm()),
+                             new UpdateValueStrategy().setConverter(new HandleLibrarySelectionConverter(selectedLibraries,
+                                                                                                        library)),
+                             new UpdateValueStrategy(UpdateValueStrategy.POLICY_NEVER));
```
briandealwis: This feels a bit of an abuse of the databinding framework.  The converter should be free of side-effects.  The real problem here is that this isn't a value-to-value binding but a value-to-a-list binding.  The solution is to define a custom `Binding`.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+                             new DependentLibrarySelected(getDisplayRealm(),
+                                                          selectedLibraries,
+                                                          library.getId(),
+                                                          true /* resultIfFound */,
+                                                          new ButtonManuallySelected(libraryButton)),
+                             new UpdateValueStrategy(UpdateValueStrategy.POLICY_NEVER),
+                             new UpdateValueStrategy());
+    // library checkbox enablement model -> UI
+    bindingContext.bindValue(libraryButtonEnablement,
+                             new DependentLibrarySelected(getDisplayRealm(),
```
briandealwis: `IsRequiredLibrary`?

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+
+  @VisibleForTesting
+  List<Button> getLibraryButtons() {
+    return libraryButtons;
+  }
+
+  /**
+   * Utility interface to return a value;
+   */
+  private static interface Getter<T> {
```
akerekes: I was thinking about it, but that declares an `Exception` thrown in the signature and I didn't like that. I was hoping there's a simple interface I forgot about instead of this.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
akerekes: I agree that it's a hack: #911

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroup.java
```diff
@@ -107,55 +104,209 @@ private static String getLibraryName(Library library) {
...
+     * @param libraries the list of libraries to search
+     * @param libraryId the id of the library to be searched for
+     * @param resultIfFound value returned by {@link #calculate()} if the library is found
+     * @param resultIfNotFound if the library is not found in the list, return the result of <code>condition.get()</code>
+     */
+    private DependentLibrarySelected(Realm realm,
+                                     IObservableList libraries,
+                                     String libraryId,
+                                     boolean resultIfFound,
+                                     Getter<Boolean> resultIfNotFound) {
```
akerekes: Probably the naming is wrong, but the intent here is that if result not found we give a chance to clients to define the result instead of defaulting to `!resultIfFound`

PR #919, opened 2016-10-31
Fix corner cases where Cloud SDK is not available

### LINKED ISSUE #847: Check for Google Cloud SDK and App Engine Java component before creating a maven project
In the new Maven App Engine Project wizard

image

Google Cloud SDK path was not provided and could not be found in any known install locations.

briandealwis:
This PR fixes a few situations that occur when a Cloud SDK is not installed or not found.

Fixes #847

Add CloudSdkProvider checks on Native and Maven new-project wizards

screen shot 2016-10-31 at 2 32 29 pm

Catch and report Cloud SDK Not Found exceptions in Standard facet install

screen shot 2016-10-31 at 3 03 39 pm

Remove unused messages


chanseokoh:
Your first screenshot seems to be a different one?

briandealwis:
Oops good catch @chanseokoh

plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenArchetypeProjectWizard.java
```diff
@@ -95,13 +103,18 @@ public void run(IProgressMonitor monitor)
...
   @Override
-  public void init(IWorkbench workbench, IStructuredSelection selection) {}
+  public void init(IWorkbench workbench, IStructuredSelection selection) {
+    if (cloudSdkLocation == null) {
+      cloudSdkLocation = CloudSdkPrompter.getCloudSdkLocation(getShell());
+      // if the user doesn't provide the Cloud SDK then we'll error in performFinish() too
+    }
+  }
 
   private boolean appEngineJavaComponentExists() {
```
elharo: static, perhaps

chanseokoh:
I think this still does not fix [the issue described](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/847#issuecomment-256488789) in #847 for the project wizards. For example, if a user has configured a wrong path for the Cloud SDK (or the user physically deleted the Cloud SDK at the configured location), the wizards do not prompt to configure it. (`CloudSdk.Builder().build()` doesn't throw `AppEngineException`.) As a result, the wizards show the misleading dialog saying the Java component is missing. Try it yourself.

briandealwis:
The current behaviour of `CloudSdk.Builder.build()` is to not validate the location.  It only raises an exception if a location was not explicitly specified and its discovery mechanisms were not able to find a Google Cloud SDK.  The default `GOOGLE_CLOUD_SDK_HOME` and `PATH`-based discovery mechanism only checks if the [specified directory exists](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/PathResolver.java#L90), the rationale being that there may be a Google Cloud SDK update that breaks [our validation](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java#L299) even though the SDK itself is ok.

So in my testing, where I don't have the Google Cloud SDK installed into one of the expected locations, I get the following results on creating a new project (bolded indicates behaviour by this patch):
  - If I run with `GOOGLE_CLOUD_SDK_HOME` set to a gibberish path, I get a prompt (since directory doesn't exist).  (*Prior to this patch: exception logged*)
  - If I run with `GOOGLE_CLOUD_SDK_HOME=/tmp`, I get a the _App Engine Java components missing_ page (since `/tmp` is an existing directory, but not a valid installation)
  - if I run with `GOOGLE_CLOUD_SDK_HOME` unset, and no path set in the _Google Cloud SDK_ prefpane, I get a prompt. (*Prior to this patch: exception logged*)
  - if I run with `GOOGLE_CLOUD_SDK_HOME` unset, and `/tmp` set in the _Google Cloud SDK_ prefpane, I get a the _App Engine Java components missing_ page (since `/tmp` is an existing directory, but not a valid installation)

And similar exceptions happen when installing a AE Standard facet.
And the prefpane doesn't allow specifying an invalid directory.

I think we should re-think how we do this validation, but that should best be done when we can support installing components as per https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/244

chanseokoh:
> - If I run with GOOGLE_CLOUD_SDK_HOME=/tmp, I get a the App Engine Java components missing page (since /tmp is an existing directory, but not a valid installation)
> - if I run with GOOGLE_CLOUD_SDK_HOME unset, and /tmp set in the Google Cloud SDK prefpane, I get a the App Engine Java components missing page (since /tmp is an existing directory, but not a valid installation)

Can there be something we can do to at least suppress the Java components missing error in these cases? These were the cases that bit me before, and my concern is that the error message is misleading. (It instructs to do `gcloud components install app-engine-java`.)

briandealwis:
But those cases only occur where the user has explicitly specified a directory, either via the `GOOGLE_CLOUD_SDK_HOME` or through the Google Cloud SDK prefpage.  If they're b0rk3d, you'll find out pretty soon when you try to `gcloud components install app-engine-java`. :-/

chanseokoh:
That's a fair point. However, my case was that I mistakenly set a wrong directory through the Google Cloud SDK prefpage, while at the same time I had a valid, working Cloud SDK installation. `gcloud components install app-engine-java` worked fine, because `gcloud` is in `$PATH`. That made me to think that our plugin is bugged or at least not behaving properly.

briandealwis:
You should have seen a warning message in the preference page when you set the invalid directory:

screen shot 2016-11-01 at 3 43 36 pm

briandealwis:
I wonder if we even need these `CloudSdkProvider` and `CloudSdk` checks since we're downloading the libraries from Maven Central rather than taking the jars from the Google Cloud SDK.  These calls are only needed during deploy.

chanseokoh:
> I wonder if we even need these CloudSdkProvider and CloudSdk checks since we're downloading the libraries from Maven Central rather than taking the jars from the Google Cloud SDK.

I thought that way too and I asked @nbashirbello the other day why we need Cloud SDK when creating new projects. IIRC, she said it's related to installing a facet. We better need to check this with her.

Issue #883, opened 2016-10-25
App Engine dev app server is not automatically selected when there are multiple server runtimes


chanseokoh:
For example, if a user has a Tomcat runtime, it will be auto-selected instead of ours.

![selection_003](https://cloud.githubusercontent.com/assets/10523105/19706246/dfbd4656-9ade-11e6-83b3-040bebc97513.png)

Steps to reproduce:

1. Add a new server runtime environment that is not ours.
  1. _Run As_ > _Run on Server_
  1. Click on _Configure runtime environments..._
  1. _Add..._
  1. Select _Apache Tomcat v7.0_. Complete adding a new runtime. (You may need to [download Tomcat](https://tomcat.apache.org/download-70.cgi) first.) 
1. _Run As_ > _Run on Server_

briandealwis:
Do you think this is because the list is sorted alphabetically?

chanseokoh:
Possibly. It may just be that _Run on Server_ selects the first available server for web apps.

briandealwis:
The UI provides a way to mark the project to always run with a server.  I wonder if we can embed this programmatically.

briandealwis:
I wonder if we can define anti-requisites on facets (e.g., certain facets are only supported by specific servers).

akerekes:
I started to look into this as part of #693 and I found 2 things:

 - an Eclipse bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=315289 where the server selection wizard page does not display the validation message, so if e.g. the project uses jst.web 3.0 and the user selects App Engine Standard (that supports 2.5), the Next/Finish buttons will be disabled, but the reason is not visible
 - there is a way to not display the incompatible servers in the wizard if the module type is not null, when the org.eclipse.wst.server.ui.internal.wizard.page.NewManualServerComposite instance is created, however I could not find an execution path, where this is true -> all servers will be listed, even those that cannot be used to run the project on.

Issue #254, opened 2016-06-15
Don't show appengine runtimes/servers for non-appengine projects


elharo:
That is, projects that don't contain either appengine-web.xml or app.yaml

Maybe with facets?

branflake2267:
This would be a good idea, otherwise you can add a project that doesn't. And you get an error that is obvious. 

This is what happens if you start with out the `appengine-web.xml`
![screen shot 2016-11-04 at 3 24 38 pm](https://cloud.githubusercontent.com/assets/1326504/20024377/775d7f82-a2a3-11e6-8524-43d720641ac2.png)

chanseokoh:
The issue seems to say App Engine projects can be run only on the App Engine runtime. If that's the case, then it's in conflict with #930, which says we allow any runtime to run App Engine projects.

BTW, I think what @branflake2267 said above is saying a different one: the App Engine server running non-App Engine projects.

chanseokoh:
The title IMO is confusing, and I believe I misunderstood this issue previously. Correct me if I'm wrong, but I've changed the title from

"Only show appengine runtimes/servers for appengine projects"

to

"Don't show appengine runtimes/servers for non-appengine projects."

chanseokoh:
So, what we did instead of hiding App Engine runtimes/servers was to prevent adding non-appengine projects to App Engine servers: #928

Perhaps this was the best we could do. I believe @akerekes couldn't find a way to make WTP hide our servers. Actually, I like the solution of #928: I think it's better to explicitly show the reason that you cannot use the App Engine server, rather than silently hiding the servers. Hiding might confuse users.

PR #925, opened 2016-11-01
Fix runtime compatibility of App Engine Java Standard faceted project

### LINKED ISSUE #693: Configure access to App Engine server via "Run As" context menu
Even when you don’t have the App Engine facet installed, you still have access to the App Engine server via “Run As”

chanseokoh:
This may potentially address #883 and #254 by explicitly requiring to use our _App Engine Standard_ runtime for projects that have _App Engine Java Standard_ facet installed.

Runtimes that don't support the Standard facet will show the following error message when selected:

selection_002

## Details
I think we have had misconfiguration around our facet and runtime. We define the extension `org.eclipse.wst.common.project.facet.core.runtimes` associated with the same facet ID (our Standard facet `com.google.cloud.tools.eclipse.appengine.facets.standard`) in two places:
  1. `.appengine.facet`: [link](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml#L36)
  2. `.appengine.localserver`: [link](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml#L56)

It seems weird that they are defined twice while they tell different things. I believe the former one says that any runtime can support our App Engine facet. It looks like the former definition overrides the restriction from the latter. For example, if I just remove

```xml
         <facet
               id="com.google.cloud.tools.eclipse.appengine.facets.standard"
               version="1">
         </facet>
```

from the latter (indicating that our _App Engine Standard_ runtime no longer supports our facet), nothing changes: I can still select and create our runtime for our App Engine projects. However, if I in turn remove the former definition entirely (as in this PR), it will then show the error message that our runtime does not support our facet (as shown in the screenshot above).

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
@@ -33,6 +33,13 @@
     </action>
   </extension>
 
+  <extension point="org.eclipse.wst.common.project.facet.core.runtimes">
+    <supported>
+      <runtime-component any="true" />
+      <facet id="com.google.cloud.tools.eclipse.appengine.facets.standard" />
+    </supported>
+  </extension>
+
   <extension point="org.eclipse.wst.common.project.facet.core.listeners">
     <listener
       class="com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardRuntimeChangeListener"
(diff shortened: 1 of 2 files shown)


elharo:
I'm not sure if I follow, but I think this is backwards. We're trying to only show the App Engine server for projects that have the Ap Engine facet, not hide App Engine projects from other servers. 

Please coordinate with @akerekes who is also looking at this.

chanseokoh:
I misunderstood #693. The description of #693 says something else. This PR doesn't close #693.

chanseokoh:
Reverted to make any runtime support our App Engine facet. As such, this PR just removes redundant facet declaration.

elharo:
I still don't see what this is doing. I suggest closing this PR and, if necessary, opening a new issue and new PR for what this is trying to do.

chanseokoh:
Will reopen this as a new PR.

PR #928, opened 2016-11-01
Fixes #693 - Configure access to App Engine server via "Run As" context menu

### LINKED ISSUE #693: Configure access to App Engine server via "Run As" context menu

akerekes:
This PR adds an additional check to verify if the module/project being added to the dev server has the App Engine Standard facet.
While this will not hide the App Engine Server runtime in case of ineligible projects, it will display an error if the user select dev app server for such a project:

image

This works if the project would be otherwise compatible with the runtime (Java 7 and Servlet 2.5). If the project is incompatible with the runtime e.g. because of different Servlet API version, no error message will be visible, but `Next` and `Finish` buttons are disabled, see: https://bugs.eclipse.org/bugs/show_bug.cgi?id=315289


plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
```diff
@@ -90,4 +123,16 @@ public void testGetRootModules() throws CoreException {
     Assert.assertEquals("module1", rootModules[0].getId());
   }
 
+  private LocalAppEngineServerDelegate getDelegatewithServer() throws Exception {
```
chanseokoh: Seems like `CoreException` works here.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
```diff
@@ -46,6 +61,24 @@ public void testCanModifyModules() {
...
+  public void testCanModifyModules_NoAppEngineStandardFacet() throws Exception {
+    delegate = getDelegatewithServer();
+    IModule[] remove = new IModule[0];
+    IModule[] add = new IModule[]{ module1 };
+    when(module1.getProject()).thenReturn(dynamicWebProject.getProject());
+    Assert.assertEquals(Status.ERROR, delegate.canModifyModules(add, remove).getSeverity());
+  }
+
+  @Test
+  public void testCanModifyModules_appEngineStandardFacet() throws Exception {
```
chanseokoh: `CoreException` would work (after fixing `getDelegateWithServer()` too).

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
```diff
@@ -46,6 +61,24 @@ public void testCanModifyModules() {
   }
 
   @Test
+  public void testCanModifyModules_NoAppEngineStandardFacet() throws Exception {
```
chanseokoh: `CoreException` would work (after fixing `getDelegateWithServer()` too).

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -56,6 +56,7 @@
   static final String DEFAULT_RUNTIME_NAME = "App Engine Standard";
   public static final String DEFAULT_APPENGINE_SDK_VERSION = "1.9.44";
   public static final String DEFAULT_GCLOUD_PLUGIN_VERSION = "2.0.9.130.v20161013";
+  public static final IProjectFacetVersion GAE_STANDARD = ProjectFacetsManager.getProjectFacet(ID).getVersion(VERSION);
```
briandealwis: Let's not make these object final unless there's evidence that they're expensive to compute.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/TestProject.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
+import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
+import org.eclipse.wst.common.project.facet.core.internal.FacetedProjectNature;
+import org.junit.rules.ExternalResource;
+
+/**
+ * TODO Duplicate of com.google.cloud.tools.eclipse.appengine.libraries.TestProject with additional option to
+ * add facets to the project. https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/771
+ */
+public final class TestProject extends ExternalResource {
```
briandealwis: This should have `Faceted` in there somewhere.  `FacetedProjectCreator`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -65,12 +68,27 @@ public IStatus canModifyModules(IModule[] add, IModule[] remove) {
...
     return Status.OK_STATUS;
   }
 
+  private IStatus hasAppEngineStandardFacet(IModule module) {
+    try {
+      if (AppEngineStandardFacet.hasAppEngineFacet(ProjectFacetsManager.create(module.getProject()))) {
+        return Status.OK_STATUS;
+      }
+    } catch (CoreException ex) {
+      return StatusUtil.error(this, Messages.FACETED_PROJECT_CREATION_FAILED, ex);
```
briandealwis: It's strange to have an error that a project could not be created from a predicate method.  Maybe _Not a faceted project_?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -65,12 +68,27 @@ public IStatus canModifyModules(IModule[] add, IModule[] remove) {
...
+          if (status != null && !status.isOK()) {
+            return status;
+          }
         }
       }
     }
     return Status.OK_STATUS;
   }
 
+  private IStatus hasAppEngineStandardFacet(IModule module) {
```
briandealwis: These status objects should embed the module name.  You can add multiple modules and it may be just one module that fails.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
elharo: public fields should be final, whether they're expensive or not.

elharo: APPENGINE_STANDARD_VERSION for consistency

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -65,12 +68,27 @@ public IStatus canModifyModules(IModule[] add, IModule[] remove) {
           if (status != null && !status.isOK()) {
             return status;
           }
+          status = hasAppEngineStandardFacet(module);
```
elharo: Reusing local variables is error prone in the face of future changes. I suggest using two different IStatus variables.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -65,12 +68,27 @@ public IStatus canModifyModules(IModule[] add, IModule[] remove) {
...
 
+  private IStatus hasAppEngineStandardFacet(IModule module) {
+    try {
+      if (AppEngineStandardFacet.hasAppEngineFacet(ProjectFacetsManager.create(module.getProject()))) {
+        return Status.OK_STATUS;
+      }
+    } catch (CoreException ex) {
+      return StatusUtil.error(this, Messages.FACETED_PROJECT_CREATION_FAILED, ex);
+    }
+    return StatusUtil.error(this, Messages.GAE_STANDARD_FACET_MISSING);
```
elharo: +1

akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/TestProject.java
akerekes: My intent is to merge the two `TestProject` classes in the to-be-created test util bundle (i.e. get rid of the other class) and use it as a general utility class for test projects. However a renaming is due.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
akerekes: done

akerekes:
@chanseokoh @briandealwis @elharo PTAL

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/TestProject.java
```diff
@@ -87,7 +87,7 @@ private void createJavaProject(String projectName) throws CoreException, JavaMod
     if (!Strings.isNullOrEmpty(containerPath)) {
       addContainerPathToRawClasspath();
     }
-    if (projectFacetVersions != null) {
+    if (!projectFacetVersions.isEmpty()) {
```
briandealwis: Ditto re the above for `addFacets()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/TestProject.java
```diff
@@ -87,7 +87,7 @@ private void createJavaProject(String projectName) throws CoreException, JavaMod
     if (!Strings.isNullOrEmpty(containerPath)) {
```
briandealwis: I know I'd find it a bit easier to read through this if you moved this check into `addContainerPathToRawClasspath()`.  It simplifies this method, plus it moves the state-checking into the method that's actually using the state and reduces the implicit context required to read through `addContainerPathToRawClasspath()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -64,29 +65,31 @@ public IStatus canModifyModules(IModule[] add, IModule[] remove) {
...
     try {
       if (AppEngineStandardFacet.hasAppEngineFacet(ProjectFacetsManager.create(module.getProject()))) {
         return Status.OK_STATUS;
+      } else {
+        return StatusUtil.error(LocalAppEngineServerDelegate.class, NLS.bind(Messages.GAE_STANDARD_FACET_MISSING,
+                                                                             module.getName()));
       }
     } catch (CoreException ex) {
-      return StatusUtil.error(this, Messages.FACETED_PROJECT_CREATION_FAILED, ex);
+      return StatusUtil.error(LocalAppEngineServerDelegate.class, Messages.NOT_FACETED_PROJECT, ex);
```
briandealwis: This should also contain the project name.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
briandealwis: > This also follows the pattern how other facets declare IProjectFacetVersion instances.

Oh ok, then I'm ok with it.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/messages.properties
```diff
@@ -1,4 +1,4 @@
 NOT_FACETED_PROJECT=Not a faceted project.
```
briandealwis: Sorry I meant the project name in NOT_FACETED_PROJECT

elharo:
just to be clear, this also disables next and finish buttons if the facet is missing, right?

akerekes:
@elharo good point, actually it does not (I suspect an eclipse bug behind it, but I'll look into it)

chanseokoh:
Hmm... it disables the next and the finish buttons in my environment.

However, the buttons are enabled when you choose _Choose an existing server_ when you already have a server, but you will notice that the next wizard screen will not contain the module you selected. There seems no way to add a module that doesn't have our AE-S facet, so it seems safe.

akerekes:
I have observed the latter behavior (haven't tested without a server). I'd
prefer to not allow the user to the next page and click finish (which will
throw an error) once the problem is discovered for an improved UX, but this
might be difficult due to how the wizard is implemented.

akerekes:
So after some digging: the wizard page fails to disable the `Next` and `Finish` buttons since [this WTP patch](https://bugs.eclipse.org/bugs/attachment.cgi?id=238703&action=diff) was applied to fix another issue regarding button enablement. I have filed https://bugs.eclipse.org/bugs/show_bug.cgi?id=506990 to address this.

For now I don't think we can do much besides sending a patch to WTP. The user will be able to click Next/Finish, on Next the module is not available to be added to the server and if the user clicks Finish, an error message will be shown: 
image
Not the most user-friendly, but at least the user is made aware of the error again (after the error message has already been shown in the title area of the dialog).

akerekes:
@elharo PTAL

Issue #932, opened 2016-11-02
Objectify javadoc not displayed


akerekes:
Objectify Javadoc is hosted on javadoc.io, which is served by CloudFlare. The default settings on CloudFlare seem to disallow user agents from JDKs: https://github.com/nodejs/node-v0.x-archive/issues/25912
I've sent an email to the contacts on javadoc.io to see if they can modify the settings.
The problem is that we cannot override the user agent used to got the javadoc pages (maybe globally for the JVM running our plugin, but that's probably not something we should do), and there doesn't seem to be another publicly hosted Objectify javadoc site.

Alternatively once #921 is implemented we won't depend on javadoc.io.

briandealwis:
#921 is the right way to go, methinks.  We can document the Javadoc resolving issue, with the workaround to modify the `eclipse.ini` and add `-Dhttp.agent=xxx`.

branflake2267:
Just an idea. If you can't reach their javadoc, you could build it yourself, for use internally. I do this with third party dependencies, and display it on some of our sites. 

This is what I do. 
1. add pom.xml to build process
2. run it with mvn clean install -Dgwt.version=2.8.0 to change the version I need
3. It builds into the target directory, then I can distribute it to the location I need. 

https://gist.github.com/branflake2267/7f3a3bfcbaef71c752fe75129c34661d - pom.xml to build. I do this with several of the GWT dependencies. They just need to be on Maven central to Javadoc, or installed.

akerekes:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=506935

elharo:
seems to have been fixed.

PR #934, opened 2016-11-02
Fixes the URL for Objectify javadoc

akerekes:
It will still not work because of #932 , but once javadoc.io maintainers change the CloudFlare settings, it will.


akerekes:
@chanseokoh that's OK, if you request for a specific resource, then it works, check with this: http://static.javadoc.io/com.googlecode.objectify/objectify/5.1.13/com/googlecode/objectify/ObjectifyFactory.html

briandealwis:
@chanseokoh you should have no problems if you launch with `-Dhttp.agent=random/1.0` or anything other than `Java/1.XXXX`.

akerekes:
@briandealwis The root directory will give error even in a browser.

briandealwis:
I wasn't sure if @chanseokoh was referring to the javadocUri base URL or from the actual Javadoc hover as they haven't fixed that.  They both return _Access Denied_ but with very different content.

PR #941, opened 2016-11-03
WIP - move classes to test bundle

akerekes:
The changes are meant to be complete except for the `XmlTest`/`AbstractXmlTest`/`PluginXmlDocument` where I'm looking for feedback whether I should go ahead with it, or make major changes.

We have an `XmlTest` in almost all of our test bundles verifying the corresponding production bundle's `plugin.xml`. The reason for the Manifest magic is that: 1. I wanted to keep the test lightweight in the sense that it does not have to be run as plugin test, 2. rely on the conventions we use now instead of explicitly specifying the location of the host bundle's `plugin.xml`.

If we agree that it's a good direction or find a better solution, I'll change the other `XmlTests` accordingly as part of this PR.


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.rules.ExternalResource;
+import org.w3c.dom.Document;
+
+/**
+ * Test utility class to obtain a {@link Document} representing the host bundle's plugin.xml.
+ * <p>
+ * Assumptions:
+ * <ul>
+ *  <li>instances are created only in test fragments bundles that are hosted by the corresponding production</li>
+ *  <li>host and test bundles are located under the same parent directory</li>
```
briandealwis: (i.e., the tests execute with the working directory in a bundle or fragment directory)

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+    DocumentBuilder builder = factory.newDocumentBuilder();
+    return builder;
+  }
+
+  private String getHostBundleName() throws IOException, FileNotFoundException {
+    String manifestPath = "META-INF/MANIFEST.MF";
+    Manifest manifest = new Manifest(new FileInputStream(manifestPath));
+    Attributes attr = manifest.getMainAttributes();
+    String testBundleName = attr.getValue("Bundle-SymbolicName");
+    List<String> bundleNameComponents = Splitter.on('.').splitToList(testBundleName);
```
briandealwis: This could do with some explanation because I'm reading it and a bit lost!  There's a few things that make this a bit error prone:
  - the BSN for a bundle or fragment with a `plugin.xml`/`fragment.xml` should have a ';singleton:=true' attribute, which you should strip off here
  - this assumption that you can strip off the last trailing component is rather brittle

Fortunately fragments specify their host bundle (`Fragment-Host`), so you could avoid most of these issues and use that instead.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+ * Test utility class to obtain a {@link Document} representing the host bundle's plugin.xml.
+ * <p>
+ * Assumptions:
+ * <ul>
+ *  <li>instances are created only in test fragments bundles that are hosted by the corresponding production</li>
+ *  <li>host and test bundles are located under the same parent directory</li>
+ *  <li>test bundle's name (Bundle-SymbolicName) is the host bundle's name postfixed with <code>.test</code>
+ * </ul>
+ */
+public class PluginXmlDocument extends ExternalResource {
```
briandealwis: I think it's worth adding a constructor to explicitly specify the bundle symbolic-name to test.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private Document doc;
+
+  @Override
+  protected void before() throws Throwable {
+    String hostBundleName = getHostBundleName();
+    String pluginXmlLocation = "../" + hostBundleName + "/plugin.xml";
+    DocumentBuilder builder = createDocumentBuilder();
+    // test fails if malformed
+    doc = builder.parse(new File(pluginXmlLocation));
```
briandealwis: It would be worth separating this plugin.xml-locating code so subclasses can override it if necessary.  Have the method return an `InputStream`.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/TestProject.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.cloud.tools.eclipse.appengine.localserver.server;
+package com.google.cloud.tools.eclipse.test.util.project;
```
briandealwis: Not a fan of this class name.  I think it should indicate that it's a faceted project.  `FacetedProjectCreator`?

Do the todos in the class comment still apply?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/service/TestServiceRegistrar.java
```diff
@@ -14,24 +14,29 @@
...
+  private ServiceRegistration<S> serviceRegistration;
+  private Class<S> serviceClass;
+  private S service;
+
+  public TestServiceRegistrar(Class<S> serviceClass) {
+    this.serviceClass = serviceClass;
+  }
 
   @Override
   protected void before() throws Throwable {
```
chanseokoh: You can remove `throws Throwable`.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+ *  <li>host and test bundles are located under the same parent directory</li>
+ *  <li>test bundle's name (Bundle-SymbolicName) is the host bundle's name postfixed with <code>.test</code>
+ * </ul>
+ */
+public class PluginXmlDocument extends ExternalResource {
+
+  private Document doc;
+
+  @Override
+  protected void before() throws Throwable {
```
chanseokoh: You can specialize `Throwable`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/ServerPortExtensionTest.java
```diff
@@ -41,28 +41,22 @@
 @RunWith(MockitoJUnitRunner.class)
 public class ServerPortExtensionTest {
 
+  @Rule ShellTestResource shellTestResource = new ShellTestResource();
```
chanseokoh: Should be `public` to use `@Rule`.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/AbstractXmlTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.google.cloud.tools.eclipse.test.util;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.w3c.dom.Document;
+
+public class AbstractXmlTest {
+
+  @Rule public PluginXmlDocument pluginXmlDocument = new PluginXmlDocument();
```
chanseokoh: Just a thought: I don't think `PluginXmlDocument` will be re-used elsewhere. Why not merge it into `AbstractXmlTest`? (I'm not strong about this.)

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+import javax.xml.parsers.ParserConfigurationException;
+import org.junit.rules.ExternalResource;
+import org.w3c.dom.Document;
+
+/**
+ * Test utility class to obtain a {@link Document} representing the host bundle's plugin.xml.
+ * <p>
+ * Assumptions:
+ * <ul>
+ *  <li>instances are created only in test fragments bundles that are hosted by the corresponding production</li>
```
elharo: test fragment bundles

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
elharo: I think you need to delete "are" but 'm having trouble parsing this sentence.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,81 @@
...
+    DocumentBuilder builder = createDocumentBuilder();
+    // test fails if malformed
+    doc = builder.parse(new File(pluginXmlLocation));
+  }
+  
+  public Document get() {
+    return doc;
+  }
+
+  private DocumentBuilder createDocumentBuilder() throws ParserConfigurationException {
```
elharo: static

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
elharo: Do we need that now?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
elharo: Yes, a comment explaining what you're doing here with an example string would be very useful.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
akerekes: I omitted it for the same reason, we can add it easily later if needed.

elharo:
FYI, I'm OK with changing the existign Xmltests in a subsequent PR if that's helpful.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/TestProject.java
akerekes: I've changed the name to `TestProjectCreator`. I'm against `FacetedProjectCreator` only because we can create non-faceted projects with this class as well.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/META-INF/MANIFEST.MF
```diff
@@ -10,6 +10,9 @@ Fragment-Host: com.google.cloud.tools.eclipse.appengine.deploy.ui
 Require-Bundle: org.hamcrest;bundle-version="1.1.0",
  org.junit;bundle-version="4.12.0"
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
+ com.google.cloud.tools.eclipse.test.util,
+ com.google.cloud.tools.eclipse.test.util.ui,
+ com.google.common.base;version="15.0.0",
```
akerekes: I changed it to 17. We have guava 17 bundled in `com.google.cloud.tools.eclipse.3rdparty.feature`. If we want to update to 20, we can do it (unless there's some dependency on 17) but should be done in a separate PR.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
akerekes: I've replaced it with @briandealwis 's suggestion which made it clear what I meant originally.

akerekes:
@chanseokoh @elharo @briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -16,52 +16,61 @@
...
 
   private Document doc;
 
   @Override
-  protected void before() throws Throwable {
-    String hostBundleName = getHostBundleName();
-    String pluginXmlLocation = "../" + hostBundleName + "/plugin.xml";
+  protected void before() throws ParserConfigurationException, SAXException, IOException {
     DocumentBuilder builder = createDocumentBuilder();
+    InputStream pluginXml = getPluginXml();
```
briandealwis: Use a try-with to ensure the stream is closed.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -72,10 +81,6 @@ private String getHostBundleName() throws IOException, FileNotFoundException {
     String manifestPath = "META-INF/MANIFEST.MF";
     Manifest manifest = new Manifest(new FileInputStream(manifestPath));
     Attributes attr = manifest.getMainAttributes();
-    String testBundleName = attr.getValue("Bundle-SymbolicName");
-    List<String> bundleNameComponents = Splitter.on('.').splitToList(testBundleName);
-    int size = bundleNameComponents.size();
-    assertThat(bundleNameComponents.get(size - 1), is("test"));
-    return Joiner.on('.').join(bundleNameComponents.subList(0, size - 1));
+    return attr.getValue("Fragment-Host");
```
briandealwis: What should happen if this isn't a fragment?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/PluginXmlDocument.java
```diff
@@ -0,0 +1,86 @@
...
+  protected void before() throws ParserConfigurationException, SAXException, IOException {
+    DocumentBuilder builder = createDocumentBuilder();
+    InputStream pluginXml = getPluginXml();
+    assertNotNull(pluginXml);
+    // test fails if malformed
+    doc = builder.parse(pluginXml);
+  }
+
+  /**
+   * Subclasses should override this method in case the plugin.xml is not at the location of
```
elharo: in case the --> if

PR #952, opened 2016-11-06
Fix cast logic error

chanseokoh:
Currently, we cast `event.getSource()` into `Button` before checking if the object is a `Button`.


chanseokoh:
I'm certain it's always a `Button`. Using `Preconditions` seems better.

PR #956, opened 2016-11-07
Fixes #800 - API attacher - Download source artifact and set its location in the classpath entry

### LINKED ISSUE #800: API attacher - Download source artifact and set its location in the classpath entry
When a source uri is defined for a library, we need to download the file pointed by the URI and add the local path to the classpath entry.
If there's no source URI defined we can try to resolve the source artifact using Maven.

akerekes:
Downloading source artifacts is synchronous for now.
#957 filed for async download.


plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/M2RepositoryServiceTest.java
```diff
@@ -53,10 +56,15 @@
 
   private M2RepositoryService m2RepositoryService;
   
+  @Before
+  public void setUp() throws Exception {
```
chanseokoh: `throws Exception` is not necessary.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -44,6 +47,8 @@
 @Creatable
 public class LibraryClasspathContainerSerializer {
 
+  private static final String APPENGINE_LIBRARIES_BUNDLE_NAME = "com.google.cloud.tools.eclipse.appengine.libraries";
```
chanseokoh: Seems like this is used only by `LibrariesBundleStateLocationProvider`. Put it there?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -154,4 +166,19 @@ public IPath getBaseLocation() {
       return new Path(MavenPlugin.getRepositoryRegistry().getLocalRepository().getBasedir().getAbsolutePath());
     }
   }
+
+  @VisibleForTesting
+  static class LibrariesBundleStateLocationProvider implements ArtifactBaseLocationProvider {
```
chanseokoh: Seems like this can be `private static`, without `@VisibleForTesting`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/SerializableClasspathEntry.java
```diff
@@ -62,18 +90,43 @@ public void setAccessRules(IAccessRule[] accessRules) {
...
-                                    new Path(sourceAttachmentPath),
+                                    sourceAttachmentPath.isEmpty() ? null : restoreSourcePath(baseDirectory, 
+                                                                                              sourceBaseDirectory),
                                     null,
                                     getAccessRules(),
                                     getAttributes(),
                                     true);
   }
 
+  protected IPath restoreSourcePath(IPath baseDirectory, IPath sourceBaseDirectory) {
```
chanseokoh: `private`, I think.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/M2RepositoryService.java
```diff
@@ -50,33 +56,65 @@
...
     return JavaCore.newLibraryEntry(new Path(artifact.getFile().getAbsolutePath()),
-                                    getSourceLocation(libraryFile),
+                                    getSourceLocation(mavenCoordinates, sourceUrl),
                                     null /*  sourceAttachmentRootPath */,
                                     getAccessRules(libraryFile.getFilters()),
                                     libraryFileClasspathAttributes,
                                     true /* isExported */);
   }
 
+  protected URL getSourceUrlFromUri(URI sourceUri) {
```
chanseokoh: `private`, I think.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/M2RepositoryService.java
```diff
@@ -98,26 +136,63 @@ private Artifact resolveArtifact(MavenCoordinates mavenCoordinates) throws Libra
...
-    } catch (CoreException | MalformedURLException ex) {
+    } catch (CoreException ex) {
       throw new LibraryRepositoryServiceException("Could not create classpath attributes", ex);
     }
   }
 
-  private IPath getSourceLocation(LibraryFile libraryFile) {
-    if (libraryFile.getSourceUri() == null) {
-      return getSourceJarLocation(libraryFile.getMavenCoordinates());
+  protected void addUriAttribute(List<IClasspathAttribute> attributes, String attributeName, URI uri) {
```
chanseokoh: `private`, I think.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/io/FileDownloaderTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();
+  
+  @Test(expected = NullPointerException.class)
+  public void testConstructor_null() {
+    new FileDownloader(null);
+  }
+
+  @Test
+  public void testConstructor_emptyPath() {
+    assertNotNull(new FileDownloader(new Path("")));
```
chanseokoh: `new FileDownloader()` can never be null, so this test seems weird.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/io/FileDownloaderTest.java
```diff
@@ -0,0 +1,69 @@
...
+  
+  @Test(expected = IOException.class)
+  public void testDownload_cannotCreateDownloadFolder() throws IOException {
+    FileDownloader fileDownloader = new FileDownloader(new Path("/dev/null/foo"));
+    fileDownloader.download(new URL("http://google.com"));
+  }
+  
+  public void testDownload_successful() throws IOException {
+    FileDownloader fileDownloader = new FileDownloader(new Path(temporaryFolder.newFolder().getAbsolutePath()));
+    IPath downloadPath = fileDownloader.download(new URL("https://www.google.com/intl/en/about.html"));
```
chanseokoh: Are you actually downloading a file from the Internet? Sounds like an integration test rather than a unit test. I think this should be mocked.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/FileDownloader.java
```diff
@@ -0,0 +1,82 @@
...
+   * <p>
+   * If the directory does not exist, it will be created on-demand when the first file is downloaded.
+   *
+   * @param downloadFolderPath path to the directory where the files must be downloaded. Cannot be <code>null</code>,
+   * but does not have to exist, it will be created on demand.
+   */
+  public FileDownloader(IPath downloadFolderPath) {
+    Preconditions.checkNotNull(downloadFolderPath, "downloadFolderPath is null");
+    File downloadFolder = downloadFolderPath.toFile();
+    Preconditions.checkArgument(!downloadFolder.exists() || downloadFolder.isDirectory());
```
chanseokoh: It'd be good check if the path is absolute. The behavior is quite unpredictable from users' perspective if a path is relative.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -73,6 +73,10 @@ public ServletClasspathProvider() {
     }
   }
 
+  // TODO this method is called e.g. when the user clicks through the project to drill down into the libraries attached
+  // to the project and resolves the servlet and jsp jars each time. So we must ensure that if the supported versions
+  // change in the future, resolved versions of these jars remain the same for a project as long as it is associated
+  // the same runtime https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/953
```
briandealwis: This comment is still a bit cryptic (i.e., I'm not sure I understand it)

plugins/com.google.cloud.tools.eclipse.test.util/.classpath
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
+	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src/"/>
```
briandealwis: You should probably revert this change.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/io/FileDownloaderTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Test(expected = NullPointerException.class)
+  public void testDownload_null() throws IOException {
+    FileDownloader fileDownloader = new FileDownloader(new Path(temporaryFolder.newFolder().getAbsolutePath()));
+    fileDownloader.download(null);
+  }
+  
+  @Test(expected = IOException.class)
+  public void testDownload_cannotCreateDownloadFolder() throws IOException {
+    FileDownloader fileDownloader = new FileDownloader(new Path("/dev/null/foo"));
+    fileDownloader.download(new URL("http://google.com"));
```
briandealwis: But what if google.com is unavailable? :-)

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/io/FileDownloaderTest.java
briandealwis: I agree.  It would also be good to have a test to verify that the content matches.  E.g., return a known set of bytes and ensure the on-disk set match.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/M2RepositoryService.java
```diff
@@ -98,26 +136,63 @@ private Artifact resolveArtifact(MavenCoordinates mavenCoordinates) throws Libra
...
-      return new Path("/downloaded/source/file"); //$NON-NLS-1$
+      return getDownloadedSourceLocation(mavenCoordinates, sourceUrl);
     }
   }
 
+  private IPath getDownloadedSourceLocation(MavenCoordinates mavenCoordinates, URL sourceUrl) {
+    
+    try {
+      IPath downloadFolder = getDownloadedFilesFolder(mavenCoordinates);
+      IPath path = new FileDownloader(downloadFolder).download(sourceUrl);
```
briandealwis: I question whether we want to go down this path.  Source jars on MC have a predictable layout, but handling arbitrary source jars is fraught.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/FileDownloader.java
```diff
@@ -0,0 +1,82 @@
...
+   * @return a path pointing to the downloaded file
+   * @throws IOException if the URL cannot be opened, the output file cannot be written or the transfer of the remote
+   * file fails
+   */
+  public IPath download(URL url) throws IOException {
+    Preconditions.checkNotNull(url, "url is null");
+    ensureDownloadFolderExists();
+
+    File downloadedFile = downloadFolderPath.append(new Path(url.getPath()).lastSegment()).toFile();
+    URLConnection connection = url.openConnection();
```
chanseokoh: I wonder if we should consider `setConnectTimeout()` and `setReadTimeout()`, even if we make this async later, because, on my machine, the default timeout for both is 0 (i.e., timeout of infinity).

elharo:
Looks like @akerekes is on leave. @briandealwis can you take this over? Is it ready to be merged?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/FileDownloader.java
akerekes: I'd rather delegate the responsibility to the user in this case as the path is provided by him. Our class works regardless of relative or absolute path and the download method will return an absolute path to the downloaded file.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/M2RepositoryService.java
akerekes: We can add the sourceAttchmentRootPath that JDT uses for classpath entries in a subsequent path.

akerekes:
@briandealwis the reason for separate serialization/deserialization is that I expect that in most cases the persisted information will be correct so no need to do a resolution each time. I expect re-resolution when the M2_REPO or the plugin state location is changed/cleared which should not happen very often.

akerekes:
@elharo @briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/M2RepositoryService.java
```diff
@@ -151,8 +226,16 @@ private ArtifactRepository getCustomRepository(String repository) throws Library
...
-  private IPath getSourceJarLocation(MavenCoordinates mavenCoordinates) {
-    return new Path("/path/to/source/jar/file/in/m2_repo/" + mavenCoordinates.getArtifactId() + "." + mavenCoordinates.getType()); //$NON-NLS-1$ //$NON-NLS-2$
+  private IPath getMavenSourceJarLocation(MavenCoordinates mavenCoordinates) {
+    try {
+      MavenCoordinates sourceMavenCoordinates = new MavenCoordinates(mavenCoordinates);
+      sourceMavenCoordinates.setClassifier("sources");
+      Artifact artifact = resolveArtifact(sourceMavenCoordinates);
+      return new Path(artifact.getFile().getAbsolutePath());
+    } catch (LibraryRepositoryServiceException exception) {
+      // source file is failed to download, this is not an error
```
elharo: source file failed

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/http/TestHttpServer.java
```diff
@@ -0,0 +1,88 @@
...
+
+    server.dumpStdErr();
+    server.start();
+  }
+
+  private void stopServer() {
+    try {
+      server.stop();
+      server.join();
+    } catch (Exception ex) {
```
elharo: IOException?

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
```diff
@@ -13,6 +13,8 @@ Export-Package: com.google.cloud.tools.eclipse.util,
  com.google.cloud.tools.eclipse.util.status,
  com.google.cloud.tools.eclipse.util.templates.appengine
 Import-Package: com.google.common.base;version="15.0.0",
```
elharo: do you want to update this too?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/http/TestHttpServer.java
akerekes: unfortunately `stop()` and `join()` `throws Exception` so we cannot catch a more specific exception.

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
akerekes: sure

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
akerekes: I tried to add more clarification, but maybe it only became more convoluted. If so, lets chat about it and I'll explain what I mean and we can find a better way to document it.

plugins/com.google.cloud.tools.eclipse.test.util/.classpath
akerekes: After merging master this line has been readded a few lines above.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/FileDownloader.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/AppEngineLibraryContainerInitializer.java
```diff
@@ -121,7 +121,9 @@ private void validateJarPaths(LibraryClasspathContainer container) throws Librar
     IClasspathEntry[] classpathEntries = container.getClasspathEntries();
     for (int i = 0; i < classpathEntries.length; i++) {
       IClasspathEntry classpathEntry = classpathEntries[i];
-      if (!classpathEntry.getPath().toFile().exists()) {
+      if (!classpathEntry.getPath().toFile().exists()
+          || (classpathEntry.getSourceAttachmentPath() != null
+              && !classpathEntry.getSourceAttachmentPath().toFile().exists())) {
         classpathEntries[i] = repositoryService.rebuildClasspathEntry(classpathEntry);
```
briandealwis: Future Speculation: I wonder if this wouldn't be better put into the repositoryService somehow — just thinking that the repository service would want to control how many of these operations are in progress at once.

Then again, it could use a `JobGroup` to manage the family of jobs.  That lets the Jobs framework manage parallelism.

PR #959, opened 2016-11-07
Temporarily disable Flex facet

nbashirbello:


chanseokoh:
Unit tests need to be fixed too.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
```diff
@@ -27,33 +27,36 @@
...
+        ProjectFacetsManager.isProjectFacetDefined("com.google.cloud.tools.eclipse.appengine.facets.flex"));
+  }
+
   public void testFlexFacetExists() {
     Assert.assertTrue(
         ProjectFacetsManager.isProjectFacetDefined("com.google.cloud.tools.eclipse.appengine.facets.flex"));
   }
 
-  @Test
   public void testHasAppEngineFacet_withFacet() {
```
elharo: I don't like having test methods that are hidden like this by omitting @Test. IN google3 someone is working on a static analysis to flag this as an almost certain mistake.

nbashirbello: What do you suggest as a fix for temporarily not running these test?

elharo: Honestly, disabling tests makes me queasy. What might be called for here is a branch with the flex development. The head branch shouldn't have these tests.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
```diff
@@ -38,26 +37,39 @@ public void testFlexFacetDoesNotExists() {
...
   }
 
-  public void testFlexFacetExists() {
-    Assert.assertTrue(
-        ProjectFacetsManager.isProjectFacetDefined("com.google.cloud.tools.eclipse.appengine.facets.flex"));
-  }
-
+  @Test
   public void testHasAppEngineFacet_withFacet() {
+    if (!ProjectFacetsManager.isProjectFacetDefined(
```
elharo: This is what the Assume class in Junit is designed for:
http://junit.sourceforge.net/javadoc/org/junit/Assume.html
It lets the test runner notate the tests as skipped rather than passed.

PR #961, opened 2016-11-07
Show grayed out workspace location by default

### LINKED ISSUE #211: show default location
In the maven based app engine project wizard, when create project in workspace is checked, show the default location in grayed out text in the Location field, like we do in the other wizard:

image

chanseokoh:
Fixes #211.

The behavior is same as that of the workspace project location field in standard wizards:
- If disabled, it always shows the current workspace location.
- When enabled for the first time, the field is empty.
- What user had entered is saved and will be restored back when the field becomes enabled again.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -75,13 +75,31 @@
   /** True if we're programmatically setting javaPackageField with an auto-generated value */
   private boolean javaPackageProgrammaticUpdate = false;
 
+  private WorkspaceLocationProvider workspaceLocationProvider;
+
+  @VisibleForTesting
+  static interface WorkspaceLocationProvider {
```
elharo: Why  add an interface here? Seems like overkill. Could we just define a class? For that matter, why a provider instead of just an IPath?

chanseokoh: Wow, I should have used an `IPath`. Thanks.

PR #963, opened 2016-11-08
Select single account in deploy dialog when possible

### LINKED ISSUE #794: select single default account
In the deployment dialog, even when exactly one account is signed in, it comes up with an empty dialog for account. In this case, it should select that account automatically.

chanseokoh:
Fixes #794.

In the deploy dialog, if the account selector has only one account signed in, always select it. (As desired, the saved account email in the model isn't affected. So, the saved account will still be auto-selected if that account is added in later.)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -161,6 +161,11 @@ public IStatus validate(Object value /* email */) {
         .setConverter(new Converter(String.class, String.class) {
       @Override
       public Object convert(Object fromObject /* email */) {
+        // If there exists only one account signed in, make it selected.
+        String singleAccountEmail = accountSelector.getSingleAccountEmail();
```
briandealwis: This converter doesn't seem the right location.  I expected to find logic in the `AccountSelector`.

chanseokoh:
I also need to fix a test.

chanseokoh:
> perhaps configured by the AccountSelector's instantiator.

@briandealwis I believe you mean `AccountSelectorObservableValue`. Doing this inside `AccountSelector` won't work because the databinding will kick in later.

briandealwis:
I think the option you're implementing is a useful one, so I think `AccountSelector` should support that option independently of databinding.  In that light, some logic would belong in `AccountSelector#selectAccount()`: if set to `null` or the empty string, AND selecting-single-account-mode, then select the single account.  If everything is done right, that should propagate back using standard databinding.

chanseokoh:
Got into a trick situation while implementing this through `AccountSelector.selectAccount()`.

[If the `modelToTarget` converter returns `null` when there is no matching account in the selector](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java#L164
), then [`AccountSelectorObservableValue.doSetValue()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorObservableValue.java#L61) is not called at all. Not sure why and how to make the databinding framework call this method when the converter returns `null`.

chanseokoh:
The reason was that the `null` value after conversion is intercepted by the `accountSelectedChecker`, which raises an validation error.

chanseokoh:
@briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -154,10 +154,21 @@ public IStatus validate(Object value /* email */) {
         return ValidationStatus.ok();
       }
     };
+    IValidator modelToTargetValidator = new IValidator() {
+      @Override
+      public IStatus validate(Object value /* email */) {
```
elharo: Generally, outer classes are easier to unit test.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -100,9 +99,26 @@ public int selectAccount(String email) {
     return index;
   }
 
+  /**
+   * Works the same way as {@code selectAccount} does, except that if there exists only one
```
elharo: there exists only one account signed in --> only one account is signed in

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -100,9 +99,26 @@ public int selectAccount(String email) {
...
+   * account signed in, that account will always be selected regardless of {@code email}.
+   *
+   * @see #selectAccount
+   */
+  public int selectAccountInSingleAccountSelectMode(String email) {
```
elharo: we really need an extra method here? N way to include this logic in selectAccount?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
elharo: I'm not sure but I'm tempted to go a third way. It looks cleaner to me (though I'm not 100% sure of this) to not add anything to AccountSelector, check if there's only one, and if so set it here. But I don't have all the code in front of me, so maybe not.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
chanseokoh: Actually, I first added the logic into `selectAccount()` by accepting an additional parameter `singleAccountSelectMode`. The problem was that it was very difficult to write a Javadoc for `selectAccount()` in a clean way due to the complexity of all the possible cases. As such, it is not trivial to understand all the behaviors of `selectAccount()` either. I think having a separate method is better; the Javadocs for both methods are clean.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
chanseokoh: Done.

chanseokoh:
> I'm not sure but I'm tempted to go a third way. It looks cleaner to me (though I'm not 100% sure of this) to not add anything to AccountSelector, check if there's only one, and if so set it here. But I don't have all the code in front of me, so maybe not.

@elharo The original commit in this PR did what you said at that code location, through databinding. If what you meant by "a third way" is explicitly selecting an account outside databinding, then it's not going to work (I just tried that), since the validation goes out of sync with this. I think values in the model and the selector should be set and loaded through the databinding framework — I think that's the supposed way to use databinding.

chanseokoh:
@briandealwis Can you also update the Javadoc of `selectAccount()` accordingly in that branch?

There's a bug, BTW. Below, the account is selected, but it shows the error message:

selection_020

I think the reason of the bug is that the selector selects the account by itself, bypassing the databinding framework. I can verify that doing
```java
    IValidator modelToTargetValidator = new IValidator() {
     @Override
     public IStatus validate(Object value /* email */) {
       // Return OK, because if only one account is signed in, accountSelector will auto-select it.
       if (accountSelector.getAccountCount() == 1) {
         return ValidationStatus.ok();
       }
       return accountSelectedChecker.validate(value);
     }
   };
```
can resolve the bug, as expected, although not sure if this is best in this context.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -465,4 +452,29 @@ public void setFont(Font font) {
...
+  };
+
+  private class ModelToAccountSelectorValidator implements IValidator {
+    @Override
+    public IStatus validate(Object value /* email */) {
+      // Return OK, because if only one account is signed in, accountSelector will auto-select it.
+      if (accountSelector.getAccountCount() == 1) {
+        return ValidationStatus.ok();
+      }
+      return new AccountSelectedChecker().validate(value);
```
chanseokoh: For simplicity, if I have to answer. I'll make it accept the instance through the constructor.

briandealwis:
I'll take this as @chanseokoh is away.  I discovered that there's a missing piece in that the auto-single-select'd account isn't propagated back to the model.

elharo:
OK. Honestly it sounds likes data binding may be causing us more problems than it's solving here.

briandealwis:
Adds a new option to `AccountSelector` to automatically select an account when none is set and there is a single known account.  Ensures such a change is propagated back to the model.  Separates the account-selection validation logic from the actual data-binding.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -141,33 +144,29 @@ private void setupDataBinding() {
...
-            return ValidationStatus.error(Messages.getString("error.account.missing.signedin"));
-          } else {
-            return ValidationStatus.error(Messages.getString("error.account.missing.signedout"));
+    AccountSelectorObservableValue accountSelectorObservableValue =
+        new AccountSelectorObservableValue(accountSelector);
+    UpdateValueStrategy modelToTarget =
+        new UpdateValueStrategy().setConverter(new Converter(String.class, String.class) {
+          @Override
+          public Object convert(Object fromObject /* email */) {
+            return accountSelector.isEmailAvailable((String) fromObject) ? fromObject : null;
```
elharo: can you rewrite this without the ?: operator. I had to read this three times before I saw it was a simple not null check. Oh hell, that isn't it either, is it? Simple if-else would make this much easier to read, at least for me.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorTest.java
```diff
@@ -208,6 +207,93 @@ public void testSelectAccount_nullOrEmptyEmail() {
...
 
   @Test
+  public void testSelectAccountInSingleAccountSelectMode_multipleAccounts() {
+    when(loginService.getAccounts())
+        .thenReturn(new HashSet<>(Arrays.asList(account1, account2, account3)));
+    AccountSelector selector =
+        new AccountSelector(shell, loginService, "<select this to login>", true);
+
+    selector.selectAccount("some-email-2@example.com");
+    assertNotEquals(-1, selector.combo.getSelectionIndex());
```
elharo: shouldn't we assertEqual on the expected index? I would expect this isn't a random value.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorTest.java
```diff
@@ -208,6 +207,93 @@ public void testSelectAccount_nullOrEmptyEmail() {
...
+
+    AccountSelector selector =
+        new AccountSelector(shell, loginService, "<select this to login>", true);
+    assertEquals(0, selector.combo.getSelectionIndex());
+    assertEquals(credential1, selector.getSelectedCredential());
+    assertEquals("some-email-1@example.com", selector.getSelectedEmail());
+
+    selector.selectAccount("some-email-1@example.com");
+
+    assertNotEquals(-1, selector.combo.getSelectionIndex());
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorTest.java
```diff
@@ -208,6 +207,93 @@ public void testSelectAccount_nullOrEmptyEmail() {
...
+    assertEquals(credential1, selector.getSelectedCredential());
+    assertEquals("some-email-1@example.com", selector.getSelectedEmail());
+
+    selector.selectAccount(null);
+
+    assertNotEquals(-1, selector.combo.getSelectionIndex());
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -37,14 +36,25 @@
   private String loginMessage;
   private Credential selectedCredential;
 
+  /**
+   * If true and if there is no selected account and there is exactly one logged-in account, then
+   * select that account.
+   */
+  private boolean selectDefaultSingleAccount = true;
```
elharo: why do we need this? When is this not true?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -77,22 +91,30 @@ public Credential getSelectedCredential() {
...
+   * item. If there is no account corresponding to the {@code email}, <b>and</b> there is exactly 1
+   * known account and {@link #selectDefaultSingleAccount} is true, then we automatically select
+   * that single account; otherwise this method does returns -1 while retaining current selection
+   * (if any).
    *
-   * @param email email address to use to select an account. If {@code null} or the empty string,
-   *     or if there is no matching account, this method does nothing
-   * @return index of the newly selected combo item; -1 if {@code email} is {@code null} or
-   *     the empty string, or if there is no matching account
+   * @param email email address to use to select an account.
```
elharo: nit: no period

chanseokoh:
The latest commit has a new critical bug. Steps to reproduce:

1. Log out all accounts.
1. Create a new project. (Doesn't seem necessary, but just in case.)
1. Open the deploy dialog.
1. Select "<Add a new account...>" and sign in an account. (Now you have one account in the combo.)
1. Select "<Add a new account...>" again and sign in another account.

Observe that the combo doesn't have the last account, i.e., it has only one account.

Expected outcome: the second account should be added to the combo and auto-selected as well.

chanseokoh:
I am guessing that the bug above is because that `selectAccount()` is being used in another place, in [`addAndSelectAccount()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java#L126). Modifying `selectAccount()` to work in the `selectDefaultSingleAccount` mode breaks `addAndSelectAccount()`.

@briandealwis I think you need to make `selectDefaultSingleAccount` a parameter of `selectAccount()` and pass `false` inside `addAndSelectAccount()`.

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorTest.java
briandealwis: The combo isn't being kept in any kind of order.  I'll make it sort alphabetically.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -60,16 +65,20 @@ public AccountSelector(Composite parent, IGoogleLoginService loginService,
     combo = new Combo(this, SWT.READ_ONLY);
     GridDataFactory.fillDefaults().grab(true, false).applyTo(combo);
 
-    for (Account account : loginService.getAccounts()) {
+    List<Account> sortedAccounts = new ArrayList<>(loginService.getAccounts());
+    Collections.sort(sortedAccounts, new Comparator<Account>() {
```
elharo: This would be straight-forward to test if it were an outer class.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -138,30 +162,24 @@ public void widgetSelected(SelectionEvent event) {
         Account account = loginService.logIn(null /* no custom dialog message */);
         if (account != null) {
           addAndSelectAccount(account);
-        } else {
-          combo.deselectAll();
+          // account is selected and selectedCredential saved
+          return;
         }
+        // XXX: shouldn't this re-select the previous account?
```
elharo: can we resolve this now?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -130,6 +140,20 @@ public int getAccountCount() {
     return combo.getItemCount() - 1;  // <Add a new account...> is always in the combo
   }
 
+  private void fireSelectionListeners() {
+    for (Object o : selectionListeners.getListeners()) {
+      ((Runnable) o).run();
```
elharo: You're not executing these in separate threads, so maybe Callable instead of Runnable?

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
```diff
@@ -35,6 +39,7 @@
   private IGoogleLoginService loginService;
   private String loginMessage;
   private Credential selectedCredential;
+  private ListenerList selectionListeners = new ListenerList();
```
elharo: why ListenerList instead of LIst<Runnable>?

briandealwis: It's thread safe.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
briandealwis: Huh, I wasn't aware of `Runnable` association with threading.  Given that we may re-use this `AccountSelector` elsewhere, I'll just create a proper `AccountSelectionListener`.

plugins/com.google.cloud.tools.eclipse.appengine.login/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelector.java
briandealwis: Sure.

briandealwis:
PTAL @elharo @chanseokoh

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -139,33 +141,35 @@ private void setupDataBinding() {
...
-            return ValidationStatus.error(Messages.getString("error.account.missing.signedin"));
-          } else {
-            return ValidationStatus.error(Messages.getString("error.account.missing.signedout"));
+    AccountSelectorObservableValue accountSelectorObservableValue =
+        new AccountSelectorObservableValue(accountSelector);
+    UpdateValueStrategy modelToTarget =
+        new UpdateValueStrategy().setConverter(new Converter(String.class, String.class) {
+          @Override
+          public Object convert(Object expectedEmail) {
+            // Expected to be an email address, but must check is a current account
```
elharo: check if it is a current account? a currently logged in account?

plugins/com.google.cloud.tools.eclipse.appengine.login.test/src/com/google/cloud/tools/eclipse/appengine/login/ui/AccountSelectorTest.java
```diff
@@ -208,6 +208,93 @@ public void testSelectAccount_nullOrEmptyEmail() {
...
+  }
+
+  @Test
+  public void testAutoSelectAccount_threeAccounts() {
+    when(loginService.getAccounts())
+        .thenReturn(new HashSet<>(Arrays.asList(account1, account2, account3)));
+
+    AccountSelector selector =
+        new AccountSelector(shell, loginService, "<select this to login>", true);
+    // should not select any element if more than 1 accounts
```
elharo: more than one account

Issue #656, opened 2016-09-01
Project Selector


elharo:
Where ever we ask for a project ID, we should consider implementing a popup that reads the available project IDs from the server and supplies them. 

We will need to filter out GCP projects that do not have App Engine applications attached (though per #505 we want to add the ability to attach a new App Engine application to an existing GCP project. We may want to design the UI of the two features at the same time.)

chanseokoh:
> We will need to filter out GCP projects that do not have App Engine applications attached

I believe the updated design doc suggests we show an indicator that a project doesn't have the application resources instead of filtering the project out.

patflynn:
+1.

elharo:
`gcloud projects list` returns all IDs, though it might be nicer to use the cloud resource manager API: https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/list

In any case this will go into appengine-plugins-core

patflynn:
On Tue, Jan 10, 2017 at 9:59 AM, Elliotte Rusty Harold <
notifications@github.com> wrote:

> gcloud projects list returns all IDs, though it might be nicer to use the
> cloud resource manager API: https://cloud.google.com/
> resource-manager/reference/rest/v1beta1/projects/list
>
> +1 don't use gcloud for things you can do easily from an available API.

> In any case this will go into appengine-plugins-core
>
If you do this you'll add the google client library dependency tree to our
maven/gradle plugins which is undesirable.

akerekes:
This issue will track the initial selector with minimal features. The selector will be a table widget displaying name+ID for each existing project ordered alphabetically on name.

akerekes:
- [ ] #1353 

- [ ] #1349 

- [ ] #1350 

- [ ] #1351 

- [ ] #1352 

- [ ] #505 

- [ ] #843

PR #966, opened 2016-11-14
fix #686

### LINKED ISSUE #686: Disable link in deploy dialog if there's no project ID
If there's no project ID set by the user, the link to App Engine versions page will not contain a project ID and the user is taken to the versions page of the last project he opened in the web console which may confuse the user.
We should probably disable the link until a project ID is set and a tooltip (on hover) should explain why it is disabled.

elharo:
@briandealwis @akerekes @chanseokoh A few things I noted while working on this bug fix:

1. It's much easier to test outer classes than inner classes and especially anonymous inner classes.

2. We may be overusing interfaces when they aren't called for. My rule of thumb is that if there aren't at least two non-test implementations, then you don't need an interface. I tried to start fixing that up in this PR but it got too complex so I reverted that work for now; but it's worth coming back to.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/ProjectIdQueryParameterProviderTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/ProjectIdQueryParameterProviderTest.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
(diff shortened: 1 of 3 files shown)


elharo:
We're not going to disable the link. It's fine for this link to go to the general versions page IMHO.

briandealwis:
> It's fine for this link to go to the general versions page IMHO.

The issue is that there is no general versions page at the moment: when a project is not specified, the console magically selects a project.

briandealwis:
> It's fine for this link to go to the general versions page IMHO.

But that's the point: there is no _general versions_ page.  Instead a project is magically selected.  But that's an issue that should be addressed by the Cloud Console, and will go away with a project selector (#656).

PR #968, opened 2016-11-15
rebuilt target platform

elharo:
@briandealwis the usual batch of apparently meaningless changes from rebuilding the target platform and creating a clean workspace. What should we do with this?

### DIFF
--- a/third_party/com.google.cloud.tools.eclipse.jdt.launching/.classpath
+++ b/third_party/com.google.cloud.tools.eclipse.jdt.launching/.classpath
@@ -2,8 +2,6 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src"/>
-	<classpathentry kind="src" path="tests"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/4"/>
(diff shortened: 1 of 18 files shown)


third_party/com.google.cloud.tools.eclipse.jdt.launching/.classpath
```diff
@@ -2,8 +2,6 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src"/>
-	<classpathentry kind="src" path="tests"/>
-	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/4"/>
```
briandealwis: This changes is "wrong": these entries are here to compile and bind the unit tests, but they're not specified in the MANIFEST.MF as the tests aren't to be added to the bundle.  PDE doesn't have a way to specify internal unit tests (e.g., like Maven's surefire and `src/test/java`).  Hmm, maybe we could teach it with something in the `build.properties`…

elharo:
I'm fine with not committing this. I just wanted to point out that we are churning when we  start a-fresh and then import the maven projects. That suggests to me that maybe all of this ought to be .gitignored instead if it's only code Eclipse generates every time we reimport.

elharo:
What do you think about .gitignoring all of this and removing the files from the repo? Maven should generate these for us automatically.

briandealwis:
I think it works for the natures that m2e knows about, but I'm not sure it would work for thinks like `org.eclipse.pde.ds.core.builder` (used in com.google.cloud.tools.eclipse.appengine.libraries).

PR #970, opened 2016-11-15
notice misconfigured Cloud SDK on deploy

### LINKED ISSUE #901: Check for Google Cloud SDK for deploy
Similar to #847 and #900, we need to check the Cloud SDK for deploy. (Probably anywhere before we use the Cloud SDK.)

If the Cloud SDK isn't found, we see a misleading error dialog. (This one bit me. I did `gcloud components install app-engine-java` numerous times and couldn't understand why this is failing.)

problem occurred _007

Steps to reproduce:
1) Update the Cloud SDK path in the preferences to an invalid directory.
2) Deploy.

elharo:
@nbashirbello @briandealwis 
fix #901


elharo:
We're better off autoinstalling our own Cloud SDK and not depending on the user having it, but until we get there this was the simplest place I could think of to do this. Maybe we could disable the menu item, but then it wouldn't be apparent to the user what the problem was.

briandealwis:
Hmm, thinking about it more, we could put it in `StandardDeployCommandHandler`.

PR #985, opened 2016-11-15
pop open console when deploying

### LINKED ISSUE #979: autoopen console on deploy and bring to front
Just maybe this already works, but find out.

elharo:
@briandealwis fix #979

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
@@ -36,6 +36,12 @@ import org.eclipse.core.runtime.jobs.IJobChangeEvent;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
 import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.window.Window;
+import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.IWorkbenchWindow;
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -142,6 +148,20 @@ public void done(IJobChangeEvent event) {
       }
     });
     deploy.schedule();
+    
+    // https://wiki.eclipse.org/FAQ_How_do_I_find_the_active_workbench_page%3F
+    IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
```
briandealwis: I've just updated that wiki page as this `PlatformUI.getWorkbench().getActiveWorkbenchWindow()` is actually an anti-pattern!

We have the `ExecutionEvent`, so just use:
```
IWorkbenchWindow window = HandlerUtil.getActiveWorkbenchWindow(event);
```

akerekes:
@briandealwis @elharo in MessageConsoleUtilities on line 76 we use `consoleManager.showConsoleView(messageConsole);` for the same effect. Are these equivalent? If so, I'd rather use `consoleManager` as it's a much shorter solution, and add it the the `createConsole` method that is used to create the DeployConsole in `StandardDeployCommandHandler`.

briandealwis:
`ConsoleManager#showConsoleView()` is almost equivalent — its implementation uses the `PlatformUI.getWorkbench().getActiveWorkbenchWindow()` (namely as `ConsoleManager` doesn't have any context to the appropriate window).

But in the interest of reducing code to maintain, I support its use!

elharo:
switched to ConsoleManager#showConsoleView()

PR #986, opened 2016-11-16
Add Webtools WST+JST dependences

### LINKED ISSUE #981: Restrict plugin to WTP
Figure out how to programmatically specify WTP as a prerequisite for our plugin. Ideally we'd like the WTP dependencies to be automatically installed along with our plugin if necessary. But failing that, it should not be possible to install our plugin into a version of Eclipse without WTP.

briandealwis:
With these dependencies, I can install our feature in the bare-bones Eclipse Platform (no JDT, no PDE), and create new projects and launch in the debugger with _Run on Server_/_Debug on Server_.

Fixes #981

### DIFF
--- a/features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
+++ b/features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
@@ -18,6 +18,16 @@
       %license.text
    </license>
 
+   <requires>
+      <import feature="org.eclipse.jdt" version="3.11.2.v20160212-1500"/>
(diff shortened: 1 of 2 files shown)


features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
```diff
@@ -18,28 +18,38 @@
       %license.text
    </license>
 
+   <requires>
+      <import feature="org.eclipse.jdt" version="3.11.2.v20160212-1500"/>
```
elharo: Is there a reason you don't have
 match="greaterOrEqual"
here?

briandealwis: Good catch.

PR #990, opened 2016-11-17
Add default index.html for gcp-repo repository

briandealwis:
Add default `index.html` that does an immediate meta-redirect to our [installation wiki page](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/)

### DIFF
--- a/gcp-repo/src/main/resources/index.html
+++ b/gcp-repo/src/main/resources/index.html
@@ -1,16 +1,14 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2//EN">
-<HTML>
-<HEAD>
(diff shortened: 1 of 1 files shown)


gcp-repo/src/main/resources/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2//EN">
```
elharo: Is there an Eclipse reason to use HTML 3.2? Google tends to recommend HTML 5 and personally I lean toward XHTML.

briandealwis: No reason at all — I just used a redirect.html skeleton I wrote years ago.

PR #993, opened 2016-11-17
don't need category.xml; have to make artifacts.jar later in process

elharo:
@briandealwis


kokoro/ubuntu/sign.sh
```diff
@@ -12,9 +12,7 @@ mkdir -p signed && chmod 777 signed
 mkdir -p signed/plugins && chmod 777 signed/plugins
 mkdir -p signed/features && chmod 777 signed/features
  
-cp artifacts.jar signed/artifacts.jar
 cp content.jar signed/content.jar
```
briandealwis: You can drop the `content.jar` too — the publishing step regenerates that too.

PR #996, opened 2016-11-19
Handle user answering No to CloudSDKPrompter in Maven Wizard

elharo:
@briandealwis also refactors out duplicate code

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/MissingComponentPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/MissingComponentPage.java
@@ -40,14 +40,14 @@ public class MissingComponentPage extends WizardPage {
    * @param pageName internal identifier
    * @param wizardType should be constant from AnalyticsEvents
    * @param title dialog title
-   * @param description short error message
+   * @param errorMessage short error message
    * @param message detailed message and instructions. Text between single quotes is made bold.
    */
-  public MissingComponentPage(String pageName, String wizardType, String title, String description,
-      String message) {
+  public MissingComponentPage(String pageName, String wizardType, String title,
+      String errorMessage, String message) {
     super(pageName);
     setTitle(title);
-    setErrorMessage(description);
+    setErrorMessage(errorMessage);
     this.wizardType = wizardType;
     this.message = message;
   }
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/MissingComponentPage.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Wizard page that displays a message that something needed in the environment is
+ * missing with instructions on how to install it. This page disables the 'Finish' button.
+ */
+public class MissingComponentPage extends WizardPage {
+
+  private String wizardType;
+  private String message;
+
+  public MissingComponentPage(String pageName, String wizardType, String title, String description,
```
briandealwis: Could you add a comment here to indicate that wizardType is expected to be one of our wizard types on `AnalyticsEvents`. 

Also worth describing that message bolds text between single quotes.  Which doesn't seem a great choice, since it will catch any contractions and possessives too?  Perhaps we should change those to be markdown-like backquotes?

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/MissingComponentPage.java
```diff
@@ -0,0 +1,83 @@
...
+      String message) {
+    super(pageName);
+    setTitle(title);
+    setDescription(description); 
+    this.wizardType = wizardType;
```
briandealwis: Can you add:
```
setErrorMessage(description);
```
That marks the page as looking having an error.
screen shot 2016-11-21 at 11 39 12 am

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/MissingComponentPage.java
elharo: Done. Agreed the bolding is weird. I hadn't even noticed that, just inherited it from the existing code. Perhaps address that in a subsequent PR.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/MissingComponentPage.java
elharo: done

elharo:
Huh, looks like Travis didn't fire. I wonder why.

PR #998, opened 2016-11-20
remove reference to no longer defined flex facet

### LINKED ISSUE #997: Cannot deploy due to missing flex facet
@nbashirbello 

I currently seem unable to deploy at head due to:

!MESSAGE Project facet com.google.cloud.tools.eclipse.appengine.facets.flex has not been defined. It is used in plugin com.google.cloud.tools.eclipse.appengine.facets.

Let's talk about this first thing Monday morning. We need to fix this ASAP and add a test that will keep this from reoccurring in the future.

elharo:
@nbashirbello fix #997

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
@@ -12,7 +12,6 @@
           <and>
             <requires facet="jst.web" version="2.5"/>
             <requires facet="java" version="1.7"/>
-            <conflicts facet="com.google.cloud.tools.eclipse.appengine.facets.flex"/>
           </and>
         </constraint>
     </project-facet-version>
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -12,7 +12,6 @@
           <and>
             <requires facet="jst.web" version="2.5"/>
             <requires facet="java" version="1.7"/>
-            <conflicts facet="com.google.cloud.tools.eclipse.appengine.facets.flex"/>
```
nbashirbello: Can we comment this section out or add a note to remind us to add this conflict statement back when we reenable the flex facet

elharo: Using comments as weak version control is an antipattern, IMHO. We should move the flex work to a separate branch until we're ready to pull it into head.

elharo:
@briandealwis since @nbashirbello is out

PR #1018, opened 2016-11-24
Guava 15 --> 17

elharo:


elharo:
@briandealwis @akerekes We can try moving to Guava 20 next, but first it would be nice to get us on a single version.

PR #1022, opened 2016-11-25
fix bad use of static builder method

elharo:
@briandealwis error in ModuleUtils found by static analysis looks like a potentially serious bug

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.eclipse.appengine.localserver.server;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.wst.server.core.IModule;
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import com.google.cloud.tools.eclipse.appengine.localserver.server.ModuleUtils;
+
+public class ModuleUtilsTest {
+
+  @Test
+  public void testGetServiceId_null() {
+    try {
+      ModuleUtils.getServiceId(null);
+      Assert.fail();
+    } catch (NullPointerException expected) {
+    }
+  }
+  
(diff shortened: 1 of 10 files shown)


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/AppEngineDescriptor.java
```diff
@@ -71,7 +76,7 @@ public String getProjectVersion() throws CoreException {
   }
 
   /**
-   * @return service ID parsed from the &lt;application&gt; element of the appengine-web.xml, or
+   * @return service ID parsed from the &lt;service&gt; element of the appengine-web.xml, or
```
briandealwis: oops, good catch.  It should also have `<module>` for backwards compatibility.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class ModuleUtilsTest {
+
+  @Before
+  public void turnOffLogging() {
+    Logger logger = Logger.getLogger(ModuleUtils.class.getName());
+    logger.setLevel(Level.OFF);
```
briandealwis: We should restore the logging level, no?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtils.java
```diff
@@ -35,22 +34,22 @@
   private static final Logger logger = Logger.getLogger(ModuleUtils.class.getName());
 
   /**
-   * Retrieve the &lt;service&gt; or &lt;module&gt; identifier from the <tt>appengine-web.xml</tt>.
-   * If an identifier is not found, or <tt>appengine-web.xml</tt> is not found, then returns
-   * "default".
+   * Retrieve the &lt;service&gt; or &lt;module&gt; identifier from <tt>appengine-web.xml</tt>.
+   * If an identifier is not found, then returns "default".
+   * If <tt>appengine-web.xml</tt> is not found, then returns null.
```
briandealwis: The original implementation was wrong, it is true.  My intent was for this method to be `@NonNull`.

But my choice here, to return _"default"_ if the file didn't exist, was deliberate: if you want the untrammelled information, use `AppEngineDescriptor` directly.  In practice our `canModify()` first verifies all modules have an App Engine facet, which should ensure they have an `appengine-web.xml`.

If you really think we should return `null` here, then we need to fix up `LocalAppEngineServerDelegate` too.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
elharo: We can, though honestly in a test suite I'm tempted to leave it off. zAs is, it just clutters the logs.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtils.java
elharo: I don't think we should return null, but that's what I noticed the current implementation doing and I wasn't sure if that was a mistake or not. I'll update it.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/AppEngineDescriptor.java
elharo: It does look at the module elements as a backup.

PR #1026, opened 2016-11-26
Remove hamcrest from .3rdparty.feature

### LINKED ISSUE #988: Hamcrest jars in release
Our release build bundles in 3 hamcrest jars. These should only be needed for tests, not for the production code. Can we get rid of them?

briandealwis:
Fixed #988


elharo:
FYI, I wasn't thinking and merged this patch rather than approving it. The record should still be clear that you did the work.

PR #1032, opened 2016-11-28
Handle when a module is possibly added to a server with the module

### LINKED ISSUE #1029: fred and fred have same app engine service ID: default
This is at head. 

image

Not sure how I got into this state.

briandealwis:
Fixes #1029


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegate.java
```diff
@@ -133,6 +133,10 @@ IStatus checkConflictingServiceIds(IModule[] current, IModule[] toBeAdded,
     }
     if (toBeAdded != null) {
       for (IModule module : toBeAdded) {
+        if (currentServiceIds.containsValue(module)) {
```
elharo: what's the equals() implementation on a module? The docs are explicitly unclear about this  http://www.eclipse.org/webtools/wst/api/org/eclipse/wst/server/core/IModule.html but if it's just checking the IDs, this would seem to render the entire loop moot.

PR #1033, opened 2016-11-28
Add M2E SDK feature to required targets

briandealwis:


briandealwis:
Oops, forgot `oxygen`.

elharo:
Is this a subset or a superset of the m2e feature? Or neither?

briandealwis:
Hmm thanks for asking: I thought it was a superset, since everything resolves fine, but looking at the definition only brings in the `.source` bundles.  The `m2e-wtp` feature brings in most of m2e, but I'll keep the explicit m2e feature to be on the safe side.

briandealwis:
Sorry, missed the `category.xml` @elharo

PR #1039, opened 2016-11-29
better error message for cloud SDK not found in configured directory

### LINKED ISSUE #1034: Error messages for Cloud SDK directory are confusing
At least on Windows:

1. Missing directory is "Invalid directory". Should be "No such directory" or some such. The problem is not that it's invalid, whatever that means. It's that it's not there.

2. No SDK Found: Validation error ... \bin\gcloud.cmd is not a file
is too long to fit in the area. Plus it's implementation specific and means nothing to end user. Tell them more directly that directory is not a cloud sdk install. Cloud SDK is not installed in dirName.

elharo:
@briandealwis fix #1034

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
@@ -203,7 +203,7 @@ public class CloudSdkPreferenceArea extends PreferenceArea {
         return false;
       }
       status = Status.OK_STATUS;
-      return getStringValue().isEmpty() || validateSdk(Paths.get(getStringValue()));
+      return directory.isEmpty() || validateSdk(Paths.get(directory));
     }
   }
 }
--- a/plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
@@ -196,8 +196,10 @@ public class CloudSdkPreferenceArea extends PreferenceArea {
 
     @Override
     protected boolean doCheckState() {
+      String directory = getStringValue();
       if (!super.doCheckState()) {
-        status = new Status(IStatus.ERROR, getClass().getName(), "Invalid directory");
+        String message = MessageFormat.format(SdkUiMessages.NoSuchDirectory, directory);
+        status = new Status(IStatus.ERROR, getClass().getName(), message);
         return false;
       }
       status = Status.OK_STATUS;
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -193,8 +196,10 @@ public void setFilterPath(File path) {
...
     @Override
     protected boolean doCheckState() {
+      String directory = getStringValue();
       if (!super.doCheckState()) {
-        status = new Status(IStatus.ERROR, getClass().getName(), "Invalid directory");
+        String message = MessageFormat.format(SdkUiMessages.NoSuchDirectory, directory);
+        status = new Status(IStatus.ERROR, getClass().getName(), message);
         return false;
       }
       status = Status.OK_STATUS;
```
chanseokoh: I think you can replace two instances of `getStringValue()` in the next line with the variable `directory`.

elharo: done

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -193,12 +196,14 @@ public void setFilterPath(File path) {
 
     @Override
     protected boolean doCheckState() {
+      String directory = getStringValue();
       if (!super.doCheckState()) {
-        status = new Status(IStatus.ERROR, getClass().getName(), "Invalid directory");
+        String message = MessageFormat.format(SdkUiMessages.NoSuchDirectory, directory);
```
briandealwis: This happens if the user specifies a file, so "No Such Directory" isn't really appropriate.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -160,19 +161,22 @@ private boolean validateSdk(Path location) {
     } catch (AppEngineException ex) {
       // accept a seemingly invalid location in case the SDK organization
       // has changed and the CloudSdk#validate() code is out of date
+      // todo: does this really make sense? In that case, wouldn't all the
```
briandealwis: `appengine-plugins-core` checks for [various path locations](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/appengine/cloudsdk/CloudSdk.java#L69) whereas we just use `gcloud` and `dev_appserver`.

elharo: I took the todo out of this PR, though I'm still not convinced this shouldn't be a hard error. It comes down to probabilities. What's more likely? That the user has not installed the app engine Java components or that the layout of the Cloud SDK has changed in a way that throws th exception but doesn't break our functionality?

PR #1043, opened 2016-12-01
upgrade to appengine-plugins-core 0.2.1

elharo:
@chanseokoh


elharo:
FYI this is failing because the mock SDK path in CloudSdkContextFunctionTest doesn't pass the new version check.

chanseokoh:
Tests seem to keep failing, and I wonder if this is related.

elharo:
PTAL @chanseokoh

PR #1053, opened 2016-12-05
Bump guava to 20.0, appengine-plugins-core to 0.2.3

### LINKED ISSUE #943: Upgrade Guava to 20
We use Guava 17 in `com.google.cloud.tools.eclipse.3rdparty.feature`, we should update to the latest unless this breaks something.

briandealwis:
Fixes #943


chanseokoh:
Don't we need to update MANIFEST files as in #1018, since 20 is the minimum requirement?

briandealwis:
You're right, we should.  And ditto for `c.g.c.t.appengine` too.

chanseokoh:
FYI, #1043 too.

PR #1064, opened 2016-12-06
Include intro/ directory in binary build

chanseokoh:
I think we need to include `intro/` (icon, CSS, and Welcome link contents) in the binary build, right?


elharo:
I'm not sure. Install the beta candidate I just posted and see if the resources are there.

chanseokoh:
Yeah, I tested it, and the resources are not there.

PR #1067, opened 2016-12-07
In dev-mode, default to the normal OAuth client values

briandealwis:


chanseokoh:
I think it is extremely unlikely that someone will run Eclispe in the dev mode, so perhaps it's okay to leave it as-is? No strong opinion here.

briandealwis:
After some further thinking, I can't think of a reason why someone would want to change the values at runtime, nor whether it's something we really want to support.  Closing.

PR #1075, opened 2016-12-08
Cannot install CT4E if GPE is already installed and vice versa

nbashirbello:
Fix for #866


nbashirbello:
Not ready for review.

nbashirbello:
Ready for review

nbashirbello:
In GPE, we would have to add a notification that will be displayed on start-up that the GPE plugin has been replaced with CT4E (listing the features CT4E has and doesn't have). Users would be able to mute the notification permanently.

briandealwis:
GPE has some notification system — I remember seeing something when there was a new update. We might be able to leverage that.

PR #1082, opened 2016-12-08
Bump appengine-plugins-core 0.2.4

elharo:


chanseokoh:
One more file: `plugins/com.google.cloud.tools.eclipse.integration.appengine/META-INF/MANIFEST.MF`

Issue #1080, opened 2016-12-08
App Engine Project in File > New menu


elharo:
Can we/should we put "App Engine Project" and "Maven App Engine Project" in the File > New submenu?

![image](https://cloud.githubusercontent.com/assets/1005544/21014924/d5e8fec4-bd2d-11e6-952c-79bd31eaa81c.png)

branflake2267:
I like the idea. That menu also shows up on right click of project too. 

![screen shot 2016-12-08 at 12 59 57 pm](https://cloud.githubusercontent.com/assets/1326504/21027592/4ebf89a4-bd46-11e6-8c7a-5884cbef530a.png)

briandealwis:
`org.eclipse.ui.perspectiveExtensions` for the win.

PR #1087, opened 2016-12-08
Add new project wizard extensions

briandealwis:
Adds the native and maven projects to the _New_ menu.  Note that, you'll need to reset your perspective to pick up these changes if you're running from an existing workspace/window.

screen shot 2016-12-08 at 4 22 00 pm
.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/plugin.xml
@@ -43,6 +43,15 @@
          </exclude>
       </validator>
    </extension>
+   <extension
+         point="org.eclipse.ui.perspectiveExtensions">
+      <perspectiveExtension
+            targetID="*">
+         <newWizardShortcut
+               id="com.google.cloud.tools.eclipse.appengine.newproject.MavenAppEngineStandard">
+         </newWizardShortcut>
+      </perspectiveExtension>
+   </extension>
    
   <extension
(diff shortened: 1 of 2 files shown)


briandealwis:
This adds the extension to all perspectives. It might be better to add it to the _Java_, _Java EE_, and _Web_ perspectives.

chanseokoh:
Good to know how to add these. Will these available in any perspective? Lastly, one concern of mine is, we may end up adding too many wizards, e.g., flex env, Gradle-based, etc.

BTW, the XML has some Git stash markers.

briandealwis:
Oh geesh. Github needs a :blush: reaction.

branflake2267:
Couple random notes from a random user's angle. 

* It might be a bit greedy to be in all perspectives. For instance, as a Java user, I'm in the Java perspective most of the time and I think it's one of those features that isn't obscure to Eclipse users. 

* Regarding labeling. I might suggest something like moving Maven next to Java, instead of Maven based. Or better yet, only use one shortcut and then specify in the dialog specify which build type. I like it either way, but looking into the future, if you add more build types might get busy. 
 - Google App Engine Standard Java Project
 - Google App Engine Standard Java Maven Project
 - [Google App Engine Standard Java Gradle|Ant|Blaze Project]

briandealwis:
Restricted the set of perspectives extended.  

Opened #1094 to consider new project wizard naming.

briandealwis:
Fixes #1080

PR #1093, opened 2016-12-09
Avoid running tests with SDK features

briandealwis:
Avoid spurious messages in our tests like the following by replacing SDK features with runtime features:
```
!ENTRY org.eclipse.equinox.registry 2 0 2016-12-09 11:37:52.205
!MESSAGE The extensions and extension-points from the bundle "org.eclipse.wst.common.infopop.source" are ignored. The bundle is not marked as singleton.

!ENTRY org.eclipse.equinox.registry 2 0 2016-12-09 11:37:52.205
!MESSAGE The extensions and extension-points from the bundle "org.eclipse.wst.validation.infopop.source" are ignored. The bundle is not marked as singleton.

!ENTRY org.eclipse.equinox.registry 2 0 2016-12-09 11:37:52.345
!MESSAGE The extensions and extension-points from the bundle "org.eclipse.wst.command.env.infopop.source" are ignored. The bundle is not marked as singleton.
```

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/pom.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/pom.xml
@@ -36,22 +36,22 @@
               </requirement>
               <requirement>
                 <type>eclipse-feature</type>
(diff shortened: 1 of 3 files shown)


chanseokoh:
Nice!

PR #1113, opened 2016-12-13
verify keywords we reference are defined

### LINKED ISSUE #206: Check keywordReferences
In BasePluginXmlTest check that keywordReference elements in plugin.xml points to an actually defined keyword.

elharo:
fix #206 @chanseokoh @briandealwis


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -54,13 +57,29 @@ public void testValidExtensionPoints() {
     
     // todo should we test that the file has at least one extension point?
     
+    IExtensionRegistry registry = RegistryFactory.getRegistry();
```
briandealwis: Might be worth pulling this out to a field.

PR #1136, opened 2016-12-15
show enter a project name on new wizard

### LINKED ISSUE #1117: Enter a project name.
Per Eclipse [guideline 5.3 "Start a wizard with a prompt, not an error message."](https://wiki.eclipse.org/User_Interface_Guidelines#Initialization_2)

When this dialog first appears the message should be "Enter a project name.", not "Create a new App Engine Standard environment project in the workspace." cf. the new Java Project wizard. Note that this is not an error message. It's Info, not warning or error. 

image

elharo:
fix #1117 @briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
@@ -12,6 +12,7 @@ Require-Bundle: org.junit;bundle-version="4.12.0",
  org.hamcrest;bundle-version="1.1.0"
 Import-Package: com.google.cloud.tools.eclipse.test.util,
  com.google.cloud.tools.eclipse.test.util.project,
+ com.google.cloud.tools.eclipse.test.util.ui,
  com.google.common.io;version="20.0.0",
  org.eclipse.core.resources,
  org.mockito;provider=google;version="1.10.19",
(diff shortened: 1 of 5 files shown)


chanseokoh:
Tests will not succeed since the tests are running in a non-UI thread: `org.eclipse.swt.SWTException: Invalid thread access`.

I know one way to fix this, by flipping the `<useUIThread>` setting in the `.appengine.newproject.test/pom.xml`.

```java
      <plugin>
        <artifactId>tycho-surefire-plugin</artifactId>
        <configuration>
          <useUIHarness>true</useUIHarness>
          <useUIThread>false</useUIThread>
          <product>org.eclipse.platform.ide</product>
        </configuration>
      </plugin>
```

elharo:
Anyone see why the tests are failing with invalid thread access? I know it's the new Shell cal but why is this test different than, for example, the one in LocalAppEngineServerWizardFragmentTest?


testCreateControl(com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandardWizardPageTest)  Time elapsed: 0.001 sec  <<< ERROR!
org.eclipse.swt.SWTException: Invalid thread access
	at org.eclipse.swt.SWT.error(SWT.java:4491)
	at org.eclipse.swt.SWT.error(SWT.java:4406)
	at org.eclipse.swt.SWT.error(SWT.java:4377)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:482)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:266)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:257)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:208)
	at org.eclipse.swt.widgets.Shell.<init>(Shell.java:144)
	at com.google.cloud.tools.eclipse.appengine.newproject.AppEngineStandardWizardPageTest.testCreateControl(AppEngineStandardWizardPageTest.java:64)

chanseokoh:
See the comment above if you've missed it.

elharo:
That seems likely. We had   &lt;useUIThread>false&lt;/useUIThread>
Testing now. I wonder if we should split our test suites into those that run on the UIThread and those that don't?

chanseokoh:
You could revert back to use `ShellTestResource`, and [`<useUIThread>` is true by default](https://eclipse.org/tycho/sitedocs/tycho-surefire/tycho-surefire-plugin/test-mojo.html#useUIThread), so that can also be removed (like in other test bundles).

Issue #1100, opened 2016-12-10
Alternate locations for webapp


elharo:
The standard facet assumes a certain layout of

src/main/webapp

for non-java resources.

This is not the layout used by all projects, and it's not the layout used  by GPE by default. 

We need some way to 

1. Configure different directories as our webapp folder.
2. Ideally recognize these alternate locations when importing/converting an existing project.

elharo:
FYI this is the code from AppEngineStandardFacet  that unconditionally sets "src/main/webapp" as the CONFIG_FOLDER:

  private static void installWebFacet(IFacetedProject facetedProject, IProgressMonitor monitor)
      throws CoreException {
    if (facetedProject.hasProjectFacet(WebFacetUtils.WEB_25)) {
      return;
    }

    IDataModel webModel = DataModelFactory.createDataModel(new WebFacetInstallDataModelProvider());
    webModel.setBooleanProperty(IJ2EEModuleFacetInstallDataModelProperties.ADD_TO_EAR, false);
    webModel.setBooleanProperty(IJ2EEFacetInstallDataModelProperties.GENERATE_DD, false);
    webModel.setBooleanProperty(IWebFacetInstallDataModelProperties.INSTALL_WEB_LIBRARY, false);
    webModel.setStringProperty(IWebFacetInstallDataModelProperties.CONFIG_FOLDER, "src/main/webapp");
    facetedProject.installProjectFacet(WebFacetUtils.WEB_25, webModel, monitor);
  }

elharo:
FYI, I do not see anyway to configure this from the Eclipse UI to change it after initial creation.

elharo:
When the facet is added the code enters StandardFacetInstallDelegate.execute so this is where we can hook in.

public class StandardFacetInstallDelegate extends AppEngineFacetInstallDelegate {
  private final static String APPENGINE_WEB_XML = "appengine-web.xml";

  @Override
  public void execute(IProject project,
                      IProjectFacetVersion version,
                      Object config,
                      IProgressMonitor monitor) throws CoreException {
    super.execute(project, version, config, monitor);
    createConfigFiles(project, monitor);
  }

The config object here seems to be null.

chanseokoh:
Even for Java, I think the src/main/... structure is ostensibly Maven.

elharo:
The UI for the Web facet does allow the selection of the content folder, though I still can't figure out how to change that after the web facet is applied through "Further configuration available...":

[...]

Issue #1101, opened 2016-12-11
Overlapping files in deploy


elharo:
This one bit me on a personal web site. I had multiple web content folders mapped in my deployment assembly. Each contained an index.html file. The wrong one got uploaded.

We should recognize and warn the user when they're mapping two different files to the same URL.

chanseokoh:
See https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1100#issuecomment-266801517

briandealwis:
This seems to be a general issue for WTP.

PR #1137, opened 2016-12-15
Pseudo-code for discussions

chanseokoh:
This PR is to have discussions about https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1100#issuecomment-267369938.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
@@ -26,7 +26,6 @@ import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
-import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.Job;
@@ -46,28 +45,37 @@ public class StandardFacetInstallDelegate extends AppEngineFacetInstallDelegate 
                       Object config,
                       IProgressMonitor monitor) throws CoreException {
     super.execute(project, version, config, monitor);
-    // If Dynamic Web facet was additionally installed together, it will have created "WebContent/"
-    // as the default webapp directory (unless the user changed the default when installing
-    // the Web facet). We detect if there already existed another webapp directory, and if so,
-    // fix the default webapp setting to point to the existing directory.
-    fixWebContentRootIfNecessary(project, monitor);
 
+    fixWebContentRootIfNecessary(project, monitor);
     createConfigFiles(project, monitor);
     installAppEngineRuntimes(project);
   }
 
+  /**
+   * Deletes the incorrect "/WebContent" (attributed to the fault of the Dynamic Web Module facet)
+   * and fixes the default webapp setting accordingly.
+   *
+   * If the Dynamic Web Module facet was installed together, it will have created the "/WebContent"
+   * folder with the default webapp structure (if the user did not change the default folder name
+   * when selecting the Web facet to install). This method detects if the project already had
+   * another genuine webapp directory, and if so, fixes the default webapp project setting
+   * to point to the existing webapp.
+   */
   private void fixWebContentRootIfNecessary(IProject project, IProgressMonitor monitor)
       throws CoreException {
-    if (WebProjectUtil.isDefaultWebContentFolderBogus(project)) {
-
-      // Fix/update webapp setting in ".settings/org.eclipse.wst.common.component" (XML). How? To figure out...
-      IVirtualComponent component = ComponentCore.createComponent(project /* of IProject */);
+    if (WebProjectUtil.hasBogusWebRootFolder(project)) {
+      IVirtualComponent component = ComponentCore.createComponent(project);
       if (component != null && component.exists()) {
         IVirtualFolder rootFolder = component.getRootFolder();
-        rootFolder.removeLink(new Path("/WebContent"), IVirtualFolder.FORCE, monitor);
+        // This removes the following entry in ".settings/org.eclipse.wst.common.component":
+        // <wb-resource deploy-path="/" source-path="/WebContent" tag="defaultRootSource"/>
+        rootFolder.removeLink(WebProjectUtil.DEFAULT_DYNAMIC_WEB_FACET_WEB_CONTENT_PATH,
+                              IVirtualFolder.FORCE, monitor);
       }
 
-      // Delete "WebContent" folder. (Skip this? But it's very confusing that this folder exists.)
+      // Delete "/WebContent" physically.
+      project.getFolder(WebProjectUtil.DEFAULT_DYNAMIC_WEB_FACET_WEB_CONTENT_PATH)
+        .delete(true /* force */, monitor);
     }
   }
 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
@@ -34,7 +34,7 @@ import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
 import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
 
 public class StandardFacetInstallDelegate extends AppEngineFacetInstallDelegate {
-  private final static String APPENGINE_WEB_XML = "appengine-web.xml";
+  private static final String APPENGINE_WEB_XML = "appengine-web.xml";
 
   @Override
   public void execute(IProject project,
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
```diff
@@ -34,18 +34,32 @@
...
+    // fix the default webapp setting to point to the existing directory.
+    fixWebContentRootIfNecessary(project);
+
     createConfigFiles(project, monitor);
     installAppEngineRuntimes(project);
   }
 
+  private void fixWebContentRootIfNecessary(IProject project) {
+    if (WebProjectUtil.isDefaultWebContentFolderBogus(project)) {
+      Fix/update webapp setting in ".settings/org.eclipse.wst.common.component" (XML). How? To figure out...
```
elharo: can we do this by setting the relevant facet property as we do now in the new project wizard? That is, 



    IDataModel webModel = DataModelFactory.createDataModel(new WebFacetInstallDataModelProvider());
    webModel.setBooleanProperty(IJ2EEModuleFacetInstallDataModelProperties.ADD_TO_EAR, false);
    webModel.setBooleanProperty(IJ2EEFacetInstallDataModelProperties.GENERATE_DD, false);
    webModel.setBooleanProperty(IWebFacetInstallDataModelProperties.INSTALL_WEB_LIBRARY, false);
    webModel.setStringProperty(IWebFacetInstallDataModelProperties.CONFIG_FOLDER, "src/main/webapp");
    facetedProject.installProjectFacet(WebFacetUtils.WEB_25, webModel, monitor);

chanseokoh: I don't think we can. For the new project wizard, it is us that freshly installs the Web facet programmatically with the code above. When a user is adding a facet (both the App Engine and Web facets) to an existing project, it is WTP that's doing the that. WTP sets "WebContent" by default.

chanseokoh: Or, maybe we could explicitly uninstall the Web facet and reinstall it by ourselves as a workaround?

chanseokoh: Okay, I tested calling `facetedProject.installProjectFacet(WebFacetUtils.WEB_25, ...)` twice consecutively in the new project wizard, but unfortunately, it throws an `IllegalArgumentException`. Calling this after uninstalling the facet works. However, we can't uninstall the Web facet in our delegate, since the Web facet is a pre-requisite.

Nice try, but seems like we can't update the properties in ".settings/org.eclipse.wst.common.component" this way.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -70,4 +92,125 @@ public static IFile findInWebInf(IProject project, IPath filePath) {
     return file.exists() ? file : null;
   }
 
+  private static List<IFolder> findAllWebInfFolders(IContainer container) {
+    List<IFolder> webInfFolders = new ArrayList<>();
```
briandealwis: Probably best to use `IResourceVisitor` / `IResource#accept()`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -70,4 +92,125 @@ public static IFile findInWebInf(IProject project, IPath filePath) {
...
+  }
+
+  /**
+   * Returns the web content path set by the Dynamic Web Module facet. May return {@code null}.
+   *
+   * The path information resides in an XML file, which a project settings file:
+   * ".settings/org.eclipse.wst.common.component". This methods effectively returns the path
+   * tagged with {@code "defaultRootSource"} in the XML file.
+   */
+  private static IPath getWtpWebContentPathSetting(IProject project) {
```
briandealwis: `getDefaultRootFolder()`?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -70,4 +92,125 @@ public static IFile findInWebInf(IProject project, IPath filePath) {
...
+        ITaggedVirtualResource resource = (ITaggedVirtualResource) rootFolder;
+        IPath defaultRootSource = resource.getFirstTaggedResource(
+            WorkbenchComponent.DEFAULT_ROOT_SOURCE_TAG /* "defaultRootSource" String literal */);
+        return defaultRootSource;
+      }
+    }
+    return null;
+  }
+
+  public static boolean isDefaultWebContentFolderBogus(IProject project) {
```
briandealwis: `hasBogusRootFolders()`?

briandealwis:
From a code-structure perspective, I don't think we should have these helper methods on `WebProjectUtil` since they're pretty specific to a particular issue.

But stepping back, I don't think our App Engine Standard facet installation is the right place to do this. 

First, I don't think we should do this fixup automatically as we may guess wrong.  At best, we could launch a `UIJob` that would prompt the user to say _it appears this project is misconfigured with multiple WEB-INF directories_ and offer to fix it.

But maybe we should instead have a builder to identify this problem, and a quick-fix to offer a workflow to fix the solution.

How will this situation occur?  From importing an existing non-WTP based GPE project, right?  It won't arise from creating a new project, or adding the AES facet to an existing project.  So can't we fix this with a better importer?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
chanseokoh: Singular (`...Folder`), I think. We want to detect if the Dynamic Web facet created a very specific directory `/WebContent`.

chanseokoh:
> It won't arise from creating a new project, or adding the AES facet to an existing project.

This happens whenever you add the Dynamic Web facet to an existing project. Adding the facet will always create "/WebContent" and make it the default. So, adding the facet to non-WTP projects (e.g., migrating GPE-generated projects) will encounter this. You can quickly check this even with our App Engine projects — try removing the Web facet and adding the facet back.

chanseokoh:
> From a code-structure perspective, I don't think we should have these helper methods on WebProjectUtil since they're pretty specific to a particular issue.

Make sense. We can refactor that later, no matter which solution we decide to use.

> At best, we could launch a `UIJob` that would prompt the user to say _it appears this project is misconfigured with multiple WEB-INF directories and offer to fix it._

That could be another solution, but my concern is that it is too late. For example, adding the Web and App Engine facets to an existing project may creates `appengine-web.xml` under `/WebContent/WEB-INF`, instead of `/src/main/webapp/WEB-INF/appengine-web.xml`.

BTW, I really have no idea, but having multiple WEB-INF directories itself isn't a problem? Users could selectively specify certain directories to assemble as a deployment, or a project may have multiple J2EE (um, or WTP?) modules?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
chanseokoh: I figured out how to update the default webapp setting in `.settings/org.eclipse.wst.common.component`, so we are good to go.

briandealwis:
I understand how `WebContent` comes about, but we can't distinguish between the cases of someone installing our AES facet into an existing DWP vs the DWP implicitly added as they added our AES.

Could we add a configuration widget, like DWP, and allow the user to instruct us to reconcile the different folders?

> BTW, I really have no idea, but having multiple WEB-INF directories itself isn't a problem? Users could selectively specify certain directories to assemble as a deployment, or a project may have multiple J2EE (um, or WTP?) modules?

That's a good question — it shouldn't make a difference except when there are conflicting files (which is the situation that bit @elharo in #1101).  But it can be puzzling for the user.

chanseokoh:
> Could we add a configuration widget, like DWP, and allow the user to instruct us to reconcile the different folders?

Do you mean popping up a confirmation dialog in `StandardFacetInstallDelegate.execute()` if it seems to need a fix? I think it's a good idea. Or, are you suggesting to add the widget somewhere else? In any case, I'd like to make sure that the users always go through this to have a chance to fix it. 

> but we can't distinguish between the cases of someone installing our AES facet into an existing DWP vs the DWP implicitly added as they added our AES.

We can try our best guess (and I think we are nevertheless forced to do so here to mitigate the issue). What I've been doing is like, if a project has more than one WEB-INF folder, one of which is `/WebContent` whose layout and contents are exactly like below (no extra file allowed), and `web.xml` contains no element.

```
/WebContent
    /WEB-INF
        /lib
        web.xml
    /META-INF
        MANIFEST.MF
```

chanseokoh:
> Do you mean popping up a confirmation dialog in StandardFacetInstallDelegate.execute() if the project seems to need a fix?

Or, another idea is to always show a configuration widget in the Delegate to enter the default webapp folder if there exist multiple WEB-INF folders. That way, we don't even need to check if `/WebContent` is bogus. It's just a fault of the Dyn Web facet by-product, and we don't care. How about that?

briandealwis:
A configuration widget for our facet ("Further configuration available…") is what I meant.  I think that's the best solution.

screen shot 2016-12-16 at 1 22 16 pm

chanseokoh:
To my understanding, the main goal of the issue #1100 is to simplify the [GPE migration story](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/wiki/Migration-from-the-Google-Plugin-for-Eclipse-%28GPE%29), with some automation. Currently, when migrating GPE projects, there is a major friction: following the instructions will create the bogus `/WebContent` directory, set it to the default webapp, and place `appengine-web.xml` into `/WebContent/WEB-INF`. To fix this, users have to jump through some additional hoops.

Making the users take additional actions (i.e., clicking the "Further configuration available…" and fixing the webapp setting before adding the facets) and adding these instructions seems to further increase the migration steps. Moreover, it's easy for users to neglect these "optional" actions, and they will see `appengine-web.xml` placed in a weird place.

Perhaps we need more brainstorming.

briandealwis:
> To my understanding, the main goal of the issue #1100 is to simplify the GPE migration story, with some automation

I don't think it's out of line to tell them to click on the _Further configuration available_ link and specify `path/to/xxx`?  Especially since that path will be fixed for GPE projects, right?

And I believe there are some people using GPEs WTP support.

chanseokoh:
#1100 closed.

PR #1141, opened 2016-12-16
Add Run As/Debug As support

### LINKED ISSUE #225: Add shortcut to Run on Appengine instead of Run on Server ...
This works, now (with the expanded archetype.)

I'd like to make this a higher level menu item though. However:

http://stackoverflow.com/questions/10504620/add-new-maven-run-configuration-to-menu-in-eclipse

Perhaps we can create a completely new menu item such as "Run Local App Engine" that just invokes the relevant maven command?

Also, debugging this doesn't seem to work (breakpoints not hit). If we go this route we need to figure out how to attach a debugger.

briandealwis:
This patch adds support for _Run As_/_Debug As_ launch shortcuts for the local App Engine development server.  Some of the text shown to the UI needs feedback/refinement.

I've backed out the menu contributions to the popup menu (which added a _Debug on App Engine_ options), but kept the launch command/handlers as they will be useful for an GPE-like toolbar drop-down.

Fixes #225

----
_Run As_/_Debug As_ from a selection of projects:
screen shot 2016-12-16 at 3 46 05 pm

Clicking on the _Debug_ toolbar button when selecting a project or on a file in the editor:

screen shot 2016-12-16 at 3 44 49 pm

----

I added a check to our `LocalAppEngineServerLaunchConfigurationDelegate#launch()` to check that the server is in a stopped state to prevent double-launches.  Launching an already-running server results in an error message:
screen shot 2016-12-16 at 3 35 14 pm

----

I added some support in `LocalAppEngineServerBehaviour#setupLaunchConfiguration()` to record a the projects corresponding to the server `IModule`s.  This isn't strictly necessary as we implement `ILaunchShortcut2` (see below), but it's probably good form.

----

The launch shortcut implementation reuses existing server definitions where possible, including a server which has other modules (e.g., you select modules A and B and click _Debug As_, then we may choose a server defined with A, B, and C). Reusing existing servers is helpful as:
    1. it's confusing to have multiple servers appear, and will become more so once we support per-server parameters,
    2. It's also helpful as the Run/Debug tooltips show details about the server.
screen shot 2016-12-16 at 3 33 49 pm

----

### Background

I had to peer deeply into the launch code to understand and work around some seemingly pathological behaviour that arises from the odd relationship between WTP's `IServer` and Debug's `ILaunchConfigurations` and `ILaunch`.  `ILaunchConfiguration` normally describes *how* to launch something, and can usually be launched multiple times; each launch results in an `ILaunch` instance.  But each WTP `IServer` instance is tightly associated with a `ILaunchConfiguration` too.  So launching a server's `ILaunchConfiguration` when that server was already started elsewhere results in really strange behaviour as the server's state is changed — and thus starts or stops the previously running server.  Thus the change in `LocalAppEngineServerLaunchConfigurationDelegate#launch()` to ensure the server is not running.

The Debug Launch framework provides two interfaces to implement these _Run As_/_Debug As_ launch shortcuts: `ILaunchShortcut` and `ILaunchShortcut2`.  With plain `ILaunchShortcut`, the framework chooses an existing configuration based on the associated resources (hence the fix to `setupLaunchConfiguration()`), and the `ILaunchShortcut` is only consulted when no matching launch configuration can be found.  `ILaunchShortcut2` provides more control over how an `ILaunchConfiguration` is chosen from existing `ILaunchConfiguration`s.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
@@ -223,6 +223,37 @@
       </fragment>
    </extension>
    <extension
+         point="org.eclipse.ui.menus">
+      <menuContribution
+            allPopups="false"
+            locationURI="popup:org.eclipse.ui.popup.any?after=additions">
+         <command
+               commandId="com.google.cloud.tools.eclipse.appengine.localserver.launch"
+               disabledIcon="platform:/plugin/org.eclipse.debug.ui/icons/full/dtool16/run_exc.png"
+               icon="platform:/plugin/org.eclipse.debug.ui/icons/full/etool16/run_exc.png"
+               label="Run in App Engine"
+               mnemonic="R"
+               style="push">
+            <parameter
+                  name="launchMode"
+                  value="run">
+            </parameter>
+         </command>
+         <command
+               commandId="com.google.cloud.tools.eclipse.appengine.localserver.launch"
+               disabledIcon="platform:/plugin/org.eclipse.debug.ui/icons/full/dtool16/debug_exc.png"
+               icon="platform:/plugin/org.eclipse.debug.ui/icons/full/etool16/debug_exc.png"
+               label="Debug in App Engine"
+               mnemonic="D"
+               style="push">
+            <parameter
+                  name="launchMode"
+                  value="debug">
+            </parameter>
+         </command>
+      </menuContribution>
+   </extension>
+   <extension
          point="org.eclipse.ui.commands">
       <command
             description="Launch on App Engine development server"
@@ -242,25 +273,22 @@
             class="com.google.cloud.tools.eclipse.appengine.localserver.ui.LaunchAppEngineStandardHandler"
             commandId="com.google.cloud.tools.eclipse.appengine.localserver.launch">
          <enabledWhen>
-            <or>
-               <iterate
-                     ifEmpty="false">
-                  <or>
-                     <adapt
-                           type="org.eclipse.wst.server.core.IModule">
-                     </adapt>
-                     <test
-                           property="org.eclipse.wst.server.ui.isRunnable">
-                     </test>
-                  </or>
-               </iterate>
-               <with
-                     variable="activeEditor">
+           <or>
+           	  <iterate>
+		        <test
+        		    property="org.eclipse.wst.common.project.facet.core.projectFacet"
+            		value="com.google.cloud.tools.eclipse.appengine.facets.standard" />
+              </iterate>
+              <with
+                     variable="activeEditorInput">
                   <adapt
-                        type="org.eclipse.ui.IEditorPart">
+                        type="org.eclipse.core.resources.IFile">
+			        <test
+            			property="org.eclipse.wst.common.project.facet.core.projectFacet"
+            			value="com.google.cloud.tools.eclipse.appengine.facets.standard" />
                   </adapt>
                </with>
-            </or>
+           </or>
          </enabledWhen>
       </handler>
    </extension>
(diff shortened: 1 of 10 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -239,29 +239,69 @@
...
-            </or>
+           </or>
          </enabledWhen>
       </handler>
    </extension>
+   <extension
+         point="org.eclipse.debug.ui.launchShortcuts">
+        <shortcut
+            id="com.google.cloud.tools.eclipse.appengine.localserver.launch.appengine.standard"
+            label="App Engine"
```
elharo: It would be nice to move user visible strings into plugin.properties

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LocalAppEngineStandardLaunchShortcut.java
```diff
@@ -0,0 +1,105 @@
+
```
elharo: copyright noticed needec

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LocalAppEngineStandardLaunchShortcut.java
```diff
@@ -0,0 +1,105 @@
...
+  private LaunchHelper launcher = new LaunchHelper();
+
+  @Override
+  public void launch(ISelection selection, String launchMode) {
+    try {
+      IModule[] modules = launcher.asModules(selection);
+      launcher.launch(modules, launchMode);
+    } catch (CoreException ex) {
+      ErrorDialog.openError(null, "Unable to launch App Engine server", ex.getLocalizedMessage(),
```
elharo: perhaps move these to messages.properties

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LocalAppEngineStandardLaunchShortcut.java
```diff
@@ -0,0 +1,105 @@
...
+    } catch (CoreException ex) {
+      // return null as we do not support these types of objects
+      return null;
+    }
+    return getLaunchConfigurations(modules);
+  }
+
+  @Override
+  public ILaunchConfiguration[] getLaunchConfigurations(IEditorPart editor) {
+    IModule[] modules = null;
```
elharo: please pull happy path into the try block so you don't have to split declaration and intiialization

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LocalAppEngineStandardLaunchShortcut.java
```diff
@@ -0,0 +1,105 @@
...
+      launcher.launch(modules, launchMode);
+    } catch (CoreException ex) {
+      ErrorDialog.openError(null, "Unable to launch App Engine server", ex.getLocalizedMessage(),
+          ex.getStatus());
+    }
+  }
+
+  @Override
+  public ILaunchConfiguration[] getLaunchConfigurations(ISelection selection) {
+    IModule[] modules = null;
```
elharo: please pull happy path into the try block so you don't have to split declaration and intiialization

briandealwis:
PTAL @elharo

PR #1146, opened 2016-12-19
Add service element to new project wizard

elharo:
@akerekes @nbashirbello


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
```diff
@@ -78,4 +79,15 @@ public void setAppEngineLibraries(Collection<Library> libraries) {
     this.appEngineLibraries = new ArrayList<>(libraries);
   }
 
+  String getServiceName() {
+    if (serviceName == null) {
+      return "default"; // NON-NLS-1
```
patflynn: the default behavior in gcloud and the app engine docs all just leave 'service' out when targetting 'default'.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
```diff
@@ -87,13 +90,28 @@ private void createPackageField(Composite container) {
...
   }
 
+  // App Engine service name
+  private void createServiceField(Composite container, ModifyListener pageValidator) {
+    Composite composite = new Composite(container, SWT.NONE);
+    // assumed that container has a single-column GridLayout
+    GridDataFactory.fillDefaults().applyTo(composite);
+
+    Label serviceNameLabel = new Label(composite, SWT.LEAD);
+    serviceNameLabel.setText("App Engine service:");
```
briandealwis: I think this would benefit from a tooltip to provide some background on services/modules.

Thinking about this some more, we would benefit from richer tooltips that offers a hyperlink into cloud.google.com for further information.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardProjectConfig.java
elharo: I think we're still doing that when generating the appengine-web.xml.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineStandardWizardPage.java
briandealwis: We'd need to write our own subclass of JFace `ToolTip`.

PR #1157, opened 2016-12-20
adjust progress monitors and line length

elharo:
@chanseokoh some minor tech debt while I dig into this bug


elharo:
This is what you need to know about monitors:

https://www.eclipse.org/articles/Article-Progress-Monitors/article.html

with the caveat that that article describes Eclipse 4.6 and we still need to work in 4.5 Hence `newChild` instead of `split`.

PR #1167, opened 2016-12-27
Fix build error caused by race condition in App Engine runtime installation

### LINKED ISSUE #1155: Build errors in new project
In Eclipse 4.7 with new standard project:

image

Maybe the runtime didn't get attached?

chanseokoh:
Fixes #1155.

One of my attempts on #1155. Not sure if this is best. We are so doomed because WTP modifies concurrently a project in a non-atomic way when installing the Dynamic Web facet.

- This fix assumes that installing the Dynamic Web facet will schedule a `ConvertJob` to install the JSDT facet.

- `AppEngineStandardFacet.installAppEngineFacet()`: Instead of using `IFacetedProject.installProjectFacet()`, we install facets in a batch using `IFacetedProject.modify()` so that we keeping holding a lock until we finish installing all the facets. This makes the first `ConvertJob` install the JSDT facet only after the batch is complete, which in turn prevents the first `ConvertJob` from scheduling the second `ConvertJob` (triggered by installing the JSDT facet.)

- `FutureJobSuspender`: Prevents the second `ConvertJob` from running until we have done attaching the App Engine runtime.

- `StandardFacetInstallDelegate.installAppEngineRuntimes`: We wait until the first `ConvertJob` completes modifying the project by checking if the JSDT facet is installed as a "fixed" facet. After that, we attach the App Engine runtime.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright 2016 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.eclipse.appengine.facets;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
+import org.junit.After;
+import org.junit.Test;
+
+public class FutureNonSystemJobSuspenderTest {
+
+  private Job job1 = new NoOpSpinJob("Test job 1");
+  private Job job2 = new NoOpSpinJob("Test job 2");
+
+  @After
+  public void tearDown() {
+    FutureNonSystemJobSuspender.resumeInternal();
+    job1.cancel();
+    job2.cancel();
+  }
+
+  @Test
+  public void testCannotSuspendConcurrently() {
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    try {
+      FutureNonSystemJobSuspender.suspendFutureJobs();
+      fail();
+    } catch (IllegalArgumentException ex) {}
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testCannotResumeIfNotSuspended() {
+    FutureNonSystemJobSuspender.resume();
+  }
+
+  @Test
+  public void testSuspendFutureJobs() {
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    job1.schedule();
+    job2.schedule(10000 /* ms */);
+    assertEquals(Job.NONE, job1.getState());
+    assertEquals(Job.NONE, job2.getState());
+  }
+
+  @Test
+  public void testScheduledJobsAreNotSuspended() {
+    job1.schedule();
+    job2.schedule(10000 /* ms */);
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    assertTrue(Job.WAITING == job1.getState() || Job.RUNNING == job1.getState());
+    assertEquals(Job.SLEEPING, job2.getState());
+  }
+
+  @Test
+  public void testSystemJobsAreNotSuspended() {
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    job1.setSystem(true);
+    job2.setSystem(true);
+    job1.schedule();
+    job2.schedule(10000 /* ms */);
+    assertTrue(Job.WAITING == job1.getState() || Job.RUNNING == job1.getState());
+    assertEquals(Job.SLEEPING, job2.getState());
+  }
+
+  @Test
+  public void testResume() {
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    job1.schedule();
+    job2.schedule(10000 /* ms */);
+    assertEquals(Job.NONE, job1.getState());
+    assertEquals(Job.NONE, job2.getState());
+
+    FutureNonSystemJobSuspender.resume();
+    assertTrue(Job.WAITING == job1.getState() || Job.RUNNING == job1.getState());
+    assertEquals(Job.SLEEPING, job2.getState());
+  }
+
+  private static class NoOpSpinJob extends Job {
+
+    private NoOpSpinJob(String name) {
+      super(name);
+    }
+
+    @Override
+    protected IStatus run(IProgressMonitor monitor) {
+      while (!monitor.isCanceled()) {
+        try {
+          Thread.sleep(100);
+        } catch (InterruptedException ex) {}
+      }
+      return Status.OK_STATUS;
+    }
+  };
+}
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
@@ -18,7 +18,6 @@ package com.google.cloud.tools.eclipse.appengine.facets;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
@@ -39,16 +38,13 @@ public class FutureNonSystemJobSuspenderTest {
     job2.cancel();
   }
 
-  @Test
+  @Test(expected = IllegalStateException.class)
   public void testCannotSuspendConcurrently() {
     FutureNonSystemJobSuspender.suspendFutureJobs();
-    try {
-      FutureNonSystemJobSuspender.suspendFutureJobs();
-      fail();
-    } catch (IllegalArgumentException expected) {}
+    FutureNonSystemJobSuspender.suspendFutureJobs();
   }
 
-  @Test(expected = IllegalArgumentException.class)
+  @Test(expected = IllegalStateException.class)
   public void testCannotResumeIfNotSuspended() {
     FutureNonSystemJobSuspender.resume();
   }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
@@ -45,7 +45,7 @@ public class FutureNonSystemJobSuspenderTest {
     try {
       FutureNonSystemJobSuspender.suspendFutureJobs();
       fail();
-    } catch (IllegalArgumentException ex) {}
+    } catch (IllegalArgumentException expected) {}
   }
 
   @Test(expected = IllegalArgumentException.class)
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
@@ -33,25 +33,25 @@ public class FutureNonSystemJobSuspenderTest {
 
   @After
   public void tearDown() {
-    FutureNonSystemJobSuspender.resumeInternal();
+    NonSystemJobSuspender.resumeInternal();
     job1.cancel();
     job2.cancel();
   }
 
   @Test(expected = IllegalStateException.class)
   public void testCannotSuspendConcurrently() {
-    FutureNonSystemJobSuspender.suspendFutureJobs();
-    FutureNonSystemJobSuspender.suspendFutureJobs();
+    NonSystemJobSuspender.suspendFutureJobs();
+    NonSystemJobSuspender.suspendFutureJobs();
   }
 
   @Test(expected = IllegalStateException.class)
   public void testCannotResumeIfNotSuspended() {
-    FutureNonSystemJobSuspender.resume();
+    NonSystemJobSuspender.resume();
   }
 
   @Test
   public void testSuspendFutureJobs() {
-    FutureNonSystemJobSuspender.suspendFutureJobs();
+    NonSystemJobSuspender.suspendFutureJobs();
     job1.schedule();
     job2.schedule(10000 /* ms */);
     assertEquals(Job.NONE, job1.getState());
@@ -62,14 +62,14 @@ public class FutureNonSystemJobSuspenderTest {
   public void testScheduledJobsAreNotSuspended() {
     job1.schedule();
     job2.schedule(10000 /* ms */);
-    FutureNonSystemJobSuspender.suspendFutureJobs();
+    NonSystemJobSuspender.suspendFutureJobs();
     assertTrue(Job.WAITING == job1.getState() || Job.RUNNING == job1.getState());
     assertEquals(Job.SLEEPING, job2.getState());
   }
 
   @Test
   public void testSystemJobsAreNotSuspended() {
-    FutureNonSystemJobSuspender.suspendFutureJobs();
+    NonSystemJobSuspender.suspendFutureJobs();
     job1.setSystem(true);
     job2.setSystem(true);
     job1.schedule();
@@ -80,13 +80,13 @@ public class FutureNonSystemJobSuspenderTest {
 
   @Test
   public void testResume() {
-    FutureNonSystemJobSuspender.suspendFutureJobs();
+    NonSystemJobSuspender.suspendFutureJobs();
     job1.schedule();
     job2.schedule(10000 /* ms */);
     assertEquals(Job.NONE, job1.getState());
     assertEquals(Job.NONE, job2.getState());
 
-    FutureNonSystemJobSuspender.resume();
+    NonSystemJobSuspender.resume();
     assertTrue(Job.WAITING == job1.getState() || Job.RUNNING == job1.getState());
     assertEquals(Job.SLEEPING, job2.getState());
   }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
@@ -123,8 +123,8 @@ public class AppEngineStandardFacet {
 
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
     // Instead of using "IFacetedProject.installProjectFacet()", we install facets in a batch
-    // using "IFacetedProject.modify()" so that we keeping holding a lock until we finish
-    // installing all the facets. This makes the first ConvertJob install the JSDT facet only
+    // using "IFacetedProject.modify()" so that we keep holding a lock until we finish installing
+    // all the facets. This ensures that the first ConvertJob starts installing the JSDT facet only
     // after the batch is complete, which in turn prevents the first ConvertJob from scheduling
     // the second ConvertJob (triggered by installing the JSDT facet.)
     Set<IFacetedProject.Action> facetInstallBatchQueue = new HashSet<>();
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
@@ -122,10 +122,11 @@ public class AppEngineStandardFacet {
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
 
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
-    // We install facets in a batch so that we continue to hold a lock until we finish installing
-    // all the facets. This makes the first ConvertJob install the JSDT facet only after the batch
-    // is complete, which in turn prevents the first ConvertJob from scheduling the second
-    // ConvertJob (triggered by installing the JSDT facet.)
+    // Instead of using "IFacetedProject.installProjectFacet()", we install facets in a batch
+    // using "IFacetedProject.modify()" so that we keeping holding a lock until we finish
+    // installing all the facets. This makes the first ConvertJob install the JSDT facet only
+    // after the batch is complete, which in turn prevents the first ConvertJob from scheduling
+    // the second ConvertJob (triggered by installing the JSDT facet.)
     Set<IFacetedProject.Action> facetInstallBatchQueue = new HashSet<>();
     // Install required App Engine facets i.e. Java 1.7 and Dynamic Web Module 2.5
     if (installDependentFacets) {
@@ -142,6 +143,7 @@ public class AppEngineStandardFacet {
       facetInstallBatchQueue.add(new IFacetedProject.Action(
           IFacetedProject.Action.Type.INSTALL, appEngineFacetVersion, config));
       facetedProject.modify(facetInstallBatchQueue, subMonitor.newChild(100));
+      facetedProject.installProjectFacet(fv, config, subMonitor);
     }
   }
 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
@@ -123,7 +123,7 @@ public class AppEngineStandardFacet {
 
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
     // Instead of using "IFacetedProject.installProjectFacet()", we install facets in a batch
-    // using "IFacetedProject.modify()" so that we keep holding a lock until we finish installing
+    // using "IFacetedProject.modify()" so that we hold the lock until we finish installing
     // all the facets. This ensures that the first ConvertJob starts installing the JSDT facet only
     // after the batch is complete, which in turn prevents the first ConvertJob from scheduling
     // the second ConvertJob (triggered by installing the JSDT facet.)
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
@@ -143,7 +143,6 @@ public class AppEngineStandardFacet {
       facetInstallBatchQueue.add(new IFacetedProject.Action(
           IFacetedProject.Action.Type.INSTALL, appEngineFacetVersion, config));
       facetedProject.modify(facetInstallBatchQueue, subMonitor.newChild(100));
-      facetedProject.installProjectFacet(fv, config, subMonitor);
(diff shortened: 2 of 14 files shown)


elharo:
Have you filed a bug against WTP for the race condition? If not, we should do so.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
```diff
@@ -0,0 +1,114 @@
...
+    job2.cancel();
+  }
+
+  @Test
+  public void testCannotSuspendConcurrently() {
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    try {
+      FutureNonSystemJobSuspender.suspendFutureJobs();
+      fail();
+    } catch (IllegalArgumentException ex) {}
```
elharo: ex --> expected

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -116,13 +118,20 @@ public static boolean isAppEngineStandardRuntime(
...
   public static void installAppEngineFacet(IFacetedProject facetedProject,
       boolean installDependentFacets, IProgressMonitor monitor) throws CoreException {
-    
+
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
-    
+
+    // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
+    // Instead of using "IFacetedProject.installProjectFacet()", we install facets in a batch
+    // using "IFacetedProject.modify()" so that we keeping holding a lock until we finish
```
elharo: keeping holding a --> hold the

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Prevents scheduling all future non-system jobs once {@link #suspendFutureJobs} is called, until
+ * {@link #resume} is called. Jobs already scheduled are not affected and will run to completion.
+ *
+ * The class is for https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155. The
+ * purpose of the class is to prevent the second ConvertJob from running. The second ConvertJob is
+ * triggered by the first ConvertJob when the latter job installs the JSDT facet.
+ *
+ * Not recommended to use for other situations, although the workings of the class is general.
```
elharo: is general --> are general

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class FutureNonSystemJobSuspender {
+  private static boolean suspended;
+
+  private static class SuspendedJob {
+    private Job job;
+    private long scheduleDelay;
```
elharo: ms?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,110 @@
...
+    private SuspendedJob(Job job, long scheduleDelay) {
+      this.job = job;
+      this.scheduleDelay = scheduleDelay;
+    }
+  }
+
+  private static List<SuspendedJob> suspendedJobs =
+      Collections.synchronizedList(new ArrayList<SuspendedJob>());
+
+  private static JobChangeAdaptor jobChangeListener = new JobChangeAdaptor() {
```
elharo: don't use an anonymous class here. Just put this logic into JobChangeAdaptor and change its name.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
```diff
@@ -46,22 +47,49 @@ public void execute(IProject project,
...
     // so schedule a job as a workaround.
     Job installJob = new Job("Install App Engine runtimes in " + project.getName()) {
       @Override
       protected IStatus run(IProgressMonitor monitor) {
+        // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
+        // Wait until the first ConvertJob installs the JSDT facet.
+        try {
+          IProjectFacet jsdtFacet = ProjectFacetsManager.getProjectFacet("wst.jsdt.web");
+          for (int times = 0;
+              times < 100 && !facetedProject.isFixedProjectFacet(jsdtFacet);
```
elharo: please use a named constant for 100

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
```diff
@@ -46,22 +47,49 @@ public void execute(IProject project,
...
+              times++) {
+            try {
+              // To prevent going into the SLEEPING state, don't use "Job.schedule(100)".
+              Thread.sleep(100 /* ms */);
+            } catch (InterruptedException ex) {}
```
elharo: empty catch blocks in non-test code require a comment

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
```diff
@@ -0,0 +1,114 @@
...
+  @Test
+  public void testCannotSuspendConcurrently() {
+    FutureNonSystemJobSuspender.suspendFutureJobs();
+    try {
+      FutureNonSystemJobSuspender.suspendFutureJobs();
+      fail();
+    } catch (IllegalArgumentException expected) {}
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
elharo: Why do we use the expected annotation here and not in the previous similar test case?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,107 @@
...
+
+  /** Once called, it is imperative to call {@link resume()} later. */
+  public static synchronized void suspendFutureJobs() {
+    Preconditions.checkArgument(!suspended, "Already suspended.");
+    suspended = true;
+    Job.getJobManager().addJobChangeListener(jobScheduleListener);
+  }
+
+  public static synchronized void resume() {
+    Preconditions.checkArgument(suspended, "Not suspended.");
```
elharo: This is not quite an idiomatic use of checkArgument. I think you want an IllegalStateException if this fails, not an IllegalArgumentException because this method takes no arguments.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private static List<SuspendedJob> suspendedJobs =
+      Collections.synchronizedList(new ArrayList<SuspendedJob>());
+
+  private static JobScheduleListener jobScheduleListener = new JobScheduleListener();
+
+  /** Once called, it is imperative to call {@link resume()} later. */
+  public static synchronized void suspendFutureJobs() {
+    Preconditions.checkArgument(!suspended, "Already suspended.");
```
elharo: IllegalStateException

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,107 @@
...
+
+    for (SuspendedJob jobInfo : suspendedJobs) {
+      jobInfo.job.schedule(jobInfo.scheduleDelay);
+    }
+    suspendedJobs.clear();
+  }
+
+  private FutureNonSystemJobSuspender() {}
+
+  /** Listens for every job being scheduled and cancel it. */
```
elharo: cancel --> cancels

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
```diff
@@ -46,24 +50,67 @@ public void execute(IProject project,
     installAppEngineRuntimes(project);
   }
 
-  private void installAppEngineRuntimes(final IProject project) {
+  private void installAppEngineRuntimes(final IProject project) throws CoreException {
```
elharo: project probably doesn't need to be final

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
chanseokoh: Oops. Changed to `Preconditions.checkState()`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
chanseokoh: Oops. Changed to `Preconditions.checkState()`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
chanseokoh: Removed `final`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private static List<SuspendedJob> suspendedJobs =
+      Collections.synchronizedList(new ArrayList<SuspendedJob>());
+
+  private static JobScheduleListener jobScheduleListener = new JobScheduleListener();
+
+  /** Once called, it is imperative to call {@link resume()} later. */
+  public static synchronized void suspendFutureJobs() {
+    Preconditions.checkState(!suspended, "Already suspended.");
```
elharo: Why do we care if someone suspends twice? Seems you could simply not add the jobScheduleListener in that case. 

If we do care, then should we also check that suspendedJobs is empty at this point?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,107 @@
...
+    resumeInternal();
+  }
+
+  @VisibleForTesting
+  static synchronized void resumeInternal() {
+    suspended = false;
+    Job.getJobManager().removeJobChangeListener(jobScheduleListener);
+
+    for (SuspendedJob jobInfo : suspendedJobs) {
+      jobInfo.job.schedule(jobInfo.scheduleDelay);
```
elharo: Since this is global static state, I'm a little nervous. Any chance this throws an exception before we're done? Would it be safer if we either:

1. remove each job after we schedule it rather than clearing the whole list later?
2. set suspendedJobs to a new empty list in suspend, rather than reusing the same static object?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
chanseokoh: You're right that there is no technical problem allowing suspending/resuming twice, but if callers could do this, it means we let another loophole for #1155. We don't want two independent suspend-resume cycles to overlap and race.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
chanseokoh: Ah, I missed something. I need to revise synchronization here.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,107 @@
...
+  private FutureNonSystemJobSuspender() {}
+
+  /** Listens for every job being scheduled and cancels it. */
+  private static class JobScheduleListener implements IJobChangeListener {
+    @Override
+    public void scheduled(IJobChangeEvent event) {
+      Job job = event.getJob();
+      if (!job.isSystem()) {
+        job.cancel();  // This will always succeed since the job is not running yet.
+        suspendedJobs.add(new SuspendedJob(job, event.getDelay()));
```
elharo: This doesn't seem to share the same lock as the synchronized methods. Is this a problem? I.e. jobs can be cancelled and added in the middle of a synchronized method. I suggest:

1. Extract these inner classes to outer classes.
2. Add a synchronized add() method to FutureNonSystemJobSuspender that this method calls instead of adding directly into suspendedJobs.

chanseokoh: I think you're not looking at the latest commit. These are properly guarded. However, extracting the method into `add()` seems a good idea.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspender.java
```diff
@@ -0,0 +1,110 @@
...
+        job.cancel();  // This will always succeed since the job is not running yet.
+        suspendedJobs.add(new SuspendedJob(job, scheduleDelay));
+      }
+    }
+  }
+
+  private FutureNonSystemJobSuspender() {}
+
+  /** Listens for every job being scheduled and cancels it. */
+  private static class JobScheduleListener implements IJobChangeListener {
```
elharo: please move to an outer class

chanseokoh: Will do. Are you okay with putting in the same Java source file, or do you want it to be in its own file? For the latter, I would rename the class to something like `FutureNonSystemJobScheduleListener`. (Can't think of a good name actually.)

elharo: Putting it in a separate file makes it easier to read and encourages looser coupling. 

re: naming, I'm not sure why we have "Future" in the name of these things. Maybe just NonSystemJobScheduleListener

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -206,7 +217,8 @@ public static IRuntime createAppEngineFacetRuntime(IProgressMonitor monitor)
   /**
    * Installs Java 1.7 facet if it doesn't already exist in <code>factedProject</code>
    */
-  private static void installJavaFacet(IFacetedProject facetedProject, IProgressMonitor monitor)
+  private static void addJavaFacetToBatchQueue(IFacetedProject facetedProject,
+      Set<IFacetedProject.Action> batchQueue)
```
elharo: Is this a queue or a set? The name should match the type. Let's not call it a queue if it isn't. I was analyzing this code under the assumption it was FIFO, and now I realize that was wrong. 

same issue for the method name. Maybe just addJavaFacetToBatch?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -218,13 +230,15 @@ private static void installJavaFacet(IFacetedProject facetedProject, IProgressMo
...
+    batchQueue.add(new IFacetedProject.Action(
+        IFacetedProject.Action.Type.INSTALL, JavaFacet.VERSION_1_7, javaConfig));
   }
 
   /**
    * Installs Dynamic Web Module 2.5 facet if it doesn't already exits in <code>factedProject</code>
    */
-  private static void installWebFacet(IFacetedProject facetedProject, IProgressMonitor monitor)
+  private static void addWebFacetToBatchQueue(IFacetedProject facetedProject,
+      Set<IFacetedProject.Action> batchQueue)
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -116,13 +118,20 @@ public static boolean isAppEngineStandardRuntime(
...
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
-    
+
+    // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
+    // Instead of using "IFacetedProject.installProjectFacet()", we install facets in a batch
+    // using "IFacetedProject.modify()" so that we hold the lock until we finish installing
+    // all the facets. This ensures that the first ConvertJob starts installing the JSDT facet only
+    // after the batch is complete, which in turn prevents the first ConvertJob from scheduling
+    // the second ConvertJob (triggered by installing the JSDT facet.)
+    Set<IFacetedProject.Action> facetInstallBatchQueue = new HashSet<>();
```
elharo: see below, but this isn't a queue, is it? Should it be?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/NonSystemJobSuspender.java
```diff
@@ -0,0 +1,86 @@
...
+ * Prevents scheduling all future non-system jobs once {@link #suspendFutureJobs} is called, until
+ * {@link #resume} is called. Jobs already scheduled are not affected and will run to completion.
+ *
+ * The class is for https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155. The
+ * purpose of the class is to prevent the second ConvertJob from running. The second ConvertJob is
+ * triggered by the first ConvertJob when the latter job installs the JSDT facet.
+ *
+ * Not recommended to use for other situations, although the workings of the class are general.
+ */
+public class NonSystemJobSuspender {
```
elharo: Every class loader (and there are many in OSGI) that sees this load a new instance of this class. Does this matter to us? Or is only a single class loader ever going to see this? 

Can we make this class non-public (package-protected). That might be a little safer.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.junit.Assert.assertTrue;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
+import org.junit.After;
+import org.junit.Test;
+
+public class FutureNonSystemJobSuspenderTest {
```
elharo: NonSystemJobSuspenderTest

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
chanseokoh: Good catch! It's a set. (WTP will determine the order of actions according to dependencies.)

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/NonSystemJobSuspender.java
chanseokoh: Made package-protected. The `StandardFacetInstallDelegate` shall be the only user of this, so I think now we're good.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/FutureNonSystemJobSuspenderTest.java
chanseokoh: Renamed the test class. Thanks.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -116,13 +118,20 @@ public static boolean isAppEngineStandardRuntime(
...
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
-    
+
+    // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
+    // Instead of using "IFacetedProject.installProjectFacet()", we install facets in a batch
+    // using "IFacetedProject.modify()" so that we hold the lock until we finish installing
+    // all the facets. This ensures that the first ConvertJob starts installing the JSDT facet only
+    // after the batch is complete, which in turn prevents the first ConvertJob from scheduling
+    // the second ConvertJob (triggered by installing the JSDT facet.)
+    Set<IFacetedProject.Action> facetInstallBatchSet = new HashSet<>();
```
elharo: I think  just facetInstallBatch, or facetInstallSet; or is there some meaning of "Batch" separate from set here?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -206,8 +217,8 @@ public static IRuntime createAppEngineFacetRuntime(IProgressMonitor monitor)
   /**
    * Installs Java 1.7 facet if it doesn't already exist in <code>factedProject</code>
    */
-  private static void installJavaFacet(IFacetedProject facetedProject, IProgressMonitor monitor)
-      throws CoreException {
+  private static void addJavaFacetToBatch(IFacetedProject facetedProject,
+      Set<IFacetedProject.Action> batchSet) {
```
elharo: perhaps just facetActions?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -218,14 +229,15 @@ private static void installJavaFacet(IFacetedProject facetedProject, IProgressMo
...
+        IFacetedProject.Action.Type.INSTALL, JavaFacet.VERSION_1_7, javaConfig));
   }
 
   /**
    * Installs Dynamic Web Module 2.5 facet if it doesn't already exits in <code>factedProject</code>
    */
-  private static void installWebFacet(IFacetedProject facetedProject, IProgressMonitor monitor)
-      throws CoreException {
+  private static void addWebFacetToBatch(IFacetedProject facetedProject,
+      Set<IFacetedProject.Action> batchSet) {
```
elharo: actions or actionSet?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -218,14 +229,15 @@ private static void installJavaFacet(IFacetedProject facetedProject, IProgressMo
     sourcePaths.add(new Path("src/main/java"));
     sourcePaths.add(new Path("src/test/java"));
     javaConfig.setSourceFolders(sourcePaths);
-    facetedProject.installProjectFacet(JavaFacet.VERSION_1_7, javaConfig, monitor);
+    batchSet.add(new IFacetedProject.Action(
+        IFacetedProject.Action.Type.INSTALL, JavaFacet.VERSION_1_7, javaConfig));
   }
 
   /**
    * Installs Dynamic Web Module 2.5 facet if it doesn't already exits in <code>factedProject</code>
```
elharo: already exist in facetedProject

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/NonSystemJobSuspender.java
```diff
@@ -0,0 +1,86 @@
...
+    private Job job;
+    private long scheduleDelay;  // ms
+
+    private SuspendedJob(Job job, long scheduleDelay) {
+      this.job = job;
+      this.scheduleDelay = scheduleDelay;
+    }
+  }
+
+  private static final List<SuspendedJob> suspendedJobs = new ArrayList<SuspendedJob>();
```
elharo: ArrayList<>()

chanseokoh: Nice. Fixed.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
chanseokoh: `facetInstallSet` sounds good. Changed to it across all the places consistently.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -215,10 +215,10 @@ public static IRuntime createAppEngineFacetRuntime(IProgressMonitor monitor)
   }
 
   /**
-   * Installs Java 1.7 facet if it doesn't already exist in <code>factedProject</code>
+   * Installs Java 1.7 facet if it doesn't already exist in {@code factedProject}.
```
elharo: facted --> faceted

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -229,15 +229,15 @@ private static void addJavaFacetToBatch(IFacetedProject facetedProject,
...
     sourcePaths.add(new Path("src/test/java"));
     javaConfig.setSourceFolders(sourcePaths);
-    batchSet.add(new IFacetedProject.Action(
+    facetInstallSet.add(new IFacetedProject.Action(
         IFacetedProject.Action.Type.INSTALL, JavaFacet.VERSION_1_7, javaConfig));
   }
 
   /**
-   * Installs Dynamic Web Module 2.5 facet if it doesn't already exits in <code>factedProject</code>
+   * Installs Dynamic Web Module 2.5 facet if it doesn't already exist in {@code factedProject}.
```
elharo: faceted

PR #1175, opened 2016-12-29
check for a different failure message

### LINKED ISSUE #1172: Java VM can't start but server stays running
I forced the VM to fail when launching a local server by supplying an undefined argument to the VM. The WTP server doesn't notice this and keeps going. 

```
INFO     2016-12-29 12:37:41,007 devappserver2.py:769] Skipping SDK update check.
INFO     2016-12-29 12:37:41,116 api_server.py:205] Starting API server at: http://localhost:52060
INFO     2016-12-29 12:37:41,123 dispatcher.py:197] Starting module "default" running at: http://localhost:8080
INFO     2016-12-29 12:37:41,127 admin_server.py:116] Starting admin server at: http://localhost:8000
ERROR    2016-12-29 12:37:42,270 http_runtime.py:396] bad runtime process port ['']
Unrecognized option: -versub2
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
ERROR    2016-12-29 12:37:42,276 http_proxy.py:116] bad runtime process port ['']
INFO     2016-12-29 12:37:42,383 module.py:788] default: "GET /_ah/warmup HTTP/1.1" 500 169
```

elharo:
fix #1172 @chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
@@ -393,7 +393,8 @@ public class LocalAppEngineServerBehaviour extends ServerBehaviourDelegate
         // An error occurred
         setServerState(IServer.STATE_STOPPED);
       } else if (line.contains("Error: A fatal exception has occurred. Program will exit")) { //$NON-NLS-1$
-        setServerState(IServer.STATE_STOPPED);      
+        // terminate the Python process
+        stop(false);
       } else if (line.contains("Starting module")  //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


chanseokoh:
Unfortunately, looks like there is another problem with #1172. The dev_appserver Python process in this case doesn't terminate. I think dev_appserver doesn't recognize that the JVM failed to start, and it continues to run (and hold the port 8080). You can verify this behavior on the command line. Ideally, the dev_appserver should recognize this JVM failure and exit accordingly. Not sure if we can explicitly kill the Python process as a workaround, or if doing so is a good idea.

chanseokoh:
Maybe we could call `stop(false)` instead of just setting the state `STOPPED` in this case. This terminates the server gracefully. It still pops up the error dialog that the server failed to start, which is nice, although not sure who pops up the error dialog.

chanseokoh:
Yeah, I think what we need is `stop(false)`. @briandealwis agreed?

briandealwis:
> @chanseokoh: Yeah, I think what we need is stop(false).

That seems right to me.

PR #1044, opened 2016-12-01
Show warning when m2e _do not update remote-dependencies_ setting is on

briandealwis:
I ended up stripping out the databinding out of the `AppEngineLibrariesSelectionGroup` as  I needed to expose the an event when a library was selected.  Although I could have exposed the `WritableList` or some other `IObservableList` and used databinding as the listening mechanism, it seemed more trouble than it was worth.  Especially since this component couldn't be embedded in a larger databinding scheme.

Fixes #830
Fixes #911

Issue #830, opened 2016-10-14
Force m2eclipse to check remote repository for latest artifact version when library is defined to use the latest version


akerekes:
If m2eclipse is set to not check for latest version automatically in remote repositories (`Window/Preferences/Maven/Do not automatically update dependencies from remote repositories`), then if the user has an older version of any of the libraries we're adding to the classpath via the API Attacher, it will use that jar, and download the latest for those that are not present locally. This can lead to having App Engine API 1.9.40 and App Engine Endpoints 1.9.44 on the classpath (assuming that App Engine API 1.9.40 was present in the local repo, no App Engine Endpoints jar was available locally and the latest version is 1.9.44)

chanseokoh:
Also, when we later download servlet-api.jar, we need to specifically pin a certain version.

akerekes:
You can specify an explicit version for a library already.

chanseokoh:
Right. The title of this issue suggests to force using latest artifact versions, so I was making a note that servlet-api.jar is an exception. It should be 2.5 for App Engine standard.

akerekes:
Title updated. Also for clarification: this issue has nothing to do with artifacts that are configured with a specific version, only with those that use 'LATEST' as version.

akerekes:
I have digged into this issue, but I could not find a good way to override the preferences setting for a given request. I don't think we should change the user's preference just to make this work, so I'm not sure how we could achieve this (apart from deleting all local artifacts of for the given Maven coordinates to ensure that the lookup will hit the remote repo, but this also seems excessive).

branflake2267:
Could a m2e configurator pick up the version on an maven update and then reconfigure the api used? I do some configuring on m2e updates. Although I've noticed once I add a lifecycle, I have to run that lifecycle in eclipse m2e.

briandealwis:
I suspect the best we can do is log the situation.

briandealwis:
The M2Eclipse preference is `MavenPreferenceConstants.P_GLOBAL_UPDATE_NEVER` (boolean), managed by `org.eclipse.m2e.core.ui.internal.preferences.MavenPreferencePage`.  This preference is used by m2e when generating its `MavenConfigurationImpl` to create an Aether repository update policy (`org.eclipse.aether.repository.RepositoryPolicy.UPDATE_POLICY_NEVER`).

[...]

PR #1176, opened 2016-12-29
remove warning about unable to check for latest version

### LINKED ISSUE #1170: Maven settings prevent checking for later versions of selected libraries
This error message when creating a new maven project and adding libraries is surprising. It sounds like even if this fails maybe we shouldn't bother the user with this?

image

elharo:
fix #1170 @akerekes

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
@@ -29,7 +29,6 @@ import java.io.IOException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Paths;
 import java.util.Collection;
-import org.eclipse.aether.repository.RepositoryPolicy;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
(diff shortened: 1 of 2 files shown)


briandealwis:
That was deliberately introduced as #1044 (and I've been bitten by this myself). Rather than backing it out entirely, can't we change `WARNING` to `INFO` and fix-up the wording?

elharo:
The question is whether anybody actually expects this to happen or is this just a best effort behind the scenes thing. Rajeev and I were both independently confused by this when we hit it. 

If we change it to info, what wording do you suggest?

I think not being able to check for the latest version is the default setting in Eclipse since neither Rajeev nor myself made the effort to change it. Is it likely this feature will work for most users out of the box?

briandealwis:
That you see the text means you will be affected by the issue (I think the original issue is #830). This issue is perhaps too subtle to explain in a single sentence, especially to provide instructions on how to fix it. Perhaps we should have a separate wizard page that can provide more detail — with an option for temporarily disabling the setting?

elharo:
I think (not yet sure) that a default install of Eclipse+m2e is going to hit this every time. 

I think a separate wizard page is overly complex.

What's the impact on the user of simply not resolving the problem and allowing them to proceed with their potentially older version of the library?

akerekes:
@elharo they can end up with a rather old library and if they want to update it, they will have to do it manually:
- remove the old one we've added to the project
- download and add the newer version to the project

elharo:
Yes,  but this warning doesn't prevent the problem or give guidance on how to fix it.

elharo:
Ping. I'm going to keep pushing on this one. We have now had *3* engineers who all said WTF when they saw this message. That's strong evidence this warning is confusing. And this happens with the default set of preferences. I do not want to show a warning to every developer who creates a maven-based project, and right now that's what we're doing.

elharo:
Let's talk about #1223 next week. I'm not sure how high a priority it is. Independently of that , though, we do need to fix this issue, and we can do it now.

PR #1183, opened 2017-01-04
Remove .appengine.facets dependency from ui.util

### LINKED ISSUE #1181: .ui.util bundle depends on .appengine.facets bundle, so the facets bundle cannot use util classes
The `c.g.c.t.e.ui.util` bundle imports a package from the `c.g.c.t.e.appengine.facets` bundle.

```
...
...
```

So, if I attempt to use a util class in `c.g.c.t.e.ui.util` bundle from `c.g.c.t.e.appengine.facets`, it won't compile with the error "a cycle was detected in the build path".

`ProjectFromSelectionHelper` in the `util` bundle seems to be only one that requires the `facets` bundle.

chanseokoh:
Fixes #1181. Basically, `ProjectFromSelectionHelper.getAppEngineStandardProject()` --> `ProjectFromSelectionHelper.getProject()`. And the method is now static.

Also, code has been simplified along the way:

- `StandardDeployCommandHandlerTest` removed. This was testing the case where `ProjectFacetsManager.create(project)` throws `CoreException`, but I think we never get this exception because our code is activated only for projects that have the App Engine facet. Moreover, `ProjectFacetsManager.create(project)` returns `null` instead of throwing `CoreException` even if the project is not faceted.
- `FacetedProjectHelper.getFacetedProject()` (which was a thin wrapper that existed only for the unit testing above) is no longer needed and removed. Tests that used the method become to use the original method.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandlerTest.java
```diff
@@ -1,61 +0,0 @@
...
-@RunWith(MockitoJUnitRunner.class)
-public class StandardDeployCommandHandlerTest {
-
-  @Mock private FacetedProjectHelper facetedProjectHelper;
-
-  @Test(expected = ExecutionException.class)
-  public void testExecute_facetedProjectCreationThrowsException() throws ExecutionException, CoreException {
-    StandardDeployCommandHandler handler = new StandardDeployCommandHandler(facetedProjectHelper);
-
-    when(facetedProjectHelper.getFacetedProject(isA(IProject.class))).thenThrow(getFakeCoreException());
```
elharo: does this code path no longer exist?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -71,40 +70,32 @@
...
-        }
-
-        IGoogleLoginService loginService = ServiceUtils.getService(event, IGoogleLoginService.class);
-        DeployPreferencesDialog dialog =
-            new DeployPreferencesDialog(HandlerUtil.getActiveShell(event), project, loginService);
-        if (dialog.open() == Window.OK) {
-          launchDeployJob(project, dialog.getCredential());
-        }
+      IProject project = ProjectFromSelectionHelper.getProject(event);
+      Preconditions.checkState(project != null, "Command enablement wrongly configured.");
```
elharo: This is a good check, but it's not a precondition.
also exception message is unclear.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -71,40 +70,32 @@
...
+      IFacetedProject facetedProject = ProjectFacetsManager.create(project);
+      Preconditions.checkState(facetedProject != null, "Command enablement wrongly configured.");
```
elharo: ditto, don't use Preconditions for normal null checks

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
chanseokoh: Yeah, what I really want is an assertion (impossible condition), but I know we don't use assertions at Google. I think I'll remove these checks and add a comment instead.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandlerTest.java
chanseokoh: That's what I was saying with the impossible condition checks. This test is simulating that attempting to get `IFacetedProject` (by calling `ProjectFacetsManager.create(project)`) throws an exception. But because the deploy menu is enabled (only) for faceted projects, it must always succeed. If it doesn't success, it implies the deploy menu was enabled on a wrong resource.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
elharo: Just use an if (project != null)

no need for a library or an assert

chanseokoh:
For `MANIFEST.MF`, it's updated, if you look at it carefully.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -71,40 +69,36 @@
...
-
-        IGoogleLoginService loginService = ServiceUtils.getService(event, IGoogleLoginService.class);
-        DeployPreferencesDialog dialog =
-            new DeployPreferencesDialog(HandlerUtil.getActiveShell(event), project, loginService);
-        if (dialog.open() == Window.OK) {
-          launchDeployJob(project, dialog.getCredential());
-        }
+      IProject project = ProjectFromSelectionHelper.getProject(event);
+      if (project == null) {
+        throw new AssertionError("Deploy menu enabled for non-project resources.");
```
elharo: this should be an exception not an error; Errors are for major VM problems only.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -71,40 +69,36 @@
...
+      }
+      IFacetedProject facetedProject = ProjectFacetsManager.create(project);
+      if (facetedProject == null) {
+        throw new AssertionError("Deploy menu enabled for non-faceted projects.");
```
elharo: this should be an exception not an error; Errors are for major VM problems only.

Issue #348, opened 2016-07-12
Convert to Google App Engine project action


elharo:
Probably in a context menu;

Remove old nature and facets; add new facets

chanseokoh:
Looks like we need more discussion, so I guess this won't happen very soon. Some ideas popped up in #1188:

- Moving this to a submenu of _Configure_ seems right. The label _Convert to (Google) App Engine Project..._ seems okay too.
![selection_005](https://cloud.githubusercontent.com/assets/10523105/21780488/cf64c370-d679-11e6-8c33-7fa6dee7e17b.png)
- Perhaps it's a good idea to have a dialog for confirming the conversion, primarily for messaging.
- Removing obsolete natures and facets? (What to remove and how?)
- Move this to a new bundle. This isn't only for GPE-compatibility, so I think a name like `projectconversion` is better suited.

- Anything else?

elharo:
No confirmation please. The user told us to do it. We do it. There are very few truly irrevocable actions where you should ask for confirmation.

I don't think you need paretheses around Google. Just "Convert to Google" App Engine Project" or maybe just "Convert to  App Engine Project" Ellipses only if it brings up a dialog. 

We should remove the natures used by GPE.

elharo:
https://cloud.google.com/eclipse/docs/converting

PR #1188, opened 2017-01-05
Add "Convert to App Engine Standard Project" menu

chanseokoh:
WIP for #348. At least it's working. However, this always creates `WEB-INF` under `src/main/webapp`, so this should be fixed.

I need to add tests, and probably need to go through more iterations.

Some screenshots:

selection_003

incompatible java facet version _004

converted to app engine standard project _005


chanseokoh:
Probably the messaging

"App Engine Standard project requires Java version 1.7, but the installed version is 1.8."

is confusing. It's the facet version we're checking here.

patflynn:
is 'Conversion' standard nomenclature for adding framework support?

Also are we informing the user of the restrictions that app engine standard
entails and warning the user for projects that couldn't work on AE std?

briandealwis:
Haven't looked at the code changes yet.  Two things that jump out:
  - Normally this type of action would appear under the _Configure_ submenu.
  - The dialogs should include the project name

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.properties
```diff
@@ -7,3 +7,8 @@ flexFacetDescription = Supply App Engine Flexible Environment Functionality
 
 standardFacetName = App Engine Java Standard Environment
 standardFacetDescription = Supply App Engine Standard Environment Functionality
+
+convertToAppEngineStandardCommandDescription = Converts an Eclipse project into a Google App \
+ Engine standard environment project.
+convertToAppEngineStandardCommandName = Convert to App Engine Standard Project
+convertToAppEngineStandardMenuLabel = Convert to App Engine Standard Project
```
elharo: Looks like the command name and the menu label are the same. Do we need two messages?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -46,4 +46,44 @@
...
+    </menuContribution>
+  </extension>
+
+  <extension point="org.eclipse.ui.handlers">
+    <handler
+      class="com.google.cloud.tools.eclipse.appengine.facets.ui.AppEngineStandardProjectConverter"
+      commandId="com.google.cloud.tools.eclipse.appengine.facets.convertToStandard">
+      <enabledWhen>
+        <and>
+          <count value="1" />  <!-- enabled when only one project is selected -->
```
elharo: only --> exactly

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -46,4 +46,44 @@
...
+          <iterate>
+            <adapt type="org.eclipse.core.resources.IProject" />
+          </iterate>
+          <not>
+            <iterate>
+              <reference definitionId="com.google.cloud.tools.eclipse.appengine.onlyInGCPProjects" />
```
briandealwis: Can we not make this test more specific by checking that that the project has a GPE nature (`com.google.appengine.clipse.core.gaeNature`) or the GPE WTP facet (`com.google.appengine.facet`)?

briandealwis:
Also need to remove any obsolete natures and facets.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
elharo: I think this is supposed to work for GPE and non-GPE projects.

chanseokoh: @briandealwis This test is negated by `<not>`. The name of the test is also misleading in that it tests (before the negation) if the project has the App Engine Standard facet.

That said, this will show the conversion menu for *any* project that doesn't have the Standard facet. Now I wonder if we need to narrow down the scope, e.g., by checking if the project has a directory `WEB-INF`.

briandealwis: > I think this is supposed to work for GPE and non-GPE projects.

Oh!  I had thought it was a forward migration tool for GPE.  General purpose makes sense.

I wonder if we should rationalize `onlyInGCPProject` and have more explicit `isGCPProject` and `isInGCPProject` tests.  I opened issue #1214 for that.

briandealwis:
> @briandealwis: I'd like this to be moved into a proper bundle. `.compat`?

I recant: I thought this was GPE-migration functionality that we might ditch eventually.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -46,4 +46,44 @@
...
+    </command>
+  </extension>
+
+  <extension point="org.eclipse.ui.menus">
+    <menuContribution
+      allPopups="false"
+      locationURI="popup:org.eclipse.ui.popup.any?after=additions">
+      <command
+        commandId="com.google.cloud.tools.eclipse.appengine.facets.convertToStandard"
+        label="%convertToAppEngineStandardMenuLabel">
```
briandealwis: You can drop this label

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/Messages.java
```diff
@@ -27,9 +28,9 @@
   private Messages() {
   }
 
-  public static String getString(String key) {
+  public static String getString(String key, Object... params) {
     try {
-      return RESOURCE_BUNDLE.getString(key);
+      return MessageFormat.format(RESOURCE_BUNDLE.getString(key), params);
```
briandealwis: Huh!  We should do this for all of our `Messages` classes

plugins/com.google.cloud.tools.eclipse.appengine.facets/META-INF/MANIFEST.MF
```diff
@@ -14,18 +14,23 @@ Require-Bundle: com.google.cloud.tools.appengine;bundle-version="0.2.5",
  org.eclipse.jst.j2ee,
  org.eclipse.jst.j2ee.web,
  org.eclipse.m2e.maven.runtime,
- org.eclipse.m2e.core;bundle-version="1.6.2"
-Import-Package: com.google.cloud.tools.eclipse.util,
+ org.eclipse.m2e.core;bundle-version="1.6.2",
```
elharo: I don't think you should have this in both Require-Bundle and Import-Package below. See what happens if you remove this one.

chanseokoh:
Still working on it: need to remove obsolete natures. @briandealwis what's the name (or ID) of the nature we want to remove?

plugins/com.google.cloud.tools.eclipse.appengine.facets/META-INF/MANIFEST.MF
chanseokoh: Yeah, I was able to remove `org.eclipse.m2.core` from both.

briandealwis:
- GPE GAE project nature: `com.google.appengine.eclipse.core.gaeNature`
- GPE WTP GAE Facet: `com.google.appengine.facet` version `1`
  - there was also a facet `com.google.appengine.facet.ear` that we aren't supporting any more
- there was also the unreleased facet `com.google.cloudsdk.facet` but it was never formally released and doesn't seem to appear in the wild

briandealwis:
Oh, and I suppose there's also the WTP server runtimes too:
  - `com.google.appengine.runtime.id1` and the corresponding _runtime-component_ `com.google.appengine.runtime.id`
  - the unreleased `com.google.cloudsdk.runtime.id` runtime shouldn't appear anywhere

chanseokoh:
There is a slight problem in removing GPE facets which are not defined in our bundles. To uninstall a facet, we need to "get" and pass `IProjectFacetVesion` to `uninstallProjectFacet()`, which is the only method I could find to remove a facet. However, trying to get an undefined facet

```java
ProjectFacetsManager.getProjectFacet("com.google.appengine.facet");
```

 throws an exception:

`java.lang.IllegalArgumentException: Project facet com.google.appengine.facet has not been defined.`

chanseokoh:
On the other hand, I think I can remove undefined runtimes by this way:

```java
      Set<IRuntime> runtimes = facetedProject.getTargetedRuntimes();
      for (IRuntime runtime : runtimes) {
        if (GPE_WTP_SERVER_RUNTIME.equals(runtime.getProperty("id"))) {
          facetedProject.removeTargetedRuntime(runtime, null /* monitor */);
        }
      }
```

briandealwis:
May need to define a dummy `com.google.appengine.facet` facet.  Perhaps in a .compat bundle? :-O

chanseokoh:
Okay, seems like we need a `.compat` bundle in the end.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/convert/AppEngineStandardProjectConvertJob.java
```diff
@@ -0,0 +1,76 @@
+package com.google.cloud.tools.eclipse.appengine.facets.convert;
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/NatureUtilsTest.java
```diff
@@ -0,0 +1,51 @@
...
+  @Rule public TestProjectCreator projectCreator = new TestProjectCreator();
+
+  @Test
+  public void testRemoveNature() throws CoreException {
+    IProject project = projectCreator.getProject();
+    // By default, project has Java nature and faceted project nature.
+    assertEquals(2, project.getDescription().getNatureIds().length);
+
+    NatureUtils.removeNature(projectCreator.getProject(), JavaCore.NATURE_ID);
+    assertEquals(1, project.getDescription().getNatureIds().length);
```
elharo: should probably test that it removed the correct one

chanseokoh:
@elharo What do you think of adding the `.compat` bundle and move the code to remove the GPE nature, runtimes, and facets to a separate PR? I expect removing GPE stuff will require several iterations.

elharo:
If a .compat bundle is what we need, that's fine. Or maybe call it .convert? It's not really about compatibility.

General principle is that the code should be shippable after every merge; so we shouldn't expose menu items that don't work yet or worse yet break things.

chanseokoh:
What @briandealwis suggested is a .compat bundle, which will be only for removing the GPE stuff. We agreed that we don't create a .convert bundle for "conversion" (i.e., adding the App Engine Standard facet to any project). In that sense, this PR can work as-is (after removing the code related to the GPE nature), although it wont' do anything about GPE. But maybe it's better to wait until I put all the GPE stuff here in this PR.

chanseokoh:
Still WIP, mostly for adding tests.

chanseokoh:
At last, the PR is ready for review.

- The conversion menu appears for any project that doesn't have the App Engine Standard facet. It appears under _Configure_. This functionality should work. I'll do some more actual testing for a while.
selection_001
It simply calls a method to add the App Engine facet. (Very simple.) The conversion is done silently. It's done via a job per @briandealwis's suggestion.

- If a project being converted does not have any of the Java 1.7 facet and the Dynamic Web Module facet 2.5, the facets will be installed together. If a project has incompatible facets, users will see an error dialog.

- The menu will also attempt to remove the obsolete GPE nature, runtime, and facets (if they exist). It required me to add many dummy definitions in `plugin.xml`. That is, our plugin defines those fixtures with the same ID that GPE uses. However, I'm not confident if this will really work. Will this hackery trick Eclipse into thinking that the dummy definitions refer to the GPE definitions? Actual testing is needed.

- I'm still largely confused regarding runtimes: there are terminologies like runtime type, runtime, and runtime component. Hope what I did in this PR is correct. @briandealwis, please look into it.

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/DummyDelegate.java
```diff
@@ -0,0 +1,33 @@
...
+
+package com.google.cloud.tools.eclipse.appengine.compat;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.wst.common.project.facet.core.IDelegate;
+import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
+
+/** A dummy delegate required for defining facet install/uinstall actions in {@code plugin.xml}. */
```
elharo: uninstall

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/DummyNature.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+ package com.google.cloud.tools.eclipse.appengine.compat;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IProjectNature;
+import org.eclipse.core.runtime.CoreException;
+
+/** A dummy nature runtime class only to enable defining a dummy nature for testing. */
```
elharo: This should be in the test tree if at all possible.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
elharo: If the project doesn't have a WEB-INF then we add one, right? along with web.xml and appengine-web.xml?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -257,4 +269,51 @@ private static void addWebFacetToBatch(IFacetedProject facetedProject,
     return appEngineRuntimes.toArray(appEngineRuntimesArray);
   }
 
+  /**
+   * Attempts to find a main web application directory, by the following logic:
+   *
+   * 1. If there is no {@code WEB-INF} folder in the {@code project}, returns {@code null}.
+   * 2. Otherwise, if there is at least one {@code WEB-INF} folder that contains {@code WEB-INF},
```
elharo: In other words WEB-INF/WEB-INF or should this be WEB-INF/web.xml?

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/DummyNature.java
chanseokoh: I tried to define the nature in the test bundle (inside `plugin.xml`), but it seems like a test bundle (which has `Fragment-Host:`) doesn't allow having a `plugin.xml`. It's very unfortunate. Right, @briandealwis?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
chanseokoh: @elharo Correct. Actually, files and folders including `WEB-INF` and `web.xml` are created when we install the Dynamic Web Module facet. If there is no `WEB-INF`, we tell the Web facet to create it under `/src/main/webapp`. After that, our App Engine facet places `appengine-web.xml` to the main `WEB-INF` directory.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
chanseokoh: The latter, `WEB-INF/web.xml`. I'll fix the comment.

chanseokoh:
FYI, I noticed that [the GPE code does more things](https://github.com/GoogleCloudPlatform/google-plugin-for-eclipse/blob/master/plugins/com.google.appengine.eclipse.core/src/com/google/appengine/eclipse/core/nature/GaeNature.java#L63) than merely removing the GPE nature. Here, I just remove the nature.

chanseokoh:
Converting a GPE project more or less works.

- The GPE runtime and the GPE facets are removed. For example, the old facet settings XML below

```xml
<faceted-project>                                                                                                                                                                                              <runtime name="Google App Engine"/>
  <installed facet="java" version="1.7"/>
  <installed facet="jst.web" version="2.5"/>
  <installed facet="com.google.appengine.facet" version="1"/>
</faceted-project>
```

becomes like this:

```xml
<faceted-project>                                                                                                                                                                                              <runtime name="App Engine Standard Runtime"/>
  <installed facet="java" version="1.7"/>
  <installed facet="jst.web" version="2.5"/>
  <installed facet="com.google.cloud.tools.eclipse.appengine.facets.standard" version="1"/>
</faceted-project>
```

- `WEB-INF` is correctly identified too after conversion:

```xml
<?xml version="1.0" encoding="UTF-8"?><project-modules id="moduleCoreId" project-version="1.5.0">
    <wb-module deploy-name="GPE">                                                                                                                                                                                    <wb-resource deploy-path="/WEB-INF/classes" source-path="/src"/>
        <wb-resource deploy-path="/" source-path="/war" tag="defaultRootSource"/>
        <property name="context-root" value="GPE"/>
        <property name="java-output-path" value="/GPE/war/WEB-INF/classes"/>
    </wb-module>
</project-modules>
```

However, I see there remains some old libraries. How can I remove these?

selection_006

We could also remove GPE preferences files under `.settings` if that's possible?

`com.google.appengine.eclipse.core.prefs`
`com.google.gdt.eclipse.core.prefs`
`org.eclipse.jdt.core.prefs`

chanseokoh:
Another problem is that we litter the available facet set by adding the two dummies:

selection_007

chanseokoh:
At least we can change the names and the description of the dummy facets:

selection_009

We can also prevent installing these dummies in their facet definitions, but I need to be able to install them for unit testing.

elharo:
@chanseokoh You don't sound very confident in this approach and it is getting quite complex. Should we consider directly modifying the relevant files (.project and .settings/org.eclipse.wst.common.project.facet.core.xml) to remove the legacy GPE content rather than forcing Eclipse to do it?

chanseokoh:
From my yesterday's testing, this works fairly well, so I have built some confidence. It's mostly that the lingering libraries issue needs to be taken care of further. (And we might discover a few more we can clean up.)

The only (and major) downside is that we litter the facet space with the dummies that are exposed to users. Hopefully changing the labels to indicate they are obsolete can be a viable option.

I wonder what other folks in our team think.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/XmlTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class XmlTest extends BasePluginXmlTest {
+
+  @Test
+  public void testFacetsDefined() {
+    checkFacetDefined("com.google.appengine.facet");
+    checkFacetDefined("com.google.appengine.facet.ear");
+  }
+
+  private void checkFacetDefined(String facetId) {
```
elharo: Can this move to a common utility method in BasePluginXmltest to avoid duplicating code?

plugins/com.google.cloud.tools.eclipse.appengine.compat/plugin.properties
```diff
@@ -0,0 +1,3 @@
+gpeGaeFacetName=Obsolete GPE App Engine Facet 
+gpeGaeEarFacetName=Obsolete GPE App Engine EAR Facet
+obsoleteGpeFacetDescription=This facet from GPE is now obsolete and should not be used.
```
elharo: GPE --> the Google Plugin for Eclipse

plugins/com.google.cloud.tools.eclipse.appengine.compat/plugin.xml
```diff
@@ -0,0 +1,65 @@
...
+      <action type="INSTALL">  <!-- Needed only for unit testing -->
+        <delegate class="com.google.cloud.tools.eclipse.appengine.compat.DummyDelegate" />
+      </action>
+      <action type="UNINSTALL">
+        <delegate class="com.google.cloud.tools.eclipse.appengine.compat.DummyDelegate" />
+      </action>
+    </project-facet-version>
+  </extension>
+
+  <!-- Dummy GPE GAE nature definition; required only for enabling unit testing -->
```
elharo: If these dummies show up in the user interface, I'm not sure they carry their weight. Is there any ay we can test without them?

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/DummyNature.java
elharo: Again, if users can find this we really shouldn't have it. Maybe we can strip these out at build time?

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
```diff
@@ -0,0 +1,68 @@
...
+  private static final String GPE_GAE_NATURE_ID = "com.google.appengine.eclipse.core.gaeNature";
+
+  private static final String GPE_WTP_GAE_FACET_ID = "com.google.appengine.facet";
+  private static final String GPE_WTP_GAE_FACET_VERSION = "1";
+  private static final String GPE_WTP_GAE_EAR_FACET_ID = "com.google.appengine.facet.ear";
+  private static final String GPE_WTP_GAE_EAR_FACET_VERSION = "1";
+
+  private static final String GPE_WTP_GAE_RUNTIME = "com.google.appengine.runtime.id";
+
+  public static void removeObsoleteGpeFixtures(
```
elharo: fixtures? Is that the right word?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    facetedProject.installProjectFacet(JavaFacet.VERSION_1_7, null, null);
+    facetedProject.installProjectFacet(WebFacetUtils.WEB_25, null, null);
+    assertTrue(facetedProject.hasProjectFacet(WebFacetUtils.WEB_25));
+
+    assertTrue(commandHandler.checkFacetCompatibilty(facetedProject, mockDialogWrapper));
+    verify(mockDialogWrapper, never()).openInformation(anyString(), anyString());
+  }
+
+  @Test
+  public void testCheckFacetCompatibilty_java1_6FacetIsIncompatible() throws CoreException {
```
elharo: Compatibility

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    verify(mockDialogWrapper, never()).openInformation(anyString(), anyString());
+  }
+
+  @Test
+  public void testCheckFacetCompatibilty_web2_5FacetIsCompatible() throws CoreException {
+    facetedProject.installProjectFacet(JavaFacet.VERSION_1_7, null, null);
+    facetedProject.installProjectFacet(WebFacetUtils.WEB_25, null, null);
+    assertTrue(facetedProject.hasProjectFacet(WebFacetUtils.WEB_25));
+
+    assertTrue(commandHandler.checkFacetCompatibilty(facetedProject, mockDialogWrapper));
```
elharo: Compatibility

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+  public void testCheckFacetCompatibilty_java1_6FacetIsIncompatible() throws CoreException {
+    facetedProject.installProjectFacet(JavaFacet.VERSION_1_6, null, null);
+    assertTrue(facetedProject.hasProjectFacet(JavaFacet.VERSION_1_6));
+
+    assertFalse(commandHandler.checkFacetCompatibilty(facetedProject, mockDialogWrapper));
+    verify(mockDialogWrapper, times(1)).openInformation(anyString(), anyString());
+  }
+
+  @Test
+  public void testCheckFacetCompatibilty_java1_8FacetIsIncompatible() throws CoreException {
```
elharo: Compatibility

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    facetedProject.installProjectFacet(JavaFacet.VERSION_1_8, null, null);
+    assertTrue(facetedProject.hasProjectFacet(JavaFacet.VERSION_1_8));
+
+    assertFalse(commandHandler.checkFacetCompatibilty(facetedProject, mockDialogWrapper));
+    verify(mockDialogWrapper, times(1)).openInformation(anyString(), anyString());
+  }
+
+  @Test
+  public void testCheckFacetCompatibilty_web2_4FacetIsIncompatible() throws CoreException {
```
elharo: Compatibility

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    facetedProject.installProjectFacet(JavaFacet.VERSION_1_7, null, null);
+    facetedProject.installProjectFacet(WebFacetUtils.WEB_24, null, null);
+    assertTrue(facetedProject.hasProjectFacet(WebFacetUtils.WEB_24));
+
+    assertFalse(commandHandler.checkFacetCompatibilty(facetedProject, mockDialogWrapper));
+    verify(mockDialogWrapper, times(1)).openInformation(anyString(), anyString());
+  }
+
+  @Test
+  public void testCheckFacetCompatibilty_web3_0FacetIsIncompatible() throws CoreException {
```
elharo: Compatibility

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -0,0 +1,122 @@
...
+    facetedProject.installProjectFacet(JavaFacet.VERSION_1_7, null, null);
+    facetedProject.installProjectFacet(WebFacetUtils.WEB_30, null, null);
+    assertTrue(facetedProject.hasProjectFacet(WebFacetUtils.WEB_30));
+
+    assertFalse(commandHandler.checkFacetCompatibilty(facetedProject, mockDialogWrapper));
```
elharo: Compatibility

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -257,4 +269,51 @@ private static void addWebFacetToBatch(IFacetedProject facetedProject,
     return appEngineRuntimes.toArray(appEngineRuntimesArray);
   }
 
+  /**
+   * Attempts to find a main web application directory, by the following logic:
+   *
+   * 1. If there is no {@code WEB-INF} folder in the {@code project}, returns {@code null}.
+   * 2. Otherwise, if there is at least one {@code WEB-INF} folder that contains {@code web.xml},
+   *     returns the parent directory of one of such {@code WEB-INF} folders.
+   * 3. Otherwise, returns the parent directory of an arbitrary {@code WEB-INF}.
```
elharo: should we also consider the presence/absence of appengine-web.xml when choosing between multiple candidates?

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
chanseokoh: > fixtures? Is that the right word?
Not really. Just something that came up in my mind. Any suggestion that can accommodate facets, natures, and runtimes?

briandealwis:
Several elements on the facet extension point supports an `ignore-problems` attribute:
> &lt;!ELEMENT and (and | or | requires | conflicts)+&gt;
> &lt;!ATTLIST and> ignore-problems CDATA #IMPLIED&gt;
> A facet constraint operator that performs a conjunction over the results of evaluation of two or more sub-expressions. This operator will return true if and only if all sub-expressions evaluate to true.
> 
> *ignore-problems* - Used to control how the framework reports problems when the entities referenced by this constraint are not defined. When a problem is encountered, the framework will prune the constraint expression in a way that maintains as much of the expression as possible while removing the part that has the problem. For instance, if a problem is found in one branch of an OR expression, only that branch is removed. Once the problem has been handled, the framework will (by default) log an error with information about the problem. The framework can be told not to log anything in specific circumstances by using the ignore-problems attribute.
> If the ignore-problems attribute is specified at multiple levels in the constraint expression, the one that's closest to the source of the problem takes affect.
> Currently, the following rules are supported. Multiple rules can be specified by separating them with a comma. The master "not-defined" rule is a short hand notation for specifying all of the *-not-defined rules at the same time.
> 
> not-defined
> facet-not-defined
> facet-version-not-defined
> group-not-defined

I wonder if we could define a `<constraint ignore-problems="not-defined"><requires facet="foo" version="0"/></constraint>` for a non-existent facet?

chanseokoh:
> I wonder if we could define a `<constraint ignore-problems="not-defined"><requires facet="foo" version="0"/></constraint>` for a non-existent facet?

I tested this, and I don't think this works for a couple of reasons. As from the description of the `ignore-problems` attribute, it's only for telling not to log anything:

> The framework can be told not to log anything in specific circumstances by using the ignore-problems attribute.

At the beginning, Eclipse ignores non-existing facets without the attribute. The only difference is whether Eclipse will log the ignore error or not.

Second, even if `ignore-problems` was working as what we wanted, an unsatisfied `<constraint>` doesn't hide facets. The facet UI will kindly tell users what constraints have not been satisfied and why.

chanseokoh:
Since we discussed explicitly modifying the WTP facet settings XML, and from my understanding we agreed that it's a viable alternative. Doesn't seem like an ideal solution, but I guess I have to go with this approach.

elharo:
Sounds good, but please do that in a separate branch in case that doesn't work and we need to come back to this one.

chanseokoh:
Directly modifying the facet metadata file more or less works, but there remains an issue related to classpath entries in order to make it really work. Classpath entries created by GPE have extra attributes specifying that the GPE facets are the owner. I think removing the classpath entries will do the job.

chanseokoh:
Now it seems working. However, I see a number of JARs under `WEB-INF/lib`. I think I'll delete `appengine-api-1.0-sdk` and `appengine-api-labs`. Perhaps not a good idea to delete others.

```
appengine-api-1.0-sdk-1.9.34.jar
appengine-api-labs.jar
appengine-endpoints-deps.jar
appengine-endpoints.jar
appengine-jsr107cache-1.9.34.jar
asm-4.0.jar
datanucleus-api-jdo-3.1.3.jar
datanucleus-api-jpa-3.1.3.jar
datanucleus-appengine-2.1.2.jar
datanucleus-core-3.1.3.jar
geronimo-jpa_2.0_spec-1.0.jar
gwt-servlet.jar
jdo-api-3.0.1.jar
jsr107cache-1.1.jar
jta-1.1.jar
```

chanseokoh:
So, I'm doing a serious of operations to remove GPE remnants:

- Removes classpath entries from GPE.
- Deletes some JARs from GPE.
- Removes the GPE nature.
- Removes the GPE runtime.
- Removes the GPE facets.

One question I'll ask is whether we need to do something such as grabbing a lock during the operations.

chanseokoh:
Hmm... maybe I shouldn't touch the JARs at all. I was a bit confused thinking `appengine-api-1.0-sdk` and `appengine-api-labs` are runtime jars we provide.

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
chanseokoh: I changed the word "fixtures" to "remnants".

chanseokoh:
Boom. Just noticed that removing the GPE runtime doesn't work anymore if we don't have the dummy GPE facets defined. I don't understand the mechanism, but somehow the GPE runtime is connected to the GPE GAE facet. `runtime.getProperty("id")` below returns `null` if we don't have the dummy facets.

```java
    Set<IRuntime> runtimes = facetedProject.getTargetedRuntimes();
    for (IRuntime runtime : runtimes) {
      if (GPE_WTP_GAE_RUNTIME.equals(runtime.getProperty("id"))) {
        facetedProject.removeTargetedRuntime(runtime, null /* monitor */);
      }
    }
 ```

**Comment by chanseokoh** [comment]
Okay, I'll just use the GPE runtime name "Google App Engine" (instead of the runtime ID `com.google.appengine.runtime.id`) to check if the runtime is the GPE one. 
(This seems legitimate, since `FacetedProject.getRuntimeFromName()` actually uses the runtime name as a unique key.)

**Comment by briandealwis** [comment]
The runtimes are connected via a `runtime-component`.

I think it's all or nothing: either we define dummies for GPE facets and runtimes, or we define nothing and do things by hand.

**Comment by chanseokoh** [comment]
I can still make it work to remove the runtime without dummies while using the standard `IFacetedProject` interface, but I think your idea to do it all by hand is actually better in our case. Will do so.

**Comment by chanseokoh** [comment]
Yeah, I noticed I had missed it. Added that in my last commit.

Still working on creating tests. There are some lingering questions too.

File: plugins/com.google.cloud.tools.eclipse.appengine.compat/xslt/wtpMetadata.xsl [hunk 23]
```diff
@@ -0,0 +1,23 @@
...
+<!--
+  This stylesheet removes the following GPE runtime and facets from the WTP facet metadata file:
+
+    <runtime name="Google App Engine"/>
+    <installed facet="com.google.appengine.facet" version="1"/>
+    <installed facet="com.google.appengine.facet.ear" version="1"/>
+
+  The metadata file is ".settings/org.eclipse.wst.common.project.facet.core.xml".
+-->
+<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
```
elharo: which processor are we using? Does it support XSLT 2? Looks to me like you're not using anything that isn't in XSLT 1.

plugins/com.google.cloud.tools.eclipse.appengine.compat/xslt/wtpMetadata.xsl
```diff
@@ -0,0 +1,23 @@
...
+
+  <xsl:template match="node()|@*">
+    <xsl:copy>
+      <xsl:apply-templates select="node()|@*"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <xsl:template match="runtime[@name='Google App Engine']"/>
+  <xsl:template match="installed[@facet='com.google.appengine.facet']"/>
+  <xsl:template match="insatlled[@facet='com.google.appengine.facet.ear']"/>
```
elharo: installed (and we should have a unit test that would catch that)

plugins/com.google.cloud.tools.eclipse.appengine.compat/xslt/wtpMetadata.xsl
chanseokoh: I don't know what the processor means here, but from what you said, I guess XSLT 1 is sufficient. Changed the version to "1.0".

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorXsltTest.java
```diff
@@ -0,0 +1,115 @@
...
+  @Test
+  public void testApplyXslt()
+      throws IOException, ParserConfigurationException, SAXException, TransformerException {
+    try (InputStream xmlStream = stringToInputStream(WTP_METADATA_XML);
+        InputStream stylesheetStream = stringToInputStream(STYLESHEET)) {
+
+      DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+      Document document = builder.parse(xmlStream);
+
+      try (InputStream inputStream = GpeMigrator.applyXslt(document, stylesheetStream)) {
```
elharo: This API seems strange. Why do we pass a stylesheet to GpeMigrator? Shouldn't this be an internal implementation detail? I.e. why should the caller know or care that GpeMigrator uses XSLT?

chanseokoh:
Now things are more or less working. For its state, this PR is ready for review. (However, I feel there still exist a few things to consider further.)

To summarize what this PR does,

- The conversion menu appears for any project that doesn't have the App Engine Standard facet. It appears under _Configure_.
selection_001
It simply calls a method to add the App Engine facet. (Very simple.) The conversion is done silently. It's done via a job per @briandealwis's suggestion. (FYI, [due to the way we install our runtimes](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1167#issue-197753071), it takes a lot of time for the entire conversion process to "complete".)

- If a project being converted does not have any of the Java 1.7 facet and the Dynamic Web Module facet 2.5, the facets will be installed together. If a project has incompatible facets, users will see an error dialog.

- Changed the way we configure the main webapp directory when we install the Dynamic Web Module facet as part of installing the App Engine Standard facet. Previously, we were always set it to `/src/main/webapp`. Now it searches for existing `WEB-INF` folders. (The logic could be improved over time.)

- Adds `.compat` bundle: this bundle is strictly for GPE stuff. The conversion menu will attempt to remove the obsolete GPE nature, runtime, and facets (if they exist). (There might be more things we could remove? We will see.)

- When removing the GPE runtime and facets, it directly modifies the WTP facet metadata settings XML file (using XSLT transfomation). The file is `.settings/org.eclipse.wst.common.project.facet.core.xml`.

- `test-projects/GPE-project.zip` is a GPE project I generated using GPE. (Slightly modified by hand too.)

PTAL @elharo @briandealwis

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorXsltTest.java
chanseokoh: This is an internal method visible only for testing (protected). GpeMigrator has only one public method, removeObsoleteGpeRemnants().

Issue #1127, opened 2016-12-14
Open most important file on Wizard Completion


elharo:
In maven and new project wizards

1. Eclipse suggests a readme.html file explaining what has been generated
2. We should open the most important file in the wizard. Maybe HelloAppEngine.java?
3. We should switch to the Java perspective.

The New Resource and Import Wizards commonly create new files, folders, and projects within the workbench. If a single file is created, the wizard should open the file in an editor in the active page. If more than one file is created, open the most important, or central file. This makes it easier to modify and save the file.

Have a readme.html file for every example project, and open that readme.html automatically upon project creation. This will give users an immediate overview of the example: what it does, prerequisites, limitations, steps to take, and so on.

branflake2267:
I might suggest proving a link in the readme.html to a guide online. I've found guides online are easier to maintain, meaning there updates might not be tied down by a release. Docs online are easier to update by writers when they see fit.

elharo:
The readme.html describes the example project, and that is tied to a specific release. It's not for the plugin as a whole.

PR #1192, opened 2017-01-07
Open pom.xml in editor on wizard completion

elharo:
@chanseokoh For the non-Maven project we open a .java file instead. I haven't yet figured out how to guess from an archetype which file is the most important we should open.


chanseokoh:
Maybe we could have a list of files and try to get them in order. (I'm not suggesting this in this PR though, because I can't find suitable files than `pom.xml` either. I see the helloworld archetype and the guestbook archetype both have `README.md`. However, the content of the README files don't fit in the IDE context.)

plugins/com.google.cloud.tools.eclipse.ui.util/META-INF/MANIFEST.MF
```diff
@@ -32,6 +32,7 @@ Import-Package: com.google.cloud.tools.eclipse.ui.util,
  org.eclipse.ui.browser;ui.workbench=split,
  org.eclipse.ui.console,
  org.eclipse.ui.handlers,
+ org.eclipse.ui.ide,
```
chanseokoh: Since we moved this util code, can we remove this from the newproject `MANIFEST.MF`?

elharo: Tried that. Turned out this is used for other reasons in other classes not in this PR.

chanseokoh:
Fixes #1127.

PR #1198, opened 2017-01-08
don't rule out 4.7

### LINKED ISSUE #1154: Mention 4.7
Name should be "Google Cloud Platform for Eclipse 4.5, 4.6, and 4.7" or 4.5 and later.

elharo:
@briandealwis fix #1154


elharo:
Until I started working with Eclipse, I never remembered which names went with which versions. I think version numbers are a little more obvious for most devs.

PR #1201, opened 2017-01-09
Don't set Java App Engine SDK path in App Engine runtime

### LINKED ISSUE #1185: Why do we set Java App Engine SDK path in App Engine runtime?
[We set Java App Engine SDK Path in the App Engine runtime](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java#L204) when we create the runtime for the first time if it doesn't exist. The actual path value is "&lt;Cloud SDK home&gt;/platform/google_appengine/google/appengine/tools/java/lib".

It doesn't seem like we make any use of this path in our code. Local run works fine too without it. ~~We could probably remove it, which also allows removing dependency on appengine-plugins-core in `.appengine.facets` bundle.~~ UPDATED: `AppEngineFacetInstallDelegate` requires appengine-plugins-core for validation.

chanseokoh:
Fixes #1185.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.eclipse.appengine.facets;
 
-import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.eclipse.util.FacetedProjectHelper;
 import com.google.common.base.Preconditions;
 import java.util.ArrayList;
@@ -52,7 +51,8 @@ public class AppEngineStandardFacet {
   public static final String ID = "com.google.cloud.tools.eclipse.appengine.facets.standard";
 
   static final String VERSION = "1";
-  static final String DEFAULT_RUNTIME_ID = "com.google.cloud.tools.eclipse.appengine.standard.runtime";
+  static final String DEFAULT_RUNTIME_ID =
+      "com.google.cloud.tools.eclipse.appengine.standard.runtime";
   static final String DEFAULT_RUNTIME_NAME = "App Engine Standard";
   public static final String DEFAULT_APPENGINE_SDK_VERSION = "1.9.46";
   public static final String DEFAULT_GCLOUD_PLUGIN_VERSION = "2.0.9.133.v201611104";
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -184,32 +184,24 @@ public static void installAllAppEngineRuntimes(IFacetedProject project, IProgres
...
+      IProgressMonitor monitor) throws CoreException {
     IRuntimeType appEngineRuntimeType =
         ServerCore.findRuntimeType(AppEngineStandardFacet.DEFAULT_RUNTIME_ID);
     if (appEngineRuntimeType == null) {
-      throw new NullPointerException("Could not find " + AppEngineStandardFacet.DEFAULT_RUNTIME_NAME + " runtime type");
+      throw new NullPointerException(
+          "Could not find " + AppEngineStandardFacet.DEFAULT_RUNTIME_NAME + " runtime type");
     }
 
     IRuntimeWorkingCopy appEngineRuntimeWorkingCopy
```
elharo: where does this runtime get the Cloud SDK from now? Looking at this PR, it seems to me I was wrong earlier the Cloud SDK is indeed required at this stage.

chanseokoh: Let's see... hope I understood your question correctly. I _think_ currently the runtime doesn't (and doesn't need to) get, set, or provide the Cloud SDK. [The `LocalAppEngineServerBehavior` directly creates an `CloudSdkAppEngineDevServer` instance.](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java#L338)

briandealwis: The runtime definition is actually in [`.localserver`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml#L51) and it uses the [`ServletClasspathProvider`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java) to obtain the necessary jars from Maven Central.

PR #1207, opened 2017-01-09
don't check for updates during stage to work avoid incorrect fix message

elharo:
@chanseokoh


chanseokoh:
Sorry, what problem does this fix specifically? I see the IntelliJ plugin isn't setting this, so I wonder what this achieves.

briandealwis:
I think it avoids a distracting message on the deploy console if their Cloud SDK is out of date.

PR #1208, opened 2017-01-09
Add bundles for App Engine project conversion

chanseokoh:
For #348, add `.appengine.conversion` bundle and its test counterpart.

This is a mere skeleton, so no user-visible impact.


plugins/com.google.cloud.tools.eclipse.appengine.conversion.test/src/com/google/cloud/tools/eclipse/appengine/conversion/PluginXmlTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+import com.google.cloud.tools.eclipse.test.util.BasePluginXmlTest;
+import org.junit.Test;
+import org.w3c.dom.NodeList;
+
+public class PluginXmlTest extends BasePluginXmlTest {
+
+  @Test
+  public void testExtensionPoint() {
+    NodeList extensions = getDocument().getElementsByTagName("extension");
```
elharo: I seem to recall that Eclipse doesn't like plugins that define no extensions

briandealwis: Maybe we should have a test to ensure there is no `plugin.xml`

chanseokoh: I'll just remove `plugin.xml` from this PR. Will add back that when needed (soon).

chanseokoh:
After discussing with @briandealwis ([related comment](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1188#issuecomment-271689159)), we agree on not adding a new bundle. Closing.

PR #1211, opened 2017-01-10
test at least one extension point exists

elharo:
@chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
@@ -43,23 +43,23 @@ public class BasePluginXmlTest {
     doc = pluginXmlDocument.get();
   }
 
-  protected Document getDocument() {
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -43,23 +46,21 @@ public void setUp() {
...
+  protected final Document getDocument() {
     return doc;
-  }
-  
-  // Generic tests that should be true of all plugin.xml files
+  } 
   
   @Test
-  public void testRootElementIsPlugin() {
+  public final void testRootElementIsPlugin() {
```
chanseokoh: Hmm... These are just tests. Is adding `final` to unit tests our new convention?

elharo: No, it isn't; but in this particular case since this is a base class I wanted to make sure these tests didn't get turned off or revised; at least not without everyone realizing that's what they were doing.

PR #1216, opened 2017-01-11
clear status on empty string

### LINKED ISSUE #1212: Warning on delete of cloud SDK location
1. Type in a nonexistent directory path into the SDK location field.
2. See the warning up top.
3. Delete the nonexistent directory path from the SDK location field.

The warning should go away or change to an info that a cloud SDK location is required. However the old warning about the now deleted path remains.

image

elharo:
@chanseokoh fix #1212


plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/CloudSdkPreferenceAreaTest.java
```diff
@@ -43,11 +43,11 @@
  */
 @RunWith(MockitoJUnitRunner.class)
 public class CloudSdkPreferenceAreaTest {
-  @Mock
-  IPreferenceStore preferences;
+  
+  @Mock IPreferenceStore preferences;
```
chanseokoh: Can be `private`.

PR #1220, opened 2017-01-11
Fix test class names

chanseokoh:
Tests don't run if not ending with `Test`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/AbstractProjectTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/AbstractProjectTest.java
@@ -30,7 +30,7 @@ import org.junit.BeforeClass;
 /**
  * Common infrastructure for workbench-based tests that create a single project.
  */
-public class AbstractProjectTests {
+public class AbstractProjectTest {
 
   protected static SWTWorkbenchBot bot;
   protected IProject project;
--- a/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/BaseProjectTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/BaseProjectTest.java
@@ -30,7 +30,7 @@ import org.junit.BeforeClass;
 /**
  * Common infrastructure for workbench-based tests that create a single project.
  */
-public class AbstractProjectTest {
+public class BaseProjectTest {
 
   protected static SWTWorkbenchBot bot;
   protected IProject project;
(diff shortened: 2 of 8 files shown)


plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/DebugNativeAppEngineStandardProject.java
```diff
@@ -52,7 +52,7 @@
  * terminate.
  */
 @RunWith(SWTBotJunit4ClassRunner.class)
-public class DebugNativeAppEngineStandardProject extends AbstractProjectTests {
+public class DebugNativeAppEngineStandardProject extends AbstractProjectTest {
```
elharo: DebugNativeAppEngineStandardProjectTest

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/AbstractProjectTest.java
```diff
@@ -30,7 +30,7 @@
 /**
  * Common infrastructure for workbench-based tests that create a single project.
  */
-public class AbstractProjectTests {
+public class AbstractProjectTest {
```
elharo: I'm surprised the name of the superclass matters, but c'et la vie. However this class should be abstract (or renamed).

akerekes:
I haven't tested it, but JUnit 4 doesn't care about the name of the class,
it only looks for the annotations to be present. The reason these tests
aren't executed still is probably the Maven Surefire plugin's default
inclusion filter:
http://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/AbstractProjectTest.java
chanseokoh: I did another test to verify this: as long as the inheriting class ends with `Test`, the name of the superclass doesn't matter.

Anyway, I renamed this to `BaseProjectTest`. (Not `Abstract`.)

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/DebugNativeAppEngineStandardProject.java
chanseokoh: Good catch.

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/BaseProjectTest.java
```diff
@@ -30,7 +30,7 @@
 /**
  * Common infrastructure for workbench-based tests that create a single project.
  */
-public class AbstractProjectTest {
+public class BaseProjectTest {
```
elharo: could still make this abstract if you like

chanseokoh: There's no abstract method in this class, so I'll keep it `BaseProjectTest`. (It's not my code, so I don't immediately see how I could convert this to an abstract class.)

elharo: abstract classes are not required to have abstract mthods

PR #1221, opened 2017-01-12
Use UrlEscapers.urlFormParameterEscaper()

### LINKED ISSUE #1071: UrlEscapers.urlFormParameterEscaper() instead of URLEncoder.encode()
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java#L276

Verify that we have no problem sending Analytics hits after this change.

elharo:
fix #1071 but not yet ready for review. Looks like we may need to fix a problem in Guava first.


plugins/com.google.cloud.tools.eclipse.usagetracker/META-INF/MANIFEST.MF
```diff
@@ -11,4 +11,5 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="3.11.1",
  org.eclipse.ui.workbench,
  com.google.cloud.tools.eclipse.preferences,
  com.google.cloud.tools.eclipse.util
-Import-Package: com.google.common.annotations;version="20.0.0"
+Import-Package: com.google.common.annotations;version="20.0.0",
+ com.google.common.net;version="15.0.0"
```
chanseokoh: Should this be `20.0.0`?

elharo:
@briandealwis do you happen to know if it's actually necessary to Import-Package com.google.common.escaper in MANIFEST.MF  here?

The code does not explicitly import anything from that package. However it calls


UrlEscapers.urlFormParameterEscaper().escape(entry.getValue()))

and UrlEscapers.urlFormParameterEscaper() returns a com.google.common.escaper.Escaper even though we never store it into a variable of that type.

chanseokoh:
I tested this at the current state, and for some reason, the class cannot be found at runtime:

problem occurred _008

```
java.lang.NoClassDefFoundError: com/google/common/net/UrlEscapers
	at com.google.cloud.tools.eclipse.usagetracker.AnalyticsPingManager.getParametersString(AnalyticsPingManager.java:274)
	at com.google.cloud.tools.eclipse.usagetracker.AnalyticsPingManager.sendPingHelper(AnalyticsPingManager.java:204)
	at com.google.cloud.tools.eclipse.usagetracker.AnalyticsPingManager.access$1(AnalyticsPingManager.java:201)
	at com.google.cloud.tools.eclipse.usagetracker.AnalyticsPingManager$2.run(AnalyticsPingManager.java:95)
Caused by: java.lang.ClassNotFoundException: com.google.common.net.UrlEscapers cannot be found by com.google.cloud.tools.eclipse.usagetracker_0.1.0.qualifier
	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:439)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:352)
	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:344)
	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:160)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:361)
	... 5 more
```

I wonder if this is related to this warning that the method is not API:

selection_009

briandealwis:
@chanseokoh the code at c535dc2 (which imports `com.google.common.escape`) works for me.

Why is importing `com.google.common.escape`necessary?  Because you're using a class (`Escaper`), the class's definition needs to be available. The JVM tries to resolve the class definition using `AnalyticsPingManager`'s class-loader, which is generated from its bundle, and thus the lookup fails without the `com.google.common.escape` package import.

The JDT complaint about API problems is a bit misleading — it's complaining that accessing the `Escaper` class definition is forbidden.  PDE generates the project's classpath by adding the `com.google.guava` bundle and adding access rules to expose the specified packages and forbidding access to all other packages.  Forbidding access to classes is usually done for non-API classes, and hence the message.

screen shot 2017-01-12 at 3 03 41 pm

chanseokoh:
@briandealwis The Analytics code is enabled only at production. (It should have the Analytics tracking ID injected and not in the Dev mode.)

As a hacky workaround only for testing, you can force `boolean analyticsEnabled = true` in `AnalyticsPingManager.getInstance()`. You should have opted in for usage reporting too. To trigger the code, open any of the two new project wizards.

briandealwis:
I copied the `UrlEscapers` line into `AnalyticsPingManager`'s constructor and verified it was executed.

chanseokoh:
@briandealwis You're right. It's working and I can see the pings delievered. Perhaps I wasn't at the latest commit.

LGTM.

PR #1224, opened 2017-01-12
Simplify code by removing StandardJobDeployJob builder

chanseokoh:
No one calls `Builder.exporter()`, `Builder.staging()`, and `Builder.deployer()`, and we always use `ExplodedWarPublisher`, `StandardProjectStaging`, and `AppEngineProjectDeployer`. In fact, I think we don't even need the fields for these. Even if we may need to provide alternatives for these in the future, it could be YAGNI.

Let me know what you think.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
@@ -60,31 +60,15 @@ public class StandardDeployJob extends WorkspaceJob {
 
   private static final Logger logger = Logger.getLogger(StandardDeployJob.class.getName());
 
-  private final ExplodedWarPublisher exporter;
-  private final StandardProjectStaging staging;
-  private AppEngineProjectDeployer deployer;
-
   //temporary way of error handling, after #439 is fixed, it'll be cleaner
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployJob.java
```diff
@@ -101,8 +85,10 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
 
       try {
         getJobManager().beginRule(config.getProject(), progress);
-        exporter.publish(config.getProject(), explodedWarDirectory, progress.newChild(10));
-        staging.stage(explodedWarDirectory, stagingDirectory, cloudSdk, progress.newChild(20));
+        new ExplodedWarPublisher().publish(
+            config.getProject(), explodedWarDirectory, progress.newChild(10));
+        new StandardProjectStaging().stage(
```
elharo: whenever I see

new Foo(/* no args */).doSomething() 

that suggests to me doSomething() should be a static method, but that's for another PR

akerekes:
The intent with the builder was to make the job independent of the actual implementation and make it testable.

chanseokoh:
> The intent with the builder was to make the job independent of the actual implementation and make it testable.

So, since we're not doing it, okay to remove them?

elharo:
YAGNI. I say remove them.

Issue #354, opened 2016-07-13
Release script


elharo:
A release script and documentation, preferably though not necessarily in the form of a maven goal, that publishes the update site. 

Note that multiple decisions need to be made (e.g. where to host the update site) before we start coding this.

akerekes:
Another point we should consider is to be able to publish a new version without hiding the latest published, something like a testing/beta update site.

PR #1225, opened 2017-01-12
Add semi-auto script for the second release part

chanseokoh:
The script covers steps from 14 in the release instructions. I used it to make the last release, so I know it's working. It's semi-automatic and interactive, as some steps need to get some values such as a path to an Eclipse executable or a release version string. The script will certainly help a lot, since the steps are long, can be error-prone and time-consuming.

The script is safe (in terms of public visibility) to cancel out anytime and re-run, unless you execute the final step which makes things visible to public.

I don't think I follow the Google-style, if it exists. I won't mind if we don't merge this. For example, this could be kept in a separate branch.

Part of #354

### DIFF
--- a/release/release_after_kokoro.sh
+++ b/release/release_after_kokoro.sh
@@ -89,12 +89,12 @@ echo "#     $LOCAL_REPO/plugins/*"
 echo "#"
 ask_proceed
 
+echo "The following steps require running an Eclipse binary."
 while true; do
-    echo "The following steps require running an Eclipse binary."
     echo -n "Enter the Eclipse home path: "
     read ECLIPSE_HOME
     if [ -x "$ECLIPSE_HOME/eclipse" ]; then break; fi
-    echo "'$ECLIPSE_HOME/eclipse' not executable. Try again."
+    echo "'$ECLIPSE_HOME/eclipse' is not an executable. Try again."
 done
 
 set -x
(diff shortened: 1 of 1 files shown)


release/release_after_kokoro.sh
```diff
@@ -0,0 +1,195 @@
...
+echo "#"
+echo "#     $LOCAL_REPO/artifacts.xml"
+echo "#     $LOCAL_REPO/content.xml"
+echo "#     $LOCAL_REPO/features/*"
+echo "#     $LOCAL_REPO/plugins/*"
+echo "#"
+ask_proceed
+
+echo "The following steps require running an Eclipse binary."
+while true; do
```
elharo: I'm tempted to suggest that ECLIPSE_HOME should be set before running the script. Maybe if not set, default to the path on Goobuntu.

release/release_after_kokoro.sh
```diff
@@ -0,0 +1,195 @@
...
+echo "#     $LOCAL_REPO/plugins/*"
+echo "#"
+echo "# into 'gs://cloud-tools-for-eclipse/<VERSION>/'."
+echo "#"
+ask_proceed
+
+while true; do
+    echo -n "Enter version: "
+    read VERSION
+    echo -n "Verify version: "
```
elharo: Is it really necessary to verify the version? Or maybe only verify the version if it looks off via regexp?

release/release_after_kokoro.sh
chanseokoh: Yeah, failing fast seems better. The version of the bundled Eclipse on my Laptop is too old (3.8), so I'd rather require explicitly setting ECLIPSE_HOME.

release/release_after_kokoro.sh
chanseokoh: Ok, reordring by putting `read VERSION` before `ask_proceed` seems sufficient.

PR #1231, opened 2017-01-14
Add "About" feature branding information

briandealwis:
Cause `com.google.cloud.tools.eclipse.appengine.ui` to be a _branding plugin_ for CTE, to contribute to the _About_ dialog.  Although this doesn't seem the best bundle to use, I used it as it's the bundle that's hosting our images.

The use of an `about.properties` is required for the `{featureVersion}` substitution to be performed.

screen shot 2017-01-13 at 11 40 14 pm

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.ui/about.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.ui/about.properties
@@ -1,0 +1,8 @@
+blurb=Google Cloud Tools for Eclipse\n\
+Version: {featureVersion}\n\
+\n\
+Copyright 2016-2017 Google Inc.\n\
+\n\
+Cloud Tools for Eclipse is a Google-sponsored plugin that adds \
+support for the App Engine standard environment to the Eclipse IDE.\n\
+Visit: https://cloud.google.com/eclipse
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.ui/about.properties
```diff
@@ -0,0 +1,8 @@
+blurb=Google Cloud Tools for Eclipse\n\
+Version: {featureVersion}\n\
+\n\
+Copyright 2016-2017 Google Inc.\n\
+\n\
+Cloud Tools for Eclipse is a Google-sponsored plugin that adds \
+support for the App Engine standard environment to the Eclipse IDE.\n\
```
akerekes: We could use a more general text here instead of this narrow scope which will need to be updated whenever we add Flex, GKE, GCE. Maybe `support for the Google Cloud Platform environments`, or something along these lines?

chanseokoh: I agree, as the name "CT4E" suggests.

elharo: I prefer the current text. Right now it supports App Engine standard environment only. The time to make the message different is when we add support for additional environments, not before.

briandealwis: I copied that text from our doc page (https://cloud.google.com/eclipse/docs).

PR #1232, opened 2017-01-14
Align fields in deploy dialog

### LINKED ISSUE #676: Align deploy preferences
Figure out how all the parts of this dialog should align with each other:

image

elharo:
@briandealwis by unnesting composites

fix #676 

image

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
@@ -117,7 +117,7 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
 
     Dialog.applyDialogFont(this);
 
-    this.setLayout(gridLayout);
+    setLayout(gridLayout);
     
     loadPreferences(project);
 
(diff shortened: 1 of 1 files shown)


briandealwis:
Looking at this dialog, the checkbox for _Specify version_ seems odd.  We could drop the checkbox and instead use `Text#setMessage()` to set a greyed helper text within the version box?

_If unspecified a version will be automatically generated_

elharo:
Personally I do not find the GridLayoutFactory code easier to read or write. I do not have the defaults at my fingertips, and prefer to be very explicit about what is being set. 

Good point about the Specify version checkbox. That's independent of these changes though so let's consider it in a separate PR.

If you're proposing a completely different approach to solving the problem, which is what I think you're doing; it would help if you could use a different branch/PR. The CLA bot seems to get confused when multiple people push to a branch,  and I don;t have an easy time sorting out the git history myself.

briandealwis:
It's not a different approach, I just tweaked some of the changes: the`widthHint` masked some missing `SWT.FILL`s.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -114,8 +117,8 @@ public StandardDeployPreferencesPanel(Composite parent, IProject project,
 
     Dialog.applyDialogFont(this);
 
-    GridLayoutFactory.fillDefaults().generateLayout(this);
-
+    this.setLayout(gridLayout);
```
chanseokoh: `this.` can be removed?

Issue #263, opened 2016-06-19
slf4j warnings


elharo:
These warnings litter our build and are annoying and confusing and pointless:

SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.

Figure out a way to get rid of them without introducing any new dependencies.

briandealwis:
Should be able to include `org.eclipse.m2e.maven.runtime.slf4j.simple` in the ide-target-platform

briandealwis:
Sadly including `org.eclipse.m2e.maven.runtime.slf4j.simple ` was not the easy fix I had hoped — it is a hack to pull in a logger.  As Logback was already in Orbit, I just pulled it in instead.

chanseokoh:
Reopening. I have a bad feeling that we will keep reopening this as we continue to add bundles.

briandealwis:
I can think of a few approaches to handling this.  One is to create our own internal minimal testing feature that includes these minimal bundles.  Another is to gather different types of tests into different directories (e.g., `/tests/workbench-based`, `/tests/simple`?) and then use the module inheritance to configure the dependencies.

PR #1237, opened 2017-01-16
Quieten Maven builds by specifying a logback configuration for tests

briandealwis:
Although bringing in Logback for the tests (#1143) did rid us of the SLF4j errors (#263), the cure was worse than the disease as Logback's [default configuration logs at the `DEBUG` level](http://logback.qos.ch/manual/configuration.html#automaticConf), and our logs are now clogged with m2e-related `DEBUG`-level detail.  This patch creates a new fragment, `com.google.cloud.tools.eclipse.test.logback`, hosted against `ch.qos.logback.classic`, that provides a `logback.xml` configuration file that sets the default logging level to `INFO`.

Logback can be configured either by setting system property to point to a `logback.xml`, or by finding the file on its classpath.  Maven doesn't provide a way to easily reference a file (requires using the _assembly_ plugin or something).  We could configure .travis.yml to reference a file, but we'd need to specify it when building locally too.

We could simplify the setup slightly by having a feature that pulls in `ch.qos.logback.slf4j` (the SLF4j binding configuration), `ch.qos.logback.classic`, and this new `com.google.cloud.tools.eclipse.test.logback` bundle, so that the test `pom.xml`s specify the feature rather than 2 plugins.  As this is a build-time testing issue, it doesn't seem much of a win.

### Asides ###

We must use SLF4j from Orbit as M2Eclipse explicitly pulls in Orbit's SLF4j by a `Require-Bundle: org.sfl4j.api`; SLF4j has since added OSGi metadata but unfortunately chose to name its bundle `slf4j.api`.  Orbit only provides SLF4j logging bindings for Logback and _log4j 1.2_; the `org.slf4j.jul`, `org.slf4j.jcl`, and `org.slf4j.log4j` bundles are logging facades.

Using _log4j 1.2_ without a configuration it produces similar output to #263, and so would also require a configuration fragment.
```
log4j:WARN No appenders could be found for logger (org.eclipse.m2e.core.internal.project.registry.ProjectRegistryRefreshJob).
log4j:WARN Please initialize the log4j system properly.
```
Since this is a build-time test-only issue anyways, and since log4j 1.2 is ancient, we'll just stay with Logback.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
@@ -1,0 +1,13 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: %pluginName
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.logback
+Bundle-Version: 0.1.0.qualifier
+Fragment-Host: ch.qos.logback.classic
+Bundle-Vendor: %providerName
+Bundle-Localization: plugin
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-ActivationPolicy: lazy
+Import-Package: org.slf4j.impl
+Eclipse-ExtensibleAPI: true
+Eclipse-BundleShape: dir
--- a/plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
@@ -9,5 +9,4 @@ Bundle-Localization: plugin
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-ActivationPolicy: lazy
 Import-Package: org.slf4j.impl
-Eclipse-ExtensibleAPI: true
 Eclipse-BundleShape: dir
--- a/plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
-Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.logback
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.logback;singleton:=true
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: ch.qos.logback.classic
 Bundle-Vendor: %providerName
--- a/plugins/com.google.cloud.tools.eclipse.test.logback/logback.xml
+++ b/plugins/com.google.cloud.tools.eclipse.test.logback/logback.xml
@@ -1,0 +1,12 @@
+<configuration scan="true"> 
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender"> 
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <logger name="com.google.cloud.tools" level="debug"/>
+  <root level="info">
+    <appender-ref ref="STDOUT" />
+  </root>
+</configuration>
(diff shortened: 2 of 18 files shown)


plugins/com.google.cloud.tools.eclipse.test.logback/logback.xml
```diff
@@ -0,0 +1,12 @@
+<configuration scan="true"> 
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender"> 
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
```
chanseokoh: I don't know much about this settings file, but I wonder if we can leave this out and use the default pattern (for the sake of not adding something we don't really need).

plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,13 @@
...
+Bundle-Name: %pluginName
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.logback
+Bundle-Version: 0.1.0.qualifier
+Fragment-Host: ch.qos.logback.classic
+Bundle-Vendor: %providerName
+Bundle-Localization: plugin
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-ActivationPolicy: lazy
+Import-Package: org.slf4j.impl
+Eclipse-ExtensibleAPI: true
```
chanseokoh: I looked up for this directive. Question: does this mean the `.test.logback` bundle allow other bundles that have `.test.logback` as a fragment host (i.e., bundles that have `Fragment-Host: ... .test.logback`) to extend APIs being exported by `.test.logback`? Or, does it allow the `.test.logback` bundle to extend APIs exposed by `ch.qos.logback.classic`?

plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,13 @@
...
+Eclipse-BundleShape: dir
```
chanseokoh: I am interested in what this means and why we need it here.

plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
briandealwis: `ExtensibleAPI` is a directive for PDE to allow `.test.logback` to see other fragments extending the same bundle host. This line is from an earlier experiment and isn't necessary; I'll remove it.

plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
briandealwis: `BundleShape: dir` tells p2 to install bundle as an exploded jar (the contents) rather than as a jar.  It means end-users can easily modify the contained `logback.xml`.  Since we're not shipping this jar, it's not so useful, but it doesn't hurt.

plugins/com.google.cloud.tools.eclipse.test.logback/logback.xml
briandealwis: Unfortunately it doesn't seem that we can.

plugins/com.google.cloud.tools.eclipse.test.logback/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,12 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: %pluginName
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.logback
```
chanseokoh: Do we need `;singleton:=true` or not?

chanseokoh: We don't have `plugin.xml`, so I guess not.

briandealwis: It can't hurt.

briandealwis: Since this bundle is for test-only purposes, it doesn't much matter.  But `singleton` isn't out of place since we don't want to have multiple `logback.xml`s.

PR #1238, opened 2017-01-16
Add favicon.ico for new native project

### LINKED ISSUE #213: favicon.ico
Add this to generated projects, if for no other reason than to stop spamming the log files and the console with 404 errors when the browser asks for it.

briandealwis:
Pull in `favicon.ico` from GPE.  [All modern browsers load `favicon.ico` if no icon `<link>` is found](http://stackoverflow.com/questions/1344122/favicon-png-vs-favicon-ico-why-should-i-use-png-instead-of-ico/1344379#1344379), so no need to add a `<link>` to the templates.

Fixes #213 

(I also fixed up how the progress monitor was calculated in `CodeTemplates` as the `createFoldersForPackage()` calls into `createChildFolder()` for each segment, and so the accumulation could be off.)


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/templates/appengine/AppEngineTemplateUtility.java
```diff
@@ -64,6 +63,22 @@ public static void createFileContent(
...
   }
 
+  public static void copyFileContent(String outputFileLocation, String sourceName)
+      throws CoreException {
+    Preconditions.checkNotNull(outputFileLocation, "output file is null");
+    Preconditions.checkNotNull(sourceName, "source file name is null");
+
+    try (InputStream inputStream =
+        AppEngineTemplateUtility.class.getResourceAsStream("/templates/appengine/" + sourceName)) {
+      File outputFile = new File(outputFileLocation);
```
elharo: you don't require a separate File object here. You can create the stream directly from the path, and then you could create this in the try() opener so it's autoclosed too.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -94,17 +93,21 @@ public static IFile materialize(IProject project, AppEngineStandardProjectConfig
...
     createChildFile("appengine-web.xml", AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
-        webinf, subMonitor, properties);
+        webinf, subMonitor.newChild(5), properties);
 
     Map<String, String> packageMap = new HashMap<>();
     String packageValue = config.getPackageName().isEmpty() ? "" : config.getPackageName() + ".";
     packageMap.put("package", packageValue);
-    createChildFile("web.xml", AppEngineTemplateUtility.WEB_XML_TEMPLATE, webinf, subMonitor,
+    createChildFile("web.xml", AppEngineTemplateUtility.WEB_XML_TEMPLATE, webinf,
+        subMonitor.newChild(5),
```
chanseokoh: Merge this line with the following line?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -94,17 +93,21 @@ public static IFile materialize(IProject project, AppEngineStandardProjectConfig
...
         packageMap);
 
-    createChildFile("index.html", AppEngineTemplateUtility.INDEX_HTML_TEMPLATE, webapp, subMonitor,
+    createChildFile("index.html", AppEngineTemplateUtility.INDEX_HTML_TEMPLATE, webapp,
+        subMonitor.newChild(5),
```
chanseokoh: Merge this line with the following line?

briandealwis:
I created [a new favicon from the blue hexagonal App Engine icon](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/6eedb37682a5d44a2a8c40fa3cd013c089448f37/plugins/com.google.cloud.tools.eclipse.util/templates/appengine/favicon.ico?raw=true) which seemed fitting since this template is specific to App Engine.  The blank hexagon looked kinda weird.  But the "Generic GCP" might work (see below)?  @arouzrokh do you have a preference?

generic-gcp

arouzrokh:
Let's keep with the GAE one please if possible. The top one @briandealwis 

Thanks!
Amir

PR #1242, opened 2017-01-18
test what we bundle in our binaries

### LINKED ISSUE #1161: Test if necessary resources are exported into final binary (i.e., included in build.properties)
For example, `.welcome/plugin.xml`, `.appengine.newproject/helpContexts.xml`, `.welcome/intro/cloud-tools-for-eclipse.xml`, etc should be specified in `build.properties` so that they are included in a final binary. At the minimum, check if some well-known files such as `plugin.xml` are always included.

Adding an integration test that covers all the bundles? Add a new test bundle?

An ideal solution would be to make a debug env reflect the final binary (i.e., a debug run can't find resources if not included very much like in the prod), but not sure if this is possible.

elharo:
fix #1161 We've forgotten to add files to bin.includes shockingly often.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.whitelist/plugin.xml
@@ -4,7 +4,7 @@
 
   <extension
       id="appEngineWhitelist"
-      name="App Engine JRE Whitelist"
+      name="%extension.name"
       point="org.eclipse.jdt.core.compilationParticipant">
      <compilationParticipant
          createsProblems="true"
--- a/plugins/com.google.cloud.tools.eclipse.preferences/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.preferences/plugin.xml
@@ -1,13 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
-   <extension-point id="com.google.cloud.tools.eclipse.preferences.areas" name="Preference Areas" schema="schema/com.google.cloud.tools.eclipse.preferences.areas.exsd"/>
+   <extension-point id="com.google.cloud.tools.eclipse.preferences.areas" 
+                    name="%extension.name" 
+                    schema="schema/com.google.cloud.tools.eclipse.preferences.areas.exsd"/>
 
    <extension
          point="org.eclipse.ui.preferencePages">
       <page
             id="com.google.cloud.tools.eclipse.preferences.main"
-            name="Google Cloud Tools"
+            name="%extension.page.name"
             class="com.google.cloud.tools.eclipse.preferences.areas.AreaBasedPreferencePage">
       </page>
    </extension>
@@ -21,7 +23,7 @@
          point="com.google.cloud.tools.eclipse.preferences.areas">
       <area
             host="com.google.cloud.tools.eclipse.preferences.main"
-            title="Usage Statistics"
+            title="%analytics.optin.area.title"
             class="com.google.cloud.tools.eclipse.preferences.AnalyticsOptInArea"
             preferences="configuration://com.google.cloud.tools.eclipse.preferences">
       </area>
--- a/plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
@@ -11,7 +11,7 @@ Require-Bundle: org.junit;bundle-version="4.12.0",
  org.eclipse.ui.ide
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
  com.google.common.base;version="20.0.0",
- com.google.common.collect;version="15.0.0",
+ com.google.common.collect;version="20.0.0",
  javax.servlet,
  org.eclipse.core.commands,
  org.eclipse.core.expressions,
--- a/plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
@@ -11,6 +11,7 @@ Require-Bundle: org.junit;bundle-version="4.12.0",
  org.eclipse.ui.ide
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
  com.google.common.base;version="20.0.0",
+ com.google.common.collect;version="15.0.0",
  javax.servlet,
  org.eclipse.core.commands,
  org.eclipse.core.expressions,
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
@@ -19,9 +19,7 @@ package com.google.cloud.tools.eclipse.test.util;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Properties;
+import java.util.Set;
 
 import org.eclipse.core.runtime.IExtensionPoint;
 import org.eclipse.core.runtime.IExtensionRegistry;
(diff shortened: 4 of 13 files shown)


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -30,14 +35,20 @@
...
 @SuppressWarnings("restriction")
-public class BasePluginXmlTest {
+public abstract class BasePluginXmlTest {
 
-  @Rule public final PluginXmlDocument pluginXmlDocument = new PluginXmlDocument();
-  @Rule public final PluginProperties pluginProperties = new PluginProperties();
+  @Rule
+  public final PluginXmlDocument pluginXmlDocument = new PluginXmlDocument();
+  @Rule
+  public final EclipseProperties pluginProperties = new EclipseProperties("plugin.properties");
```
briandealwis: How does this work if no `plugin.properties` was required?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -87,6 +98,33 @@ public final void testKeywordsDefined() {
...
+    Assert.assertTrue(includes.contains("plugin.xml"));
+    Assert.assertTrue(includes.contains("plugin.properties"));
+    Assert.assertTrue(includes.contains("."));
+    Assert.assertTrue(includes.contains("META-INF/"));
+    
+    testIncludedIfPresent(includes, "helpContexts.xml");
+    testIncludedIfPresent(includes, "icons/");
+    testIncludedIfPresent(includes, "lib/");
+    testIncludedIfPresent(includes, "README.md/");
+    testIncludedIfPresent(includes, "epl-v10.html/");
```
briandealwis: `epl-v10.html` is a file, not a directory.

Checking for these files is expanding the scope of what I'd expect from`BasePluginXmlTests`, but I really like the idea — it's like a PDE-lint.  We could do some more sophisticated checking.  For example, if `epl-v10.html` is found, ensure it has a [`Bundle-License` header](https://www.osgi.org/bundle-headers-reference/) pointing to `https://www.eclipse.org/legal/epl-v10.html`.  If it has a `Bundle-License`, ensure it points to a valid library.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -87,6 +98,33 @@ public final void testKeywordsDefined() {
...
+    testIncludedIfPresent(includes, "OSGI-INF/");
+    testIncludedIfPresent(includes, "fragment.xml");
```
briandealwis: And test for `fragment.properties`?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -110,7 +148,7 @@ private void assertPropertiesDefined(Element element) {
 
   private void assertPropertyDefined(String name) {
```
briandealwis: `assertPropertyExternalized?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
elharo: Effectively it's now required.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
elharo: done

elharo: yes, we can certainly expand the tests here. I think first I may want to break out testing plugin.xml from testing build.properties into separate classes.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
elharo: I'm not sure "externalized" is the right word here. This test is checking that a property used in plugin.xml is present in plugin.properties with no typos or misspellings. build.properties isn't involved in this one.

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/plugin.xml
```diff
@@ -4,7 +4,7 @@
 
   <extension
       id="appEngineWhitelist"
-      name="App Engine JRE Whitelist"
+      name="%extension.name"
```
chanseokoh: Does this work? I see `Bundle-Localization: plugin` in `MANIFEST.MF` in many bundles, but not in this bundle. I may be wrong about this though.

plugins/com.google.cloud.tools.eclipse.preferences/plugin.xml
```diff
@@ -1,13 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
-   <extension-point id="com.google.cloud.tools.eclipse.preferences.areas" name="Preference Areas" schema="schema/com.google.cloud.tools.eclipse.preferences.areas.exsd"/>
+   <extension-point id="com.google.cloud.tools.eclipse.preferences.areas" 
+                    name="%extension.name"
```
chanseokoh: Yeah, this doesn't work. You need `Bundle-Localization: plugin`.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
briandealwis: Since you're looking for the leading '%', it's about externalized labels.

plugins/com.google.cloud.tools.eclipse.preferences/plugin.xml
briandealwis: That should be tested too :-)

elharo: #1257

plugins/com.google.cloud.tools.eclipse.appengine.whitelist/plugin.xml
elharo: done

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
elharo: To my way of thinking a test that a property is externalized would look on the relevant element and verify that it begins with a %.

This test is finding externalized strings that do begin with a percent are defined in the plugin.properties file.

chanseokoh:
You need `Bundle-Localization: plugin` in `com.google.cloud.tools.eclipse.preferences` too.

plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
```diff
@@ -11,6 +11,7 @@ Require-Bundle: org.junit;bundle-version="4.12.0",
  org.eclipse.ui.ide
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
  com.google.common.base;version="20.0.0",
+ com.google.common.collect;version="15.0.0",
```
chanseokoh: I wonder if this can be "20.0.0". Why does Eclipse keep suggesting 15?

Issue #1243, opened 2017-01-18
Shrunk input fields in native project wizard 


chanseokoh:
![selection_001](https://cloud.githubusercontent.com/assets/10523105/22052518/a6a7f3da-dd18-11e6-91fe-5e96be1fdff3.png)

PR #1240, opened 2017-01-17
align two fields in new project wizard

elharo:
@briandealwis @hray2003 I'm not 100% sure this one's worth it, since I haven't figured out how to line up with the widgets we inherit from WizardNewProjectCreationPage, but it maybe looks a little better:

image

PR #1244, opened 2017-01-18
Fix input field length in native wizard

chanseokoh:
The problem started with #1240, when the line was removed (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1240#discussion_r96426626). I believe the removed line was setting `GridData` and not `GridLayout`, so it is necessary.


chanseokoh:
Fixes #1243.

PR #1252, opened 2017-01-19
Show unique validation UI message as INFO when bucket name is empty, while disabling dialog submit

### LINKED ISSUE #1120: Invalid Bucket Name
When no bucket name is entered, this message should say "Enter a bucket name" not "Invalid bucket name" and it should not be an error, though it should prevent the info from being applied.

image

chanseokoh:
Fixes #1120. The message is "Empty bucket name", following the pattern "Empty project ID".

I did some hackery to achieve showing the validation error as INFO while disabling the submit buttons. Not sure if it's worth doing so.

- Empty string:
deploy to app engine standard _002

- Invalid string:
deploy to app engine standard _005

- Empty string:
properties for a _003

- Invalid string:
properties for a _004


elharo:
The staging bucket is optional. What if we simply treat an empty staging bucket as no value and proceed as normal?

chanseokoh:
> The staging bucket is optional. What if we simply treat an empty staging bucket as no value and proceed as normal?

That can be another option, with which we can get away from this kind of hackery. We will probably remove the checkbox too? (I remember I saw a recent issue to remove the checkbox for project ID, but can't find it now.)

Another option to avoid this hackery is to just let the empty bucket message be shown as an error — just the current state.

I have a slight preference for whichever option to avoid this hackery.

elharo:
An Info level problem should not disable the button. However, the more I think about this, the more I think an empty field should just be the default bucket.

briandealwis:
+1

chanseokoh:
Okay, so let's back out and allow submitting empty version (and bucket) strings. (That is, we just show an INFO message, not an ERROR message.) No complication in code.

> Maybe they're just better called _Overrides_.

I vaguely remember they were originally called _Overrides_, but later we changed the labels. (Not 100% sure.)

elharo:
I don't think we even need an Info message for this. The default covers most users. We don't want to give them the idea that they should change this unless they really know what they're doing.

chanseokoh:
Closing per https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1120#issuecomment-275755416. It's much cleaner to open a new PR.

Issue #1253, opened 2017-01-19
Add signing verification steps


briandealwis:
Our post-signing build steps should verify that our signed jars are signed.  Something like this should work:

```bash
for jar in path/to/final/plugins/com.google.cloud.tools.eclipse.*.jar \
      path/to/final/features/com.google.cloud.tools.eclipse.*.jar; do
   if ! jarsigner -strict -verify "$jar" >/dev/null; then
      echo "unsigned artifact: $jar"
      exit 1
   fi
done
```

We can test installing the feature by installing with the p2 directory with the `eclipse.p2.unsignedPolicy=fail` system property with the following:
```bash
$ eclipse -application org.eclipse.equinox.p2.director \
   -repository file:$PWD/gcp-repo/target/repository,http://download.eclipse.org/releases/mars \
   -destination /tmp/test.signedness \
   -profile SDKProfile \
   -installIU com.google.cloud.tools.eclipse.suite.e45.feature.feature.group \
   -vmargs -Declipse.p2.unsignedPolicy=fail

org.eclipse.m2e.logback.configuration: The org.eclipse.m2e.logback.configuration bundle was activated before the state location was initialized.  Will retry after the state location is initialized.
Installing com.google.cloud.tools.eclipse.suite.e45.feature.feature.group 0.1.0.201701181923.
Installation failed.
An error occurred during the org.eclipse.equinox.internal.p2.engine.phases.CheckTrust phase.
 session context was:(profile=SDKProfile, phase=org.eclipse.equinox.internal.p2.engine.phases.CheckTrust, operand=, action=).
[...]

PR #1255, opened 2017-01-19
Verify if jars are all signed

chanseokoh:
Part of #1253. I noticed verifying jars takes quite some time, so I decided to let `jarsigner` print verification results.

I also automated verification of constant injection.
 
Output looks like below:

```
Operation completed over 29 objects/6.4 MiB.
+ set +x
#
# Now verifying whether jars are all signed.
#
com.google.cloud.tools.eclipse.appengine.deploy_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.deploy.ui_0.1.0.201701161800.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.facets_0.1.0.201701112147.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.libraries_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.localserver_0.1.0.201701112147.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.login_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.newproject_0.1.0.201701081354.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.newproject.maven_0.1.0.201701081354.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.ui_0.1.0.201701160540.jar: jar verified.
com.google.cloud.tools.eclipse.appengine.whitelist_0.1.0.201610241636.jar: jar verified.
com.google.cloud.tools.eclipse.jdt.launching_0.1.0.201610262044.jar: jar verified.
com.google.cloud.tools.eclipse.jst.server.core_0.1.0.201611181529.jar: jar verified.
com.google.cloud.tools.eclipse.preferences_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.sdk_0.1.0.201701112147.jar: jar verified.
com.google.cloud.tools.eclipse.sdk.ui_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.ui.util_0.1.0.201701111857.jar: jar verified.
com.google.cloud.tools.eclipse.usagetracker_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.util_0.1.0.201701122020.jar: jar verified.
com.google.cloud.tools.eclipse.welcome_0.1.0.201701171430.jar: jar verified.
com.google.cloud.tools.eclipse.3rdparty.feature_0.1.0.201612060008.jar: jar verified.
com.google.cloud.tools.eclipse.suite.e45.feature_0.1.0.201701171430.jar: jar verified.

#
# Verify if constants have been injected...
#
Proceed ([Y]/n)? 
  private static final java.lang.String OAUTH_CLIENT_ID = "... redacted ....";
  private static final java.lang.String OAUTH_CLIENT_SECRET = "... redacted ...";
  public static final java.lang.String ANALYTICS_TRACKING_ID = "... redacted ...;

Looks good, but check the output above once more. Proceed ([Y]/n)? 
```

### DIFF
--- a/release/release_after_kokoro.sh
+++ b/release/release_after_kokoro.sh
@@ -69,6 +69,18 @@ if [ 0 -eq $( ls -1 $SIGNED_DIR | wc -l ) ]; then
     exit 1
 fi
 
+echo "#"
+echo "# Now verifying whether jars are all signed."
+echo "#"
+for jar in $SIGNED_DIR/plugins/com.google.cloud.tools.eclipse.*.jar \
+		$SIGNED_DIR/features/com.google.cloud.tools.eclipse.*.jar; do
+    echo -n "$( basename $jar ): "
+	if ! jarsigner -strict -verify $jar; then
+        echo "unsigned artifact. Halting."
+        exit 1
+	fi
+done
+
 ###############################################################################
 echo
 echo "#"
@@ -76,14 +88,24 @@ echo "# Verify if constants have been injected..."
 echo "#"
 ask_proceed
 
-set -x
-javap -private -classpath $SIGNED_DIR/plugins/com.google.cloud.tools.eclipse.appengine.login_0.1.0.*.jar \
-       -constants com.google.cloud.tools.eclipse.appengine.login.Constants
-javap -classpath $SIGNED_DIR/plugins/com.google.cloud.tools.eclipse.usagetracker_0.1.0.*.jar \
-       -constants com.google.cloud.tools.eclipse.usagetracker.Constants
-set +x
+LOGIN_CONSTANTS=$( javap -private -classpath \
+    $SIGNED_DIR/plugins/com.google.cloud.tools.eclipse.appengine.login_*.jar \
+    -constants com.google.cloud.tools.eclipse.appengine.login.Constants \
+    | grep OAUTH_CLIENT_ )
+ANALYTICS_CONSTANT=$( javap -classpath \
+    $SIGNED_DIR/plugins/com.google.cloud.tools.eclipse.usagetracker_*.jar \
+    -constants com.google.cloud.tools.eclipse.usagetracker.Constants \
+    | grep ANALYTICS_TRACKING_ID )
+echo -e "$LOGIN_CONSTANTS\n$ANALYTICS_CONSTANT"
 
-echo "Check the output above. "
+if echo -e "$LOGIN_CONSTANTS\n$ANALYTICS_CONSTANT" | \
+        grep -q '@oauth.client\|@ga.tracking.id@'; then
+	echo "Some constant(s) have not been injected. Halting."
+	exit 1
+fi
+
+echo
+echo -n "Looks good, but check the output above once more. "
 ask_proceed
 
 ###############################################################################
--- a/release/release_after_kokoro.sh
+++ b/release/release_after_kokoro.sh
@@ -3,22 +3,26 @@
 # This script starts with Step 14 in the release process (i.e., after Kokoro
 # has built and signed the binaries.)
 #
-# It is safe to re-run this script again when something goes wrong, provided
+# It is safe to rerun this script again when something goes wrong, provided
 # that you didn't complete the final step, which means you are done. (The final
-# step grants public access to files uploaded to cloud-tools-for-eclipse.)
+# step grants public access to the files uploaded to cloud-tools-for-eclipse.)
 #
 # Note: this script clears and creates "$HOME/CT4E_release_work" (after
-# confirmation) for temporary work directory, which can be inspected later
+# confirmation) for a temporary work directory, which can be inspected later
 # and/or deleted safely anytime.
 
-if [ -z "$ECLIPSE_HOME" ]; then
-    echo 'The release steps require running an Eclipse binary.'
-    echo 'Set $ECLIPSE_HOME before running this script.'
+die() {
+    for line in "$@"; do
+        echo $line
+    done
     exit 1
-elif [ ! -x "$ECLIPSE_HOME/eclipse" ]; then
-    echo "'$ECLIPSE_HOME/eclipse' is not an executable. Halting."
-    exit 1
-fi
+}
+
+[ -z "$ECLIPSE_HOME" ] && \
+    die 'The release steps require running an Eclipse binary.' \
+        'Set $ECLIPSE_HOME before running this script.'
+[ ! -x "$ECLIPSE_HOME/eclipse" ] && \
+    die "'$ECLIPSE_HOME/eclipse' is not an executable. Halting."
 
 WORK_DIR=$HOME/CT4E_release_work
 SIGNED_DIR=$WORK_DIR/signed
@@ -65,9 +69,8 @@ mkdir $SIGNED_DIR && \
 set +x
 
 if [ 0 -eq $( ls -1 $SIGNED_DIR | wc -l ) ]; then
-    echo "'$SIGNED_DIR' is empty. Wrong GCS bucket URL provided?"
-    echo "Fix the problem and rerun the script. Halting."
-    exit 1
+    die "'$SIGNED_DIR' is empty. Wrong GCS bucket URL provided?" \
+        "Fix the problem and rerun the script. Halting."
 fi
 
 echo "#"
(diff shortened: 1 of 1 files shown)


release/release_after_kokoro.sh
```diff
@@ -69,21 +70,62 @@ if [ 0 -eq $( ls -1 $SIGNED_DIR | wc -l ) ]; then
...
+    grep OAUTH_CLIENT_ )
+ANALYTICS_CONSTANT=$( javap -classpath \
+    $SIGNED_DIR/plugins/com.google.cloud.tools.eclipse.usagetracker_*.jar \
+    -constants com.google.cloud.tools.eclipse.usagetracker.Constants | \
+    grep ANALYTICS_TRACKING_ID )
+echo -e "$LOGIN_CONSTANTS\n$ANALYTICS_CONSTANT"
+
+# Verify if the constants have been injected.
+if echo -e "$LOGIN_CONSTANTS\n$ANALYTICS_CONSTANT" | \
+        grep -q '@oauth.client\|@ga.tracking.id@'; then
```
briandealwis: I thought `|` is only in extended REs and so requires `egrep`?  In this case, you might be better off using `expr`
```
if [ `expr "${LOGIN_CONSTANTS}${ANALYTICS_CONSTANT}" : @" = 0 ]; then
```
Or instead perform validation using known patterns (below).

release/release_after_kokoro.sh
```diff
@@ -69,21 +70,62 @@ if [ 0 -eq $( ls -1 $SIGNED_DIR | wc -l ) ]; then
...
+    echo "Some constant(s) have not been injected. Halting."
+    exit 1
+fi
 
-echo "Check the output above. "
+# Verify if the injected constants have the right lengths.
```
briandealwis: Checking lengths seems a bit risky. A better check might be to `curl` against a real endpoint on `accounts.google.com` and `google-analytics.com`?

release/release_after_kokoro.sh
```diff
@@ -69,21 +70,62 @@ if [ 0 -eq $( ls -1 $SIGNED_DIR | wc -l ) ]; then
...
+OAUTH_CLIENT_ID=$( echo "$LOGIN_CONSTANTS" | grep "OAUTH_CLIENT_ID" | \
+    sed -e 's/.* = "\(.*\)"; */\1/' )
+OAUTH_CLIENT_SECRET=$( echo "$LOGIN_CONSTANTS" | grep "OAUTH_CLIENT_SECRET" | \
+    sed -e 's/.* = "\(.*\)"; */\1/' )
+ANALYTICS_TRACKING_ID=$( echo "$ANALYTICS_CONSTANT" | \
+    sed -e 's/.* = "\(.*\)"; */\1/' )
+if [ ${#OAUTH_CLIENT_ID} -ne 72 ]; then
+    echo "The length of injected OAUTH_CLIENT_ID is not 72. Halting."
```
briandealwis: `OAUTH_CLIENT_ID` seems to be a pseudo domain-name matching `^[0-9]*-[0-9a-z]*\.apps\.googleusercontent\.com$`
`OAUTH_CLIENT_SECRET` seems to be alphanumeric.
`ANALYTICS_TRACKING_ID` matches `^UA-[0-9]*-1`

release/release_after_kokoro.sh
chanseokoh: AFAIK, the OAUTH_ strings are merely identifiers and have nothing to do with actual endpoints. They are unique, fixed identifiers generated only once for our product, CT4E.

For the Analytics tracking ID, there is no way to check if a certain ID exists or works online. It cannot be probed. Injecting the correct ID is so important, so I would like to make sure that the length is correct, to catch the situation where any of us slipped one character when copy-pasting.

I think we are okay with these brittle checks. If we had no problem exposing the constants to the public, I would have embedded the constants in the script to compare if the values are exact-match.

release/release_after_kokoro.sh
chanseokoh: That'll work, but I really like to check the length. I think stricter checks are better; if any of us missed even one character when copy-pasting these constants, users won't be able to log in.

release/release_after_kokoro.sh
chanseokoh: You can use `|` in regular REs. No need for extended REs.

Your use of `expr` seems very interesting, so I'll tinker with it.

chanseokoh: Hmm... [`expr` does _anchored_ pattern matching](http://stackoverflow.com/questions/16251579/expr-string-matching-mixed-number-and-chars), so it needs some tweaks. But TIL from the URL that bash natively supports RE matching.

briandealwis: Good point re: `expr` anchoring — my tests all had '@' at the beginning or not at all. :blush:

Hmm, something strange is going on as `|` is an ordinary character for basic REs.  Here's what I see on both OS X and Ubuntu:
```sh
$ echo wheels | grep 'foo|ls'
$ echo wheels | egrep 'foo|ls'
wheels
```

But your script uses /bin/sh so avoid bashisms.

release/release_after_kokoro.sh
briandealwis: Really what we want to check is that the builder didn't mistakenly use the codes from CT4IJ.  Maybe we should compare an SHA256 hash?

release/release_after_kokoro.sh
chanseokoh: For `grep`, you need to escape `|`, like `grep 'foo\|ls'`. But then I see `egrep` looks better. I'll use `egrep`. Also, I'll avoid the bash RE matching support.

release/release_after_kokoro.sh
chanseokoh: The builder doesn't fetch the codes. We manually drag-select, copy, and paste the codes into the command-line when running `mvn clean install`. We could use SHA256 for perfection, but I think that's a bit overkill, considering how we copy-paste the codes. What do you think?

elharo:
What does it look like if the verification fails? The general UI principle here is that we should show a big obvious error if something fails, and nothing at all if it succeeds and no further attention is required. I'm not sure how easy that is to accomplish in bash though.

chanseokoh:
If the verification fails, the script halt immediately at the first jar that is unsigned like below:

```
com.google.cloud.tools.eclipse.preferences_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.sdk_0.1.0.201701112147.jar: jar verified.
com.google.cloud.tools.eclipse.sdk.ui_0.1.0.201701132149.jar: jar verified.
com.google.cloud.tools.eclipse.ui.util_0.1.0.201701111857.jar: jar verified.
com.google.cloud.tools.eclipse.usagetracker_0.1.0.201612081359.jar: jar is unsigned.
Unsigned artifact found. Halting.
```

If some constants are not injected or have wrong length, the script will halt with the proper message at that point too:
```
#
# Verify if constants have been injected...
#
Proceed ([Y]/n)? 
  private static final java.lang.String OAUTH_CLIENT_ID = "@oauth.client.id@";
  private static final java.lang.String OAUTH_CLIENT_SECRET = "@oauth.client.secret@";
  public static final java.lang.String ANALYTICS_TRACKING_ID = "... <redacted> ...";
Some constant(s) have not been injected. Halting.
```

```
#
# Verify if constants have been injected...
#
Proceed ([Y]/n)? 
  private static final java.lang.String OAUTH_CLIENT_ID = "... <redacted> ...";
  private static final java.lang.String OAUTH_CLIENT_SECRET = "... <redacted> ...";
  public static final java.lang.String ANALYTICS_TRACKING_ID = "... <redacted> ...";
OAUTH_CLIENT_SECRET is not of length 24. Halting.
```

release/release_after_kokoro.sh
```diff
@@ -102,7 +102,7 @@ ANALYTICS_CONSTANT=$( javap -classpath \
 echo -e "$LOGIN_CONSTANTS\n$ANALYTICS_CONSTANT"
 
 # Verify if the constants have been injected.
-if [[ "${LOGIN_CONSTANTS}${ANALYTICS_CONSTANT}" =~ "@" ]]; then
```
briandealwis: FTR, I'm happy with bash'isms for this script — that's nice and compact. Just be sure to change the first line to `#!/bin/bash` though.

PR #1272, opened 2017-01-23
try jacoco

elharo:


elharo:
@briandealwis

briandealwis:
That looks positive!  If you add `.travis.yml` hook, we should see if the end-to-end hookup works, right?

briandealwis:
> That looks positive! If you add .travis.yml hook, we should see if the end-to-end hookup works, right?

Erm never mind: I see that's already there.  But it's [apparently not quite enough](https://codecov.io/gh/GoogleCloudPlatform/google-cloud-eclipse/commit/1c464165d106f68d2953ea050a35e75063db3dc2).

elharo:
The hook's already checked in.

briandealwis:
- JaCoCo configures the `tycho.argLine` property to specify its run-agent but it is ignored as we specify an explicit `<argLine>`.  So we explicitly bring it in.
  - I separated out the JaCoCo into a separate profile to simplify testing this, and then added it explicitly for the Oxygen build.
  - Tycho Surefire is bound to the `verify` goal, not `test`.

With this, I now see JaCoCo reports:
```
$ mvn -Pjacoco verify
…
[INFO] All tests passed!
[INFO] 
[INFO] --- jacoco-maven-plugin:0.7.8:report (report) @ com.google.cloud.tools.eclipse.appengine.deploy.test ---
[INFO] Loading execution data file /Users/bsd/Manumitting/Projects/GooglePluginEclipse/repo-gcloud-eclipse/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/target/jacoco.exec
[INFO] Analyzed bundle 'com.google.cloud.tools.eclipse.appengine.deploy.test' with 10 classes
```

briandealwis:
Hmm, [we seem to be getting coverage assessments of the test code](https://codecov.io/gh/GoogleCloudPlatform/google-cloud-eclipse/tree/128daf7842e0bfe0d43ee8d566875b84af15d3f0/plugins), rather than the code tested.

briandealwis:
[From the JaCoCo wiki](https://github.com/jacoco/jacoco/wiki/MavenMultiModule):
> Current Situation
> 
> The current JaCoCo Maven goals work on single modules only: Tests are executed within the module and contribute coverage only to code within the same module. Coverage reports are created for each module separately. There is no built-in support for cross-module coverage or combined reports for multiple modules.

briandealwis:
The underlying issue is that codecov.io needs generated JaCoCo's XML reports, and JaCoCo reporting needs to be able to access the source.

Found a recent [article on hooking up JaCoCo with a Tycho project](http://www.lorenzobettini.it/2017/02/jacoco-code-coverage-and-report-of-multiple-eclipse-plug-in-projects/).  Unfortunately it requires explicitly listing the set of bundles and test.

There is a way to run the reports via an [Ant task](https://dzone.com/articles/jacoco-maven-multi-module) but it still requires enumerating the source and class locations.

briandealwis:
This latest commit uses the `report-aggregate` goal with an explicit list of dependencies.  I included a script to generate a list of dependencies with appropriate scope. 

Ideally [`jacoco-maven-plugin:report-aggregate`](https://github.com/jacoco/jacoco/blob/master/jacoco-maven-plugin/src/org/jacoco/maven/ReportAggregateMojo.java#L162) would extract the reactor projects that match some filter.  But that's a fair bit of work.

build/jacoco/pom.xml
```diff
@@ -0,0 +1,319 @@
...
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.google.cloud.tools.eclipse</groupId>
+    <artifactId>build</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
+    <relativePath>../</relativePath>
+  </parent>
+  <artifactId>jacoco</artifactId>
+  <version>0.1.0-SNAPSHOT</version>
```
elharo: do we need to keep using 0.1.0 in new bundles?

briandealwis: It's not a bundle, and the result isn't meant for outside consumption.  Using < 1.0.0 indicates it's not API.

briandealwis:
> @elharo: I don't know why but github isn't letting me approve this. :-(

I think it's because you're the original author.  I'll approve it on your behalf.

PR #1278, opened 2017-01-24
Fixes #1274 - App Engine libraries in wrong order on JVM 8

### LINKED ISSUE #1274: App Engine libraries in wrong order on JVM 8
If the Eclipse instance was launched with JVM 7, the App Engine libraries are ordered as follows:

image

If the Eclipse instance was launched with JVM 8, the App Engine libraries are ordered as follows:

image

Notice `App Engine API` being the last for JVM 8.
The reason for this is that the libraries are stored in a Hashmap in the wizard which will not maintain insertion order. It should be a LinkedHashmap (or something other that maintains predictable ordering).

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
```diff
@@ -66,6 +69,29 @@ public void run() {
...
 
   @Test
+  public void testButtonOrder() {
+    syncExec(new Runnable() {
+
+      @Override
+      public void run() {
+        Control groupAsControl = shell.getChildren()[0];
+        assertThat(groupAsControl, instanceOf(Group.class));
+        Control[] buttonsAsControls = ((Group)groupAsControl).getChildren();
```
chanseokoh: One space after the `(Group)` casting.

plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
```diff
@@ -66,6 +69,29 @@ public void run() {
...
+        String[] expectedLibraryOrder = new String[]{ "appengine-api", "appengine-endpoints", "objectify" };
+        for (int i = 0; i < buttonsAsControls.length; i++) {
+          Control control = buttonsAsControls[i];
+          assertThat(control, instanceOf(Button.class));
+          Button button = (Button)control;
```
chanseokoh: Ditto.

plugins/com.google.cloud.tools.eclipse.appengine.ui.test/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineLibrariesSelectorGroupTest.java
```diff
@@ -66,6 +69,29 @@ public void run() {
...
+          assertNotNull(button.getData());
+          assertThat(button.getData(), instanceOf(Library.class));
```
chanseokoh: Since we have this, do we really need `assertNotNull()` above?

akerekes: The minor difference is that `assertNotNull` throws an AssertionError while without it we would get a NullPointerException. The former suggests to me that the developer thought about that case, while the latter suggests that null wasn't considered as a possible value.

chanseokoh: Hmm... `null instanceof(Object.class)` doesn't throw `NullPointerException`. I guess then `assertThat` can throw it?

akerekes: you're right, it also throws AssertionError (I assumed there would be an NPE which was false). I remove the nullcheck.

PR #1279, opened 2017-01-24
Use maven wrapper to use Maven 3.3.9

### LINKED ISSUE #1276: Test failures do not make Travis build fail
https://travis-ci.org/GoogleCloudPlatform/google-cloud-eclipse/jobs/194884620#L6687

Possible cause:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=490890

This is how Code Recommenders worked around it: http://git.eclipse.org/c/recommenders/org.eclipse.recommenders.git/commit/?id=5cd08804815230b9e6fb42002de0b6437eef5151

Another possibility is that we're running an older Tycho that is using the older Equinox.
Upgrading Tycho has a problem with our toolchain config.
From @briandealwis :

>  Problem is that Tycho 0.26 (0.25?) requires Java 8. And so our tycho.toolchains won't work on Travis.
> Ah, we can have a .travis.yml rule to copy in a configured ~/.m2/toolchains.xml, and use the docker image to figure out what needs to be done
> I have part of this done as I explored going pomless a while back but hit this issue

Also Travis uses an older Maven version.

briandealwis:
Fixes #1276

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 sudo: false
 language: java
 install: 
+  # install the maven wrapper
+  - mvn -N io.takari:maven:wrapper -Dmaven=3.3.9
   # download Cloud SDK
   - wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-139.0.1-linux-x86_64.tar.gz
   - tar -xzf google-cloud-sdk-139.0.1-linux-x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


chanseokoh:
Give me a second. I'm testing on Travis if this actually fails with the original test @akerekes wrote.

.travis.yml
```diff
@@ -1,6 +1,8 @@
 sudo: false
 language: java
 install: 
+  # install the maven wrapper
+  - mvn -N io.takari:maven:wrapper -Dmaven=3.3.9
```
elharo: Could we have a comment explaining why we do this and what it is? I know in 6 months I won't remember.

Issue #910, opened 2016-10-28
Avoid `Display#syncExec()` in tests


briandealwis:
Requiring our tests to use `syncExec()` adds extraneous information to the tests.  There must be a better way to ensure the test is run within the UI thread.

diegomarquezp:
Closing as not planned.

PR #1280, opened 2017-01-24
Remove syncexec from AppEngineLibrariesSelectorGroupTest

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.ui.test/pom.xml
```diff
@@ -18,6 +18,11 @@
       <plugin>
         <groupId>org.eclipse.tycho</groupId>
         <artifactId>tycho-surefire-plugin</artifactId>
+        <configuration>
+          <useUIHarness>true</useUIHarness>
+          <useUIThread>true</useUIThread>
```
chanseokoh: FYI, this is true be default, but I think it's informative to have this line here.

chanseokoh:
#910

PR #1282, opened 2017-01-25
Put version bounds on Guava and Gson

briandealwis:
With Guava 21 making some substantial API/ABI changes, we should put upper bounds on our Guava imports and requires.  For good measure, I also added bounds to `com.google.gson` too.


chanseokoh:
So I see it's `version` if `Import-Package`, and `bundle-version` if `Require-Bundle`.

I wonder if the gson version should be `[2.7.0,3.0.0)`, since its OSGi bundle version is 2.7.0 while its Maven version is 2.7. Maybe OSGi is smart enough to ignore the patch version.

briandealwis:
> Maybe OSGi is smart enough to ignore patch versions (i.e., the last digit).

It is.

PR #1288, opened 2017-01-26
separate plugin.xml and build.properties tests

### LINKED ISSUE #1281: Cannot test build.properties if bundle doesn't have plugin.xml
If a bundle doesn't have `plugin.xml`, it cannot test `build.properties` with [`BasePluginXmlTest.testBuildProperties`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java#L101), because `BasePluginXmlTest` will outright fail if `plugin.xml` doesn't exist. We may need to factor out the `build.properties` portion to its own class.

elharo:
fix #1281 @chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/BuildPropertiesTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/BuildPropertiesTest.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2016 Google Inc.
+ * Copyright 2017 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *    http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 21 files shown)


chanseokoh:
> Could the tests instead use assume()

Interesting idea.

chanseokoh:
> > Could the tests instead use assume()
>
> Interesting idea.

But then we can't test if `plugin.xml` is missing (although not having `plugin.xml` outright means the bundle is useless, which would be obvious to discover).

elharo:
The larger these get the more sense it makes to split them out. Shoving the properties test into the same base class was a bad hack I committed to avoid having to write subclasses. The real issue is that as @chanseokoh originally pointed out these are really independent things. Conditional test logic is a code smell: http://xunitpatterns.com/Conditional%20Test%20Logic.html

briandealwis:
Agreed: my second approach explicitly encodes whether a `plugin.xml` should exist in a field (_ shouldHavePluginXml_) that can be used to validate that the bundle matches the intended state.

briandealwis:
The plugin.xml tests may be independent of build.properties tests right now, but as we develop these further they won't be.

The ideal to my mind would be a single class that I create that verifies this is a well-formed bundle, where I have to explicitly toggle off things that I feel don't apply.  So it should test that there's a `Bundle-Activation: lazy` header, that if `Bundle-{Name,Vendor,Description}` start with an `%` then there should be a `Bundle-Localization` header, if there's a `Bundle-Localization: XXX` header then there's a `XXX.properties` file and it's in the `build.properties`.

PR #1294, opened 2017-01-26
Generate web.xml when installing Dynamic Web Module facet

### LINKED ISSUE #1291: "Convert to App Engine Standard Project" needs to add missing web.xml
I used our new "Convert to App Engine Standard Project" menu item to convert a new empty, plain vanilla Java project. It added WEB-INF and appengine-web.xml but not web.xml. So when I went to the now enabled "Run As > App Engine" menu item:

image

chanseokoh:
Fixes #1291.

WTP doesn't override if `web.xml` already exists, as shown below (WTP source code):
```java
    private void createWeb25DeploymentDescriptor(final IProject project, final IProjectFacetVersion fv, 
                                               IFolder webinfFolder, IProgressMonitor monitor) throws CoreException {
       // Create the deployment descriptor (web.xml) if one doesn't exist
       IFile webxmlFile = webinfFolder.getFile("web.xml"); //$NON-NLS-1$
       if (!webxmlFile.exists()) {
           try {
               // Create a minimal web.xml file, so the model can be initialized
               final String webXmlContents = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<web-app id=\"WebApp_ID\" version=\"2.5\" xmlns=\"http://java.sun.com/xml/ns/javaee\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xsi:schemaLocation=\"http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd\">\n</web-app>"; //$NON-NLS-1$
               webxmlFile.create(new ByteArrayInputStream(webXmlContents.getBytes("UTF-8")), true, monitor); //$NON-NLS-1$
           
               populateDefaultContent(project, fv);
           } catch (UnsupportedEncodingException e) {
        	   WebPlugin.logError(e);
           }
       }
   }
```

I verified that our new (native and Maven) projects do have `web.xml` we provide, and that the "Convert to App Engine project" menu adds the WTP default `web.xml`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallationTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallationTest.java
@@ -21,16 +21,23 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 import com.google.cloud.tools.eclipse.test.util.project.ProjectUtils;
+import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.List;
+import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IFile;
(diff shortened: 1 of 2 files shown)


chanseokoh:
It doesn't override existing `web.xml`. I verified it. See the original comment above for details.

chanseokoh:
Okay, let me try.

PR #1298, opened 2017-01-27
update objectify

### LINKED ISSUE #1297: Guava 18 --> 20 for Objectify
Adding Objectify to a non-maven project adds Guava 18.

image

elharo:
@akerekes fix #1297


akerekes:
Objectify 5.1.14 depends still on Guava 18: https://github.com/objectify/objectify/blob/17e07c195bad9007f047711eddfbed0de2bee5d7/pom.xml#L184

elharo:
Guess I was looking at HEAD. In any case it should work with 20:
I.e. they updated without changing anything else:

https://github.com/objectify/objectify/commit/39a7d9cf4f4bcf74d01742afc5c21d1a77cae6d4

I don't think we want to put an old version of Guava in users's classpath if we can help it.

PR #1303, opened 2017-01-27
No override check boxes and allow empty version and bucket on deploy preference dialog

### LINKED ISSUE #1120: Invalid Bucket Name

chanseokoh:
Fixes #1120.

However, this introduces a bug that I can't figure out why it's happening or how to fix it, so I need some help. Everything works as expected, except that the deploy dialog doesn't display the error message "select an account" when no account was selected, but only initially. Funny things about this bug are:

1. The changes in this PR have nothing to do with the account selector or its validation. I guess something was off from the beginning.
1. Although the error message is not shown, the dialog is correctly prevented from being submitted; only the error message is not displayed. So I'm roughly guessing that the dialog gets the error status but somehow the dialog message is reset by someone.
1. If you make any change to any of the fields or check boxes, the bug disappears onward.

@akerekes @briandealwis any ideas?


chanseokoh:
The issue that the error message is not shown intially even though there is a validation error is that `TitleAreaDialogSupport` thinks UI has never changed, which is true:

From `TitleAreaDialogSupport.handleStatusChanged()`:
```
		if (type == IMessageProvider.ERROR) {
			dialog.setMessage(null);
			dialog.setErrorMessage(uiChanged ? message : null);
			if (currentStatus != null && currentStatusHasException()) {
				handleStatusException();
			}
```

`uiChanged` is `null` at the beginning, so it passes `null` to `setErrorMessage()`.

briandealwis:
The Eclipse UI guidelines §5.5 suggests that:
> Display a prompt when information is absent, and an error when information is invalid.

chanseokoh:
Okay, I fixed the message display issue.

No check boxes, and no message when nothing is entered, per [this](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1120#issuecomment-275755416) and [this](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1252#issuecomment-275487937). (BTW, @elharo what do you think of [this](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1303#issuecomment-276161518)?)

deploy to app engine standard _001

PTAL @akerekes @briandealwis.

akerekes:
It seems a bit strange to have all text field labels being a noun except for the version field where it's a command/action.
Also I liked the expressiveness of the previous version more. Could we add a hint text to the version and bucket fields to indicate they are optional, and user should set them only if she wants to override them?

chanseokoh:
> Could we add a hint text to the version and bucket fields to indicate they are optional.

Maybe we could say `Version (optional):`, although `Staging Bucket (optional):` gets a bit longer.

akerekes:
So why not add it into the text field using `setMessage(...)`?

chanseokoh:
> So why not add it into the text field using setMessage(...)?

Yeah, we have been discussing this for quite a while. (You can see it by following the thread in #1252.) We had such messages, but later folks thought that's not really messages, and then it looks like Eclipse guideline recommends to say something...

chanseokoh:
> So why not add it into the text field using setMessage(...)?

Ah, I misunderstood it. I got it.

chanseokoh:
TIL `setMessage()` is for this. How about these messages?

selection_004

briandealwis:
Seeing them in place, I think I prefer _(optional)_: the text requires some effort to read but doesn't provide enough detail to understand what's necessary, and we provide detail in the tooltips.

chanseokoh:
That's also true. @elharo what do you think?

elharo:
I still prefer "default" though this isn't critical.

chanseokoh:
"default" seems to be a reasonable compromise.

PR #1315, opened 2017-01-30
remove internal interface reference we don't seem to need

elharo:
@briandealwis also clean up JavaDoc


chanseokoh:
Closing per #1316.

Issue #1196, opened 2017-01-08
Flakiness in StandardFacetInstallationTest


elharo:
Likely because we're loading a remote schema we shouldn't be loading in the first place:

```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 7.794 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallationTest
testStandardFacetInstallation(com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallationTest)  Time elapsed: 7.791 sec  <<< FAILURE!
java.lang.AssertionError: Imported projects have errors
/test-dynamic-web-project/war/WEB-INF/web.xml:1: Referenced file contains errors (http://java.sun.com/xml/ns/javaee/javaee_5.xsd).  For more information, right click on the message in the Problems View and select "Show Details..."
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.google.cloud.tools.eclipse.test.util.project.ProjectUtils.failIfBuildErrors(ProjectUtils.java:131)
	at com.google.cloud.tools.eclipse.test.util.project.ProjectUtils.failIfBuildErrors(ProjectUtils.java:125)
	at com.google.cloud.tools.eclipse.test.util.project.ProjectUtils.importProjects(ProjectUtils.java:112)
	at com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallationTest.testStandardFacetInstallation(StandardFacetInstallationTest.java:54)
```

briandealwis:
This is after the souped-up `waitForIdle()` added as part of #1300.
```
testStandardFacetInstallation(com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallationTest)  Time elapsed: 15.832 sec  <<< ERROR!
java.lang.IllegalArgumentException: Element not found: /test-dynamic-web-project/.settings.
	at org.eclipse.core.internal.watson.ElementTree.elementNotFound(ElementTree.java:256)
	at org.eclipse.core.internal.watson.ElementTree.deleteElement(ElementTree.java:248)
	at org.eclipse.core.internal.resources.Workspace.deleteResource(Workspace.java:1427)
	at org.eclipse.core.internal.resources.Resource.deleteResource(Resource.java:876)
[...]

PR #1318, opened 2017-01-30
Add diagnostics for StandardFacetInstallationTest flakiness

briandealwis:
From the test errors, it looks like there are multiple things deleting the project?  I wonder if we can `jstack` ourselves at runtime.

#1196


chanseokoh:
This is so weird. There is only one test that unzips and opens `test-dynamic-web-project.zip`; there cannot ever be more than one test that tears it down.

Do you think logging only the current job is sufficient? How about all jobs?

briandealwis:
`JobManager#toString()` seems to list the jobs, but I wasn't sure if listed the current job.

PR #1319, opened 2017-01-30
Avoid potential NPE in CloudSdkPreferenceAreaTest

briandealwis:
Introduced with the Win10 test changes in #1300.


plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/CloudSdkPreferenceAreaTest.java
```diff
@@ -78,7 +76,7 @@ public void testInvalidPath() {
         root = r;
       }
     }
-    assertTrue("No root directory!?", root.exists());
+    assertTrue("No root directory!?", root != null && root.exists());
```
chanseokoh: I think separating the conditions has a benefit of immediately knowing which one failed.
```
assertNotNull(root);
assertTrue(root.exists());
```

PR #1320, opened 2017-01-30
Fix compilation errors in .appengine.compat and .test.util

briandealwis:
I'm seeing reference errors in my Eclipse as `GpeMigrator` and `NatureUtils` are referencing `FacetedProject.METADATA_FILE` and
`FacetedProjectNature.NATURE_ID` but are not importing
`org.eclipse.wst.common.project.facet.core.internal`.

Not sure why this isn't causing errors in the Travis build.  I do see testing errors in my local build, though it didn't fail the compilation process.


chanseokoh:
Yeah, I had this too on my laptop when I first pull the compat bundles for the first time. Wasn't sure why because I dis not have these errors on my workstation. I ended up adjusting the error level (lowering to an error) through the Eclipse preferences on my laptop.

briandealwis:
I wonder if we should use `Require-Bundle` since we're depending on internals.

chanseokoh:
> I wonder if we should use Require-Bundle since we're depending on internals.

I have no idea. I guess you have the best expertise among us regarding this.

briandealwis:
Yes, we could create our own strings.

briandealwis:
Replaced with inlined strings instead. PTAL @elharo (assuming it builds)

PR #1323, opened 2017-01-30
Change label: "Specify version:" --> "Version:"

### LINKED ISSUE #1322: Make labels in deploy preferences consistent
selection_002

The only text input that's using a verb is the version field. Probably just "Version:".

chanseokoh:
Fixes #1322.


briandealwis:
Should this be bundled up with #1303?

It seems odd that there's no indication that the version is an optional field.
```
   version.setMessage("(optional)");
```
or
```
   version.setMessage("generated if unspecified");
```
?

chanseokoh:
I think the problem #1322 stands itself regardless of #1303. They are not consistent.

Issue #1058, opened 2016-12-06
Integration test for testing CT4E links in Welcome intro


chanseokoh:
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.welcome/intro/cloud-tools-for-eclipse.xml#L8

For example, if we change the title of the Wiki docs, the link will be broken. This may not be relevant if we link docs on google.com, so might not worth putting effort to set the integration test.

chanseokoh:
The links are now all under `google.com`, so I think we are fine without this. Closing. Feel free to object.

PR #1325, opened 2017-01-30
Point Migration link to google doc

chanseokoh:
The doc is live now. However, the new link doesn't say anything about GWT.


chanseokoh:
With this, I think we can also kill #1058.

PR #1329, opened 2017-02-01
Add support for WTP IURLProvider

### LINKED ISSUE #916: Provide IURLProvider adapter for LocalAppEngineServerBehaviour
Goal read the web url either from the console stream or IURLProvider adapter.

SocketListenMultiConnectorProcess.getStreamsProxy() returns null. Tomcat provides a streams proxy. The proxy streams the content that goes into the console. I was going to use it as a generic method to into and get the urls. Since the web server could be on an odd port I thought it might be nifty to provide that into the GWT Dev View.

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/c75fa7ab422a716a07a8763572afae5a1e57175b/third_party/com.google.cloud.tools.eclipse.jdt.launching/src/com/google/cloud/tools/eclipse/jdt/internal/launching/SocketListenMultiConnectorProcess.java#L265

briandealwis:
Adds support for exposing the per-service URLs from the`dev_appserver`. This is used by GWT's URL launcher, and could be used by other tools too. The interface supports being provided a `null` which I've mapped to the `dev_appserver` admin page.

Fixes #916


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -432,4 +462,32 @@ public IPath getPublishDirectory(IModule[] module) {
...
+      return new URL(getServiceURL(serviceId));
+    } catch (MalformedURLException ex) {
+      /* ignore */
+    }
+    return null;
+
+  }
+
+  @VisibleForTesting
+  String getServiceURL(String serviceId) {
```
chanseokoh: `Url`, for this one.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -432,4 +462,32 @@ public IPath getPublishDirectory(IModule[] module) {
...
+        return null;
+      }
+      if (module == null) {
+        String adminPage = "http://" + getServer().getHost() + ":" + adminPort; //$NON-NLS-1$ //$NON-NLS-2$
+        return new URL(adminPage);
+      }
+      String serviceId = ModuleUtils.getServiceId(module); // never null
+      return new URL(getServiceURL(serviceId));
+    } catch (MalformedURLException ex) {
+      /* ignore */
```
chanseokoh: Never happening, right? What do you think of logging?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -408,18 +427,29 @@ public void onOutputLine(String line) {
...
+        } catch (NumberFormatException e) {
+          // ignore
         }
-
-      } else if (line.contains("Starting admin server at: http://")) { //$NON-NLS-1$
-        if (serverPort == 0) {  // We assume we will no longer see URLs for modules from now on.
-          serverPort = serverPortCandidate;
-        }
-        if (adminPort == 0) {
-          adminPort = extractPortFromServerUrlOutput(line);
```
chanseokoh: We should remove this method entirely.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -391,11 +401,20 @@ static int extractPortFromServerUrlOutput(String line) {
...
+    private Pattern moduleStartedPattern = Pattern.compile(
+        ".*Starting module \"(?<service>[^\"]+)\" running at: (?<url>http://.+:(?<port>[0-9]+))$");
+    private Pattern adminStartedPattern =
+        Pattern.compile(".*Starting admin server at: (?<url>http://.+:(?<port>[0-9]+))$");
 
     private int serverPortCandidate = 0;
 
     @Override
     public void onOutputLine(String line) {
+      Matcher matcher;
```
chanseokoh: How about declaring two matchers, one for module lines and the other for admin lines?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -82,6 +90,8 @@
   @VisibleForTesting int adminPort = -1;
 
   private DevAppServerOutputListener serverOutputListener;
+  
+  private Map<String, String> moduleToUrlMap = Maps.newHashMap();
```
chanseokoh: One thing I noticed is that this map is created once when a server is first created. That is, restarting a server will carry the values extracted from the previous run. Most of the time this will be okay because the values will be overwritten, but I see a slight chance of this going out of sync when a user modifies `appengine-web.xml`. So maybe it'd better to reset this map at the start of a server.

briandealwis: Very good catch.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
briandealwis: I tried and it doesn't seem to help readability (IMHO).

chanseokoh: Understood.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
```diff
@@ -93,51 +98,36 @@ public void testCheckProjectFacets_appEngineStandardFacet() throws CoreException
   @Test
   public void testCheckConflictingId_defaultServiceIds() throws CoreException {
     delegate = getDelegateWithServer();
-    Function<IModule, String> alwaysDefault = new Function<IModule, String>() {
-      @Override
-      public String apply(IModule module) {
-        return "default";
-      }
-    };
+    // Functions.constant() should have inferred the return type too
```
briandealwis: You're right that comment is wrong.  It should be:
```
// Functions.constant() should parameterize both the input and output types
```
This is what it should have been:
```
  private <InputType, OutputType> Function<InputType, OutputType> constant(final OutputType result) {
    return new Function<InputType, OutputType>() {
      @Override
      public OutputType apply(InputType input) {
        return result;
      }
    };
  }
```

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -391,11 +401,20 @@ static int extractPortFromServerUrlOutput(String line) {
...
    */
   public class DevAppServerOutputListener implements ProcessOutputLineListener {
+    // DevAppServer outputs the following for module-started and admin line (on one line):
+    // <<HEADER>> Starting module "default" running at: http://localhost:8080
+    // <<HEADER>> Starting admin server at: http://localhost:8000
+    // where <<HEADER>> = INFO 2017-01-31 21:00:40,700 dispatcher.py:197]
+    private Pattern moduleStartedPattern = Pattern.compile(
+        ".*Starting module \"(?<service>[^\"]+)\" running at: (?<url>http://.+:(?<port>[0-9]+))$");
+    private Pattern adminStartedPattern =
+        Pattern.compile(".*Starting admin server at: (?<url>http://.+:(?<port>[0-9]+))$");
```
briandealwis: Hmm, test failures and my tests show that the leading `.*` is needed.

PR #1330, opened 2017-02-01
Add "Accounts" section to the GCP toolbar

briandealwis:
- `.appengine.ui`: adds an `account` placeholder in the GCP toolbar at the end
- `.login`: adds a login contribution to the account placeholder

We should move the GCP toolbar to a more central location than `.appengine.ui`.

screen shot 2017-02-01 at 10 20 00 am


chanseokoh:
Cool. Actually, I have been wondering if we want to remove the "G" icon at the bottom-right corner, since we have the toolbar icon.

PR #1333, opened 2017-02-02
another try

elharo:
@briandealwis This PR definitely doesn't work. That is, the compile breaks. The question is why? What else needs to change?

Where is com.google.cloud.tools.eclipse.category defined such, specifically the requirement that it needs 'com.google.cloud.tools.eclipse.suite.e45.feature.feature.group [0.1.0,1.0.0]'

That sounds like a MANIFEST.MF problem, not a pom.xml problem. 

[ERROR]   Software being installed: com.google.cloud.tools.eclipse.dist 1.0.0.qualifier
[ERROR]   Missing requirement: com.google.cloud.tools.eclipse.category 1.0.0 requires 'com.google.cloud.tools.eclipse.suite.e45.feature.feature.group [0.1.0,1.0.0]' but it could not be found
[ERROR]   Cannot satisfy dependency: com.google.cloud.tools.eclipse.dist 1.0.0.qualifier depends on: com.google.cloud.tools.eclipse.category [1.0.0]
[ERROR] 
[ERROR] See http://wiki.eclipse.org/Tycho/Dependency_Resolution_Troubleshooting for help.
[ERROR] Cannot resolve dependencies of MavenProject: com.google.cloud.tools.eclipse:google-cloud-plugin-eclipse.repo:0.1.0-SNAPSHOT @ /Users/elharo/google-cloud-eclipse/gcp-repo/pom.xml: See log for details -> [Help 1]
[ERROR]

### DIFF
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -36,5 +36,5 @@ Please visit the following URL for details:\n\n\
 units.1.licenses.0.location = https://www.apache.org/licenses/LICENSE-2.0
 units.1.requires.1.namespace=org.eclipse.equinox.p2.iu
 units.1.requires.1.name=com.google.cloud.tools.eclipse.suite.e45.feature.feature.group
-units.1.requires.1.range=[1.0.0,1.0.0]
+units.1.requires.1.range=[1.0.0,1.0.1)
 units.1.requires.1.greedy=true
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -36,5 +36,4 @@ Please visit the following URL for details:\n\n\
 units.1.licenses.0.location = https://www.apache.org/licenses/LICENSE-2.0
 units.1.requires.1.namespace=org.eclipse.equinox.p2.iu
 units.1.requires.1.name=com.google.cloud.tools.eclipse.suite.e45.feature.feature.group
-units.1.requires.1.range=[1.0.0,1.0.1)
 units.1.requires.1.greedy=true
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -36,5 +36,5 @@ Please visit the following URL for details:\n\n\
 units.1.licenses.0.location = https://www.apache.org/licenses/LICENSE-2.0
 units.1.requires.1.namespace=org.eclipse.equinox.p2.iu
 units.1.requires.1.name=com.google.cloud.tools.eclipse.suite.e45.feature.feature.group
-units.1.requires.1.range=[0.1.0,1.0.0]
+units.1.requires.1.range=[1.0.0]
 units.1.requires.1.greedy=true
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -36,5 +36,5 @@ Please visit the following URL for details:\n\n\
 units.1.licenses.0.location = https://www.apache.org/licenses/LICENSE-2.0
 units.1.requires.1.namespace=org.eclipse.equinox.p2.iu
 units.1.requires.1.name=com.google.cloud.tools.eclipse.suite.e45.feature.feature.group
-units.1.requires.1.range=[1.0.0]
+units.1.requires.1.range=[1.0.0,1.0.0]
 units.1.requires.1.greedy=true
--- a/features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
+++ b/features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="com.google.cloud.tools.eclipse.suite.e45.feature"
       label="Google Cloud Platform for Eclipse 4.5 and later"
-      version="0.1.0.qualifier"
+      version="1.0.0.qualifier"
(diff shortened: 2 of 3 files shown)


briandealwis:
It's the `requires` range on the `.suite` feature — 1.0.0 < 1.0.0.qualifier.
```diff
-units.1.requires.1.range=[0.1.0,1.0.0]
+units.1.requires.1.range=[0.1.0,1.0.1)
```

briandealwis:
Or we make the feature `1.0.0` rather than `1.0.0.qualifier`/`1.0.0-SNAPSHOT`.

akerekes:
I seem to recall, that is we leave out the .qualifier postfix, then the
resulting build will not contain the Eclipse convention timestamp, e.g. in
case of

com.google.appengine.eclipse.core_3.9.5.v20160930-1755.jar, the '

v20160930-1755' part.

briandealwis:
That is exactly right.  I'm not sure that really much matters though for the formal release as we'll then move onto 1.0.1 (or something else) post-release.

elharo:
shouldn't 

units.1.requires.1.range=[0.1.0,1.0.0]
units.1.requires.1.range=[0.1.0,1.0.1)

be the same? I.e. the first is right inclusive and the second is right exclusive so both allow 1.0.0?

briandealwis:
In OSGi and p2, 1.0.0 < 1.0.0.X.  So 1.0.0.X ⊄ [0.0.0,1.0.0].  But 1.0.0.X < 1.0.1 so 1.0.0.X ⊂ [0.0.0,1.0.1). 

We might be able to drop the range entirely.  In fact, given that this _metadata.product_ exists entirely as a convenience, we can safely make the range [0.0.0,999999].

briandealwis:
Dropping the range from the .p2.inf works:
```diff
diff --git a/gcp-repo/metadata.p2.inf b/gcp-repo/metadata.p2.inf
index 3afca09..5b9def1 100644
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -36,5 +36,4 @@ Please visit the following URL for details:\n\n\
 units.1.licenses.0.location = https://www.apache.org/licenses/LICENSE-2.0
 units.1.requires.1.namespace=org.eclipse.equinox.p2.iu
 units.1.requires.1.name=com.google.cloud.tools.eclipse.suite.e45.feature.feature.group
-units.1.requires.1.range=[0.1.0,1.0.0]
 units.1.requires.1.greedy=true
```

elharo:
Interesting. This seems to be passing on travis though it wouldn't even compile locally.

elharo:
Interestingly I get different failures on Mac and Linux.

elharo:
FYI my Linux compile failure likely has nothing to do with this. Mac still failing though.

briandealwis:
Odd, as it builds fine for me on OS X.  What error are you seeing?

elharo:
Fails when it tries to build gcp-repo:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce (default) on project google-cloud-plugin-eclipse.repo: Execution default of goal org.apache.maven.plugins:maven-enforcer-plugin:1.4.1:enforce failed: org.apache.maven.shared.dependency.graph.DependencyGraphBuilderException: Could not resolve following dependencies: [com.google.cloud.tools.eclipse:com.google.cloud.tools.eclipse.suite.e45.feature:jar:1.0.0 (provided)]: Could not resolve dependencies for project com.google.cloud.tools.eclipse:google-cloud-plugin-eclipse.repo:eclipse-repository:0.1.0-SNAPSHOT: Could not find artifact com.google.cloud.tools.eclipse:com.google.cloud.tools.eclipse.suite.e45.feature:jar:1.0.0 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
[ERROR]

akerekes:
@elharo same here

elharo:
@akerekes You're seeing that error on Linux, right?

akerekes:
Yes, on linux. However I think this is expected. When I execute `mvn compile` on a fresh checkout, I get the following exception:
```
Could not find artifact com.google.cloud.tools.eclipse:com.google.licenses.apache_v2:jar:0.1.0-SNAPSHOT
```
which is expected as the `tycho-packaging-plugin` creates a jar from a feature only in the `package` phase

akerekes:
`mvn package` builds successfully

elharo:
PTAL @briandealwis @akerekes I think we understand what's going on here now, so we can merge this and then start cutting the next release.

gcp-repo/metadata.p2.inf
```diff
@@ -36,5 +36,5 @@ Please visit the following URL for details:\n\n\
 units.1.licenses.0.location = https://www.apache.org/licenses/LICENSE-2.0
 units.1.requires.1.namespace=org.eclipse.equinox.p2.iu
 units.1.requires.1.name=com.google.cloud.tools.eclipse.suite.e45.feature.feature.group
-units.1.requires.1.range=[1.0.0,1.0.0]
+units.1.requires.1.range=[1.0.0,1.0.1)
```
briandealwis: Or you could drop this line entirely.  The danger is if you have a version in your local repository, but that can be avoided by setting the `tycho.localArtifacts` property (as per the _travis_ profile).  We probably should set up a _release_ profile.

PR #1346, opened 2017-02-06
Return a copy of account set

chanseokoh:
Now I see no harm in returning a copy of `Accounts`.


plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/GoogleLoginService.java
```diff
@@ -148,10 +146,7 @@ public boolean hasAccounts() {
   @Override
   public Set<Account> getAccounts() {
     synchronized (loginState) {
-      // 'accounts' is a reference to a copy of Accounts maintained in 'loginState'.
-      // ('loginState.listAccounts()' returns a copy.) We intend to return this
-      // reference to callers, while never modifying the set itself.
-      return accounts;
+      return new HashSet<Account>(accounts);
```
akerekes: Is the comment at the `accounts` field declaration still valid in light of this change? If so, I'm not fully understanding what it means.

chanseokoh: Good catch. I'll remove the comments at the field declaration.

Issue #488, opened 2016-08-02
Enhanced UI for login: show avatar images and account names


chanseokoh:
Changing icons, showing emails and avatars, showing more information such as projects, showing a list of added accounts, etc.

chanseokoh:
I'm thinking of adding `ImageStore` (`ImageCache`, `ImageRepository`, or whatever) to fetch pictures (in a background job) and keep them in-memory. This will be kind of a util class, which can be completely independent of login stuff. Should be a single instance, so I guess it's better to define it as an OSGi service? @briandealwis

briandealwis:
Please don't define a global image cache; it's almost never going to be needed. It's particularly bad on Windows where UI resources are at a premium.  You could use a JFace `ImageRegistry`.

PR #1358, opened 2017-02-07
Show names of logged-in accounts in AccountsPanel

chanseokoh:
#488
selection_002

This required to update the ide-login JAR to the latest version.


plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanelTest.java
```diff
@@ -106,13 +122,22 @@ public void testAccountsArea_threeAccounts() {
...
+    assertEquals(3, panel.emailLabels.size());
+    verifyLabelsContains(panel.emailLabels, "alice@example.com");
+    verifyLabelsContains(panel.emailLabels, "bob@example.com");
+    verifyLabelsContains(panel.emailLabels, "charlie@example.com");
+    verifyLabelsContains(panel.nameLabels, "Alice");
+    verifyLabelsContains(panel.nameLabels, "");
+    verifyLabelsContains(panel.nameLabels, "Charlie");
+  }
+
+  private void verifyLabelsContains(List<Label> labels, String text) {
```
elharo: You don't need this. emailLabels is ordered.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
```diff
@@ -29,25 +30,24 @@
...
  * logging out all accounts.
  */
 public class AccountsPanel extends PopupDialog {
 
-  private IGoogleLoginService loginService;
+  private final IGoogleLoginService loginService;
 
   @VisibleForTesting Button logOutButton;
-  @VisibleForTesting List<Label> accountLabels = new ArrayList<>();
+  @VisibleForTesting final List<Label> nameLabels = new ArrayList<>();
```
akerekes: I'm somewhat hesitant about these label lists. They don't have any other function than making tests easier, they're not only visible for test, but they exist only because of tests. We could still verify the label texts by iterating over the labels obtained by traversing the widget hierarchy.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanelTest.java
akerekes: Alternatively you could change the loop to:
```
assertThat(labels, CoreMatchers.hasItem(HasPropertyWithValue.<String>hasProperty("text", text)));
```

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
chanseokoh: Agreed. I prefer removing the fields entirely.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Require-Bundle: org.hamcrest;bundle-version="1.1.0",
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
  com.google.cloud.tools.eclipse.test.util,
  com.google.cloud.tools.eclipse.test.util.ui,
+ com.google.cloud.tools.login,
```
briandealwis: Hmm, this would seem to have been required on the host bundle (`c.g.c.t.e.ae.deploy.ui`), and thus pulled in from there?

plugins/com.google.cloud.tools.eclipse.login/META-INF/MANIFEST.MF
```diff
@@ -15,19 +15,19 @@ Require-Bundle: com.google.gson;bundle-version="[2.7,3.0)",
...
  lib/google-oauth-client-1.22.0.jar,
  lib/google-oauth-client-java6-1.22.0.jar,
  lib/google-oauth-client-jetty-1.22.0.jar,
+ lib/plugins-login-common-0.1.1-SNAPSHOT.jar,
  .
 Export-Package: com.google.api.client.auth.oauth2;x-friends:="com.google.cloud.tools.eclipse.appengine.deploy,com.google.cloud.tools.eclipse.appengine.deploy.ui",
  com.google.cloud.tools.eclipse.login,
  com.google.cloud.tools.eclipse.login.ui,
- com.google.cloud.tools.ide.login
+ com.google.cloud.tools.login
```
briandealwis: We should version this package since it's actually coming from an embedded dependency.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
briandealwis: If you want to test that items are present, set an id on the labels (to identify their label type), and use SWTBot to pull out labels.  You can either use the SWTBot default ID key (`org.eclipse.swtbot.widget.key`) or use the E4 CSS class key (`org.eclipse.e4.ui.css.CssClassName`).  Then you should be able to use SWTBot's `labelWithId(key,value,index)` to pull out values and check order.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
```diff
@@ -83,9 +83,18 @@ protected Control createDialogArea(Composite parent) {
   @VisibleForTesting
   void createAccountsPane(Composite container) {
     for (Account account : loginService.getAccounts()) {
```
briandealwis: Images will be done separately?

chanseokoh: Image will be done separately.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanelTest.java
chanseokoh: `emailLabels` is ordered, but the problem is that the order that the labels are created is non-deterministic (created in iterations on `Set`).

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
chanseokoh: Tried to avoid setting an id on the labels, since we are not solving the problem of having code only for testing, but I don't really find a good alternative. I considered different things, but they have downsides.

briandealwis: Hence the suggestion for the CSS class name.  At least then someone can write custom CSS to render email addresses in Courier, etc.:
```
composite.setData("org.eclipse.e4.ui.css.CssClassName", "accounts");
emailLabel.setData("org.eclipse.e4.ui.css.CssClassName", "email");
nameLabel.setData("org.eclipse.e4.ui.css.CssClassName", "name");
```

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanelTest.java
elharo: That feels like a bug in itself. Should the order of these things be deterministic?

Even if it isn't there are better ways to test for this, rather than looping repeatedly.

chanseokoh: `assertThat(labels, CoreMatchers.hasItem(HasPropertyWithValue.<String>hasProperty("text", text)));`

I get a compile error that `assertThat` doesn't accept `labels` is of type `List<SWTBotLabel>`. (It does accept type `List<String>` though.)

chanseokoh: > That feels like a bug in itself. Should the order of these things be deterministic?

Maybe we should sort the accounts shown in the panel, but that'd be the next time. Currently, I didn't enforce any particular rule.

chanseokoh: @akerekes NVM about `assertThat`, I got it to work.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/META-INF/MANIFEST.MF
chanseokoh: Good catch. Removed it.

plugins/com.google.cloud.tools.eclipse.login/META-INF/MANIFEST.MF
chanseokoh: Seems like the version recognized for this package is "0", probably because we didn't add this through OGSi. Are you suggesting we should add `com.google.cloud.tools.login;version="0"`?

chanseokoh: Ah, I guess I got what you're saying; tagging a version in `Export-Pacakge:` from the `.login` bundle, right?

chanseokoh:
Test failures don't make sense: they fail at different locations. Something's fishy with either the SWT bot or the Hamcrest thing.

chanseokoh:
I guess Hamcrest was fishy. Now looking good. PTAL.

akerekes:
The last build error was a timeout:

chanseokoh:
What does not make sense is the errors in [this build](https://travis-ci.org/GoogleCloudPlatform/google-cloud-eclipse/builds/199800881). The failure comes from the same test, `testAccountsArea_threeAccounts`. Nothing's wrong in that it fails, but what's really weird is that they all fail at different assertions: lines 154, 155, and 156.

Mars:
```
java.lang.Exception: Unexpected exception, expected<java.lang.IndexOutOfBoundsException> but was<java.lang.AssertionError>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at com.google.cloud.tools.eclipse.login.ui.AccountsPanelTest.verifyLabelsContains(AccountsPanelTest.java:162)
	at com.google.cloud.tools.eclipse.login.ui.AccountsPanelTest.testAccountsArea_threeAccounts(AccountsPanelTest.java:156)
```

Neon:
```
java.lang.Exception: Unexpected exception, expected<java.lang.IndexOutOfBoundsException> but was<java.lang.AssertionError>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at com.google.cloud.tools.eclipse.login.ui.AccountsPanelTest.verifyLabelsContains(AccountsPanelTest.java:162)
	at com.google.cloud.tools.eclipse.login.ui.AccountsPanelTest.testAccountsArea_threeAccounts(AccountsPanelTest.java:154)
```

Oxygen:
```
java.lang.Exception: Unexpected exception, expected<java.lang.IndexOutOfBoundsException> but was<java.lang.AssertionError>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at com.google.cloud.tools.eclipse.login.ui.AccountsPanelTest.verifyLabelsContains(AccountsPanelTest.java:162)
	at com.google.cloud.tools.eclipse.login.ui.AccountsPanelTest.testAccountsArea_threeAccounts(AccountsPanelTest.java:155
```

akerekes:
I agree it looks interesting, but once you fixed the index, the test using
the Hamcrest matcher passed in all 3 builds. So it maybe SwtBot that was
not returning/finding the labels in the same order for different Eclipse
versions.

chanseokoh:
Yeah, come to think of it, it's possible that it's the SWT bot that gives different widget orders on different Eclipse versions.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/META-INF/MANIFEST.MF
```diff
@@ -26,7 +26,7 @@ Import-Package: com.google.api.client.auth.oauth2,
  com.google.cloud.tools.eclipse.usagetracker,
  com.google.cloud.tools.eclipse.util,
  com.google.cloud.tools.eclipse.util.status,
- com.google.cloud.tools.ide.login,
+ com.google.cloud.tools.login;version="0.1.1.snapshot",
```
briandealwis: Just `0.1.1`.

plugins/com.google.cloud.tools.eclipse.login/META-INF/MANIFEST.MF
```diff
@@ -15,19 +15,19 @@ Require-Bundle: com.google.gson;bundle-version="[2.7,3.0)",
...
  lib/google-oauth-client-1.22.0.jar,
  lib/google-oauth-client-java6-1.22.0.jar,
  lib/google-oauth-client-jetty-1.22.0.jar,
+ lib/plugins-login-common-0.1.1-SNAPSHOT.jar,
  .
 Export-Package: com.google.api.client.auth.oauth2;x-friends:="com.google.cloud.tools.eclipse.appengine.deploy,com.google.cloud.tools.eclipse.appengine.deploy.ui",
  com.google.cloud.tools.eclipse.login,
  com.google.cloud.tools.eclipse.login.ui,
- com.google.cloud.tools.ide.login
+ com.google.cloud.tools.login;version="0.1.1.snapshot"
```
briandealwis: Just `0.1.1`.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
```diff
@@ -83,9 +77,18 @@ protected Control createDialogArea(Composite parent) {
...
-      Label label = new Label(container, SWT.NONE);
-      label.setText(account.getEmail());
-      accountLabels.add(label);
+      Label name = new Label(container, SWT.LEAD);
+      if (account.getName() != null) {
+        name.setText(account.getName());
+      }
+      name.setData("org.eclipse.e4.ui.css.CssClassName", "accountName");
```
briandealwis: Would it be worth extracting this to a package-protected constant and referencing it from your tests?

chanseokoh: Makes sense.

PR #1360, opened 2017-02-08
Fix test build bug / fix broken tests / make things consistent

### LINKED ISSUE #1361: Tests in .welcome.test and .compat.test are not being executed
`<packaging>` in `pom.xml` is `eclipse-plugin`; it should be `eclipse-test-plugin`. When turning them on, there are test failures. Fixing them in #1360.

chanseokoh:
UPDATED: fixes #1361.

Critical bugs:
- `<packaging>` of test bundles: `eclipse-test-plugin`, not `eclipse-plugin`. This prevented tests from running.
- Fixing a bug in `.compat`: the facet metadata filename was wrong.

Broken tests:
- Fixing tests in `.welcome`. particularly tycho needs to explicitly pull in `org.eclipse.ui.intro`.

Consistency:
- Some bundles have `Bundle-ClassPath: .` (which is unnecessary if `.` is the only entry), while others don't.
- Some bundles have `<sourceDirectory>src</sourceDirectory>`. (Unnecessary)
- When `<useUIHarness>` is true, some bundles miss `<useUIThread>true` (which isn't really necessary but we opted to explicitly define it — the `.integeration` bundle is an exception to this, whose `<useUIThread>` is false). 
- A few other things to make things consistent across bundles.

### DIFF
--- a/eclipse/ide-target-platform/pom.xml
+++ b/eclipse/ide-target-platform/pom.xml
@@ -7,7 +7,7 @@
     <groupId>com.google.cloud.tools.eclipse</groupId>
     <artifactId>trunk</artifactId>
     <version>0.1.0-SNAPSHOT</version>
-    <relativePath>../..</relativePath>
+    <relativePath>../../</relativePath>
   </parent>
   <artifactId>ide-target-platform.repo</artifactId>
   <version>0.1.0-SNAPSHOT</version>
--- a/plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
@@ -65,8 +65,7 @@ public class GpeMigrator {
   private static final String WTP_METADATA_XSLT = "/xslt/wtpMetadata.xsl";
 
   // FacetedProject.METADATA_FILE = ".settings/" + FacetCorePlugin.PLUGIN_ID + ".xml";
-  @VisibleForTesting
-  static final String FACETS_METADATA_FILE =
+  private static final String FACETS_METADATA_FILE =
       ".settings/org.eclipse.wst.common.project.facet.core.xml";
 
   /**
--- a/plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
@@ -66,7 +66,8 @@ public class GpeMigrator {
 
   // FacetedProject.METADATA_FILE = ".settings/" + FacetCorePlugin.PLUGIN_ID + ".xml";
   @VisibleForTesting
-  static final String FACETS_METADATA_FILE = ".settings/org.eclipse.wst.common.component.xml";
+  static final String FACETS_METADATA_FILE =
+      ".settings/org.eclipse.wst.common.project.facet.core.xml";
 
   /**
    * Removes various GPE-related remnants: classpath entries, nature, runtime, and facets. Any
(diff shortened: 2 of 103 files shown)


chanseokoh:
Interesting. There are test failures.

chanseokoh:
I guess tests in the `.welcome.test` and `.compat.test` bundles have not been executed from the beginning because it had a wrong `<packaging>` type!

briandealwis:
> When <useUIHarness> is true, some bundles miss <useUIThread> (which isn't really necessary but we opted to explicitly define it).

`useUIThread=true` is for tests that drive the UI directly like some of our New Project wizard page tests.  They want control over the SWT thread.

SWTBot tests definitely don't want `useUIThread=true` as they orchestrate things from a different thread.

eclipse/ide-target-platform/pom.xml
```diff
@@ -7,7 +7,7 @@
     <groupId>com.google.cloud.tools.eclipse</groupId>
     <artifactId>trunk</artifactId>
     <version>0.1.0-SNAPSHOT</version>
-    <relativePath>../..</relativePath>
+    <relativePath>../../</relativePath>
```
briandealwis: Did you see a warning or recommendation about this?

chanseokoh: No, just for consistency. The majority of them has the trailing slash.

chanseokoh:
> SWTBot tests definitely don't want useUIThread=true as they orchestrate things from a different thread.

AFAIK, useUIThread=true is the default, so having it or not having it doesn't matter. That said, we should probably explicitly set it to false for SWTBot tests? This could be tricky, because I know we have been using SWTBot in non-integration tests too.

chanseokoh:
https://eclipse.org/tycho/sitedocs/tycho-surefire/tycho-surefire-plugin/test-mojo.html#useUIThread

chanseokoh:
The following assert in `BasePluginXmlTest` that tests if `.welcome` has a valid extension point fails, but I don't understand why. The extension point is defined in `plugin.xml`, and the test can actually walk through the XML to get the extension point attribute. It's only that `registry.getExtensionPoint()` returns null while it shouldn't. 

```
      IExtensionPoint extensionPoint = registry.getExtensionPoint(point);
      Assert.assertNotNull(extensionPoint);
```

Moreover, the test does not fail locally when run inside Eclipse.

chanseokoh:
Fixed the `BasePluginXmlTest` failure thanks to @briandealwis' advice. Making tycho pull `org.eclipse.ui.intro` in did the trick.

This PR is ready for review. (The `useUIThread` issue is settled.)

plugins/com.google.cloud.tools.eclipse.sdk.test/pom.xml
```diff
@@ -1,10 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elharo: Interesting that here you're adding something that can be defaulted, and elsewhere you're taking it away. My personal preference is to include things that can be defaulted since I rarely remember what the defaults are. It also makes it easier to change to a non-default value when necessary.

chanseokoh: Okay, I'll keep `Bundle-ClassPath: .` and add it to all bundles that miss this.

I think I'll still skip `<sourceDirectory>src</sourceDirectory>`, since otherwise there will be too many, and FWIW, the Maven philosophy is to not specify things unless it's not default. The "src" directory is too obvious too.

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
```diff
@@ -66,7 +66,8 @@
 
   // FacetedProject.METADATA_FILE = ".settings/" + FacetCorePlugin.PLUGIN_ID + ".xml";
   @VisibleForTesting
```
elharo: Generally constants do not need to be @VisibleForTesting. You can always use the literal form.

chanseokoh: Good idea. Done.

chanseokoh:
Builds have very flaky, after merging with master.

Issue #1345, opened 2017-02-06
Builds failing with deadlocks on Oxygen


briandealwis:
Oxygen builds are frequently failing on Travis:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

From the [failed build logs](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/files/753310/failedbuild.txt), it seems to be happening in the integration tests, and there are a lot of puzzling output.

It may be related to #1314

Update: turns out to be a deadlock introduced in WTP in Oxygen M4/M5 ([bug 511793](https://bugs.eclipse.org/511793)).  A workaround is to temporarily disable its background dependency graph updating.

briandealwis:
Running the tests under docker, it seems like we hit a couple of issues:
- our use of SWTBot's `resetWorkbench()` tickles [bug 511729](https://bugs.eclipse.org/bugs/show_bug.cgi?id=511729) in Oxygen.
- there are some hangs in `SwtBotWorkbenchActions.waitForIdle()` as it loops repeatedly on the while loop for `!Job.getJobManager().isIdle()`, and there are some long-lived jobs
  - there is some slowdown when creating a project, as it triggers a _Update Java Indexes_ job which can take a while (in my docker image, at least)
  - when I ran in Docker, I once saw a hang in `DebugNativeAppEngineStandardProjectTest` where it was waiting on a _Waiting for VM to connect_ process. This process should have been terminated.

chanseokoh:
FYI, I know a lot of the following `IllegalArgumentException: Argument not valid` errors during the integrate test are happening in Mars and Neon too. Those errors may be deceiving for the matter of the Oxygen build timeouts.

```
!ENTRY org.eclipse.ui 4 0 2017-02-05 16:02:12.679
!MESSAGE Unhandled event loop exception
!STACK 0
org.eclipse.e4.core.di.InjectionException: java.lang.IllegalArgumentException: Argument not valid
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:65)
	at org.eclipse.e4.ui.internal.di.UIEventObjectSupplier$UIEventHandler$1.run(UIEventObjectSupplier.java:61)
[...]

PR #1375, opened 2017-02-09
Disable WTP Dependency Graph updates during faceted project changes

briandealwis:
Reduces possibility of deadlocks with Oxygen M5.


briandealwis:
Towards #1345

Issue #1373, opened 2017-02-09
Deadlock in NonSystemJobSuspender 


briandealwis:
NonSystemJobSuspender is performing job operations with its monitor held.  These job operations may fire events.
```
+-------------------------------------------------------------------------------
| STACK DUMP @ 4.001 min: testCheckConflictingId_differentServiceIds(com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEngineServerDelegateTest)
|
| "main" [1] BLOCKED on java.lang.Class@453674ff owned by 'Worker-3 [id:37]
|     at com.google.cloud.tools.eclipse.appengine.facets.NonSystemJobSuspender.suspendJob(NonSystemJobSuspender.java:77)
|     - blocked on java.lang.Class@453674ff
|     at com.google.cloud.tools.eclipse.appengine.facets.NonSystemJobScheduleListener.scheduled(NonSystemJobScheduleListener.java:29)
|     at org.eclipse.core.internal.jobs.JobListeners$5.notify(JobListeners.java:54)
|     at org.eclipse.core.internal.jobs.JobListeners.doNotify(JobListeners.java:102)
|     at org.eclipse.core.internal.jobs.JobListeners.scheduled(JobListeners.java:161)
|     at org.eclipse.core.internal.jobs.JobManager.schedule(JobManager.java:1259)
|     at org.eclipse.core.internal.jobs.JobManager.endJob(JobManager.java:697)
|     at org.eclipse.core.internal.jobs.InternalJob.done(InternalJob.java:198)
|     at org.eclipse.core.runtime.jobs.Job.done(Job.java:253)
|     at org.eclipse.ui.progress.UIJob$1.run(UIJob.java:110)
|     at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:35)
|     at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:135)
|     - locked org.eclipse.swt.widgets.RunnableLock@5807c62e
|     at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4024)
|     at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3700)
|     at org.eclipse.ui.internal.dialogs.EventLoopProgressMonitor.runEventLoop(EventLoopProgressMonitor.java:125)
[...]

PR #1382, opened 2017-02-10
Move BuildPath to com.google.cloud.tools.eclipse.appengine.libraries

elharo:
I'm moving BuildPath to com.google.cloud.tools.eclipse.appengine.libraries in preparation for using it after project creation to add a library to an existing project, not just when a new project is created. 

Pay special attention to how the context is created.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -40,9 +40,10 @@
 import com.google.cloud.tools.eclipse.appengine.libraries.LibraryClasspathContainerResolverJob;
 import com.google.cloud.tools.eclipse.appengine.libraries.model.Library;
 
-class BuildPath {
+public class BuildPath {
 
-  static void addLibraries(IProject project, List<Library> libraries, IProgressMonitor monitor)
+  public static void addLibraries(IProject project, List<Library> libraries, 
+      IProgressMonitor monitor, Class<?> classFromBundle)
```
chanseokoh: I think it's customary to have the monitor as a last parameter.

briandealwis:
Oxygen failure from #1373.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -73,12 +74,13 @@ static void addLibraries(IProject project, List<Library> libraries, IProgressMon
...
+    IEclipseContext context = EclipseContextFactory.getServiceContext(
+        FrameworkUtil.getBundle(classFromBundle).getBundleContext());
+    
+    runContainerResolverJob(javaProject, context);
   }
   
-  private static void runContainerResolverJob(IJavaProject javaProject) {
-    IEclipseContext context = EclipseContextFactory.getServiceContext(
-        FrameworkUtil.getBundle(CreateAppEngineStandardWtpProject.class).getBundleContext());
+  private static void runContainerResolverJob(IJavaProject javaProject, IEclipseContext context) {
```
briandealwis: @chanseokoh's right: we don't need the class.  Just get the service context for this class' bundle.  I would revert this code to how it was.

The class/bundle is only important as it determines which bundle's lifecycle will be used to clean up any referenced OSGi services.

elharo: I can certainly do that. As currently written, this move is a fairly safe refactoring. However if I don't pass the class, then we're changing (I think) which bundle is used which does not seem to be a no-op. As long as that's OK, I can do it.

PR #1391, opened 2017-02-12
extract common XSLT code

### LINKED ISSUE #1389: Refactor duplicate code in GpeMigrator and ApplicationQuickFix for applying XSLT

elharo:
fix #1389


plugins/com.google.cloud.tools.eclipse.appengine.validation.test/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,8 @@ Bundle-Vendor: Google Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.validation
 Require-Bundle: org.junit,
- org.eclipse.equinox.registry
+ org.eclipse.equinox.registry,
+ com.google.cloud.tools.eclipse.test.dependencies
```
chanseokoh: AFAIK, if we add this bundle as `Required-Bundle`, we can remove all the test-related packages it imports from `Import-Package`. That is, you can remove `org.mockito;provider=google` below.

This has never been streamlined across the entire test bundles, so we might want to do so in the future.

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
```diff
@@ -13,7 +13,6 @@ Export-Package: com.google.cloud.tools.eclipse.util,
  com.google.cloud.tools.eclipse.util.templates.appengine
 Import-Package: com.google.common.base;version="[20.0.0,21.0.0)",
  com.google.common.io;version="[20.0.0,21.0.0)",
- com.google.common.net;version="[20.0.0,21.0.0)",
```
chanseokoh: I get a compile error if I don't have this. (For you, I think removing this will increase the number of warnings?)

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/Xslt.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private Xslt() {}
+  
+  private static final TransformerFactory factory = TransformerFactory.newInstance();
+
+  public static void transformInPlace(IFile metadataFile, URL xslt) 
+      throws IOException, CoreException, TransformerException {
+    try (InputStream metadataStream = metadataFile.getContents();
+        InputStream stylesheetStream = xslt.openStream()) {
+      try (InputStream resultStream = applyXslt(metadataStream, stylesheetStream)) {
```
chanseokoh: This `try` block can be merged into the outer `try` block.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorTest.java
```diff
@@ -105,9 +98,7 @@ public void testRemoveGpeRuntimeAndFacets_doNotLogIfMetadataFileDoesNotExist()
     assertFalse(metadataFile.exists());
 
     IFacetedProject facetedProject = ProjectFacetsManager.create(gpeProject);
-    Logger logger = mock(Logger.class);
-    GpeMigrator.removeGpeRuntimeAndFacets(facetedProject, logger);
-    verify(logger, never()).log(any(Level.class), anyString(), any(Throwable.class));
```
chanseokoh: I feel the test method should be renamed to something like `_doesNotErrorOutIfMetadataFileDoesNotExist()`.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ApplicationQuickFixTest.java
```diff
@@ -26,37 +26,57 @@
...
-      Document document = builder.parse(xmlStream);
+  public void testRemoveApplicationElements() throws IOException, ParserConfigurationException,
+      SAXException, TransformerException, CoreException {
+
+    IProject project = projectCreator.getProject();
+    IFile file = project.getFile("testdata.xml");
+    file.create(stringToInputStream(APPLICATION_XML), IFile.FORCE, null);
+    
+    IMarker marker = Mockito.mock(IMarker.class);
+    Mockito.when(marker.getResource()).thenReturn(file);
```
chanseokoh: Doesn't like static imports for `when` and `mock`?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ApplicationQuickFixTest.java
```diff
@@ -26,37 +26,57 @@
...
+/**
+ * Must be run as a plugin test.
+ */
 public class ApplicationQuickFixTest {
   
   private static final String APPLICATION_XML =
       "<appengine-web-app xmlns='http://appengine.google.com/ns/1.0'>"
       + "<application>"
       + "</application>"
       + "</appengine-web-app>";
```
chanseokoh: If we have some other top level elements than just `<application>`, I think the test below will give more confidence about the stylesheet XML.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/META-INF/MANIFEST.MF
elharo: My current thinking is that we should prefer Import-Package to Require-Bundle. Import-Package is the OSGi standard. Require-Bundle is an Eclipse extension that seems lightly deprecated.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ApplicationQuickFixTest.java
elharo: Worth adding but outside the scope of this PR.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ApplicationQuickFixTest.java
elharo: No, I don't. I like to know where my methods come from :-)

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
elharo: Do you get a compile error after applying this PR? Can you find the place where we're importing from here? I didn't see one.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/Xslt.java
elharo: You're right; done.

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
chanseokoh: It's `HttpHeaders.USER_AGENT` in `FileDownloader.java:86`: "Access restriction: The field 'HttpHeaders.USER_AGENT' is not API". Perhaps the reason you don't have the compile error is that this type of error is to set to a warning in your Eclipse. (It compiles anyway as can be seen in Travis. Happened before too: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1320.)

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/META-INF/MANIFEST.MF
chanseokoh: Then want to try reverting the addition of `com.google.cloud.tools.eclipse.test.dependencies` to `Require-Bundle`?

elharo: The test dependencies are one of the weird corner cases where we seem to need Require-Bundle due to the conflicting versions of Mockito and Hamcrest. :-(

chanseokoh: Fair enough. Yeah, I noticed that too. For example, I learned in a hard way that simply adding `org.mockito` doesn't work and it should be `org.mockito;provider=google`.

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
elharo: I just love pulling in a big library for one string constant. :-) Done.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/META-INF/MANIFEST.MF
briandealwis: Import-Package works when you're using the API.  Require-Bundle is when you depend on the implementation.  When we're using `provider=google', we're implicitly using the implementation.

On the flip side, `Import-Package` helps pare down dependencies since you have to explicitly bring them in.  It helps avoid problems from kitchen-sink type bundles (like our `.test.dependencies`, but also `org.eclipse.core.runtime`) that pull in the world.

So I'm +1 to using Import-Package here for the Mockito deps.

PR #1392, opened 2017-02-13
Remove unused variables, imports, duplicate entries, etc

chanseokoh:
Remove some warnings, and correct minor things.


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardRuntimeChangeListener.java
```diff
@@ -107,7 +106,6 @@ public void handleEvent(IFacetedProjectEvent event) {
     }
 
     // Add the App Engine facet
-    IProject project = facetedProject.getProject();
```
elharo: Does getProject actually do nothing? I have this funny feeling that it may cause something to happen via side effects even of we don't sue the return value.

briandealwis: `FacetedProject#getProject()` is just a getter, and a `FacetedProject` is created from an `IProject`.  So this is free of side-effects.

chanseokoh: This is a vestige of refactoring done in commit d337a34335cc2838efb02bf1a4a119a834fb4053. `InstallAppEngineFacetJob` will retrieve the project inside the job.

Issue #1381, opened 2017-02-09
Set application name when making API calls


chanseokoh:
We are starting to add new API calls (getting GCP projects, user names, profile pictures, etc), and we need to ensure that we always set an application name identifying our tool CT4E. I believe APIs will then put (perhaps append) the given application name in the User-Agent HTTP header.

elharo:
@chanseokoh We can close this one now, right?

chanseokoh:
For now, we are done.

PR #1393, opened 2017-02-13
Set application name for login API

chanseokoh:
#1381

Setting it for the test constructor isn't really necessary, but to be consistent.


chanseokoh:
It's not so feasible here, unless we check the actual HTTP request being created, but that's buried down deep at this abstraction level. On the other hand, the ide-login library does have a test to check the actual HTTP request, so I think we are covered.

PR #1394, opened 2017-02-13
Ensure we only use Guava 20

briandealwis:
Verified build breaks by modifying `com.google.cloud.tools.eclipse.util` to import a package from Guava 15.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -263,6 +263,22 @@
                     <exclude>commons-collections:commons-collections:[3.2.1]</exclude>
                   </excludes>
                 </bannedDependencies>
+                <bannedDependencies>
+                  <!--
+                    Ensure our bundles only use Guava 20.0.0. 
+                    Tycho identifies p2-resolved bundles are of form:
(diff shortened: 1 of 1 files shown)


briandealwis:
Verified that this catches `Import-Package`s without a version range, if there is no other dependency that pulls in the bundle.

Unfortunately can't catch if a package version range exceeds the wanted range like:
```
 Import-Package: com.google.common.collections;version="[20,22)"
```
as the enforcement is done by examining the resolved dependencies.

chanseokoh:
> Unfortunately can't catch if a package version range exceeds the wanted range like:

So it can't catch `version="20.0.0"`, since it has no upper bound, right?

briandealwis:
Unfortunately no.  The maven-enforcer-plugin seems to applies its configured rules to the project's dependencies (`mvn dependency:tree`).  Those dependencies are configured by Tycho from the `Import-Packages` and `Require-Bundles`.  So the enforcer plugin never sees the package version ranges.

PR #1399, opened 2017-02-14
Streamline Messages bundle classes

chanseokoh:
Making the classes consistent.


chanseokoh:
Okay, this reveals a test failure. I'll fix it.

chanseokoh:
Wow, [`MessageFormat.format()` removes single quotes](http://stackoverflow.com/questions/189889/java-messageformat-how-can-i-insert-values-between-single-quotes). Another reason to be consistent; we should always call `MessageFormat.format()`.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/messages.properties
```diff
@@ -1,4 +1,4 @@
 OPT_IN_DIALOG_TITLE=Google Cloud Tools for Eclipse
 OPT_IN_DIALOG_TEXT=Share anonymous usage statistics of Cloud Tools for Eclipse with Google?
 OPT_IN_BUTTON=Share
-OPT_OUT_BUTTON=Don't Share
\ No newline at end of file
+OPT_OUT_BUTTON=Don''t Share
```
briandealwis: I wonder if this should be _Do Not Share_?

chanseokoh: That sounds more formal. +1

PR #1412, opened 2017-02-15
Hide mutable public field

elharo:


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/standard/StandardDeployPreferences.java
```diff
@@ -27,7 +27,7 @@
 public class StandardDeployPreferences {
 
   public static final String PREFERENCE_STORE_QUALIFIER =
-      "com.google.cloud.tools.eclipse.appengine.deploy";
+      "com.google.cloud.tools.eclipse.appengine.deploy";
```
briandealwis: Some stray spaces introduced?

PR #1451, opened 2017-02-21
Explicitly prompt user to save dirty editors before launch

### LINKED ISSUE #1427: 'Run as App Engine' action should wait for user to save changes
Steps to reproduce:
1. user modifies the source, but does not save the changes
1. Right click on project/Run as App Engine
1. the local run will start while Eclipse displays a dialog to offer the user to save the changes

Expected: the local run should wait until the user closes the dialog about saving the changes.

briandealwis:
Explicitly call out to `ServerUIPlugin.saveEditors()` before starting the server with `IServer#start()`. Using `ServerUIPlugin.saveEditors()` respects the user's setting in _Preferences > Server > Launching > Save dirty editors before starting the server_. This is the same approach used by the _Servers_ view. There is one downside: if the user says _No_ to saving dirty editors, then the user will see the same puzzling behaviour as they will be prompted a second time as the server launch continues in the background.  But this same behaviour is seen when using the _Servers_ view.

Fixes #1427

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
@@ -47,6 +47,7 @@ import org.eclipse.wst.server.core.IServerType;
 import org.eclipse.wst.server.core.IServerWorkingCopy;
 import org.eclipse.wst.server.core.ServerCore;
 import org.eclipse.wst.server.core.ServerUtil;
+import org.eclipse.wst.server.ui.internal.ServerUIPlugin;
 
 /**
  * A helper class for launching modules on a server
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
```diff
@@ -127,9 +128,15 @@ private IServer createServer(IModule[] modules, SubMonitor progress) throws Core
     return serverWorkingCopy.save(false, progress.newChild(2));
   }
 
-  @VisibleForTesting
   protected void launch(IServer server, String launchMode, SubMonitor progress)
       throws CoreException {
+    // Explicitly offer to save dirty editors to avoid the puzzling prompt-to-save in
+    // IServer#start() that prompts the user <em>as the server continues to launch</em>.
```
elharo: since this isn't Javadoc; maybe just **as the server continues to launch**

elharo:
Travis is taking a long time to kick off. I wonder if our quota got reverted again? let me email support.

Issue #136, opened 2016-05-05
Support shared referenced projects


elharo:
This is a bug from the old plugin we need to check that we don't reproduce in the new plugin:

From:

https://code.google.com/p/googleappengine/issues/detail?id=10652&q=wtp&colspec=ID%20Type%20Component%20Status%20Stars%20Summary%20Language%20Priority%20Owner%20Log

The Eclipse WTP plugin does not support shared referenced projects (shared code) when the projects are referenced from multiple WAR modules.  The referenced project is correctly compiled into a common.jar file, however it's only placed in the WEB-INF/lib directory on the first WAR, and not the others.  You'll get a ClassNotFound exception in other projects.

Steps to reproduce:
1. Create a multi-module EAR project
2. Create a new "common" project (or Utility Project)
3. Reference the "common" project in all modules.
4. Access some classes in the common code project from each module (e.g. some simple method from a Servlet).
5. Run using the WTP server and/or deploy to App Engine.  The common code will work in one module, but not in the others (ClassNotFound).

Expected behaviour:

The "common" code should be accessible in all referenced WAR modules.

What version of the product are you using? On what operating system?
- Eclipse GAE WTP plugin 1.9.0
- Both Mac and Windows

Temporary Workaround:
- Use Eclipse's "Linked Source" directory support to reference shared code between applications.

briandealwis:
Note that _Eclipse WTP Plugin_ in the above is referencing the GPE WTP plugin.

briandealwis:
Our `LocalAppEnginePublishingOperation` isn't handling child modules.  Fortunately it looks relatively straightforward; I'm a bit surprised that WTP doesn't provide a base version out-of-the-box.

briandealwis:
I've pushed up a fix in the branch [i136](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/i136).  I haven't yet figured out how to express the dependencies for a self-contained test.  I'll instead use the GWT `modular-webapp` archetype.

PR #1452, opened 2017-02-22
Implement LaunchConfigurationDelegate#getProjectsForProblemSearch()

### LINKED ISSUE #1423: Run As > App Engine with build failures
We block deployment when the project doesn't build, but not when we only run locally.

image

Should we block local run too? I'm not sure. For regular Java apps, Eclipse does let you run with build errors, but it asks you first:

image

We should probably do something similar.

briandealwis:
Platform/Debug uses _LaunchConfigurationDelegate#getProjectsForProblemSearch()_ to return the list of projects to check for build errors.  Our launch configuration is not quite configured as expected by our superclass's implementation in `AbstractJavaLaunchConfigurationDelegate` and so we seemingly have no projects to check.  This fix provides an implementation of _getProjectsForProblemSearch()_ and _getBuildOrder()_ as derived from the server's configured modules.  The implementation requires obtaining the full list of modules used in a server definition (modules can have child modules — seen particularly in maven projects as per #136).

screen shot 2017-02-21 at 9 43 01 pm

Fixes #1423

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.tools.eclipse.appengine.localserver.server;
 
-import static org.mockito.Matchers.any;
-
 import java.io.InputStream;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -26,22 +24,25 @@ import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
-import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.wst.server.core.IModule;
-import org.eclipse.wst.server.core.IServer;
-import org.hamcrest.Matchers;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
-import org.mockito.AdditionalMatchers;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
 import org.mockito.Mockito;
+import org.mockito.runners.MockitoJUnitRunner;
 
+@RunWith(MockitoJUnitRunner.class)
 public class ModuleUtilsTest {
 
-  private IModule module = Mockito.mock(IModule.class);
-  private IFile descriptorFile = Mockito.mock(IFile.class);
-  private IFolder webinf = Mockito.mock(IFolder.class);
+  @Mock
+  private IModule module;
+  @Mock
+  private IFile descriptorFile;
+  @Mock
+  private IFolder webinf;
 
   @Before
   public void turnOffLogging() {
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtils.java
```diff
@@ -56,4 +60,27 @@ public static String getServiceId(IModule module) {
...
+      modules.add(module);
+      addChildModules(server, new IModule[] {module}, modules);
+    }
+    return modules.toArray(new IModule[modules.size()]);
+  }
+
+  /** Recursively walk the children from {@code modulePath}. */
+  private static void addChildModules(IServer server, IModule[] modulePath, Set<IModule> modules) {
+    IModule[] newModulePath = Arrays.copyOf(modulePath, modulePath.length + 1);
+    for (IModule child : server.getChildModules(modulePath, null)) {
```
elharo: This looks like it may have combinatorial explosion. I.e. every child module is going to check all the modules in modulePath again rather than just the newly added module. Or is there some way this doesn't happen?

elharo:
FYI, there may be an issue in the webhooks that connect Travis to Github. Clicking through, all checks seem to have passed but Github hasn't been updated.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
```diff
@@ -96,4 +102,42 @@ private void mockAppEngineWebXml(String testfile) throws CoreException {
     Mockito.when(descriptorFile.getContents()).thenReturn(in);
   }
 
+  @Test
+  public void testGetAllModules_single() {
+    IServer server = Mockito.mock(IServer.class);
```
akerekes: You can extract this to a @Mock field.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
```diff
@@ -24,11 +26,15 @@
...
 import org.eclipse.wst.server.core.IModule;
+import org.eclipse.wst.server.core.IServer;
+import org.hamcrest.Matchers;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.AdditionalMatchers;
 import org.mockito.Mockito;
 
 public class ModuleUtilsTest {
```
akerekes: all fields could be annotated with @Mock instead of explicitly initializing them.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
```diff
@@ -96,4 +102,42 @@ private void mockAppEngineWebXml(String testfile) throws CoreException {
...
+        any(IProgressMonitor.class))).thenReturn(new IModule[] {module2a, module2b});
+
+    IModule module3 = Mockito.mock(IModule.class, "module3");
+    Mockito.when(server.getChildModules(AdditionalMatchers.aryEq(new IModule[] {module, module2b}),
+        any(IProgressMonitor.class))).thenReturn(new IModule[] {module3});
+
+    IModule[] result = ModuleUtils.getAllModules(server);
+    Assert.assertNotNull(result);
+    Assert.assertEquals(4, result.length);
+    Assert.assertThat(result[0], Matchers.sameInstance(module));
```
akerekes: You could use `org.hamcrest.collection.IsArrayContainingInAnyOrder` or `org.hamcrest.collection.IsArrayContainingInOrder` instead of checking one-by-one

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtils.java
briandealwis: This code follows the same approach used by the `Server#visit()`/`#visitModules()` that is used in its internal `Server#getAllModules()` (unfortunately not exposed via `IServer`).  So it's no worse that what's currently used internally.

I honestly don't expect server module trees to be deep.  But given that we're just accumulating individual modules, we can shortcircuit the descent.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
briandealwis: Thanks for the pointer!

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/ModuleUtilsTest.java
briandealwis: I should extract these into a new test class.

briandealwis:
PTAL

Issue #517, opened 2016-08-09
Exception catching and reporting


elharo:
Eclipse provides some sort of functionality for catching and reporting unexpected exceptions from plug-ins. That is, I've seen these reports for WTP. 

We need to figure out what this is and determine if we can hook into this service so that we get these reports for our plug-ins.

Make sure we don't unintentionally collect PII, but a pure stack trace (no exception messages, no method arguments) should be safe.

elharo:
https://blog.ctrlflow.com/byo-aeri-in-eclipse-neon/

briandealwis:
It requires either using their server or somehow rolling our own or making it fit in with the existing infrastructure.

briandealwis:
I missed the bottom of that blog post — it looks surprisingly easy.  Some more details at https://wiki.eclipse.org/EPP/Logging#Is_the_server-side_available_too.3F

chanseokoh:
@elharo Did WTP send exception reporting our internal feedback server (which is what IntelliJ is currently doing)?

I wonder what we will want to do eventually is BYO AERI + our internal server?

elharo:
Possibly with Crash, see googlefeedback.jar in IntelliJ repo

elharo:
It appears that the  GoogleFeedback.jar in IntelliJ repo was written for Android Studio and depends heavily in IntelliJ classes. Probably not an option for us. We'd need to reimplement to speak directly to the server.

briandealwis:
`GoogleFeedback` seems clean; class analysis shows it just depends on `protobuf` and otherwise uses stock `java.net`.  With @akerekes finishing off the async download, I'm taking this on.  Will have something by EOB.

elharo:
We need to understand this jar better before we depend on it. Where does it come from? Where does the source code live? What's the license?

briandealwis:
I'm just hooking up AERI and won't actually submit anything.  I'm assuming something needs to be done to register the product and package code.

briandealwis:
I've put up an AERI prototype on the [`aeri`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/aeri) branch.  Although currently disabled, it has some code to call out to the `AnonymousFeedback` class in `GoogleFeedback.jar`; it actually does work, and I get a response token back.

This prototype is pretty simple:
  - Currently issues are submitted under package name of `com.google.gct.eclipse` (IntelliJ is `com.google.gct.idea`).
[...]

PR #1453, opened 2017-02-22
Add util to send "multipart/form-data" HTTP POST

chanseokoh:
In preparation for #517.

Due to some bug on the internal exception reporting system, now I have to employ a workaround to send a multipart/form-data HTTP POST.


plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/io/HttpUtilWithServerTest.java
```diff
@@ -0,0 +1,74 @@
...
+  public void testSendPostMultipart() throws IOException {
+    Map<String, String> parameters = new HashMap<>();
+    parameters.put("dp", "/virtual/some-event-type/some-event-name");
+    parameters.put("exception_info",
+        "Exception in thread \"main\" java.lang.IllegalArgumentException: Wrong argument!\n\tat B.main(B.java:73)");
+    parameters.put("product", "CT4E");
+
+    assertEquals(200, HttpUtil.sendPostMultipart(server.getAddress(), parameters));
+    assertEquals("POST", server.getRequestMethod());
+    assertTrue(server.getRequestHeaders().get("User-Agent").startsWith("gcloud-eclipse-tools/0.1.0."));
```
elharo: where's the 0.1.0 coming from? I'd love to change that to 1.0.0 but it's probably too painful

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/HttpUtil.java
```diff
@@ -31,6 +41,37 @@
   public static final int DEFAULT_CONNECT_TIMEOUT_MS = 3000;
   public static final int DEFAULT_READ_TIMEOUT_MS = 3000;
 
+  private static final String MULTIPART_BOUNDARY =
+      "---------------------------45224ee4-f3c1-4b23-8df1-4012f722218c"; // some random UUID
+
+  private static HttpTransport transport = new NetHttpTransport();
```
elharo: final

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/HttpUtil.java
```diff
@@ -31,6 +41,37 @@
   public static final int DEFAULT_CONNECT_TIMEOUT_MS = 3000;
   public static final int DEFAULT_READ_TIMEOUT_MS = 3000;
 
+  private static final String MULTIPART_BOUNDARY =
```
elharo: Where did this come from? Should we be using the same one every time?

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/io/HttpUtilWithServerTest.java
chanseokoh: #1454. We should think about this very soon.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/HttpUtil.java
chanseokoh: It's some arbitrary UUID I generated. It's perfectly fine to reuse the same boundary every time, as long as it gets its job done to separate multiple parts (i.e., it's a unique boundary value that never appears in any actual parameter values).

chanseokoh: Theoretically, we should ensure parameter values don't have this particular boundary value, but the chance that there will be an overlap is basically 0 in practice, so I think just using this boundary is an acceptable compromise.

PR #1466, opened 2017-02-23
Better Name for Blacklist Marker

### LINKED ISSUE #1430: Give a more descriptive name to 'Blacklist Marker'
We should add a more descriptive name to the marker, e.g. 'App Engine Blacklist Marker' or something similar that indicates to the user that it probably comes from our plugin.

image

abbiefarr:
fixes #1430 

@akerekes


plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.properties
```diff
@@ -1,5 +1,5 @@
 extension.name=App Engine JRE Whitelist
 appengine.web=App Engine Web XML Validator
-blacklist.marker=Blacklist Marker
+appengine.blacklist.marker=App Engine Element Problem
```
akerekes: I think something that refers to the xml would be better, maybe `App Engine Configuration Problem`?

PR #1471, opened 2017-02-23
Created the flex facet 

nbashirbello:
The flex facet is in a new bundle "com.google.cloud.tools.eclipse.appengine.flex", that will not be included in the build along with the appropriate test bundle "com.google.cloud.tools.eclipse.appengine.flex.test".  This way flex feature would not be visible to customers

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.flex.test/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.flex.test/META-INF/MANIFEST.MF
@@ -1,0 +1,18 @@
+Manifest-Version: 1.0
+Bundle-ClassPath: .
+Bundle-ManifestVersion: 2
+Bundle-Name: Front End App Engine Flexible Support Tests
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.flex.test
+Bundle-Version: 0.1.0.qualifier
+Bundle-Vendor: Google Inc.
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Require-Bundle: org.eclipse.wst.common.project.facet.core,
+ org.junit,
+ org.eclipse.core.resources,
+ org.eclipse.core.runtime,
+ com.google.guava,
+ com.google.cloud.tools.eclipse.test.dependencies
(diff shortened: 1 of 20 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.flex.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,18 @@
+Manifest-Version: 1.0
+Bundle-ClassPath: .
+Bundle-ManifestVersion: 2
+Bundle-Name: Front End App Engine Flexible Support Tests
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.flex.test
+Bundle-Version: 0.1.0.qualifier
```
elharo: could we move this to 1.0.0 or would that just break everything and be a huge ball of mess?

briandealwis:
Best practice in OSGi is to have unique packages per bundle otherwise you hit the dreaded [_package-uses_ problem](https://spring.io/blog/2008/11/22/diagnosing-osgi-uses-conflicts/).

That said, we could make the flex bundle a fragment, then sharing a package would be fine.

nbashirbello:
The flex facet definition is added the to new bundle com/google/cloud/tools/eclipse/appengine/facets/flex so that it would not be visible in subsequent builds. After GA, the flex facet definition and the rest of the UI definitions (e.g. new project wizard for flex) that will be in the new bundle, will be added to the correct bundles and this bundle will be moved. 

The flex feature is supposed to be complete by the end of March, hence why I am doing it this way, so as to meet this deadline

elharo:
We need to hide user visible features such as menu items. I'm not sure we actually need to hide the facet though. You have to look really hard to find this in Eclipse. I'd much prefer to put it in the correct place from the beginning.

nbashirbello:
It is not hard to find the facet in eclipse. It is exposed everytime you open the project properties. If we are hiding the flex features, I think we should hide all the flex features.

elharo:
But if the flex facet is added in advance of any other features, it's a noop right?

nbashirbello:
I am beginning to add the flex features and they are dependent on the flex facet. This is the first cl in a long line of cls to come.

nbashirbello:
@elharo Yes that is correct.

PR #1473, opened 2017-02-23
Util class to send exception to internal server

chanseokoh:
- Introduces a single method ExceptionSender.sendException(), a kind of static util method that logs an exception report to our internal system.

Note that this PR just adds the above method only. No one calls the method, and nothing is being sent to us.

(BTW, the commit history got really dirty because I tried to preserve @briandealwis' initial commits from his branch.)


plugins/com.google.cloud.tools.eclipse.aeri/src/com/google/cloud/tools/eclipse/aeri/ExceptionSender.java
```diff
@@ -0,0 +1,108 @@
...
+                     + "userComment: " + nullOrEmptyToNone(userComment);
+    // "comments" seems to be the only viable place where we can put extra info.
+    parameters.put("comments", extraInfo);
+
+    try {
+      // Internal system expects the product ID and version in URL too.
+      String urlParameters = "?product=" + CloudToolsInfo.EXCEPTION_REPORT_PRODUCT_ID
+          + "&version=" + CloudToolsInfo.getToolsVersion();
+      int code = HttpUtil.sendPostMultipart(endpointUrl + urlParameters, parameters);
+      if (code != 200) {
```
akerekes: `com.google.api.client.http.HttpStatusCodes.STATUS_CODE_OK`

plugins/com.google.cloud.tools.eclipse.aeri/src/com/google/cloud/tools/eclipse/aeri/ExceptionSender.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @VisibleForTesting
+  ExceptionSender(String endpointUrl) {
+    this.endpointUrl = endpointUrl;
+  }
+
+  void sendException(IThrowable exception, String eclipseBuildId, String javaVersion,
+      String os, String osVersion, String userSeverity, String userComment) {
+    Map<String, String> parameters = new HashMap<>();
+    parameters.put("product", CloudToolsInfo.EXCEPTION_REPORT_PRODUCT_ID);
```
akerekes: Extract parameter names into string constants?

plugins/com.google.cloud.tools.eclipse.aeri.test/src/com/google/cloud/tools/eclipse/aeri/ExceptionSenderWithServerTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.eclipse.epp.logging.aeri.core.IThrowable;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class ExceptionSenderWithServerTest {
+
+  @Rule public TestHttpServer server = new TestHttpServer("", "");
+
+  private static final String EXPECTED_POST_BODY =
+      "-----------------------------45224ee4-f3c1-4b23-8df1-4012f722218c" + "\r\n" +
```
akerekes: I wish there was a better way to verify this...

chanseokoh: There seems to be, but I'll hold off doing it for now.

plugins/com.google.cloud.tools.eclipse.aeri/src/com/google/cloud/tools/eclipse/aeri/ExceptionSender.java
chanseokoh: Good idea, but no need to pull google-http-java-client into this bundle. I will use `java.net.HttpURLConnection.HTTP_OK`.

chanseokoh:
The build failure on Mars seems to be because it can't load AERI v2. There should be a way to exclude the `.aeri` and `.aeri.test` bundles at the first place.

plugins/com.google.cloud.tools.eclipse.aeri/src/com/google/cloud/tools/eclipse/aeri/ExceptionSender.java
```diff
@@ -0,0 +1,120 @@
...
+    // "comments" seems to be the only viable place where we can put extra info.
+    parameters.put(COMMENTS_KEY, extraInfo);
+
+    try {
+      // Internal system expects the product ID and version in URL too.
+      String urlParameters = "?product=" + CloudToolsInfo.EXCEPTION_REPORT_PRODUCT_ID
+          + "&version=" + CloudToolsInfo.getToolsVersion();
+      int code = HttpUtil.sendPostMultipart(endpointUrl + urlParameters, parameters);
+      if (code != HttpURLConnection.HTTP_OK) {
+        logger.log(Level.WARNING, "Failed to send exception report. HTTP response code: " + code);
```
briandealwis: This would seem worthy of `ERROR`, no?

plugins/com.google.cloud.tools.eclipse.aeri/src/com/google/cloud/tools/eclipse/aeri/ExceptionSender.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  /** Format the modeled stack trace. */
+  @VisibleForTesting
+  static String formatStackTrace(IThrowable exception) {
+    if (exception == null) {
+      return NONE_MARKER;
+    }
+
+    // We need this manual formatting because exception of type "Exception" isn't anonymized.
```
briandealwis: I don't see the anonymization?

plugins/com.google.cloud.tools.eclipse.aeri/META-INF/MANIFEST.MF
```diff
@@ -7,3 +7,10 @@ Bundle-Version: 0.1.0.qualifier
 Bundle-Vendor: Google Inc.
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Bundle-ActivationPolicy: lazy
+Import-Package: com.google.cloud.tools.eclipse.util,
+ com.google.cloud.tools.eclipse.util.io,
+ com.google.common.annotations;version="[20.0.0,21.0.0)",
+ org.eclipse.emf.common.util,
+ org.eclipse.epp.logging.aeri.core
+Require-Bundle: org.eclipse.core.runtime;bundle-version="3.11.0"
```
briandealwis: Do we really need org.eclipse.core.runtime?

plugins/com.google.cloud.tools.eclipse.aeri/src/com/google/cloud/tools/eclipse/aeri/ExceptionSender.java
chanseokoh: No code to do anonymization here. I will update the comment to clarify. What I am saying is like this:

"We should send a stack trace from AERI's `IThrowable`, not a stack trace from the standard Java `Throwable`, because `Throwable` is never anonymized while `IThrowable` might have been. And because AERI's `IThrowable` doesn't implement an equivalent of `Throwable.printStackTrace()`, we have to manually implement it."

chanseokoh:
> The build failure on Mars seems to be because it can't load AERI v2. There should be a way to exclude the .aeri and .aeri.test bundles at the first place.

Interestingly, `mvn verify`, which targets Mars AFAICT, doesn't fail locally, for the reason related to loading AERI v2. Maybe builds on Travis have a slightly different environment?

(I know there is an actual test failure.)

chanseokoh:
Let me close this for now and reopen later when the time comes.

PR #1488, opened 2017-02-27
Web.xml source validation and quick assist

abbiefarr:
Implementation for source validation is the same as for appengine-web.xml. But if I open web.xml in the editor, `connect(IDocument document)` in `AbstractXmlSourceValidator` is never called.

I suspect something is wrong with plugin.xml but I'm not sure. It seems like the file name isn't recognized so validation never starts. The file name is referenced in extension point `org.eclipse.core.contenttype.contentTypes`.


plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
```diff
@@ -99,19 +99,39 @@
...
   <extension point="org.eclipse.core.contenttype.contentTypes"> 
     <content-type 
         name="appengine-web.xml"
         id="appengine-web.xml"
         base-type="org.eclipse.core.runtime.xml"      
     	file-names="appengine-web.xml">
+    </content-type>
+    <content-type 
+        name="web.xml"
+        id="web.xml"
```
elharo: Try adding priority="high" here. See if that helps.

abbiefarr: No luck :/

briandealwis:
I found two small configuration issues.

First, your `.validation` bundle needs to be marked as lazily-activated by adding the following header to your `MANIFEST.MF`:
```
Bundle-ActivationPolicy: lazy
```
Required as the Extension Registry only exposes extensions from plugins that are either started or marked as lazily activated.

Second, `web.xml` already has a content-type id `org.eclipse.jst.jee.ee5webDD` defined in [`org.eclipse.jst.j2ee`](http://git.eclipse.org/c/jeetools/webtools.javaee.git/tree/plugins/org.eclipse.jst.j2ee/plugin.xml?h=master#n865).  It has `priority=high` which is why you probably weren't able to override it.  Remove the `web.xml` content type and use `org.eclipse.jst.jee.ee5webDD` for the validator.

Note that Eclipse content types generally follow the reversed domain-name convention.  So our `appengine-web.xml` content type should probably be something like `com.google.appengine.dd.xml`; I'm assuming `app.yml` should be `com.google.appengine.dd.yaml`.  I didn't include the `cloud.tools.eclipse` because these content-types are global, and can't be overridden (as we saw above).

abbiefarr:
Ready for review

plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
```diff
@@ -118,7 +118,7 @@
 		  scope="total"
 		  class="com.google.cloud.tools.eclipse.appengine.validation.AppEngineWebXmlSourceValidator"
 		  id="appEngineWebXmlSourceValidator">
-		<contentTypeIdentifier id="appengine-web.xml">
+		<contentTypeIdentifier id="com.google.appengine.dd.xml">
```
elharo: What's dd? deployment descriptor? That usually refers to web.xml, not appengine-web.xml:

https://cloud.google.com/appengine/docs/standard/java/config/webxml

If this is validating web.xml then the class/ID should probably be  WebXmlSourceValidator rather than AppEngineWebXmlSourceValidator

abbiefarr: This is validating appengine-web.xml. I changed it based on Brian's suggestion. Should probably have been in a separate PR.

elharo: In that case why the "dd"?

abbiefarr: Not sure. @briandealwis ?

chanseokoh:
> We need to verify that the project is for App Engine Standard, right?

This is important to confrim. We must warn only if it is our App Engine standard env project.

plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
briandealwis: It seemed like an App Engine deployment descriptor to me!  I'm open to better names.

elharo: Per https://cloud.google.com/appengine/docs/standard/java/config/appref it's called a "configuration file".

briandealwis: When I proposed this to @abbiefarr I thought it would be good to have a similar name for the `app.yaml` file.  So would `com.google.appengine.conf.xml` and `com.google.appengine.conf.yaml` seem ok?

elharo: I'd prefer config instead of conf, or maybe even configuration; but otherwise it sounds fine

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
```diff
@@ -44,15 +55,21 @@
...
    */
   public void validate(IValidationContext helper, IReporter reporter) throws ValidationException {
+    IProject project = getProject(helper);
     try {
-      String encoding = getDocumentEncoding(document);
-      byte[] bytes = document.get().getBytes(encoding);
-      this.validate(reporter, bytes);
+      IFacetedProject facetedProject = ProjectFacetsManager.create(project);
+      if (AppEngineStandardFacet.hasAppEngineFacet(facetedProject)) {
+        hasValidated = true;
```
elharo: not sure it matters, but I'd expect this to come after this.validate in case validate() throws an exception

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
```diff
@@ -79,6 +96,38 @@ void createMessage(IReporter reporter, BannedElement element, int elementOffset,
...
+    }
+    return null;
+  }
+  
+  /**
+   * Returns the IFile for a given URI or null if the file does
+   * not exist in the workspace.
+   */
+  IFile getFile(String fileUri) {
+    IPath path = new Path(fileUri);
```
elharo: right here: if this is a URI (begins with a file:) then this shouldn't work. If it's a path it should.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
```diff
@@ -44,15 +55,21 @@
       AbstractXmlSourceValidator.class.getName());
   
   private IDocument document;
+  private boolean hasValidated;
```
elharo: Is this field needed for anything except tests?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
abbiefarr: Yep, IncrementalHelper.setURI() takes in a string, and I was using file.getLocationURI().toString(), which began with file:, when I actually needed the path.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
abbiefarr: nope

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
```diff
@@ -79,6 +96,38 @@ void createMessage(IReporter reporter, BannedElement element, int elementOffset,
...
+      }
+    }
+    return null;
+  }
+  
+  /**
+   * Returns the IFile for a given URI or null if the file does
+   * not exist in the workspace.
+   */
+  IFile getFile(String filePath) {
```
elharo: Looks like this can be static.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlSourceValidator.java
```diff
@@ -79,6 +96,38 @@ void createMessage(IReporter reporter, BannedElement element, int elementOffset,
     reporter.addMessage(this, message);
   }
   
+  /**
+   * Returns the underlying IProject from a given IValidationContext or
+   * null if the IValidationContext does not return any files that need
+   * to be validated.
+   */
+  IProject getProject(IValidationContext helper) {
```
elharo: Looks like this can be static.

PR #1489, opened 2017-02-27
Cause CloudToolsInfo.getToolsVersion() to return feature version

briandealwis:


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
```diff
@@ -16,20 +16,38 @@
...
+   */
+  private static final String CLOUD_TOOLS_FOR_ECLIPSE_FEATURE_ID =
+      "com.google.cloud.tools.eclipse.suite.e45.feature";
+
   // Don't change the value; this name is used as an originating "application" of usage metrics.
   public static String METRICS_NAME = "gcloud-eclipse-tools";
 
   public static String USER_AGENT = METRICS_NAME + "/" + getToolsVersion();
 
+  /** Return the version of associated feature. May be slow. */
```
akerekes: What may cause the slowness

akerekes: Would it make sense to cache the value? Or initialize it on plugin/bundle startup?

briandealwis: I think I'm being overly cautious.  It's certainly not a network fetch.  I think caching it is fine: realistically if someone updates the feature, they'll restart.

briandealwis:
Tests are failing because the version string is embedded in a few places. I'll have to fix those up.

chanseokoh:
> Tests are failing because the version string is embedded in a few places. I'll have to fix those up.

I recommend to merge the current HEAD on master, as I made some changes in some related places.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/CloudToolsInfo.java
```diff
@@ -16,20 +16,38 @@
...
   public static String getToolsVersion() {
-    return FrameworkUtil.getBundle(CloudToolsInfo.class).getVersion().toString();
+    for(IBundleGroupProvider provider : Platform.getBundleGroupProviders()) {
+      for(IBundleGroup feature : provider.getBundleGroups()) {
+        if(CLOUD_TOOLS_FOR_ECLIPSE_FEATURE_ID.equals(feature.getIdentifier())) {
```
chanseokoh: Space needed after `for` and `if`.

```java
for (...
  for (...
    if (...
```

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/CloudToolsInfoTest.java
```diff
@@ -16,18 +16,43 @@
...
+        .thenReturn(new IBundleGroup[] {bundleGroup});
+    Mockito.when(bundleGroup.getIdentifier())
+        .thenReturn(CloudToolsInfo.CLOUD_TOOLS_FOR_ECLIPSE_FEATURE_ID);
+    Mockito.when(bundleGroup.getVersion()).thenReturn("123.456.789");
+    Assert.assertEquals("123.456.789",
+        CloudToolsInfo.getToolsVersion(new IBundleGroupProvider[] {bundleGroupProvider}));
+  }
+
+
+  // Assert.assertTrue(CloudToolsInfo.getToolsVersion().startsWith("0.1.0."));
```
chanseokoh: Should go away!

briandealwis:
Short of building, no.

briandealwis:
I had hoped that switching the launch to be feature based might work, but it doesn't include our source features, just binary features.

PR #1490, opened 2017-02-27
Fix inconsistent class hierarchy errors within Eclipse

briandealwis:
I'm seeing inconsistent-class-hierarchy errors within Eclipse as the compiler requires some JFace class definitions when processing `XsltQuickFix` and `ValidationTestUtils`.  We can make them available with the `additional.bundles` directive.

screen shot 2017-02-27 at 4 44 55 pm


chanseokoh:
Weird. I don't see these errors.

briandealwis:
I'm kinda baffled why @chanseokoh and @abbiefarr aren't seeing these issues, but I can't make them go away!

PR #1495, opened 2017-02-28
reduce public visibility

elharo:


elharo:
@akerekes

elharo:
Ping @akerekes

elharo:
Ping ping ping

elharo:
The reason is general code safety. Last week I spent two days hunting down a flaky bug in a PR that occurred because the test was mutating internal private state.

PR #1504, opened 2017-03-01
Uncheck "Stop previous version" when not promoting in deploy preferences panel

### LINKED ISSUE #1229: "Stop Previous Version" should not be checked when disabled
When the "Stop Previous Version" checkbox in the deployment dialog is disabled, we will not stop the previous version. However if the checkbox was previously checked when enabled, it still is when disabled. 

image

chanseokoh:
Fixes #1229.

@briandealwis thinks #1229 may not be a bug in a certain sense, but I see there is no harm in unchecking the disabled button. This will eliminate any possible confusion — it will convey the message in a clear way that it will not stop any previous version.

BTW, it was surprisingly painful to figure out how to fix this seemingly trivial bug.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -88,9 +89,11 @@
 
   private Text version;
 
-  private Button autoPromoteButton;
+  @VisibleForTesting
+  Button autoPromoteButton;
```
elharo: There should be a way to locate the button in the widget by starting at the parent that does not require access to the private fields. Just walk down the widget tree looking for a Button with the right name/ID. Not sure it's worth it though. Doesn't look like SWT supports this as well as it should.

http://stackoverflow.com/questions/17945611/how-to-assign-unique-id-to-swt-widget-and-get-it-in-swtbot

chanseokoh: Sweet. Thanks for the ref. I can make it work.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -213,24 +214,53 @@ private void setupProjectVersionDataBinding(DataBindingContext context) {
...
+      protected Object calculate() {
+        boolean buttonEnabled = (boolean) stopPreviousVersionEnablement.getValue();
+        boolean buttonValue = (boolean) stopPreviousVersion.getValue();
+        boolean currentValue = (boolean) currentStopPreviousVersionChoice.getValue();
+        if (buttonEnabled) {
+          return buttonValue;  // Remember the button state as the latest choice if it is enabled.
+        }
+        return currentValue;  // Otherwise, retain the latest (current) user choice.
+      }
+    }, stopPreviousVersionModel, null, new UpdateValueStrategy(UpdateValueStrategy.POLICY_NEVER));
```
briandealwis: Ugh.  This is the situation for Neon's _SideEffect_.  I wonder if it's worth rolling our own for Mars.

PR #1506, opened 2017-03-01
Refactor BannedElement

abbiefarr:
Creates subclasses of BannedElement that contain a marker ID and severity, to allow validators to add more than one type of marker to a resource during validation.

@elharo @meltsufin

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
@@ -27,7 +27,7 @@ public class BannedElementTest {
 
   @Test(expected = NullPointerException.class)
   public void testBannedElementConstructor_nullLocation() {
-    new BannedElement("test", "", 1, null, 0);
+    new BannedElement("test", "org.eclipse.core.resources.problemmarker", 1, null, 0);
   }
 
 }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidator.java
@@ -67,10 +67,10 @@ public abstract class AbstractXmlValidator extends AbstractValidator {
   /**
    * Creates a marker from a given {@link BannedElement}
    */
-  static void createMarker(IResource resource, BannedElement element,
-      int elementOffset, String markerId, int severity) throws CoreException {
-    IMarker marker = resource.createMarker(markerId);
-    marker.setAttribute(IMarker.SEVERITY, severity);
+  static void createMarker(IResource resource, BannedElement element, int elementOffset)
+      throws CoreException {
+    IMarker marker = resource.createMarker(element.getMarkerId());
+    marker.setAttribute(IMarker.SEVERITY, element.getSeverity());
     marker.setAttribute(IMarker.MESSAGE, element.getMessage());
     marker.setAttribute(IMarker.LOCATION, "line " + element.getStart().getLineNumber());
     marker.setAttribute(IMarker.CHAR_START, elementOffset);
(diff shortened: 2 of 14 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
```diff
@@ -27,7 +27,7 @@ public void testBannedElementConstructor_nullElementName() {
 
   @Test(expected = NullPointerException.class)
   public void testBannedElementConstructor_nullLocation() {
-    new BannedElement("test", null, 0);
+    new BannedElement("test", "", 1, null, 0);
```
elharo: I would expect that "" would be flagged as an illegal marker

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidator.java
```diff
@@ -42,10 +41,10 @@ protected void validate(IResource resource, byte[] bytes)
       SaxParserResults parserResults = WebXmlSaxParser.readXml(bytes);
       Map<BannedElement, Integer> bannedElementOffsetMap =
           ValidationUtils.getOffsetMap(bytes, parserResults);
-      String markerId = "com.google.cloud.tools.eclipse.appengine.validation.servletMarker";
       for (Map.Entry<BannedElement, Integer> entry : bannedElementOffsetMap.entrySet()) {
-        createMarker(resource, entry.getKey(), entry.getValue(),
-            markerId, IMarker.SEVERITY_ERROR);
+        BannedElement element = entry.getKey();
+        createMarker(resource, element, entry.getValue(),
```
elharo: You don't need to pass the markerId and severity separately. createMarker can simply call the getter methods on element

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
abbiefarr: Marker IDs are only used when searching the resource for markers for quick fixes. Should I add my own check? IResource.createMarker(marker ID) doesn't have any checks.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidator.java
```diff
@@ -42,9 +42,7 @@ protected void validate(IResource resource, byte[] bytes)
       Map<BannedElement, Integer> bannedElementOffsetMap =
```
elharo: separate issue, but I just realized that since we're using BannedElement as a key in a map, we should give it equals and hashCode methods: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1507

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidator.java
```diff
@@ -67,10 +67,10 @@ static void deleteMarkers(IResource resource) throws CoreException {
   /**
    * Creates a marker from a given {@link BannedElement}
    */
-  static void createMarker(IResource resource, BannedElement element,
-      int elementOffset, String markerId, int severity) throws CoreException {
-    IMarker marker = resource.createMarker(markerId);
-    marker.setAttribute(IMarker.SEVERITY, severity);
+  static void createMarker(IResource resource, BannedElement element, int elementOffset)
```
elharo: Do we need to pass elementOffset here? Is it different than 

element.getStart().getLineNumber()?

abbiefarr: Yes, elementOffset character offset

PR #1511, opened 2017-03-02
some tests for  AppEngineLibrariesPage

elharo:
@abbiefarr @briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/MessagesTest.java
```diff
@@ -32,5 +32,12 @@ public void testDescription() {
         "Additional jars commonly used in App Engine Standard Environment applications", 
         Messages.getString("description"));
   }
+  
+
+  @Test
+  public void testUnknownMessage() {
```
akerekes: I don't see much value in this test.

elharo: I would still like to figure out how to move the Messages code and out tests for the same to a common location. In the meantime I happened to notice in the jacoco reports that we had never actually tested this. It's a cheap test.

PR #1513, opened 2017-03-02
Fixes #1439 - Verify if we need to handle proxy settings in com.google.cloud.tools.eclipse.projectselector.GoogleApiFactory

### LINKED ISSUE #1439: Verify if we need to handle proxy settings in com.google.cloud.tools.eclipse.projectselector.GoogleApiFactory
We should verify whether the proxy settings (system.workspace/project level) are used when accessing API services

akerekes:
I've turned `GoogleApiFactory` into a service.
A proxy settings change listener is used to ensure we respect those changes.

A test for proxy setting changes is still pending, but the rest is ready for review.


plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/GoogleApiFactory.java
```diff
@@ -22,31 +22,85 @@
...
  * Class to obtain various Google cloud Platform related APIs.
- *
- * TODO move this class into a separate API bundle
- * https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1438
  */
-public class GoogleApiFactory {
+@Component
+public class GoogleApiFactory implements IGoogleApiFactory {
 
+  private static final String GOOGLEAPIS_URL = "https://appengine.googleapis.com";
```
chanseokoh: It doesn't feel right that this one contains `appengine`.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/GoogleApiFactory.java
```diff
@@ -22,31 +22,85 @@
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.jackson.JacksonFactory;
+import com.google.api.client.repackaged.com.google.common.base.Preconditions;
```
chanseokoh: This is from a wrong dependency.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/GoogleApiFactory.java
```diff
@@ -22,31 +22,85 @@
...
+  GoogleApiFactory(String googleApiUrl) {
+    this.googleApiUrl = googleApiUrl;
+  }
+
+  @Activate
+  public void init() {
+    Preconditions.checkNotNull(proxyService, "proxyService is null");
+    jsonFactory = new JacksonFactory();
+    buildTransport();
+    proxyService.addProxyChangeListener(new IProxyChangeListener() {
```
chanseokoh: I'm concerned about thread-safety here. It's not a UI widget, so I think this can be executed in any thread.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  private void buildTransports() {
+    synchronized (transports) {
+      try {
+        buildAppEngineAdminTransport();
+        buildCloudResourceManagerTransport();
+      } catch (URISyntaxException ex) {
+        logger.log(Level.SEVERE, "Could not create transport using the proxy settings", ex);
+        transports = null;
```
chanseokoh: I think this will cause NPE in many places onward. Even if you initialize it with an empty map, it will cause trouble in places where you do `transports.get(...)`. Probably it's better to leave `transport` as-is? At least it will keep previous, working transports.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+   * @return the CloudResourceManager/Projects API
+   */
+  @Override
+  public Projects newProjectsApi(Credential credential) {
+    synchronized (transports) {
+      checkFields();
+      CloudResourceManager resourceManager =
+          new CloudResourceManager.Builder(transports.get(googleApiUrls.cloudResourceManagerUrl()),
+                                           jsonFactory, credential)
+          .setApplicationName(CloudToolsInfo.USER_AGENT).build();
```
chanseokoh: Nit: indent +4

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+   * @return the Appengine/Apps API
+   */
+  @Override
+  public Apps newAppsApi(Credential credential) {
+    synchronized (transports) {
+      checkFields();
+      Appengine appengine =
+          new Appengine.Builder(transports.get(googleApiUrls.appEngineAdminUrl()), jsonFactory,
+                                credential)
+          .setApplicationName(CloudToolsInfo.USER_AGENT).build();
```
chanseokoh: Nit: indent +4

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
chanseokoh: Speaking of which, should we be prepared for when `transports.get(...)` returns null too?

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+ */
+@Component
+public class GoogleApiFactory implements IGoogleApiFactory {
+
+  private static final int DEFAULT_TIMEOUT_MS = 1000;
+
+  private static final Logger logger = Logger.getLogger(GoogleApiFactory.class.getName());
+
+  private JsonFactory jsonFactory;
+  private Map<String, HttpTransport> transports = new HashMap<>();
```
briandealwis: Are these HttpTransports expensive?  Do we really want them living forever?  Would we be better off making this a Guava `Cache` instead?

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+                   new NetHttpTransport.Builder().setConnectionFactory(connectionFactory).build());
+  }
+
+  private Proxy createProxy(String url) throws URISyntaxException {
+    Preconditions.checkNotNull(proxyService, "proxyService is null");
+    Preconditions.checkNotNull(googleApiUrls, "googleApiUrl is null");
+
+    IProxyData[] proxyData = proxyService.select(new URI(url));
+    for (final IProxyData iProxyData : proxyData) {
+      if (IProxyData.HTTPS_PROXY_TYPE.equals(iProxyData.getType())) {
```
briandealwis: Not supporting SOCKS proxies?  Hmm, we should log something if the IProxyService returned some proxy data (i.e., we should use a proxy) and we're not supporting it — otherwise we silently fail.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+    ConnectionFactory connectionFactory =
+        new TimeoutAwareConnectionFactory(createProxy(url), DEFAULT_TIMEOUT_MS, DEFAULT_TIMEOUT_MS);
+    transports.put(url,
+                   new NetHttpTransport.Builder().setConnectionFactory(connectionFactory).build());
+  }
+
+  private Proxy createProxy(String url) throws URISyntaxException {
+    Preconditions.checkNotNull(proxyService, "proxyService is null");
+    Preconditions.checkNotNull(googleApiUrls, "googleApiUrl is null");
+
```
briandealwis: Since none of the Google APIs support HTTP anymore (right?) then we could assert here that the URL does not start with 'http://', so we should never see a proxy-data of type HTTP.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,185 @@
...
+  GoogleApiFactory(GoogleApiUrls googleApiUrls) {
+    this.googleApiUrls = googleApiUrls;
+  }
+
+  @Activate
+  public void init() {
+    Preconditions.checkNotNull(proxyService, "proxyService is null");
+    jsonFactory = new JacksonFactory();
+    buildTransports();
+    proxyService.addProxyChangeListener(new IProxyChangeListener() {
```
briandealwis: I think this would be better if we made this service optionally dependent on the IProxyService (i.e., `ReferenceCardinality.OPTIONAL`: 0..1).  If there is no proxy service, we just treat everything like DIRECT.  (I believe the IProxyService is actually registered programmatically in `org.eclipse.core.net`'s activator and so if that hasn't been activated — which is unlikely, but possible — then our service won't be available.)

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: The reason to keep them around is because of the Javadoc of [NetHttpTransport](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/javanet/NetHttpTransport):

> Implementation is thread-safe. For maximum efficiency, applications should use a single globally-shared instance of the HTTP transport.

briandealwis: It's your call, but I'd prefer to have them GC'd when they're unused.  Guava Caches are thread-safe, so we can probably get rid of synchronization here, and the CacheLoader will simplify some of the logic here.  And using weakValues() means the instances will stick around as long as they're being used.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -150,22 +152,42 @@ private void buildTransportForUrl(String url) throws URISyntaxException {
...
-        return new Proxy(Type.HTTP, new InetSocketAddress(iProxyData.getHost(),
-                                                          iProxyData.getPort()));
+      switch (iProxyData.getType()) {
+        case IProxyData.HTTPS_PROXY_TYPE:
+          return new Proxy(Type.HTTP, new InetSocketAddress(iProxyData.getHost(),
+                                                            iProxyData.getPort()));
+        case IProxyData.SOCKS_PROXY_TYPE:
+          return new Proxy(Type.SOCKS, new InetSocketAddress(iProxyData.getHost(),
+                                                             iProxyData.getPort()));
+        default:
```
briandealwis: This code isn't handling a proxy directive, and we should at least log something.

briandealwis: In fact, in the case of the unhandled, we should continue looping through any other proxyData items (e.g., IProxyService returns { HTTP_PROXY_TYPE, HTTPS_PROXY_TYPE } — we don't do the first, but we will do the second).  The default case should happen outside of the `for` loop.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -150,22 +152,42 @@ private void buildTransportForUrl(String url) throws URISyntaxException {
...
-    Preconditions.checkNotNull(proxyService, "proxyService is null");
     Preconditions.checkNotNull(googleApiUrls, "googleApiUrl is null");
+    Preconditions.checkArgument(!url.startsWith("http://"), "http is not supported schema");
+
+    if (proxyService == null) {
+      return Proxy.NO_PROXY;
+    }
 
     IProxyData[] proxyData = proxyService.select(new URI(url));
     for (final IProxyData iProxyData : proxyData) {
```
briandealwis: plural: proxyDatum?  Can we come up with a better name than `iProxyData`?

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

chanseokoh: I think @briandealwis was suggesting `iProxyData` --> `proxyDatum`. (Technically `Datum` is a singular form, but I guess nowadays people just use `data` for anything.)

briandealwis: I was referring to iProxyData.  And oops re: datum being singular!

plugins/com.google.cloud.tools.eclipse.googleapis/build.properties
```diff
@@ -3,6 +3,7 @@ output.. = target/classes/
 bin.includes = META-INF/,\
                .,\
                plugin.properties,\
-               lib/
+               lib/,\
+               OSGI-INF/com.google.cloud.tools.eclipse.googleapis.internal.GoogleApiFactory.xml
```
briandealwis: Might just want to make this `OSGI-INF/` rather than the specific file.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,208 @@
...
+        new TimeoutAwareConnectionFactory(createProxy(url), DEFAULT_TIMEOUT_MS, DEFAULT_TIMEOUT_MS);
+    transports.put(url,
+                   new NetHttpTransport.Builder().setConnectionFactory(connectionFactory).build());
+  }
+
+  private Proxy createProxy(String url) throws URISyntaxException {
+    Preconditions.checkNotNull(googleApiUrls, "googleApiUrl is null");
+    Preconditions.checkArgument(!url.startsWith("http://"), "http is not supported schema");
+
+    if (proxyService == null) {
```
briandealwis: I'm sorry for suggesting you make this dynamic.  You should operate on a copy of the proxyService reference as it may be unbound before the call to `select()` below.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,208 @@
...
+          break;
+      }
+    }
+    return Proxy.NO_PROXY;
+  }
+
+  @Reference(policy=ReferencePolicy.DYNAMIC, cardinality=ReferenceCardinality.OPTIONAL)
+  public void setProxyService(IProxyService proxyService) {
+    this.proxyService = proxyService;
+    this.proxyService.addProxyChangeListener(proxyChangeListener);
```
briandealwis: I'm pretty sure OSGi ensures you won't be unbound while being bound, so this field access is safe.

akerekes:
@briandealwis @elharo @chanseokoh PTAL
FYI tests are coming, but functionality-wise it's ready for another round of review.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -55,153 +47,85 @@
...
     }
-    return Proxy.NO_PROXY;
   }
 
   @Reference(policy=ReferencePolicy.DYNAMIC, cardinality=ReferenceCardinality.OPTIONAL)
   public void setProxyService(IProxyService proxyService) {
     this.proxyService = proxyService;
     this.proxyService.addProxyChangeListener(proxyChangeListener);
-    buildTransports();
+    if (proxyFactory != null) {
```
briandealwis: Is `proxyFactory` ever null?  Would it make sense to make it _final_ and set in the constructor?

plugins/com.google.cloud.tools.eclipse.googleapis.test/src/com/google/cloud/tools/eclipse/googleapis/internal/TimeoutAwareConnectionFactoryTest.java
```diff
@@ -0,0 +1,86 @@
...
+                  new InetSocketAddress(proxyServer.getHostname(), proxyServer.getPort()));
+    HttpURLConnection connection =
+        new TimeoutAwareConnectionFactory(proxy, 1764, 3528).openConnection(new URL(NONEXISTENTDOMAIN));
+    connectReadAndDisconnect(connection);
+    assertThat(connection.getConnectTimeout(), is(1764));
+    assertThat(connection.getReadTimeout(), is(3528));
+  }
+
+  private void connectReadAndDisconnect(HttpURLConnection connection) throws IOException {
+    connection.connect();
```
chanseokoh: I think it's concise and also better to just check

```java
assertThat(connection.getResponseCode(), is(200));
```

and remove the try block altogether (while leaving the `finally` block.)

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public void proxyInfoChanged(IProxyChangeEvent event) {
+      if (transportCache != null) {
+        transportCache.invalidateAll();
+      }
+    }
+  };
+
+  public GoogleApiFactory() {
+    proxyFactory = new ProxyFactory();
```
chanseokoh: Personally, I prefer `this(new ProxyFactory())`. This way, it's easy to tell if there are difference between the two constructors, and if so, how they differ.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/TransportCacheLoader.java
```diff
@@ -0,0 +1,38 @@
...
+package com.google.cloud.tools.eclipse.googleapis.internal;
+
+import com.google.api.client.http.HttpTransport;
+import com.google.api.client.http.javanet.ConnectionFactory;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.common.cache.CacheLoader;
+
+class TransportCacheLoader extends CacheLoader<GoogleApiUrl, HttpTransport> {
+  
+  private ProxyFactory proxyFactory;
```
chanseokoh: Could be `final`.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
```diff
@@ -121,7 +122,7 @@ public AppEngine getAppEngineApplication(Credential credential, String projectId
       // just in case the API changes and exception with 404 won't be
       // used to indicate a missing application
       return AppEngine.withId(application.getId());
-    } catch (IOException ex) {
+    } catch (IOException | GoogleApiException ex) {
```
chanseokoh: Now we're at it, I think it's better to

```java
} catch (GoogleJsonResponseException ex) {
  ...
} catch (IOException | GoogleApiException ex) {
  ...
}
```

and remove the `if (ex instanceof GoogleJsonResponseException)` {.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiUrl.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.google.cloud.tools.eclipse.googleapis.internal;
+
+public enum GoogleApiUrl {
+
+  APPENGINE_ADMIN_API("https://appengine.googleapis.com"),
+  CLOUDRESOURCE_MANAGER_API("https://cloudresourcemanager.googleapis.com");
+
+  private String url;
```
chanseokoh: `final`?

plugins/com.google.cloud.tools.eclipse.googleapis.test/src/com/google/cloud/tools/eclipse/googleapis/internal/TimeoutAwareConnectionFactoryTest.java
akerekes: My intent was to check the response and fail if we accidentally connected to an existing site. I've changed the code to make this more explicit.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/TransportCacheLoader.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiUrl.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/IGoogleApiFactory.java
```diff
@@ -16,30 +16,23 @@
...
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.services.appengine.v1.Appengine.Apps;
 import com.google.api.services.cloudresourcemanager.CloudResourceManager.Projects;
-import com.google.cloud.tools.eclipse.util.CloudToolsInfo;
-import java.io.IOException;
-import org.junit.Test;
 
-public class GoogleApiFactoryTest {
+/**
+ *
```
briandealwis: Should be filled in

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
```diff
@@ -0,0 +1,146 @@
...
+
+  @VisibleForTesting
+  public GoogleApiFactory(ProxyFactory proxyFactory) {
+    Preconditions.checkNotNull(proxyFactory, "proxyFactory is null");
+    this.proxyFactory = proxyFactory;
+  }
+
+  @Activate
+  public void init() {
+    jsonFactory = new JacksonFactory();
```
briandealwis: Any reason you're not initializing this and making it _final_?  You set mocks in the tests, but don't seem to do anything with the mocks.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/TimeoutAwareConnectionFactory.java
```diff
@@ -14,29 +14,40 @@
...
 
 /**
  * Extends the {@link DefaultConnectionFactory} with option to specify connect and read timeout.
- *
- * TODO move this class into a separate API bundle
- * https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1438
  */
 class TimeoutAwareConnectionFactory extends DefaultConnectionFactory {
+
+  private static final int DEFAULT_TIMEOUT_MS = 1000;
```
briandealwis: I wonder if this should be a tuneable parameter.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/IGoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/GoogleApiFactory.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/TimeoutAwareConnectionFactory.java
akerekes: We can change it when it becomes necessary.

PR #1528, opened 2017-03-06
Move HttpUtil to .usagetracker bundle

### LINKED ISSUE #1493: Remove embedded google-http-client-1.22.jar from .util bundle
Now that we have the `.googleapis` bundle which embeds `google-http-client-1.22.jar`, we shouldn't duplicate the same JAR in the `.util` bundle.

Unfortunately, I realized that we can't make `.util` depend on `.googleapis` at the moment, because `.googleapis` is already dependent on `.util` — it will result in a build cycle.

The reason that `.googleapis` depends on `.util` is only because of its use of `CloudToolsInfo.USER_AGENT` to set an application name when calling APIs.

There may be a few options:

- Remove the `.util` dep from `.googleapis` by making `GoogleApiFactory` get a user-agent string from outside, possibly through a constructor.
- Move `CloudToolsInfo` into a new standalone bundle. (Luckily, `CloudToolsInfo` does not depend on `.util`.)
- Move `CloudToolsInfo` into `.googleapis`. (I don't like this).
- ...

chanseokoh:
Fixes #1493.

I manually verified Analytics pings are actually being sent as expected.


chanseokoh:
> are DEFAULT_CONNECT_TIMEOUT_MS ... still needed in HttpUtil?

Nice catch, but `HttpUtil` uses them too, so they are needed: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/HttpUtil.java#L86

PR #1530, opened 2017-03-06
Generate files for new flex project wizard

nbashirbello:


nbashirbello:
Not ready for review

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -36,16 +36,44 @@
...
+    return materialize(project, config, false /* isStandardProject */, monitor);
+  }
+
+  /**
+   * Creates files for a sample App Engine project in the supplied Eclipse project.
+   *
+   * @param project the Eclipse project to be filled with templated code
+   * @param config replacement values
+   * @param isStandardProject true if project should be configured to have the App Engine Standard
+   *   configuration files and false is project should have the App Engine Flexible configuration
```
elharo: "if project"

Might this be an enum? Or should we wait until we have a third option?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -92,8 +120,12 @@ public static IFile materialize(IProject project, AppEngineProjectConfig config,
       properties.put("service", service);  //$NON-NLS-1$
     }
 
-    createChildFile("appengine-web.xml", AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
-        webinf, subMonitor.newChild(5), properties);
+    if (isStandardProject) {
+      createChildFile("appengine-web.xml", AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
+          webinf, subMonitor.newChild(5), properties);
+    } else {
+      copyChildFile("app.yaml", webinf, subMonitor.newChild(5));
```
elharo: curious. Why create one and copy the other?

nbashirbello:
This PR is not ready for review :)

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
nbashirbello: Appengine-web.xml requires input from the user while app.yaml does not.

nbashirbello:
PR ready for review

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
```diff
@@ -86,20 +97,34 @@ public void testMaterialize()
...
-    Assert.assertEquals("true", threadsafe);
-    NodeList sessionsEnabledElements 
-        = doc.getDocumentElement().getElementsByTagName("sessions-enabled");
-    Assert.assertEquals("Must have exactly one sessions-enabled", 
-        1, sessionsEnabledElements.getLength());
-    String sessionsEnabled = sessionsEnabledElements.item(0).getTextContent();
-    Assert.assertEquals("false", sessionsEnabled);
-    
+
+    if (isStandardProject) {
```
elharo: It's best to avoid conditional logic in tests so it's clearer what runs and no assertions are accidentally skipped. Here it's probably easiest to not share helper methods.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -36,16 +36,44 @@
...
+   *
+   * @param project the Eclipse project to be filled with templated code
+   * @param config replacement values
+   * @param isStandardProject true if project should be configured to have the App Engine Standard
+   *   configuration files and false if project should have the App Engine Flexible configuration
+   *   files.
+   * @param monitor progress monitor
+   * @return the most important file created that should be opened in an editor
+   */
+  private static IFile materialize(IProject project, AppEngineProjectConfig config,
```
elharo: Separate methods for flex and standard would be clearer than an extra boolean argument; even more so as we dd other templates for neither of these in the future.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
nbashirbello: Not quite clear. Are you suggesting to not reuse code during testing? In this case, not reusing the validateFiles method?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
nbashirbello: The methods for flex and standard are separate via materializeAppEngineStandardFiles and materializeAppEngineFlexFiles. materialize is a private method that they both utilize.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
elharo: The antipattern is the use of `if`. You can extract common code into shared helper methods, but each test should generally run all statements from start to finish without branching to avoid accidentally skipping tests. See

http://softwareengineering.stackexchange.com/questions/213116/are-branches-in-unit-tests-bad-practice

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
elharo: Yes, and I am suggesting this should be two methods too. Consider using methods as the way to choose between two code paths rather than conditionals.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
nbashirbello: Hey, is this the link you meant to share because this is talking about using the correct test data sets.

elharo: yes, the last answer is particularly on point

PR #1534, opened 2017-03-07
Verify actual Analytics event HTTP requests / refactor AnalyticsPingManager

### LINKED ISSUE #1437: Add a test for AnalyticsPingManager and HttpUtil that actually makes a local HTTP request

chanseokoh:
Fixes #1437.

I also added more sanity checks for `sendPing()` arguments, added more tests, and simplified `AnalyticsPingManager`.

Some notes about `analyticsEnabled` and `unitTestMode`:
- `analyticsEnabled` is now implied by whether `endpointUrl` is null or not. (`endpointUrl` is not null (i.e., Analytics is enabled) only in production or in unit tests.)
- `unitTestMode` should have been removed; `analyticsEnabled` was already true in unit tests. (`analyticsEnabled` was introduced after `unitTestMode` to disable Analytics in integration tests, as the opt-in dialog was interfering with the SWT bot.)


plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -124,7 +125,8 @@ private static boolean isTrackingIdDefined() {
         && Constants.ANALYTICS_TRACKING_ID.startsWith("UA-");
   }
 
-  private String getAnonymizedClientId() {
+  @VisibleForTesting
+  String getAnonymizedClientId() {
     String clientId = preferences.get(AnalyticsPreferences.ANALYTICS_CLIENT_ID, null);
     if (clientId == null) {
       clientId = UUID.randomUUID().toString();
```
elharo: any thread safety issues here? Could we have analytics pinging from two separate operations at once? E.g. user starts a local server and then starts a deploy before the server has started up?

briandealwis: Or since the preferences comes from ConfigurationScope (shared across all instances of this product), two Eclipse workbenches.  (I often have 2-3 workbenches at once.)  Could compute this in the constructor?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -95,26 +97,25 @@ protected IStatus run(IProgressMonitor monitor) {
...
   public static synchronized AnalyticsPingManager getInstance() {
     if (instance == null) {
       IEclipsePreferences preferences = ConfigurationScope.INSTANCE.getNode(PREFERENCES_PLUGIN_ID);
-      if (preferences == null) {
-        throw new NullPointerException("Preference store cannot be null.");
-      }
-
-      boolean analyticsEnabled = !Platform.inDevelopmentMode() && isTrackingIdDefined();
-      Display display = !analyticsEnabled ? null : PlatformUI.getWorkbench().getDisplay();
+      Display display = PlatformUI.getWorkbench().getDisplay();
```
briandealwis: It would be nice to compute this on the fly, with `PlatformUI.getWorkbench()` used as a fallback if `Display.getCurrent() == null` on the off-chance that we're being called on the SWT thread.  And should check that the `PlatformUI.isWorkbenchRunning()` too.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: > Could compute this in the constructor?

I like the constructor idea. It's already synchronized.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -178,10 +153,62 @@ public void testSendPingScheduled_optInRegisteredAndAlreadyOptedIn() {
...
     verify(pingEventQueue, verificationMode).add(any(PingEvent.class));
   }
+
+  @Test
+  public void testGetAnonymizedClientId_generateNewId() {
+    when(preferences.get(eq(AnalyticsPreferences.ANALYTICS_CLIENT_ID), anyString()))
+        .thenReturn(null);  // Simulate that client ID has never been generated.
+    String clientId = pingManager.getAnonymizedClientId();
+    assertTrue(!clientId.isEmpty());
+    verify(preferences, times(1)).put(eq(AnalyticsPreferences.ANALYTICS_CLIENT_ID), eq(clientId));
```
akerekes: using `times(1)` is optional

also you may be able to remove the `eq()` calls when using actual values (instead of mixing values and matchers like `any()`)

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -178,10 +153,62 @@ public void testSendPingScheduled_optInRegisteredAndAlreadyOptedIn() {
...
     pingManager.sendPing("eventName", "metadataKey", "metadataValue");
     verify(pingEventQueue, verificationMode).add(any(PingEvent.class));
   }
+
+  @Test
+  public void testGetAnonymizedClientId_generateNewId() {
+    when(preferences.get(eq(AnalyticsPreferences.ANALYTICS_CLIENT_ID), anyString()))
+        .thenReturn(null);  // Simulate that client ID has never been generated.
+    String clientId = pingManager.getAnonymizedClientId();
+    assertTrue(!clientId.isEmpty());
```
akerekes: `assertFalse(clientId.isEmpty());` ?

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -178,10 +153,62 @@ public void testSendPingScheduled_optInRegisteredAndAlreadyOptedIn() {
...
+    verify(preferences, times(1)).put(eq(AnalyticsPreferences.ANALYTICS_CLIENT_ID), eq(clientId));
+  }
+
+  @Test
+  public void testGetAnonymizedClientId_useSavedId() {
+    when(preferences.get(eq(AnalyticsPreferences.ANALYTICS_CLIENT_ID), anyString()))
+        .thenReturn("some-unique-client-id");
+    String clientId = pingManager.getAnonymizedClientId();
+    assertEquals("some-unique-client-id", clientId);
+    verify(preferences, never()).put(eq(AnalyticsPreferences.ANALYTICS_CLIENT_ID), eq(clientId));
```
akerekes: you may be able to remove the `eq()` calls when using actual values (instead of mixing values and matchers like `any()`)

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -178,10 +153,62 @@ public void testSendPingScheduled_optInRegisteredAndAlreadyOptedIn() {
...
+  }
+
+  @Test
+  public void testSendPing_validArguments() {
+    pingManager.sendPing("eventName", "metadataKey", "metadataValue");
+    // Should not throw any exception.
```
akerekes: Is this comment necessary? If it needs to be mentioned, you could add it to the method name.
You also could verify if `pingEventQueue.add()` has been called

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
```diff
@@ -178,10 +153,62 @@ public void testSendPingScheduled_optInRegisteredAndAlreadyOptedIn() {
...
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testSendPing_emptyMetadataKeyWithNonNullValue() {
+    pingManager.sendPing("eventName", "", "metadataValue");
+  }
+
+  @Test
+  public void testSendPing_nullMetadataValue() {
+    pingManager.sendPing("eventName", "metadataKey", null);
+    // Should not throw any exception.
```
akerekes: Is this comment necessary? If it needs to be mentioned, you could add it to the method name.
You also could verify if `pingEventQueue.add()` has been called

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerWithServerTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class AnalyticsPingManagerWithServerTest {
+
+  @Mock private IEclipsePreferences preferences;
+  @Mock private Display display;
+  @Mock private ConcurrentLinkedQueue<PingEvent> pingEventQueue;
```
akerekes: I'm curious: how does this test work? the `pingEventQueue` is a mock and I don't see any behavior defined for it, so how will the ping sender job get the event?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -76,13 +78,13 @@
...
   // Preference store (should be configuration scoped) from which we get UUID, opt-in status, etc.
-  private IEclipsePreferences preferences;
-  private Display display;
-  private boolean analyticsEnabled;
+  private final IEclipsePreferences preferences;
+  private final Display display;
 
-  private ConcurrentLinkedQueue<PingEvent> pingEventQueue;
-  private Job eventFlushJob = new Job("Analytics Event Submission") {
+  private final ConcurrentLinkedQueue<PingEvent> pingEventQueue;
```
akerekes: Could be declared as `Queue`, but if we insist in making it a concurrent queue, we could use `ConcurrentArrayQueue` instead, whose javadoc says

>This class is a drop-in replacement for ConcurrentLinkedQueue, with similar performance but producing less garbage because arrays are used to store elements rather than nodes.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerWithServerTest.java
chanseokoh: Currently there is no test to see if the job is picking up events. Tests in this class are to see if events are being queued and/or the opt-in dialog is presented depending on the opt-in states. The mock queue also prevents actually making HTTP requests in this test class.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
chanseokoh: Looks good. I'll use `ConcurrentArrayQueue`.

chanseokoh: Wait, `ConcurrentArrayQueue` is not in JDK. It's from a Jetty library. I'll keep `ConcurrentLinkedQueue`.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerWithServerTest.java
akerekes: I thought the reason for adding the TestHttpServer is to test whether pings are being sent.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
akerekes: Indeed, sorry, I missed that.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerWithServerTest.java
chanseokoh: Ah, my bad. I was talking about the other test class.

In this class, I just call `sendPingHelper()` directly. In other words, it doesn't have an e2e test. But adding e2e tests seems a good idea. I'll try.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerTest.java
chanseokoh: I'll change the method name. 

> You also could verify if pingEventQueue.add() has been called

Enqeueing depends on two opt-in states, and this is actually tested thoroughly in other tests. I'll change the test name accordingly to reflect that this is only verifying arguments.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManagerWithServerTest.java
chanseokoh: It was easy to convert the tests to E2E tests. `sendPingHelper()` is no longer `@VisibleForTesting`.

PR #1535, opened 2017-03-07
Add HttpUtil test

chanseokoh:
Part of #1437.


plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/HttpUtilWithServerTest.java
```diff
@@ -33,19 +33,36 @@
   private static final String LONG_PARAMETER = "Exception in thread \"main\""
       + " java.lang.IllegalArgumentException: Wrong argument!\n\tat B.main(B.java:73)";
 
+  private static final Map<String, String> testParameters = new HashMap<String, String>() {
```
elharo: HashMap<>()

chanseokoh: I realized I can't use the diamond syntax when I'm using the [double-brace initialization](http://stackoverflow.com/questions/1372113/meaning-of-new-class-initialization-idiom/1372124#1372124). I could have used Guava, but this bundle is free of Guava, so I decided not to do so.

Now I think I'll change this to plain static initialization, because the double-brace initialization gives me the serial ID warning.

akerekes:
LGTM once the builds are successful

PR #1537, opened 2017-03-07
require version on maven coordinates

### LINKED ISSUE #1522: Inconsistent library versions when adding libraries through Build Path > Add Libraries... on Maven project
I am not certain if this is intended or if we have opened an issue for this.

When I added the App Engine API, Objectify, and Endpoints through _Build Path_ > _Add Libraries..._, Objectify uses a fixed version, but the others use `LATEST`.

```xml
<dependency>
<groupId>com.google.appengine</groupId>
<artifactId>appengine-endpoints</artifactId>
<version>LATEST</version>
</dependency>
<dependency>
<groupId>com.google.appengine</groupId>
<artifactId>appengine-api-1.0-sdk</artifactId>
<version>LATEST</version>
</dependency>
<dependency>
<groupId>com.googlecode.objectify</groupId>
<artifactId>objectify</artifactId>
<version>5.1.14</version>
</dependency>
```

### LINKED ISSUE #1540: LATEST not supported in maven 3
version="LATEST" is not supported in Maven 3. Can we require a version everywhere? Or allow the version element to be omitted?

elharo:
fix #1522 fix #1540 @akerekes @chanseokoh 

This works for now. Going forward, I wonder if we want to pull the latest version from the Maven repo and use that instead? This is not the same as pulling the latest jar. We would need to retrieve the version number of the artifact.


plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -251,23 +251,21 @@ build path <a href="http://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.
          according to the schema, so these are removed here -->
          <xs:attribute name="repositoryUri" type="xs:string">
             <xs:annotation>
-               <xs:documentation>
-                  URI pointing to a repository location or an ID of a repository defined in the POM or the Maven
-settings. If the value starts with [protocol]:// then it is treated as a URI, otherwise an ID of a repository that is
-expected to be defined. Defaults to 'central' denoting the Maven Central repository.
+			  <xs:documentation>
```
chanseokoh: Indentation became off.

elharo:
It's not at all clear that LATEST actually works. Maven documentation suggests otherwise as of Maven 3.x.

If we want a LATEST functionality I think we need to figure out how to query Maven central for the latest available number, then send that back to them.

briandealwis:
The Maven resolvers support using _LATEST_ and _RELEASE_ for resolving artifacts (cf _org.apache.maven.repository.internal.DefaultVersionResolver_), but Maven does not support their use in `pom.xml`.

briandealwis:
> If we want a LATEST functionality I think we need to figure out how to query Maven central for the latest available number, then send that back to them.

That's what we currently do. But because of m2e's defaults, we hit issue #830.  So I think explicitly specifying a version, as you're doing here, is the right thing to do.  @chanseokoh's point is that we still allow LATEST in our code, and we probably shouldn't.

chanseokoh:
> @chanseokoh's point is that we still allow LATEST in our code, and we probably shouldn't.

I don't have a strong opinion here. As I said originally (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1522#issue-212196808), I was basically asking a question if the mix of LATEST and a fixed version in the resulting `pom.xml` is intended.

akerekes:
We probably can query the latest version from Maven Central, according to https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/830#issuecomment-275237055

Using `LATEST` frees us from tracking the latest versions of these libraries in our code. If this would cause a problem, we could allow the users to select a version explicitly in the library selector.

elharo:
OK, sounds like we can keep querying for LATEST. However we do need to not put that string into pom.xml, which is what this PR fixes.

elharo:
Ping

akerekes:
This PR removes the `LATEST` from plugin.xml, besides 
>we do need to not put that string into pom.xml, which is what this PR fixes.

if I understand it correctly.

elharo:
Yes, we want specific version numbers in plugin.xml and everywhere we use maven coordinates.

patflynn:
+1 to how you ended up handling this. I like the approach of using definite
version numbers in user code, but trying to stay up to date in terms of
what we generate.  Random question, can you specify latest within a major
version?

akerekes:
@patflynn that's an interesting idea. That would require us to query the versions from the remote repository and figure out the latest for the given major version. I think it's doable. Also we could give users the choice to select a version/version alias (i.e. latest from a given major, latest from a given major.minor) and keep the attached library up to date according to the alias.

PR #1542, opened 2017-03-07
Order our New menu to distinguish App Engine Standard from Flexible

### LINKED ISSUE #1536: Order of new project menu
Do we want standard before flex or vice versa? Make a deliberate choice here. Let's make sure it's not random or alphabetical or some such. 

image

briandealwis:
Adds placeholders to our GCP Toolbar's _New_ menu to separate _Standard_ options from _Flexible_.

screen shot 2017-03-07 at 1 37 35 pm

Fixes #1536


akerekes:
I'd use the separator until they're get combined.

briandealwis:
I'm with @akerekes — I found I was having to hunt more without the separator.

PR #1546, opened 2017-03-07
Fixes #1415 - Project ID not selected when real issue is no GCP access

### LINKED ISSUE #1415: Project ID not selected when real issue is no GCP access
The issue in this case is that no project IDs are available; in this case because the logged in account does not have GCP access. 

image

Is there anyway we can improve the warning message in this case?

akerekes:
Wrap the logic to check whether the project list is empty into a `Supplier` to avoid adding an import package for `ProjectSelector` to `c.g.c.t.e.ui.util` bundle.

Fixes #1415


chanseokoh:
Unfortunately, the message is shown even if there are projects.

deploy to app engine standard _038

akerekes:
The reason for the error: after login the account was selected and the databinding updated the validator for account selector that did not give error anymore. Therefore the validation status provider of the dialog (`MaxSeverityValidationStatusProvider`) checked it's other validation status providers and picked the first error which was from the project selection validator from the time before the login happened (i.e. the login+account selection did not trigger a revalidation of the other fields including the project selector, thus the validator still contained the error from the first validation). By triggering the validation on account selection, the error is now correctly about no project selected.

However I'm not 100% suer if this is the correct way to solve this problem. Maybe adding a binding between the account selector and the project selector that does nothing but trigger validation when the account selector changes should be the solution? @briandealwis What do you think?

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectSelector.java
```diff
@@ -127,4 +127,11 @@ private void setTooltip(String tooltip) {
   public void clearStatusLink() {
     setStatusLink("", "");
   }
+
+  /**
+   * @return true if there are projects in the selector, false otherwise
+   */
+  public Boolean hasProjects() {
```
elharo: boolean per Effective Java, Item 49

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/ProjectSelectorValidator.java
```diff
@@ -17,14 +17,28 @@
...
 import org.eclipse.core.databinding.validation.IValidator;
 import org.eclipse.core.databinding.validation.ValidationStatus;
 import org.eclipse.core.runtime.IStatus;
 
 public class ProjectSelectorValidator implements IValidator {
 
+  private Supplier<Boolean> projectsFound;
+
+  /**
+   * @param projectsFound should return true if there are projects in the project selector,
```
elharo: should return --> returns

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -464,6 +463,38 @@ public Object convert(Object fromObject) {
...
+    }
+
+    @Override
+    protected IStatus validate() {
+      // this access is recorded and ensures that changes are tracked
+      Collection<?> projects = (Collection<?>) projectInput.getValue();
+      // this access is recorded and ensures that changes are tracked, don't move it inside the if
+      Object selectedProject = projectSelection.getValue();
+      if (projects.isEmpty()) {
+        return ValidationStatus.info(Messages.getString("projectselector.no.projects"));
```
chanseokoh: I see one problem when returning INFO. The _Deploy_ button is enabled.

deploy to app engine standard _039

However, if we change to return ERROR, it should go into `if (requirevalues)`, so that users can save settings in the preference page.

chanseokoh:
The "No projects found" message does not appear in the preference page _initially_, if an account having no project is already selected and saved. Seems like validation is not triggered initially. I feel like I ran into a similar issue before.

akerekes:
The reason is pretty complex. The property page does not allow to display messages until it evaluates whether the project has correct facet configuration. i.e.:
 - `canSetMessage` defaults to false
 - `setMessage() is overriden to prevent setting a message when `canSetMessage` is true
 - the value of `canSetMessage` can change when `setVisible() is called
 - the databinding&validation logic runs before `setVisible()` is called

chanseokoh:
This may not help, but does it have any similarity to https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1303#issuecomment-276145573? I vaguely remember `uiChanged` was null in my case because `setVisible()` was false too (although not sure).

akerekes:
No, this is different issue. Here the validation happens, the message would be set, but the code we wrote prevents it.

chanseokoh:
> No, this is different issue. Here the validation happens, the message would be set, but the code we wrote prevents it.

Okay, then the only difference is that it is our code here, but it wasn't our code in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1303#issuecomment-276145573. I guess if it's our code, there may be an easy fix?

akerekes:
Yes, either that or a difficult fix :)

chanseokoh:
The integration test failures seem to be related to that the "No projects found" message is displayed even though no account is selected, which covers the preference page heading:

selection_002

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanelTest.java
```diff
@@ -204,7 +187,65 @@ public void testProjectSavedInPreferencesSelected() throws ProjectRepositoryExce
...
 
-  private void initializeProjectRepository(ProjectRepository projectRepository)
+  @Test
+  public void testProjectNotSelectedIsAnErrorWhenRequireValuesIsTrue() {
+    deployPanel = createPanel(true /* requireValues */);
+    assertThat(getProjectSelectionValidator().getSeverity(), is(IStatus.ERROR));
+  }
+
+  @Test
+  public void testProjectNotSelectedIsAnErrorWhenRequireValuesIsFalse() {
```
chanseokoh: `IsAnError` --> `IsNotAnError`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -202,7 +202,7 @@ private void evaluateFacetConfiguration() {
   private void updatePageBasedOnFacetConfig(boolean hasCorrectFacetConfiguration) {
     if (hasCorrectFacetConfiguration) {
       canSetMessage = true;
-      setErrorMessage(null);
+      content.getDataBindingContext().updateTargets();
```
chanseokoh: I think it's safe to call `setErrorMessage(null)` first as originally. I remember @nbashirbello was trying to fix #783 with #880. During the course, a subtle bug surfaced (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880#issuecomment-256149028), and @nbashirbello had to put quite some effort to fix that bug. (For example, `canSetMessage` and its related logic were introduced after the discovery of that subtle bug.) Her intention with `setErrorMessage(null)` was probably to clear the error message set below when switching between the standard preference page and the flex page back and forth. We need to be careful not to make a regression on #783.

akerekes: Thanks @chanseokoh I was trying to figure out the reason for the logic, but I missed this. I'm now trying to figure out what was fixed, because I still can see the error mentioned in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880#issuecomment-256149028 :
 - checkout `master` and launch the plugin
 - open properties of a standard project
 - open the deploy preferences
 - remove standard and add flex facet
 - go back to the deploy preferences page
 - type some illegal version name (e.g.: `#`)

The `Apply` and `OK` buttons disable, no error message is shown regarding the version, only that this page is valid for standard projects only. When I try to switch to another page, I get the message that there are invalid values on the page I cannot switch.

chanseokoh: I am unsure too. If this is not a regression, I think we don't need to address it in this PR.

chanseokoh: BTW, I remember https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880#issuecomment-256149028 was happening even though if I did not switch facets, so it was critical to fix it. Now it happens only after switching facets, so it is a slightly different bug.

akerekes: Further update on this: the bug referenced in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880#issuecomment-256149028 is a gtk3 bug. If we run the plugin at the commit just prior to the comment and open the deploy preferences page, then click the `override version` checkbox (to set it), we expect to see an error message at the top of the page, that the version number is invalid (because it is an empty string currently). Based on the GTK version we see:
 - GTK2 will display the error message
 - GTK3 will not display the error message

So apparently PR #880 worked around the GTK3 bug without realizing that it was the underlying issue. I'm investigating which commit was the accidental workaround.

akerekes: For the records: to reproduce https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880#issuecomment-256149028:

1. checkout `ca7bc9d90e0329ee204850e5e627e2ac698f8ed6`
1. launch the plugin
1. create a new GAE Standard project
1. Right click on the project, select Properties > Google Cloud Platform > App Engine Deployment
1. Click on the `Override generated version` checkbox (to mark it set)
1. Depending on the GTK version you use:
    1. On GTK2 the error message is visible instead of the title of the page, the page is invalid
    1. On GTK3 the title of the page remains visible, but the page is invalid because the Apply/OK buttons are disabled and switching to another preference page shows an error, that there are invalid values on the page

akerekes: I believe I found the culprit. This change: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880/commits/00e9aec21df0f6a7b373ea1f082cae50b7311008#diff-7829288eff77ee655857fff4a023693fL158

What happens here is that the error message is set to null when the paint listener is executed. However the paint listener is never called on GTK2. This is the code that decides based on the GTK version:
[Widget.java lines 1940-1951](https://github.com/eclipse/eclipse.platform.swt/blob/8be1ba1520291fd294aff6542831ec2b8d9fb0c6/bundles/org.eclipse.swt/Eclipse%20SWT/gtk/org/eclipse/swt/widgets/Widget.java#L1936)
```
if (OS.GTK3) {
  if (OS.GTK_VERSION >= OS.VERSION (3, 9, 0) && OS.GTK_IS_CONTAINER (handle)) {
    return gtk_draw (handle, arg0);
  }
} else {
  GdkEventExpose gdkEvent = new GdkEventExpose ();
  OS.memmove (gdkEvent, arg0, GdkEventExpose.sizeof);
  long /*int*/ paintWindow = paintWindow ();
  long /*int*/ window = gdkEvent.window;
  if (window != paintWindow) return 0;
    return (state & OBSCURED) != 0 ? 1 : 0;
}
```

So when the referenced commit removed the `setErrorMessage(null)` call GTK3 did not remove the error message anymore in the paint listener.

akerekes: To summarize: the issue in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/880#issuecomment-256149028 was caused by the difference between GTK2 and GTK3, and the eventual removal of `setErrorMessage(null)` and the moving the logic from the paint listener to the `setVisible()` call (to be precise the latter is sufficient) fixed it. (Along with the problem that on GTK2 the error message regarding the invalid facet configuration for the deploy page did not appear (for the same reason: it was set in the paint listener which was not executed on GTK2)).

So I'm pretty confident that the error in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1546#discussion_r105149487 is not a regression, and was present before this PR.

akerekes: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1564 filed to record the bug.

akerekes:
@elharo @chanseokoh @briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -156,7 +155,7 @@ private void setupDataBinding() {
     bindingContext = new DataBindingContext();
```
briandealwis: Do we need to pass in the DBC when it's accessible from the `bindingContext` field?

Issue #1455, opened 2017-02-22
Associate GCP Projects with perspective


elharo:
Here's what happens when you create a new Java Project with a wizard and you're not in the Java perspective:

![image](https://cloud.githubusercontent.com/assets/1005544/23227704/ce3ae7a8-f907-11e6-92f9-e42d8963b41d.png)

Should we associate App Engine projects with a particular perspective, and if so which one?

elharo:
PlatformUI.getWorkbench().showPerspective("perspective.id", PlatformUI.getWorkbench().getActiveWorkbenchWindow());

akerekes:
JavaEE seems to me the one we should associate it with, until we come up with our perspective with cool support of GCP features.

briandealwis:
This is set on the `org.eclipse.ui.newWizards / finalPerspective` or `preferredPerspectives`.

PR #1552, opened 2017-03-08
Make ProjectUtils.waitForProjects() more robust

briandealwis:
`ProjectUtils#waitForProjects()` has a few issues:
  - I encountered a live lock problem with #1455 where `ProjectUtils#waitForProjects()` would spin trying to join on the workbench's Auto-Build job.
  - Separately @chanseokoh pointed me at a [hung Travis build ](https://travis-ci.org/GoogleCloudPlatform/google-cloud-eclipse/jobs/208831940).
  - It would spin until all build errors were resolved, which results in tests eventually timing out if those errors are never resolved (e.g., because of network resolution errors outside of our control)

This change ensures:
  - we only wait on runnable `Job`s
  - we track whether the build errors have reached a fixpoint
  - we output debug information after 10s have elapsed (could tune this further)

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
@@ -28,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Enumeration;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
@@ -59,7 +60,7 @@ import org.osgi.framework.FrameworkUtil;
  * A set of utility methods for dealing with projects.
  */
 public class ProjectUtils {
-  private static boolean DEBUG = true;
+  private static boolean DEBUG = false;
 
   /**
    * Import the Eclipse projects found within the bundle containing {@code clazz} at the
@@ -137,13 +138,13 @@ public class ProjectUtils {
 
   /** Fail if there are any build errors on the specified projects. */
   public static void failIfBuildErrors(String message, IProject... projects) throws CoreException {
-    List<String> errors = getAllBuildErrors(projects);
+    Set<String> errors = getAllBuildErrors(projects);
     assertTrue(message + "\n" + Joiner.on("\n").join(errors), errors.isEmpty());
   }
 
   /** Return a list of all build errors on the specified projects. */
-  public static List<String> getAllBuildErrors(IProject... projects) throws CoreException {
-    List<String> errors = new ArrayList<>();
+  public static Set<String> getAllBuildErrors(IProject... projects) throws CoreException {
+    Set<String> errors = new LinkedHashSet<>();
     for (IProject project : projects) {
       IMarker[] problems = project.findMarkers(IMarker.PROBLEM, true /* includeSubtypes */,
           IResource.DEPTH_INFINITE);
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -202,25 +208,50 @@ public static void waitForProjects(Runnable delayTactic, IProject... projects) {
...
+          Job job = iter.next();
+          if (job.getState() != Job.RUNNING && job.getState() != Job.WAITING) {
+            iter.remove();
+          }
+        }
+
+        // track whether we've reached a fixpoint in the build errors
         allBuildErrors = getAllBuildErrors(projects);
-        if (DEBUG) {
+        buildErrorsChanging =
```
akerekes: Would it make sense to try to check whether the build errors are the same as in the previous round instead of just the change in the count?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -184,9 +187,12 @@ public void run() {
   /** Wait for any spawned jobs and builds to complete (e.g., validation jobs). */
   public static void waitForProjects(Runnable delayTactic, IProject... projects) {
```
akerekes: This method is pretty long, I suggest to break out parts into helper methods:
 - collecting jobs (including filtering on line 214-219
 - logic when more than 60 seconds has passed

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -202,25 +208,50 @@ public static void waitForProjects(Runnable delayTactic, IProject... projects) {
         Collections.addAll(jobs, jobManager.find("org.eclipse.wst.server.core.family"));
         Collections.addAll(jobs, jobManager.find("org.eclipse.wst.server.ui.family"));
         Collections.addAll(jobs, jobManager.find(ValidationBuilder.FAMILY_VALIDATION_JOB));
+
+        // Remove SLEEPING jobs as they may not run for a long time.
```
chanseokoh: Tests are failing reliably, and I wonder if it has something to do with this. The Javadoc of `Job.schedule()` says

```
	 * Schedules this job to be run after a specified delay.  The job is put in the
	 * {@link #SLEEPING} state until the specified delay has elapsed, after which
	 * the job is added to a queue of {@link #WAITING} jobs. 
```
So, if an important job is scheduled with a delay, there is a chance we don't catch it. Not sure though.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -202,25 +208,50 @@ public static void waitForProjects(Runnable delayTactic, IProject... projects) {
...
                 allBuildErrors.size());
-          } else {
+            // don't report unchanged errors
+            if (buildErrorsChanging) {
+              for (String buildError : allBuildErrors) {
+                System.err.printf("  error: " + buildError);
+              }
+            }
+          }
+          if (timer.elapsed(TimeUnit.SECONDS) > 60) {
```
chanseokoh: So, if we go over 60 seconds, I guess we will print the state very frequently, every time we loop? A small interval might be a good idea.

briandealwis:
There's something special about the _autobuild_ jobs.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
briandealwis: It would.  I should just put them into a `Set` and do a comparison with `equals()`.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
briandealwis: Good point.

briandealwis:
PTAL @akerekes @chanseokoh

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -183,49 +186,67 @@ public void run() {
...
+
+        // identify any pending build-related jobs
+        jobs = findPendingBuildJobs();
+
+        // track whether we've reached a fixpoint in the build errors
+        Set<String> currentBuildErrors = getAllBuildErrors(projects);
+        buildErrorsChanging = !previousBuildErrors.equals(currentBuildErrors);
+        previousBuildErrors = currentBuildErrors;
+
+        if (DEBUG || timer.elapsed(TimeUnit.SECONDS) > 10) {
```
chanseokoh: I hope this doesn't print out too many messages. Once we go over 10 seconds, we log every time we loop.

briandealwis: But we sleep, and we only log for the conditions that keep us in the loop which is important info.

PR #1560, opened 2017-03-09
Make sure "Run As" > "App Engine" shows only on App Engine projects.

### LINKED ISSUE #1556: Run As > App Engine menu item appears without App Engine facet
This was on a Maven Dynamic WTP project (bookshelf).

Probably we shouldn't show that context menu item unless the App Engine Standard factet is present; but if we do show it, it should work. 

image

chanseokoh:
Fixes #1556, per https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1556#issuecomment-285190341.


chanseokoh:
Asking @briandealwis to review, since he initially set up the enablement conditions.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -287,24 +287,15 @@
...
-                        <adapt type="org.eclipse.core.resources.IProject">
-                           <test
-                                 property="org.eclipse.wst.common.project.facet.core.projectFacet"
-                                 value="com.google.cloud.tools.eclipse.appengine.facets.standard" />
-                           </adapt>
-                     </iterate>
-                  </with>
-                  <with variable="activeEditorInput">
-                     <adapt type="org.eclipse.core.resources.IFile">
+               <with variable="selection">
```
briandealwis: We can drop this outer `<with variable="selection">` and just use the `<iterate>…</iterate>` on value that's passed in (that is likely — but not necessarily — the selection).

chanseokoh: Cool, it does work without `<with>`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/RunAppEngineShortcutTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    IWorkbench workbench = PlatformUI.getWorkbench();
+    assertNotNull(WorkbenchUtil.openInEditor(workbench, file));
+
+    IProject javaProject = javaProjectCreator.getProject();
+    assertFalse(appEngineMenuExists(javaProject));
+  }
+
+  // We need regex matching, since the actual menu name is "<number> App Engine".
+  private boolean appEngineMenuExists(IProject project) {
```
elharo: static

PR #1565, opened 2017-03-09
update version to 1.0.1

elharo:
@chanseokoh @briandealwis 

Do we need to update this anywhere else?

### DIFF
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -10,11 +10,11 @@
 
 requires.1.namespace=org.eclipse.equinox.p2.iu
 requires.1.name=com.google.cloud.tools.eclipse.category
-requires.1.range=[1.0.0,1.0.0]
+requires.1.range=[1.0.0,1.0.1]
 requires.1.greedy=true
 
 units.1.id=com.google.cloud.tools.eclipse.category
-units.1.version=1.0.0
+units.1.version=1.0.1
 units.1.properties.1.name=org.eclipse.equinox.p2.type.category
 units.1.properties.1.value=true
 units.1.properties.2.name=org.eclipse.equinox.p2.name
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -27,7 +27,7 @@ units.1.properties.5.name=org.eclipse.equinox.p2.doc.url
 units.1.properties.5.value=https://cloud.google.com/eclipse/docs/
 units.1.provides.1.namespace=org.eclipse.equinox.p2.iu
 units.1.provides.1.name=com.google.cloud.tools.eclipse.category
-units.1.provides.1.version=1.0.0
+units.1.provides.1.version=1.0.1
 units.1.copyright =Copyright 2016, 2017 Google Inc.
 units.1.licenses.0 =\
 Cloud Tools for Eclipse is made available under the Apache License, Version 2.0. \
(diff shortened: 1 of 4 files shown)


chanseokoh:
I see a product version "1.0.0" in `gcp-repo/product`, but I'm not sure what that means or how it relates to a feature version.

I also see a category version "1.0.0" in `gcp-repo/metadata.p2.inf`, but I get the feeling that a category is a different kind that may keep its own version. Not sure though.

briandealwis:
The category in `gcp-repo/metadata.p2.inf` may be worth changing since it's reflected in the installation UI:

screen shot 2017-03-09 at 4 35 55 pm

The `gcp-repo/metadata.product` can be left untouched: it won't be visible except to those who know how to use p2 tools.

chanseokoh:
> The gcp-repo/metadata.product can be left untouched: it won't be visible except to those who know how to use p2 tools.

Could be useful to put a comment inside `gcp-repo/metadata.product`.

elharo:
PTAL. Updated both.

gcp-repo/metadata.p2.inf
```diff
@@ -27,7 +27,7 @@ units.1.properties.5.name=org.eclipse.equinox.p2.doc.url
 units.1.properties.5.value=https://cloud.google.com/eclipse/docs/
 units.1.provides.1.namespace=org.eclipse.equinox.p2.iu
 units.1.provides.1.name=com.google.cloud.tools.eclipse.category
-units.1.provides.1.version=1.0.0
+units.1.provides.1.version=1.0.1
```
briandealwis: I think you need to change the `requires.1.range` up above.  This .p2.inf format hurts to read.  Basically this file associates p2 metadata to the `com.google.cloud.tools.eclipse.dist` product defined in the `gcp-repo/metadata.product` file.  
  - the initial `requires.1` adds a requirement on the product IU for an IU with certain capabilities; think of the _namespace_ as being a the capability _type_, and the _name_ is the capability identifier.  So this `requires.1` says the product requires our new category
  - the `units.N` defines new IUs; the `units.N.provides.M` describes the capabilities of this IU. In this case this IU provides a category definition.  The category values are in the properties.
  - the category IU also has some requirements to pull in a feature

chanseokoh: @briandealwis there is also `units.1.version=1.0.0` above. That is left untouched?

briandealwis: I don't know! I think we have to try it and maybe change the unimportant versions to 999.999.999 to indicate that they don't matter.

elharo:
PTAL

gcp-repo/metadata.p2.inf
briandealwis: The `units.1.version` determines what's shown the user.
```diff
diff --git a/gcp-repo/metadata.p2.inf b/gcp-repo/metadata.p2.inf
index 010b3c0..c234b59 100644
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -14,7 +14,7 @@ requires.1.range=[1.0.0,1.0.1]
 requires.1.greedy=true
 
 units.1.id=com.google.cloud.tools.eclipse.category
-units.1.version=1.0.1
+units.1.version=999
 units.1.properties.1.name=org.eclipse.equinox.p2.type.category
 units.1.properties.1.value=true
 units.1.properties.2.name=org.eclipse.equinox.p2.name
```
produces:
screen shot 2017-03-09 at 11 12 48 pm

PR #1571, opened 2017-03-10
Error for undefined servlet class

### LINKED ISSUE #1478: error when web.xml references undefined servlet
I.e. the class in servlet-class is not in the project:

    <servlet-class>com.google.example.HelloAppEngine</servlet-class>

abbiefarr:
Fix #1478 
Verifies that web.xml has a servlet class that actually exists in the project.

@elharo @meltsufin


plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -54,9 +56,12 @@
...
     createFolders(project, new Path("src/main/webapp/WEB-INF"));
-    IFile webXml = project.getFile("src/main/webapp/WEB-INF/web.xml");
-    webXml.create(new ByteArrayInputStream(new byte[0]), true, null);
-    resource = webXml;
+    resource = project.getFile("src/main/webapp/WEB-INF/web.xml");
+    resource.create(new ByteArrayInputStream(new byte[0]), true, null);
+    createFolders(project, new Path("src/main/java"));
+    IFile file = project.getFile("src/main/java/ServletClass.java");
+    file.create(new ByteArrayInputStream(
+        "public class ServletClass {}".getBytes(StandardCharsets.UTF_8)), true, null);
```
elharo: We could have a separate check that the referenced class is actually a Servlet, though I'm not sure how often that comes up in practice.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -70,18 +75,54 @@ public void tearDown() throws CoreException {
...
+    validator.validate(resource, bytes);
+    IMarker[] markers = resource.findMarkers(UNDEFINED_SERVLET_MARKER, true, IResource.DEPTH_ZERO);
+    assertEquals(0, markers.length);
+  }
+  
+  @Test
+  public void testValidate_nonexistingServletClass()
+      throws IOException, CoreException, ParserConfigurationException {
+    String xml = "<web-app xmlns='http://java.sun.com/xml/ns/javaee' version='2.5'>"
+        + "<servlet-class>DNE</servlet-class>"
```
elharo: DNE --> DoesNotExist

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -70,18 +75,54 @@ public void tearDown() throws CoreException {
...
   }
   
+  @Test
+  public void testValidate_withExistingServletClass()
+      throws IOException, CoreException, ParserConfigurationException {
+    String xml = "<web-app xmlns='http://java.sun.com/xml/ns/javaee' version='2.5'>"
+        + "<servlet-class>ServletClass</servlet-class>"
+        + "</web-app>";
+    byte[] bytes = xml.getBytes(StandardCharsets.UTF_8);
+    WebXmlValidator validator = new WebXmlValidator();
```
elharo: You could make this a field shared across test methods.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/UndefinedServletElement.java
```diff
@@ -0,0 +1,40 @@
...
+ * A servlet class element that will receive an undefined servlet marker. 
+ */
+public class UndefinedServletElement extends BannedElement {
+
+  private static final String markerId = 
+      "com.google.cloud.tools.eclipse.appengine.validation.undefinedServletMarker";
+  private static final int severity = IMarker.SEVERITY_ERROR;
+  private String servletClass;
+  
+  public UndefinedServletElement(String servletClass, DocumentLocation start, int length) {
```
elharo: servletClass --> servletClassName since servletClass sounds like this will be a java.lang.Class rather than a String

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +44,30 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+      servletClassContents.append(ch, start, length);
+    }
+  }
+  
+  @Override
+  public void endElement(String uri, String localName, String qName) throws SAXException {
+    if ("servlet-class".equalsIgnoreCase(localName)) {
+      insideServletClass = false;
+      String servletClass = servletClassContents.toString();
+      addToBlacklist(new UndefinedServletElement(
```
elharo: don't we want to add it to the blacklist only if it's not defined?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidator.java
```diff
@@ -42,11 +62,55 @@ protected void validate(IResource resource, byte[] bytes)
...
+    List<SearchMatch> results = requestor.getResults();
+    return results.isEmpty() ? null : (IType) results.get(0).getElement();
+  }
+  
+  /**
+   * Performs the search for a class of a given pattern within the project.
+   */
+  static void performSearch(SearchPattern pattern, IJavaSearchScope scope, SearchRequestor requestor,
+      IProgressMonitor monitor) {
+    SearchEngine searchEngine = new SearchEngine();
```
elharo: Do we need/want to create a new one of these each time? If it's not expensive, fine; but if this is an expensive and thread-safe object then maybe create once in a field.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/messages.properties
```diff
@@ -2,4 +2,5 @@ application.element=Project ID should be specified at deploy time
 remove.application.element=Remove application element
 web.xml.version=App Engine Standard does not support this servlet version
 convert.servlet=Convert to Java Servlet API 2.5
-maven.plugin=This App Engine Maven plugin has been deprecated
\ No newline at end of file
+maven.plugin=This App Engine Maven plugin has been deprecated
+undefined.servlet.class=This class does not exist in the project
```
elharo: We should include the class name here.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -64,9 +80,44 @@ public void characters(char ch[], int start, int length) throws SAXException {
...
+        SearchPattern.R_EXACT_MATCH | SearchPattern.R_ERASURE_MATCH);
+    IJavaSearchScope scope = SearchEngine.createWorkspaceScope();
+    CollectingSearchRequestor requestor = new CollectingSearchRequestor();
+    performSearch(pattern, scope, requestor, null);
+    List<SearchMatch> results = requestor.getResults();
+    return results.isEmpty() ? null : (IType) results.get(0).getElement();
+  }
+  
+  /**
+   * Performs the search for a class of a given pattern within the project.
```
elharo: performs the search for a class of a given pattern --> Searches for a class that matches a pattern

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -102,19 +97,14 @@ public void testValidate_withExistingServletClass()
...
     validator.validate(resource, bytes);
     IMarker[] markers = resource.findMarkers(UNDEFINED_SERVLET_MARKER, true, IResource.DEPTH_ZERO);
     assertEquals(1, markers.length);
-  }
-  
-  @Test
-  public void testFindClass() {
-    IType type = WebXmlValidator.findClass("ServletClass", project);
-    assertNotNull(type);
+    String expectedMessage = "DoesNotExist " + Messages.getString("undefined.servlet.class");
```
elharo: in the test just use the literal string; no need to load the message

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/UndefinedServletElement.java
```diff
@@ -26,15 +26,16 @@
...
       "com.google.cloud.tools.eclipse.appengine.validation.undefinedServletMarker";
   private static final int severity = IMarker.SEVERITY_ERROR;
-  private String servletClass;
+  private String servletClassName;
   
-  public UndefinedServletElement(String servletClass, DocumentLocation start, int length) {
-    super(Messages.getString("undefined.servlet.class"), markerId, severity, start, length);
-    this.servletClass = servletClass;
+  public UndefinedServletElement(String servletClassName, DocumentLocation start, int length) {
+    super(servletClassName + " " + Messages.getString("undefined.servlet.class"),
```
elharo: For I18N purposes this should be done with a message format rather than string concatenation  in case the class name is not the first word in some non-English language. E.g.

Messages.getString("%s undefined.servlet.class", servletClassName)

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/messages.properties
```diff
@@ -3,4 +3,4 @@ remove.application.element=Remove application element
 web.xml.version=App Engine Standard does not support this servlet version
 convert.servlet=Convert to Java Servlet API 2.5
 maven.plugin=This App Engine Maven plugin has been deprecated
-undefined.servlet.class=This class does not exist in the project
\ No newline at end of file
+undefined.servlet.class=does not exist in the project
```
elharo: {0} does not exist in the project

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+      servletClassContents.append(ch, start, length);
+    }
+  }
+  
+  @Override
+  public void endElement(String uri, String localName, String qName) throws SAXException {
+    if ("servlet-class".equalsIgnoreCase(localName)) {
+      insideServletClass = false;
+      String servletClassName = servletClassContents.toString();
+      if (findClass(servletClassName) == null) {
```
akerekes: If we don't need the `IType` object returned from `findClass()` it could just return a `boolean` and possibly be renamed to `hasClass` or `classExists` or `isClassPresent`.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+      return null;
+    }
+    SearchPattern pattern = SearchPattern.createPattern(typeName, 
+        IJavaSearchConstants.CLASS,
+        IJavaSearchConstants.DECLARATIONS,
+        SearchPattern.R_EXACT_MATCH | SearchPattern.R_ERASURE_MATCH);
+    IJavaSearchScope scope = SearchEngine.createWorkspaceScope();
+    CollectingSearchRequestor requestor = new CollectingSearchRequestor();
+    performSearch(pattern, scope, requestor, null);
+    List<SearchMatch> results = requestor.getResults();
```
akerekes: This and the following line could be moved into `performSearch` as we don't really do anything with them after `performSearch` returns. And to align with my previous comment it could just return a boolean (i.e. `!getResult().isEmpty()`)

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+    return results.isEmpty() ? null : (IType) results.get(0).getElement();
+  }
+  
+  /**
+   * Performs the search for a class of a given pattern within the project.
+   */
+  static void performSearch(SearchPattern pattern, IJavaSearchScope scope, SearchRequestor requestor,
+      IProgressMonitor monitor) {
+    SearchEngine searchEngine = new SearchEngine();
```
akerekes: If we don't need `searchEngine` outside of the `try`'s body, this line could be moved there. Also this variable can be inlined.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+   */
+  static IType findClass(String typeName) {
+    if (typeName == null || "".equals(typeName)) {
+      return null;
+    }
+    SearchPattern pattern = SearchPattern.createPattern(typeName, 
+        IJavaSearchConstants.CLASS,
+        IJavaSearchConstants.DECLARATIONS,
+        SearchPattern.R_EXACT_MATCH | SearchPattern.R_ERASURE_MATCH);
+    IJavaSearchScope scope = SearchEngine.createWorkspaceScope();
```
akerekes: I'm wondering if workspace scope is the best here. For sure it has the best chance to contain the class we're looking for, but it could be in a project that is not a dependency of the one containing the `web.xml`. In this case we may not give a warning even though the configuration of the web application is incorrect.

After looking at the `SearchEngine` class the method `org.eclipse.jdt.core.search.SearchEngine.createJavaSearchScope(IJavaElement[])` might be the one we need.

This may require to make the validator aware of which project it is processing.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+            servletClassName, servletClassLocation, localName.length() + 2));
+      }
+    }
+  }
+  
+  /**
+   * Checks for the given class name within the project.
+   */
+  static IType findClass(String typeName) {
+    if (typeName == null || "".equals(typeName)) {
```
akerekes: Could use Guava's `String.isNullOrEmpty()` if Guava is already a dependency of this bundle.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+  }
+  
+  @Override
+  public void endElement(String uri, String localName, String qName) throws SAXException {
+    if ("servlet-class".equalsIgnoreCase(localName)) {
+      insideServletClass = false;
+      String servletClassName = servletClassContents.toString();
+      if (findClass(servletClassName) == null) {
+        addToBlacklist(new UndefinedServletElement(
+            servletClassName, servletClassLocation, localName.length() + 2));
```
akerekes: Can you add a comment explaining the `+2`? Or if it improves readability just create a constant with descriptive name?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -25,6 +40,11 @@
  */
 class WebXmlScanner extends AbstractScanner {
 
+  private static final Logger logger = Logger.getLogger(WebXmlScanner.class.getName());
+  private boolean insideServletClass;
```
akerekes: `insideServletClassElement`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -25,6 +40,11 @@
...
+  private StringBuilder servletClassContents;
```
akerekes: `servletClassElementContents`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -25,6 +40,11 @@
...
+  private DocumentLocation servletClassLocation;
```
akerekes: `servletClassElementLocation`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/UndefinedServletElement.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * A servlet class element that will receive an undefined servlet marker. 
+ */
+public class UndefinedServletElement extends BannedElement {
+
+  private static final String markerId = 
+      "com.google.cloud.tools.eclipse.appengine.validation.undefinedServletMarker";
+  private static final int severity = IMarker.SEVERITY_ERROR;
+  private String servletClassName;
```
akerekes: `final`

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -70,18 +72,47 @@ public void tearDown() throws CoreException {
...
+        + "</web-app>";
+    byte[] bytes = xml.getBytes(StandardCharsets.UTF_8);
+    validator.validate(resource, bytes);
+    IMarker[] markers = resource.findMarkers(UNDEFINED_SERVLET_MARKER, true, IResource.DEPTH_ZERO);
+    assertEquals(1, markers.length);
+    String expectedMessage = "DoesNotExist " + Messages.getString("undefined.servlet.class");
+    assertEquals(expectedMessage, markers[0].getAttribute(IMarker.MESSAGE));
+  }
+  
   private static void createFolders(IContainer parent, IPath path) throws CoreException {
```
akerekes: We already have similar methods in `com.google.cloud.tools.eclipse.appengine.newproject.CodeTemplates`. We could extract it into a util class and use it here instead of reimplementing. In a separate PR.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -70,18 +72,47 @@ public void tearDown() throws CoreException {
...
+        + " version='3.1'></web-app>";
+    byte[] bytes = xml.getBytes(StandardCharsets.UTF_8);
     validator.validate(resource, bytes);
     IMarker[] markers = resource.findMarkers(SERVLET_MARKER, true, IResource.DEPTH_ZERO);
     assertEquals(1, markers.length);
     assertEquals(ELEMENT_MESSAGE, (String) markers[0].getAttribute(IMarker.MESSAGE));
   }
   
+  @Test
+  public void testValidate_withExistingServletClass()
```
akerekes: Can you add a test where the ServletClass is in a package and used as such in the web.xml? E.g. `<servlet-class>foo.bar.ServletClass</servlet-class>`

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -64,4 +96,20 @@ public void testStartElement_correctVersion() throws SAXException {
     assertEquals(0, scanner.getBlacklist().size());
   }
   
+  @Test
+  public void testFindClass() {
+    IType type = WebXmlScanner.findClass("ServletClass");
+    assertNotNull(type);
+  }
+  
+  private static void createFolders(IContainer parent, IPath path) throws CoreException {
```
akerekes: We already have similar methods in `com.google.cloud.tools.eclipse.appengine.newproject.CodeTemplates`. We could extract it into a util class and use it here instead of reimplementing. In a separate PR.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -39,27 +38,30 @@
 
 public class WebXmlValidatorTest {
```
akerekes: It seems to me that these tests are duplicate and should rather be in the `WebXmlScannerTest` class and test `WebXmlScanner`. By looking at the `WebXmlValidator` class the tests in `WebXmlValidatorTest` should exercise the code in that class without any real validation, i.e. mock out the IResource and WebXmlSaxParser instances.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
chanseokoh: Ideally, it's best to search all the dependencies (which can include JARs) of the project containing the `web.xml`. Searching only the current project seems not enough. This could be tricky though.

akerekes: The method I suggested above includes the dependencies in the search scope.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
chanseokoh: I'll file an issue for this. This `createFolders` method is contagious.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
briandealwis: I think it should just be `servletClassName.length()`.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +60,65 @@ public void startElement(String uri, String localName, String qName, Attributes
         addToBlacklist(new JavaServletElement(Messages.getString("web.xml.version"), start, 0));
       }
     }
+    if ("servlet-class".equalsIgnoreCase(localName)) {
+      Locator2 locator = getLocator();
+      insideServletClass = true;
+      servletClassContents = new StringBuilder();
+      servletClassLocation = new DocumentLocation(locator.getLineNumber(),
```
briandealwis: This location doesn't seem right.  Shouldn't it just be `getColumnNumber()`?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -43,25 +47,34 @@
...
+  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator().withClasspathContainerPath("src/main/java");
   
   @BeforeClass
   public static void setUpBeforeClass() throws CoreException {
-    project = projectCreator.getProject();
+    IProject project = projectCreator.getProject();
+    javaProject = JavaCore.create(project);
+    
+    IPackageFragmentRoot root = javaProject.getPackageFragmentRoot("src/main/java");
+    root.createPackageFragment("com.test", false, null);
```
abbiefarr: I'm having trouble programmatically adding a package to an IProject in WebXmlScannerTest. This line throws `Java Model Exception: Java Model Status [java is read-only]`

Tried adding attributes to folders in createFolders() so they weren't read-only, but no luck.

chanseokoh: First, create the folder `src/main/java`.

Next, pass `project.getFolder(new Path("src/main/java"))` to `getPackageFragmentRoot()`.

chanseokoh: You can also use `projectCreator.getJavaProject()` instead of `JavaCore.create()`.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -64,4 +109,30 @@ public void testStartElement_correctVersion() throws SAXException {
...
+  @Test
+  public void testFindClass() {
+    assertFalse(WebXmlScanner.classExists(javaProject, "DoesNotExist"));
+    assertFalse(WebXmlScanner.classExists(null, null));
+    assertFalse(WebXmlScanner.classExists(null, ""));
+    assertTrue(WebXmlScanner.classExists(javaProject, "ServletClass"));
+  }
+  
+  @Test
+  public void testFindClass_inPackage() throws JavaModelException {
```
chanseokoh: No one throws `JavaModelException`, so remove it.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -17,19 +17,64 @@
...
+
 public class WebXmlScannerTest {
 
-  private WebXmlScanner scanner = new WebXmlScanner();
+  private static IFile resource;
+  private static IFile servletClass;
+  private static IJavaProject javaProject;
+  private WebXmlScanner scanner;
+  
+  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator().withClasspathContainerPath("src/main/java");
```
chanseokoh: Let's remove `withClasspathContainerPath()`. You're in the right path in that you will do something with project classpath entries to designate Java source folders, but this line is not going to work as you hoped.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
chanseokoh: It turns out that you don't even need to do anything. The default Java source folder was the project root, and as such, you were already creating `ServletClass` within the package `src.main.java`. That is, `assertTrue(WebXmlScanner.classExists(javaProject, "src.main.java.ServletClass"));` succeeds.

Now, there does exist a way to change the default Java source folder to something else, but this will required quite some effort. Not sure if it's worth setting up the project that way.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -17,19 +17,64 @@
...
+    IPackageFragmentRoot root = javaProject.getPackageFragmentRoot("src/main/java");
+    root.createPackageFragment("com.test", false, null);
+    createFolders(project, new Path("src/main/webapp/WEB-INF"));
+    resource = project.getFile("src/main/webapp/WEB-INF/web.xml");
+    resource.create(new ByteArrayInputStream(new byte[0]), true, null);
+    createFolders(project, new Path("src/main/java"));
+    
+    servletClass = project.getFile("src/main/java/ServletClass.java");
+    servletClass.create(new ByteArrayInputStream(
+        "public class ServletClass {}".getBytes(StandardCharsets.UTF_8)), true, null);
```
chanseokoh: As I said above, the Java source should have been like below, if you want to be absolutely correct:

```java
package src.main.java; public class ServletClass {}
```

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -64,4 +109,30 @@ public void testStartElement_correctVersion() throws SAXException {
     assertEquals(0, scanner.getBlacklist().size());
   }
   
+  @Test
+  public void testFindClass() {
+    assertFalse(WebXmlScanner.classExists(javaProject, "DoesNotExist"));
+    assertFalse(WebXmlScanner.classExists(null, null));
+    assertFalse(WebXmlScanner.classExists(null, ""));
+    assertTrue(WebXmlScanner.classExists(javaProject, "ServletClass"));
```
chanseokoh: As I said above, the following assertion will pass. (Not that you should change it.)

```java
assertTrue(WebXmlScanner.classExists(javaProject, "src.main.java.ServletClass"));
```

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -17,19 +17,64 @@
...
 import org.xml.sax.ext.Locator2Impl;
 import org.xml.sax.helpers.AttributesImpl;
 
+import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
+
 public class WebXmlScannerTest {
 
-  private WebXmlScanner scanner = new WebXmlScanner();
+  private static IFile resource;
+  private static IFile servletClass;
```
chanseokoh: No need for a field, as it's used only in `setUpBeforeClass()`.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ClassSearchRequestor.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jdt.core.search.SearchMatch;
+import org.eclipse.jdt.core.search.SearchRequestor;
+
+public class ClassSearchRequestor extends SearchRequestor {
+
+  private List<SearchMatch> matches = new ArrayList<>();
```
chanseokoh: `final`.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -40,6 +72,68 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+    try {
+      searchEngine.search(pattern,
+          new SearchParticipant[] { SearchEngine.getDefaultSearchParticipant() },
+          scope, requestor, monitor);
+      List<SearchMatch> results = requestor.getResults();
+      return !results.isEmpty();
+    } catch (CoreException ex) {
+      logger.log(Level.SEVERE, ex.getMessage());
+    }
+    return false;
```
chanseokoh: Move this inside the `catch` block.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -64,4 +109,30 @@ public void testStartElement_correctVersion() throws SAXException {
...
+    assertFalse(WebXmlScanner.classExists(null, null));
+    assertFalse(WebXmlScanner.classExists(null, ""));
+    assertTrue(WebXmlScanner.classExists(javaProject, "ServletClass"));
+  }
+  
+  @Test
+  public void testFindClass_inPackage() throws JavaModelException {
+    assertFalse(WebXmlScanner.classExists(javaProject, "DoesNotExist"));
+    assertFalse(WebXmlScanner.classExists(null, null));
+    assertFalse(WebXmlScanner.classExists(null, ""));
```
chanseokoh: These three `assertFalse` assertions are very obvious repetition from `testFindClass()`, so I'd just remove them here.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -17,19 +17,64 @@
...
+  private static IFile servletClass;
+  private static IJavaProject javaProject;
+  private WebXmlScanner scanner;
+  
+  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator().withClasspathContainerPath("src/main/java");
+  
+  @BeforeClass
+  public static void setUpBeforeClass() throws CoreException {
+    IProject project = projectCreator.getProject();
+    javaProject = JavaCore.create(project);
```
chanseokoh: `projectCreator.getJavaProject()`.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
akerekes: I'd recommend using a different package name. Using `src/main/java` as a source root folder is a Maven convention, which is fine (though as @chanseokoh mentioned, the default in an Eclipse project is `src`. To add the class to a package, I suggest to use a package name like `com.example` or `com.google.cloud.tools.eclipse.testing`.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -16,30 +16,122 @@
...
+      String servletClassName = servletClassElementContents.toString();
+      if (!classExists(project, servletClassName)) {
+    	// Adds <servlet-class> element to the queue if the class doesn't exist in the project.
+        addToBlacklist(new UndefinedServletElement(
+            servletClassName, servletClassElementLocation, servletClassName.length()));
+      }
+    }
+  }
+  
+  static boolean classExists(IJavaProject project, String typeName) {
```
akerekes: Is this `@VisibleForTesting`?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
akerekes: Interesting that both asserts would pass.

abbiefarr:
Ready for another review

@elharo

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidatorTest.java
```diff
@@ -55,14 +55,14 @@ public static void setUp() throws CoreException {
     createFolders(project, new Path("src/main/webapp/WEB-INF"));
     IFile webXml = project.getFile("src/main/webapp/WEB-INF/web.xml");
```
elharo: you don't need the webXML variable. just do 

webXmlFile = project...

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/PomXmlValidatorTest.java
```diff
@@ -58,15 +58,15 @@
...
+    webXmlFile = project.getFile("src/main/webapp/WEB-INF/web.xml");
+    ((IFile) webXmlFile).create(new ByteArrayInputStream(new byte[0]), true, null);
   }
   
   
   @After
   public void tearDown() throws CoreException {
-    if (resource != null) {
-      resource.deleteMarkers(IMarker.PROBLEM, true, IResource.DEPTH_ZERO);
+    if (webXmlFile != null) {
```
elharo: rather than cleaning up the file at the end of the test, can we make this a non-static instance variable? Static test data risks polluting one test with the next.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidatorTest.java
```diff
@@ -55,14 +55,14 @@ public static void setUp() throws CoreException {
...
-    resource = webXml;
+    webXmlFile = webXml;
   }
 
 
   @After
   public void tearDown() throws CoreException {
-    if (resource != null) {
-      resource.deleteMarkers(IMarker.PROBLEM, true, IResource.DEPTH_ZERO);
+    if (webXmlFile != null) {
```
elharo: rather than cleaning up the file at the end of the test, can we make this a non-static instance variable? Static test data risks polluting one test with the next.

plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.properties
```diff
@@ -5,3 +5,4 @@ appengine.blacklist.marker=App Engine Configuration Problem
 web.xml=Web XML Validator
 servlet.marker=Servlet Marker
 maven.plugin.marker=Obsolete Maven App Engine Plugin
+undefined.servlet.marker=Undefined Servlet Class
```
elharo: If possible, we should parameterize this with {0} to include the name of the undefined class.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ClassSearchRequestor.java
```diff
@@ -0,0 +1,39 @@
...
+
+  private final List<SearchMatch> matches = new ArrayList<>();
+  
+  @Override
+  public void acceptSearchMatch(SearchMatch match) throws CoreException {
+    matches.add(match);
+  }
+  
+  List<SearchMatch> getResults() {
+    return matches;
```
elharo: this exposes mutable private state. Consider making a copy of the list.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ClassSearchRequestor.java
```diff
@@ -0,0 +1,39 @@
...
+package com.google.cloud.tools.eclipse.appengine.validation;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jdt.core.search.SearchMatch;
+import org.eclipse.jdt.core.search.SearchRequestor;
+
+public class ClassSearchRequestor extends SearchRequestor {
```
elharo: does this class have to be public?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/UndefinedServletElement.java
```diff
@@ -0,0 +1,41 @@
...
+import org.eclipse.core.resources.IMarker;
+
+/**
+ * A servlet class element that will receive an undefined servlet marker. 
+ */
+public class UndefinedServletElement extends BannedElement {
+
+  private static final String markerId = 
+      "com.google.cloud.tools.eclipse.appengine.validation.undefinedServletMarker";
+  private static final int severity = IMarker.SEVERITY_ERROR;
```
elharo: I suggest inlining this since the field is actually less descriptive than the value.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -34,12 +67,74 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+      Locator2 locator = getLocator();
+      insideServletClassElement = true;
+      servletClassElementContents = new StringBuilder();
+      servletClassElementLocation =
+          new DocumentLocation(locator.getLineNumber(), locator.getColumnNumber());
+    }
+  }
+  
+  @Override
+  public void characters(char ch[], int start, int length) throws SAXException {
```
elharo: char[] ch

is slightly more common in Java

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidatorTest.java
```diff
@@ -55,14 +55,14 @@ public static void setUp() throws CoreException {
...
+      webXmlFile.deleteMarkers(IMarker.PROBLEM, true, IResource.DEPTH_ZERO);
```
briandealwis: Shouldn't you delete the `webXmlFile`?  That would take care of the markers, right?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/PomXmlValidatorTest.java
```diff
@@ -58,15 +58,15 @@
   public static void setUp() throws CoreException {
     project = projectCreator.getProject();
     createFolders(project, new Path("src/main/webapp/WEB-INF"));
-    resource = project.getFile("src/main/webapp/WEB-INF/web.xml");
-    ((IFile) resource).create(new ByteArrayInputStream(new byte[0]), true, null);
+    webXmlFile = project.getFile("src/main/webapp/WEB-INF/web.xml");
+    ((IFile) webXmlFile).create(new ByteArrayInputStream(new byte[0]), true, null);
```
briandealwis: Don't need the cast here.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/PomXmlValidatorTest.java
```diff
@@ -58,15 +58,15 @@
...
   }
   
   
   @After
   public void tearDown() throws CoreException {
-    if (resource != null) {
-      resource.deleteMarkers(IMarker.PROBLEM, true, IResource.DEPTH_ZERO);
+    if (webXmlFile != null) {
+      webXmlFile.deleteMarkers(IMarker.PROBLEM, true, IResource.DEPTH_ZERO);
```
briandealwis: And ditto here: just delete the file.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -17,20 +17,64 @@
...
+import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
+
 public class WebXmlScannerTest {
 
-  private WebXmlScanner scanner = new WebXmlScanner();
+  private static IFile resource;
+  private static IJavaProject javaProject;
+  private WebXmlScanner scanner;
+  
+  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator();
```
briandealwis: With the recent change to `TestProjectCreator`, you must explicitly provide the JavaFacet, and I think you'll also want the Dynamic Web Facet:
```
new TestProjectCreator().withFacetVersions(JavaFacet.VERSION_1_7, WebFacetUtils.WEB_25);
```
Note that the Dynamic Web Facet puts `web.xml` under `WebContent/` rather than `src/main/webapp` (which is Maven-ish).

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
```diff
@@ -16,73 +16,43 @@
...
+     throws CoreException, IOException, ParserConfigurationException {
+   String xml = 
+       "<web-app xmlns='http://java.sun.com/xml/ns/javaee' version='2.5'></web-app>";
+   byte[] bytes = xml.getBytes(StandardCharsets.UTF_8);
+   WebXmlValidator validator = new WebXmlValidator();
+   validator.validate(file, bytes);
+   String markerId = "com.google.cloud.tools.eclipse.appengine.validation.servletMarker";
+   Mockito.verify(file, Mockito.times(0)).createMarker(markerId);
+ }
+
```
briandealwis: Should we have a test that web.xml validation _doesn't_ happen on non-Dynamic Web Facet projects?

plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
```diff
@@ -81,6 +81,31 @@
...
+  </extension>
+  
+  <extension 
+      name="%web.xml"
+      point="org.eclipse.wst.validation.validatorV2" 
+      id="webXmlValidator">
+    <validator
+    	class="com.google.cloud.tools.eclipse.appengine.validation.WebXmlValidator"
+    	markerId="com.google.cloud.tools.eclipse.appengine.validation.undefinedServletMarker">
+      <include>
```
briandealwis: Maybe add `<facet id="jst.web" version="*"/>`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ClassSearchRequestor.java
briandealwis: Maybe better named as a `TypeSearchRequestor`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScanner.java
```diff
@@ -34,12 +67,74 @@ public void startElement(String uri, String localName, String qName, Attributes
...
+  @VisibleForTesting
+  static boolean performSearch(SearchPattern pattern, IJavaSearchScope scope,
+      IProgressMonitor monitor) {
+    try {
+      SearchEngine searchEngine = new SearchEngine();
+      ClassSearchRequestor requestor = new ClassSearchRequestor();
+      searchEngine.search(pattern,
+          new SearchParticipant[] { SearchEngine.getDefaultSearchParticipant() },
+          scope, requestor, monitor);
+      List<SearchMatch> results = requestor.getResults();
```
briandealwis: Is it worth accumulating the results if you're not actually using it?  `CountingSearchRequestor` with `count()`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/messages.properties
```diff
@@ -2,4 +2,5 @@ application.element=Project ID should be specified at deploy time
 remove.application.element=Remove application element
 web.xml.version=App Engine Standard does not support this servlet version
 convert.servlet=Convert to Java Servlet API 2.5
-maven.plugin=This App Engine Maven plugin has been deprecated
\ No newline at end of file
+maven.plugin=This App Engine Maven plugin has been deprecated
+undefined.servlet.class={0} does not exist in the project
```
briandealwis: The class could be pulled from a library.  `{0} could not be found`?  `could not be resolved`?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidatorTest.java
abbiefarr: Would it also be better to make project non-static and use Before instead of BeforeClass? I wasn't sure about that because then I'm creating a new project for every test, but there aren't that many tests so it's probably fine?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ClassSearchRequestor.java
abbiefarr: No, it does not. Good catch. Bad habit

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorTest.java
abbiefarr: Should that test be for AbstractXmlValidator? Are you talking about something similar to `testValidate_dynamicWebProject()` in AbstractXmlSourceValidatorTest?

briandealwis: That's what I was thinking of.  Great.

plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.properties
abbiefarr: This message appears as the error type in the problems view. For a similar error ("_____ cannot be resolved to a type") Eclipse gives the type as "Java Problem." But I'm fine to include the class name if you think it would be more descriptive

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -29,58 +33,99 @@
...
-  private final String ELEMENT_MESSAGE = "Project ID should be specified at deploy time.";
-  private static IResource resource;
-  private static IProject project;
+  private static final IProjectFacetVersion APPENGINE_STANDARD_FACET_VERSION_1 =
+      ProjectFacetsManager.getProjectFacet(AppEngineStandardFacet.ID).getVersion("1");;
+  private IFile resource;
 
-  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator();
+  @ClassRule public static TestProjectCreator appEngineStandardProjectCreator =
+      new TestProjectCreator().withFacetVersions(Lists.newArrayList(JavaFacet.VERSION_1_7,
```
briandealwis: Up to you: You could drop the `Lists.newArrayList()` since `withFacetVersions()` supports a varargs style too.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -29,58 +33,99 @@
...
+          WebFacetUtils.WEB_25, APPENGINE_STANDARD_FACET_VERSION_1));
+  
+  @ClassRule public static TestProjectCreator dynamicWebProjectCreator =
+      new TestProjectCreator().withFacetVersions(Lists.newArrayList(JavaFacet.VERSION_1_7,
```
briandealwis: Ditto

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -29,58 +33,99 @@
...
+    IFile file = project.getFile("src/main/webapp/WEB-INF/web.xml");
+    ValidationFramework framework = ValidationFramework.getDefault();
+    Validator[] validators = framework.getValidatorsFor(file);
+    for (Validator validator : validators) {
+      if ("com.google.cloud.tools.eclipse.appengine.validation.webXmlValidator"
+          .equals(validator.getId())) {
+        return;
+      }
     }
+    fail();
```
briandealwis: Might be useful to provide a description here: "webXmlValidator missing from AppEngine Standard project"

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -29,58 +33,99 @@
...
-
+  
+  @Test
+  public void testValidate_dynamicWebProject() throws CoreException {
+    ValidationFramework framework = ValidationFramework.getDefault();
+    Validator[] validators = framework.getValidatorsFor(resource);
+    for (Validator validator : validators) {
+      if ("com.google.cloud.tools.eclipse.appengine.validation.webXmlValidator"
+          .equals(validator.getId())) {
+        fail();
```
briandealwis: "webXmlValidator should not be found for jst.web projects"

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidatorTest.java
```diff
@@ -44,26 +45,24 @@
...
       "com.google.cloud.tools.eclipse.appengine.validation.appEngineBlacklistMarker";
-  private static IFile webXmlFile;
-  private static IProject project;
+  private IFile webXmlFile;
+  private IProject project;
 
-  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator();
+  @ClassRule public static TestProjectCreator projectCreator =
+      new TestProjectCreator()
+      .withFacetVersions(Lists.newArrayList(JavaFacet.VERSION_1_7, WebFacetUtils.WEB_25));
```
briandealwis: Ditto

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/PomXmlValidatorTest.java
```diff
@@ -49,25 +49,19 @@
...
       "com.google.cloud.tools.eclipse.appengine.validation.mavenPluginMarker";
-  private static IFile webXmlFile;
-  private static IProject project;
+  private IFile webXmlFile;
+  private IProject project;
   
-  @ClassRule public static TestProjectCreator projectCreator = new TestProjectCreator();
+  @ClassRule public static TestProjectCreator projectCreator =
+      new TestProjectCreator()
+      .withFacetVersions(Lists.newArrayList(JavaFacet.VERSION_1_7, WebFacetUtils.WEB_25));
```
briandealwis: Ditto

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -29,58 +33,99 @@
...
+      new TestProjectCreator().withFacetVersions(Lists.newArrayList(JavaFacet.VERSION_1_7,
+                                                                    WebFacetUtils.WEB_25));
 
-  @BeforeClass
-  public static void setUp() throws CoreException {
-    project = projectCreator.getProject();
+  @Before
+  public void setUp() throws CoreException {
+    IProject project = dynamicWebProjectCreator.getProject();
     createFolders(project, new Path("src/main/webapp/WEB-INF"));
```
briandealwis: So just be aware that new Dynamic Web Projects actually put the DD in `WebContent/WEB-INF/web.xml`, not `src/main/webapp/WEB-INF/web.xml`.  It doesn't matter for your validator as you're applying to all web.xml files, but maybe the tests should either make this explicit with a comment?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
```diff
@@ -17,20 +17,68 @@
...
+    resource.create(new ByteArrayInputStream(new byte[0]), true, null);
+    
+    ValidationTestUtils.createFolders(project, new Path("src/main/java"));
+    IFile servletClass = project.getFile("src/main/java/ServletClass.java");
+    servletClass.create(
+    	new ByteArrayInputStream("public class ServletClass {}".getBytes(StandardCharsets.UTF_8)),
+    	true, null);
+    
+    ValidationTestUtils.createFolders(project, new Path("com/example"));
+    IFile servletClassInPackage = project.getFile("com/example/ServletClassInPackage.java");
```
briandealwis: I didn't look at this closely enough: this needs to be in `src` for a jst.web.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
chanseokoh: AFAIK, one needs to specify a config to create the DD (`web.xml`) when adding the Dynamic Web facet. The following returns true, true, and false:

```java
    System.out.println("WebContent: " + project.getFolder("WebContent").exists());
    System.out.println("WebContent/WEB-INF: " + project.getFolder("WebContent/WEB-INF").exists());
    System.out.println("WebContent/WEB-INF/web.xml: " + project.getFolder("WebContent/WEB-INF/web.xml").exists());
```

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlScannerTest.java
abbiefarr: It seemed to me and @akerekes that the default source folder was the main project folder. So if I put the file in src/com/example, the package is src.com.example not com.example. Am I missing something?

chanseokoh: I'll clarify. Before #1591, the Java source folder was always the project root. After #1591, if you add a Java facet, the source folder is "src".

#1591 has been marged. @abbiefarr, you need to merge master into this PR and move the source files under "src".

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
chanseokoh: Nah, I was wrong. I was calling `.getFolder()` on `web.xml`.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -17,72 +17,93 @@
...
+    ValidationFramework framework = ValidationFramework.getDefault();
+    Validator[] validators = framework.getValidatorsFor(file);
+    for (Validator validator : validators) {
+      if ("com.google.cloud.tools.eclipse.appengine.validation.webXmlValidator"
+          .equals(validator.getId())) {
+        return;
+      }
     }
+    // webXmlValidator isn't applied to web.xml
+    fail();
```
chanseokoh: We typically do `fail("webXmlValidator isn't applied to web.xml")`.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AbstractXmlValidatorTest.java
```diff
@@ -17,72 +17,93 @@
...
+  
+  @Test
+  public void testValidate_dynamicWebProject() throws CoreException {
+    ValidationFramework framework = ValidationFramework.getDefault();
+    Validator[] validators = framework.getValidatorsFor(resource);
+    for (Validator validator : validators) {
+      if ("com.google.cloud.tools.eclipse.appengine.validation.webXmlValidator"
+          .equals(validator.getId())) {
+        // webXmlValidator should not be applied in jst.web project
+        fail();
```
chanseokoh: Ditto.

PR #1580, opened 2017-03-13
add menu item

elharo:
@briandealwis Not working or ready for check in yet.

Right now I'm having trouble figuring out how to get the menu item to show up in the build path menu rather than the top level menu. Anyone have an idea?


briandealwis:
I don't see that `GenerateBuildPathActionGroup` supports contributions: it has to opt-in by registering its menus with the workbench's `IMenuService`.

briandealwis:
I think the right place to add this is either our own _App Engine_ menu or in the _Configure_ menu.

Issue #1166, opened 2016-12-27
web.xml:4: cvc-elt.1: Cannot find the declaration of element 'web-app'


elharo:
Note that this is a false positive. Eclipse is incorrectly flagging an XML file without a DOCTYPE declaration as an error. It shouldn't do this. It's a huge bug in Eclipse, but we may need to work around it anyway. 

```
com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest
testPublishingSubmodules(com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest)  Time elapsed: 8.087 sec  <<< FAILURE!
java.lang.AssertionError: Imported projects have errors
/sox-server/WebContent/WEB-INF/web.xml:4: cvc-elt.1: Cannot find the declaration of element 'web-app'.
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at com.google.cloud.tools.eclipse.test.util.project.ProjectUtils.failIfBuildErrors(ProjectUtils.java:131)
	at com.google.cloud.tools.eclipse.test.util.project.ProjectUtils.failIfBuildErrors(ProjectUtils.java:125)
	at com.google.cloud.tools.eclipse.test.util.project.ProjectUtils.importProjects(ProjectUtils.java:112)
	at com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest.setUp(LocalAppEnginePublishOperationTest.java:51)
```

elharo:
Here's another flake in this test at a different place:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/189386937/log.txt

Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.607 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest
testPublishingSubmodules(com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest)  Time elapsed: 2.604 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:712)
[...]

PR #1586, opened 2017-03-13
Wait for build errors when importing zip project

chanseokoh:
About the second issue in #1166.


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -117,7 +117,7 @@
     }
 
     // wait for any post-import operations too
-    waitForProjects();
+    waitForProjects(projects.toArray(new IProject[0]));
```
briandealwis: Should we add a variant to support collections?

briandealwis:
LGTM.

PR #1587, opened 2017-03-13
Simplify library code

chanseokoh:
I was looking at the code in .libraries and thought I could improve readability a bit, mainly by using `List` in place of primitive Java arrays.

@akerekes what do you think?

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
@@ -17,23 +17,24 @@
 package com.google.cloud.tools.eclipse.appengine.libraries;
 
 import com.google.common.base.Preconditions;
+import java.util.List;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.jdt.core.IClasspathContainer;
 import org.eclipse.jdt.core.IClasspathEntry;
 
 public class LibraryClasspathContainer implements IClasspathContainer {
-  
+
   private final IPath containerPath;
   private final String description;
-  private final IClasspathEntry[] classpathEntries;
+  private final List<IClasspathEntry> classpathEntries;
 
   public LibraryClasspathContainer(IPath path, String description,
-      IClasspathEntry[] classpathEntries) {
+      List<IClasspathEntry> classpathEntries) {
     Preconditions.checkNotNull(path, "path is null");
     Preconditions.checkNotNull(description, "description is null");
     Preconditions.checkArgument(!description.isEmpty(), "description is empty");
     Preconditions.checkNotNull(classpathEntries, "classpathEntries is null");
-    
+
     this.containerPath = path;
     this.description = description;
     this.classpathEntries = classpathEntries;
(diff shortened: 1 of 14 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
```diff
@@ -66,6 +67,6 @@ public String getDescription() {
 
   @Override
   public IClasspathEntry[] getClasspathEntries() {
-    return classpathEntries;
+    return classpathEntries.toArray(new IClasspathEntry[0]);
```
elharo: I think this is why the arrays are used. A lot of the Eclipse API around this uses lists.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/SourceAttacherJob.java
```diff
@@ -54,37 +56,36 @@ protected IStatus run(IProgressMonitor monitor) {
...
       }
+
+      LibraryClasspathContainer newContainer =
+          libraryClasspathContainer.copyWithNewEntries(newClasspathEntries);
+      JavaCore.setClasspathContainer(containerPath, new IJavaProject[]{ javaProject },
+                                     new IClasspathContainer[]{ newContainer }, monitor);
+      serializer.saveContainer(javaProject, newContainer);
+      return Status.OK_STATUS;
+
     } catch (Exception ex) {
```
elharo: This looks like something we should fix if possible. (not in this PR) Why are we catching java.lang.Exception here?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
chanseokoh: Yeah, we need extra conversion here, but overall, looks like using a `List` is a win.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/SourceAttacherJob.java
```diff
@@ -54,37 +56,36 @@ protected IStatus run(IProgressMonitor monitor) {
     try {
       IClasspathContainer container = JavaCore.getClasspathContainer(containerPath, javaProject);
       if (container instanceof LibraryClasspathContainer) {
```
akerekes: Shouldn't the condition be negated? If so, and tests are still passing can you add one that fails on this before fixing it?

chanseokoh: Wow, nice catch!

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/SourceAttacherJob.java
akerekes: `sourceArtifactPathProvider` is a `Callable` and this `call()` declares Exception to be thrown. We can either replace it with a `com.google.common.base.Supplier, swallow any exception that happens inside and return a `null` or create an interface similar to `Callable` and declare `CoreException` to be thrown.

PR #1592, opened 2017-03-14
Fixes #1567 - Update the deployment preferences page according to project facet configuration

### LINKED ISSUE #1567: Update the deployment preferences page according to project facet configuration
Currently when the preferences of a project is opened (right click on the project > Properties) and the project's facet configuration is changed in incompatible way with the deployment preferences (e.g. the deployment preferences page is opened, then the project's current App Engine facet is removed or replaced with a different one), we display an error message at the top of the deploy preferences page informing the user that it's no longer valid for the current configuration, see #880 

It would be better if we actually updated the deploy preferences page based on the project's facet configuration, e.g. displaying std deploy preferences when the project has the AE Standard facet, but switch to flex deployment preferences when the user reconfigures the project.

akerekes:
Create 3 panels (flex, std, blank when no App Engine Facet as present) and displays the one corresponding to the project's facet configuration.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
```diff
@@ -61,9 +61,6 @@
         id="com.google.cloud.tools.eclipse"
         name="Google Cloud Platform"
         class="com.google.cloud.tools.eclipse.properties.GooglePropertyPage">
-       <enabledWhen>
-         <reference definitionId="com.google.cloud.tools.eclipse.appengine.onlyInGCPProjects" />
```
chanseokoh: This will now show the Google Cloud Platform root preference on any project (e.g., plain Java project). I'm not sure if we want that.

(BTW, if I'm not mistaken, the name `onlyInGCPProjects` is misleading in that it actually means `onlyInProjectsWithAppEngineStandardFacet` at the moment.)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/BlankDeployPreferencesPanel.java
```diff
@@ -0,0 +1,30 @@
...
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Label;
+
+public class BlankDeployPreferencesPanel extends DeployPreferencesPanel {
+
+  BlankDeployPreferencesPanel(Composite parent) {
+    super(parent, SWT.NONE);
+    Label noGAEProjectLabel = new Label(this, SWT.NONE);
+    noGAEProjectLabel.setText("Not an App Engine Application");
```
chanseokoh: `Gae`, not `GAE`. But then I'd do `new Label(...).setText(...)`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -105,69 +110,30 @@ public void run() {
...
 
   @Override
   public void dispose() {
-    if (content != null) {
-      // Content can be null if Properties Page is opened and App Engine facets are uninstalled
-      content.dispose();
-    }
+    blankPreferencesPanel.dispose();
+    standardPreferencesPanel.dispose();
+    flexPreferencesPanel.dispose();
```
chanseokoh: These panels are `Composite`s, which are anchored to a `PropertyPage`. I think it's not a general practice to explicitly dispose child widgets we create. (For example, we don't dispose `Label`s or `Button`s.) They will be disposed automatically when any of its ancestors is disposed.

OTOH, `databindingSupport` needs to be disposed explicitly.

briandealwis: Within the panels, you could do the DBC dispose from a dispose-listener.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -69,18 +72,20 @@ protected Control createContents(Composite parent) {
       return container;
```
nbashirbello: Should the BlankDeployPreferencesPanel be returned here since it has the appropriate error message?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: If we show it on all projects, then we can work around the problem that if a project does not have a GAE facet, then the deployment preferences item won't be shown on the left. If the user opens the preferences and adds either of the GAE facets the deployment preferences will not appear.

(You're correct about the `onlyInGCPProjects`'s incorrect naming.)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -105,69 +110,30 @@ public void run() {
...
-    } else {
-      logger.log(Level.WARNING,
-          "App Engine Deployment property page is only visible if project contains an App Engine facet");
-      return null;
-    }
+  private DeployPreferencesPanel getActivePanel() {
+    return (DeployPreferencesPanel) stackLayout.topControl;
   }
 
   @Override
```
nbashirbello: do we need to update the panels when setVisible(false)? I am thinking that when setVisible(false) is called, it is either the panel was already invisible (which we don't care too much about) or it was inimitably visible, which at that point it would have the correct configuration. And then the next time setVisible(true) is called, the configuration will be updated if necessary.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesPanel.java
```diff
@@ -48,7 +48,7 @@ protected FormToolkit getFormToolkit() {
     return formToolkit;
   }
 
-  private void initializeFormToolkit() {
+  protected void initializeFormToolkit() {
```
nbashirbello: Small - initializeFormToolkit() does't seem to be used outside this class. Its called in the DeployPreferencesPanel constructor, the GAE panel inherit.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
nbashirbello: onlyInGCPProjects would work well here, since the only GCP project the plugin supports publicly is Standard.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: Currently the panel for Standard disposes the DBC in it's dispose(), the other two do not dispose it. So I need to wire those up as well.
What's better: override dispose() and call DBC.dispose() or add a listener? The listener would come handy if the DBCs would be injected into the panels, right?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
briandealwis: One of the things I found annoying about GPE was how it contributed a `Google` properties page to **everything**.  We should only show property pages where they belong.

The problem with `onlyInGCPProjects` is it's not currently easily extensible for new project types without changing its definition.  We can come up with ways to fix that, but we should continue to use it.

How to make `onlyInGCPProjects` extensible? If our facets are mutually exclusive, we could install a nature. Or we could create a custom property test that checks the project's facets IDs matches `com.google.cloud.tools.**`, or against an extension point listing our known GCP facets.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
briandealwis: Don't override dispose(), but use `addDisposeListener()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -46,17 +45,21 @@
...
 
-  private DeployPreferencesPanel content;
-  private boolean canSetMessage = true;
-  private boolean isStandardPanel;
   private IFacetedProject facetedProject = null;
-  private String invalidFacetConfigErrorMessage = "";
   private static final Logger logger = Logger.getLogger(DeployPropertyPage.class.getName());
+  private FlexDeployPreferencesPanel flexPreferencesPanel;
+  private StandardDeployPreferencesPanel standardPreferencesPanel;
+  private BlankDeployPreferencesPanel blankPreferencesPanel;
```
briandealwis: Could you mark somewhere with a TODO to decouple this class from the specific preference panels?  One possibility is to use an extension point where we associate a panel class with a facet type:
```
<extension point="com.google.cloud.tools.deploy.ui.properties">
  <facet id="com.google.cloud.tools.eclipse.appengine.facets.flex" version="1"
    panelClass="com.google.cloud.tools.eclipse.appengine.deploy.ui.FlexDeployPreferencesPanel" />
<extension>
```

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/BlankDeployPreferencesPanel.java
```diff
@@ -0,0 +1,30 @@
...
+    return new DataBindingContext();
+  }
+
+  @Override
+  void resetToDefaults() {
+  }
+
+  @Override
+  boolean savePreferences() {
+    return true;
```
akerekes: We use the return value of the method to enable or prevent the OK button's click action to be continued. If we return false here, the users won't be able to close the dialog is this page is open. The deployment preferences manipulated by the other panels won't be affected.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: I've updated the PR that it will show the `BlankDeployPreferencesPanel` however not by explicitly returning it here.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/BlankDeployPreferencesPanel.java
akerekes: done

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: I understand the `GPE` complaint and I've looked into it whether we could dynamically update the preference page list on the left hand side of the dialog, however we cannot.
So with the risk of annoying some users, I'd rather keep the preferences page always visible to avoid forcing users to close and reopen the dialog in case they add one of the supported facets and want to change any GCP related preferences.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: I've added dispose listener to both `DeployPropertyPage` and `DeployPreferencesPanel` (the parent class of all the deploy preferences panels)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: added condition to update panels only when `setVisible(true)` is called.

akerekes:
Screenshot of the blank page

Blank deploy preferences page

akerekes:
@briandealwis @nbashirbello @chanseokoh PTAL

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
briandealwis: Let's at least make it specific to faceted projects.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -105,109 +116,92 @@ public void run() {
...
-      updatePageBasedOnFacetConfig(false);
-    } else {
-      invalidFacetConfigErrorMessage = "";
-      updatePageBasedOnFacetConfig(true);
+  private void createBlankPanelIfNeeded() {
+    if (blankPreferencesPanel == null) {
+      blankPreferencesPanel = new BlankDeployPreferencesPanel(container);
+      blankPreferencesPanel.setFacetPageSelector(new Runnable() {
+        @Override
+        public void run() {
```
chanseokoh: Can we move this logic into `BlankDeployPreferencesPanel`? It's the `BlankDeployPreferencesPanel` that cares about selecting the facet page, so I feel it's more coherent to move it to the panel.

Nits: `d` --> `dialog`, and there should be a space after type casting.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.xml
akerekes: I'll restrict it to projects. Non-faceted projects still have the `Project Facets` page in their preferences, so if the user opens that, converts it to faceted, then adds the GAE Standard (or Flexible) facet, they can immediately configure deployment settings if they want to.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: The reason I've put it here is that the panel is not aware of the `PreferenceDialog`, so it would take some `getParent()` and `instanceof` calls to find where it is.

chanseokoh: I guessed that. What if we pass `PreferenceDialog` (i.e., `setPreferenceDialog()`) instead of `Runnable()`? Passing `Runnable()` seems to add complexity which seems not really needed, and more importantly, it feels a bit off that this logic got out of `BlankDeployPreferencesPanel`.

However, I'm not strong about this, so I'll leave it up to you.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
```diff
@@ -105,109 +116,93 @@ public void run() {
...
+          }
+        }
+      });
     }
+  }
 
+  private void createStandardPanelIfNeeded() {
+    if (standardPreferencesPanel == null) {
+      IGoogleLoginService loginService =
+          PlatformUI.getWorkbench().getService(IGoogleLoginService.class);
```
briandealwis: If you like, you could use `ServiceContextFactory` to create this class and `@Inject` these values.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -66,7 +66,9 @@ flex.page.title=App Engine Deployment - Flexible Environment
 use.config.values=Use these values during deployment
 
 #Deploy properties page
-invalid.deploy.page.state=Only appropriate for projects with an {0} facet.
+add.facet.link=<a>Open the Project Facets page</a> and add the App Engine Java Standard \
+Environment or the App Engine Java Flexible Environment facet to enable the App Engine Deployment \
+settings.
```
briandealwis: The wording seems a bit odd?

> Deployment settings only available for projects with <a>facets</a> for App Engine Java Standard Environment or App Engine Java Flexible Environment.

chanseokoh: +1, since the current wording may give the impression that users have to take an immediate action, while they won't understand what it means most of the time. I also agree we don't really need to provide a shortcut link to the facet panel.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java
akerekes: I've removed the link, so all this logic has also been removed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPage.java

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
akerekes: done

PR #1599, opened 2017-03-15
Configure Travis to cache Google Cloud SDK

briandealwis:
Rather than download and install the Google Cloud SDK & App Engine Java components for each build, cache the installation and only update when we configure a new version.

This is different than the previous behaviour which always did a `gcloud components update` to update to the latest version.

It's unclear if this will increase performance much, but it will insulate us from failures to download from google.com.

Oh, and this updates to 147.0.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,18 +7,10 @@ before_install:
 install: 
   # use 3.3.9 to ensure faulty test exit states fail the build (#1276)
   - mvn -N io.takari:maven:wrapper -Dmaven=3.3.9
-  # download Cloud SDK (into .m2 so that it's cached)
-  - if [ ! -f .m2/google-cloud-sdk-${CLOUDSDK_VERSION}-linux-x86_64.tar.gz ]; then
-      wget -o .m2/google-cloud-sdk-145.0.0-linux-x86_64.tar.gz https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${CLOUDSDK_VERSION}-linux-x86_64.tar.gz
-    fi
-  - tar -xzf .m2/google-cloud-sdk-${CLOUDSDK_VERSION}-linux-x86_64.tar.gz
-  # update all Cloud SDK components
-  - gcloud components update --quiet
-  # add App Engine component to Cloud SDK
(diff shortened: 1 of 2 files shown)


chanseokoh:
Is this working as intended? Looks like every build in this PR downloads the Cloud SDK.

.travis.yml
```diff
@@ -7,13 +7,8 @@ before_install:
...
   - mvn -N io.takari:maven:wrapper -Dmaven=3.3.9
-  # download Cloud SDK
-  - wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-145.0.0-linux-x86_64.tar.gz
-  - tar -xzf google-cloud-sdk-145.0.0-linux-x86_64.tar.gz
-  # update all Cloud SDK components
-  - gcloud components update --quiet
-  # add App Engine component to Cloud SDK
-  - gcloud components install app-engine-java --quiet
+  # download and install Cloud SDK (may be cached)
+  - build/install-cloudsdk.sh 147.0.0 $HOME
```
elharo: I think 150.0.0 is current.

build/install-cloudsdk.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh
+# Download and install the Google Cloud SDK
+
+if [ $# -ne 2 ]; then
+    echo "Ensure Google Cloud SDK in <install-dir>/google-cloud-sdk is"
+    echo "at desired version"
+    echo "use: $0 <cloudsdk-version> <install-dir>"
+    echo "ex: $0 147.0.0 /tmp"
```
elharo: 150.0.0

briandealwis:
> Is this working as intended? Looks like every build in this PR downloads the Cloud SDK.

Because the builds were unsuccessful and so the results were not yet cached ;-)

briandealwis:
I was worried that caches may be poisoned by PRs, but Travis creates a cache per-PR and per-branch, and if a cache isn't found then it fetches the cache for the `master` branch.  You can see the caches on the [Travis page](https://travis-ci.org/GoogleCloudPlatform/google-cloud-eclipse/caches).

PR #1604, opened 2017-03-16
Fixes #1532 - Add refresh.png

### LINKED ISSUE #1532: Add refresh.png
`c.g.c.t.e.ui.util` depends on `org.eclipse.search` for the refresh.gif/png (depending on the Eclipse version). We should simply add the image into the bundle and remove the dependency on the `o.e.search` bundle.

akerekes:
Remove dependency on `org.eclipse.search` and add the `refresh.png` directly into the code. Should the icon not be found, the deploy settings panel will display a Unicode symbol instead, see screenshots:

image

image


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -87,6 +87,7 @@
       "https://console.cloud.google.com/appengine/versions";
   private static final String CREATE_GCP_PROJECT_WITH_GAE_URL =
       "https://console.cloud.google.com/projectselector/appengine/create?lang=java";
+  private static final String UNICODE_REFRESH_SYMBOL = "\uD83D\uDD04";
```
elharo: neat

akerekes: Credit to @briandealwis, as he suggested it

chanseokoh: I remember our style guide instructs to write unicode symbols directly and not by codes. Let me look for it.

briandealwis:
TBH: I prefer the unicode symbol.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -355,7 +356,11 @@ private void createProjectIdSection() {
         .applyTo(projectSelector);
 
     final Button refreshProjectsButton = new Button(projectSelectorComposite, SWT.NONE);
-    refreshProjectsButton.setImage(refreshIcon);
+    if (refreshIcon == null) {
```
briandealwis: Since you're now bundling in the icon, this if-not-present case never happens.

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/images/SharedImages.java
```diff
@@ -40,9 +34,7 @@
    * @return the image or null if the image was not found
    */
   public static Image createRefreshIcon(Device device) {
```
briandealwis: This class is … a bit unconventional :-)  The SWT convention is that the creator of an image is responsible for disposing of it.  In this case, this class is creating the image, and the panel is just using it.  Conventions can be broken of course.  The `org.eclipse.ui.IWorkbenchImages` provides `ImageDescriptor`s; it also provides `Image` instances from the central workbench image registry, which are good for the lifetime of the workbench.

This doesn't need to be changed in this PR.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
chanseokoh: https://google.github.io/styleguide/javaguide.html#s2.3.3-non-ascii-characters

chanseokoh:
Is it possible to remove `org.eclipse.search` from `MANIFEST.MF`?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/images/SharedImages.java
```diff
@@ -19,17 +19,11 @@
...
 import org.eclipse.swt.graphics.Image;
-import org.eclipse.ui.plugin.AbstractUIPlugin;
 
 public class SharedImages {
 
-  private static final ImageDescriptor refreshGifImageDescriptor =
-      AbstractUIPlugin.imageDescriptorFromPlugin("org.eclipse.search",
-                                                 "icons/full/elcl16/refresh.gif");
-
   private static final ImageDescriptor refreshPngImageDescriptor =
```
chanseokoh: I guess there is no need to put `Png` in the name anymore.

akerekes:
@elharo we'll need something similar to [this](http://git.eclipse.org/c/platform/eclipse.platform.ua.git/tree/org.eclipse.ui.intro.universal/about.html?h=master#n27) to explain that the refresh.png is EPL licensed.

akerekes:
For licensing reasons the images will go into a bundle under `third_party` accompanied by an EPL license file (similarly to how the other bundles under `third_party` contain a license file)

briandealwis:
I just discovered that there is a refresh image in `org.eclipse.debug.ui`, which we depend on and so is safe to reference.  `platform:/plugin/org.eclipse.debug.ui/icons/full/obj16/refresh_tab.png`

PR #1606, opened 2017-03-16
Simplify generating ide-target-platform

briandealwis:
Add a `pom.xml` to `eclipse/` to allow generating the `ide-target-platform` without having to build the entire project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,17 +122,23 @@ through to set up a working development environment.
 The Eclipse IDE and Tycho both use a _Target Platform_ to manage
 the dependencies for the source bundles and features under development.
 Although Tycho can pull dependencies directly from Maven-style
-repositories, Eclipse cannot.  So we use Tycho to cobble together
-a target platform suitable for the Eclipse IDE.
+repositories (like [Maven Central](https://search.maven.org)), Eclipse
+cannot.  So we use Tycho to cobble together
+a target platform suitable for the Eclipse IDE with the following command.
 ```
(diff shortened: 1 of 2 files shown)


chanseokoh:
Nice. I see there are profiles `build-eclipse-mars`, `build-eclipse-neon`, and `build-eclipse-oxygen` in the root `pom.xml`. Do we use them? I wonder if we can remove them.

README.md
```diff
@@ -122,17 +122,23 @@ through to set up a working development environment.
...
 the dependencies for the source bundles and features under development.
 Although Tycho can pull dependencies directly from Maven-style
-repositories, Eclipse cannot.  So we use Tycho to cobble together
-a target platform suitable for the Eclipse IDE.
+repositories (like [Maven Central](https://search.maven.org)), Eclipse
+cannot.  So we use Tycho to cobble together
+a target platform suitable for the Eclipse IDE with the following command.
 ```
-$ mvn -Pide-target-platform package
+$ (cd eclipse; mvn -Pide-target-platform package)
```
chanseokoh: `-Pide-target-platform` needed?

briandealwis: You're right, that's not needed.

But we still need the `build-eclipse-*` profiles: they're the profiles that actually configure the appropriate target platform based on the `eclipse.target` properties.

chanseokoh: Ah, didn't notice these profiles had `<activation>`.

PR #1608, opened 2017-03-16
Fixes #1589 - Add authUser to Cloud Console URLs

### LINKED ISSUE #1589: Add authUser to Cloud Console URLs
Cloud Console URLs support the `authuser=foo@gmail.com` parameter to specify which user should be selected when the URL is opened in the browser.

`ProjectSelectorSelectionChangedListener.java`
`StandardDeployPreferencesPanel.java (2 matches)`

akerekes:


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/ProjectSelectorSelectionChangedListenerTest.java
```diff
@@ -42,7 +42,8 @@
 public class ProjectSelectorSelectionChangedListenerTest {
 
   private static final String EXPECTED_LINK =
-      "https://console.cloud.google.com/appengine/create?lang=java&project=projectId";
+      "https://console.cloud.google.com/appengine/create?lang=java&project=projectId"
+      + "&authuser=user@example.com";
```
chanseokoh: I did some investigation, and my conclusion is that, although `@` may work in most systems and browsers, but it should be encoded.

PR #1610, opened 2017-03-17
Fix NPE in LocalAppEngineServerLaunchConfigurationDelegate

### LINKED ISSUE #1609: Possible NPE in LocalAppEngineServerLaunchConfigurationDelgate.getLaunch()
This is from https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1603#issuecomment-287168721.

`ILaunch.getLaunchConfiguration()` can return `null` (which is also mentioned in its Javadoc). I was able to reproduce this NPE once in a normal environment outside the context of #1603, so this can really happen.

chanseokoh:
Fixes #1609.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
@@ -86,7 +86,7 @@ public class LocalAppEngineServerLaunchConfigurationDelegate
 
   /**
    * Returns {@code value} unless it's null or empty, then returns {@code nullOrEmptyValue}.
-   * 
+   *
    * @see Strings#isNullOrEmpty(String)
    */
   private static String ifEmptyOrNull(String value, String nullOrEmptyValue) {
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -117,10 +117,17 @@ private static void validateCloudSdk() throws CoreException  {
   @Override
   public ILaunch getLaunch(ILaunchConfiguration configuration, String mode) throws CoreException {
     IServer thisServer = ServerUtil.getServer(configuration);
-    DefaultRunConfiguration thisConfig = generateServerRunConfiguration(configuration, thisServer);
+    return getLaunchInternal(configuration, mode, thisServer, getLaunchManager());
```
briandealwis: Could we make this less test-centric and instead something like:
```
IServer thisServer = ServerUtil.getServer(configuration);
checkConflictingLaunches(configuration, thisServer, launchManager.getLaunches()); 
return super.getLaunch(configuration, mode);
```
Could even pass in the `DefaultRunConfiguration` instance instead and avoid passing in the `IServer`.

chanseokoh: Good idea! I was doing this mechanically and syntactically, not trying to understand what this part does.

PR #1643, opened 2017-03-22
Add menu item for users to report bug

chanseokoh:
selection_003

It just opens up our repo issue tracker in an external browser. This assumes users will know they should log in to GitHub and press the green 'New Issue' button.


plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
+
+public class BugReportCommandHandler extends AbstractHandler {
+
+  @VisibleForTesting
+  static final String BUG_REPORT_URL =
+      "https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues";
```
akerekes: We could point the directly to the create new issue page by adding `/new` at the end of the URL.

plugins/com.google.cloud.tools.eclipse.bugreport/plugin.xml
```diff
@@ -0,0 +1,30 @@
...
+    <handler
+        class="com.google.cloud.tools.eclipse.bugreport.ui.BugReportCommandHandler"
+        commandId="com.google.cloud.tools.eclipse.bugreport.bugReportCommand">
+    </handler>
+  </extension>
+
+  <extension point="org.eclipse.ui.menus">
+    <menuContribution
+        allPopups="false"
+        locationURI="menu:com.google.cloud.tools.eclipse.appengine.actions?after=bugReport">
```
akerekes: Is this URI correct? Where is `bugReport` defined?

plugins/com.google.cloud.tools.eclipse.appengine.ui/plugin.xml
```diff
@@ -106,6 +106,10 @@
               name="accounts"
               visible="true">
         </separator>
+        <separator
```
briandealwis: Hmm, this seems overly specific.  How about "feedback"?

plugins/com.google.cloud.tools.eclipse.bugreport/plugin.properties
```diff
@@ -0,0 +1 @@
+bugReportMenu=Report Cloud Tools for Eclipse bug...
```
briandealwis: This doesn't read quite right — the _bug_ is too far away.

_Report bug about Cloud Tools for Eclipse…_?

Or since it's in our GCP menu, could it just be _Report Bug…_?

plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public Object execute(ExecutionEvent event) throws ExecutionException {
+    try {
+      IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
+      browserSupport.getExternalBrowser().openURL(new URL(BUG_REPORT_URL));
```
briandealwis: If you want to force the external browser, then it's simpler to use SWT directly:
```
import org.eclipse.swt.Program;
Program.launch(BUG_REPORT_URL);
```

akerekes:
We could customize the URL to prefill some fields of the issue, e.g.
labels, body:

http://stackoverflow.com/questions/11679087/send-user-to-create-new-github-issue-with-pre-filled-bug-data

Additionally​ we should create issue templates to help users provide the
necessary information for bug reports/feature requests (separate PR):

https://github.com/blog/2111-issue-and-pull-request-templates

plugins/com.google.cloud.tools.eclipse.bugreport/plugin.properties
akerekes: I agree, the shorter looks better. If we customize the link we could provide 2 menu items one for bugs one for feature requests and the issues created would have the proper label (bug/enhancement) attached.

chanseokoh: > I'm wondering if we want this to be Feedback rather than Bug Report? Or Report Issue? What do you think?

Using both: _Submit issue or feedback_?

plugins/com.google.cloud.tools.eclipse.bugreport/plugin.xml
chanseokoh: It's what I define as a separator in `.appengine.ui/plugin.xml`.

plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java
chanseokoh: Nice. Seems like we can have some level of customization even just with URL.

chanseokoh:
From the [official doc](https://developer.github.com/v3/issues/#create-an-issue), `title` and `body` seem to be the only usable options for outside users when creating a new issue.

chanseokoh:
So what I did now is to pre-fill some info in the issue body like this:

image

plugins/com.google.cloud.tools.eclipse.appengine.ui/plugin.xml
chanseokoh: Done.

PR #1645, opened 2017-03-22
move integration tests to end of list

### LINKED ISSUE #1641: Make integration tests run last
Since they take a long time to run and are our flakiest tests, put them last in test order so failures in other tests are found sooner.

elharo:
@briandealwis fix #1641 

This may not absolutely guarantee the integration tests run last, but it does have that effect on my system.


akerekes:
You could add a comment explaining the position before a well intended
developer moves it to its alphabetical position.

PR #1649, opened 2017-03-23
Don't create folder that already exists

### LINKED ISSUE #1646: StandardFacetInstallDelegate nullpointerexception
Not yet seen in production but I think there's a bug here:

```java
  /**
   * Creates an appengine-web.xml file in the WEB-INF folder if it doesn't exist
   */
  private void createConfigFiles(IProject project, IProgressMonitor monitor)
      throws CoreException {
    SubMonitor progress = SubMonitor.convert(monitor, 10);

    // The virtual component model is very flexible, but we assume that
    // the WEB-INF/appengine-web.xml isn't a virtual file remapped elsewhere
    IFolder webInfDir = WebProjectUtil.getWebInfDirectory(project);
    IFile appEngineWebXml = webInfDir.getFile(APPENGINE_WEB_XML);

    if (appEngineWebXml.exists()) {
      return;
    }

    ResourceUtils.createFolders(webInfDir, progress.newChild(2));

    appEngineWebXml.create(new ByteArrayInputStream(new byte[0]), true, progress.newChild(2));
    String configFileLocation = appEngineWebXml.getLocation().toString();
    AppEngineTemplateUtility.createFileContent(
        configFileLocation, AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
        Collections.<String, String>emptyMap());
    progress.worked(6);
  }
```

If `webInfDir` doesn't exist, then `WebProjectUtil.getWebInfDirectory(project);` returns null.

If `webInfDir` does exist, then there's no need to call `ResourceUtils.createFolders(webInfDir, progress.newChild(2));`

elharo:
fix #1646

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegateTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegateTest.java
@@ -16,7 +16,12 @@
 
 package com.google.cloud.tools.eclipse.appengine.facets;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
@@ -26,6 +31,7 @@ import org.junit.Rule;
 import org.junit.Test;
 
 import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
+import com.google.cloud.tools.eclipse.util.io.ResourceUtils;
 
 public class StandardFacetInstallDelegateTest {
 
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
```diff
@@ -113,32 +116,39 @@ protected IStatus run(IProgressMonitor monitor) {
...
+    SubMonitor progress = SubMonitor.convert(monitor, 9);
 
     // The virtual component model is very flexible, but we assume that
     // the WEB-INF/appengine-web.xml isn't a virtual file remapped elsewhere
     IFolder webInfDir = WebProjectUtil.getWebInfDirectory(project);
+    
+    if (webInfDir == null) {
+      webInfDir =
+          project.getFolder(WebProjectUtil.DEFAULT_WEB_PATH).getFolder(WebProjectUtil.WEB_INF);
+      ResourceUtils.createFolders(webInfDir, new NullProgressMonitor());
```
chanseokoh: The javadoc of `createFolders` says you can pass `null` for the monitor.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegateTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
+
+public class StandardFacetInstallDelegateTest {
+
+  private StandardFacetInstallDelegate delegate = new StandardFacetInstallDelegate();
+  private IProgressMonitor monitor = new NullProgressMonitor(); 
+  private IProject project;
```
chanseokoh: Not a big deal, but this doesn't need to be a field, at least for now.

chanseokoh:
The new test seems flaky. The error looks interesting to me.

```
testCreateConfigFiles(com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallDelegateTest)  Time elapsed: 2.094 sec  <<< ERROR!
org.eclipse.core.internal.resources.ResourceException: Resource is out of sync with the file system: '/test0.10732607351401313/src/main/webapp/WEB-INF/appengine-web.xml'.
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.read(FileSystemResourceManager.java:830)
	at org.eclipse.core.internal.resources.File.getContents(File.java:269)
	at org.eclipse.core.internal.resources.File.getContents(File.java:260)
	at com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallDelegateTest.testCreateConfigFiles(StandardFacetInstallDelegateTest.java:61)
```

elharo:
I think this is what's going on:

https://wiki.eclipse.org/FAQ_When_should_I_use_refreshLocal%3F

chanseokoh:
That seems like the case! We actually have some other places we modify files without using Eclipse API (e.g., XSLT transformation in App Engine Standard project conversion and XML validation), so we need to revisit them too.

elharo:
Maybe we should look into createConfigFiles to see if it needs to call refereshLocal

elharo:
FYI, I've rerun this three times now and have seen any flakiness since the latest changes.

chanseokoh:
> Maybe we should look into createConfigFiles to see if it needs to call refereshLocal

I'm also thinking if `refreshLocal` shouldn't be the right way.

elharo:
Ping. This needs an approval.

Issue #1442, opened 2017-02-17
Do API calls in separate thread to keep the UI responsive


akerekes:
Currently the list of projects and the check for App Engine application is synchronous call on the main thread. This means:
* if the user clicks on the deploy action, the dialog will not be displayed until the project list is retrieved
* if the user selects a project the dialog will not respond to other input until the response is received to the check for App Engine application

PR #1653, opened 2017-03-23
Fetch GCP projects in background for deploy preferences

chanseokoh:
First part of #1442: fetching GCP projects via API in a background job.


chanseokoh:
I noticed a slight problem... working on it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/GcpProjectQueryJob.java
```diff
@@ -0,0 +1,90 @@
...
+      final List<GcpProject> projects = projectRepository.getProjects(credential);
+
+      // The selector may have been disposed (i.e., dialog closed); check it in the UI thread.
+      display.syncExec(new Runnable() {
+        @Override
+        public void run() {
+          if (!projectSelector.isDisposed()
+              && isLatestQueryJob.apply(thisJob) /* intentionally checking in UI context */) {
+            projectSelector.setProjects(projects);
+            dataBindingContext.updateTargets();  // Select saved choice, if any.
```
akerekes: Is this necessary, because `projectSelector.setProjects()` only makes the selection but doesn't update the UI?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/GcpProjectQueryJob.java
```diff
@@ -0,0 +1,90 @@
...
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.swt.widgets.Display;
+
+/**
+ * Generic {@link Job} that queries GCP projects of given {@link Credential} through
+ * {@link ProjectRepository} and updates {@link ProjectSelector} asynchronously.
+ */
+public class GcpProjectQueryJob extends Job {
```
akerekes: Could this class extend the `UIJob` class instead?

chanseokoh:
The problem is when we auto-select a single account. Blocked on #1656.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanelTest.java
```diff
@@ -163,28 +169,35 @@ public void testUncheckStopPreviousVersionButtonWhenDisabled() {
...
         new ProjectScope(project).getNode(StandardDeployPreferences.PREFERENCE_STORE_QUALIFIER);
     node.put("project.id", "projectId1");
     node.put("account.email", EMAIL_1);
     initializeProjectRepository();
     when(loginService.getAccounts()).thenReturn(new HashSet<>(Arrays.asList(account1, account2)));
     deployPanel = createPanel(true /* requireValues */);
+    deployPanel.latestGcpProjectQueryJob.join();
+    node.clear();
+
+    ProjectSelector projectSelector = getProjectSelector();
```
akerekes: You should `assertNotNull` the `projectSelector`

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanelTest.java
```diff
@@ -163,28 +169,35 @@ public void testUncheckStopPreviousVersionButtonWhenDisabled() {
...
+      throws ProjectRepositoryException, InterruptedException, BackingStoreException {
     IEclipsePreferences node =
         new ProjectScope(project).getNode(StandardDeployPreferences.PREFERENCE_STORE_QUALIFIER);
     node.put("project.id", "projectId1");
     node.put("account.email", EMAIL_1);
     initializeProjectRepository();
     when(loginService.getAccounts()).thenReturn(new HashSet<>(Arrays.asList(account1, account2)));
     deployPanel = createPanel(true /* requireValues */);
+    deployPanel.latestGcpProjectQueryJob.join();
+    node.clear();
```
akerekes: If the purpose is to cleanup then this should be in either a `finally` block or a `@After` method to ensure that it's executed in case an exception is thrown before the method execution gets to this line.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanelTest.java
```diff
@@ -163,28 +169,35 @@ public void testUncheckStopPreviousVersionButtonWhenDisabled() {
...
-        assertThat(((GcpProject) selection.getFirstElement()).getId(), is("projectId1"));
-        return;
+        return (ProjectSelector) control;
       } else if (control instanceof Composite) {
         children.addAll(Arrays.asList(((Composite) control).getChildren()));
       }
-    };
-    fail("Did not find ProjectSelector widget");
+    }
+    return null;
```
akerekes: I'd keep the `fail()` here as there's not such test case where `null` is an expected outcome.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
```diff
@@ -128,9 +130,10 @@ public int selectAccount(String email) {
       combo.select(index);
       selectedAccount = (Account) combo.getData(email);
     }
-    // Need to fire even if nothing is selected (which is an error state), to show the error
-    // message: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1303
-    fireSelectionListeners();
+
+    if (oldIndex != index) {
+      fireSelectionListeners();
+    }
```
akerekes: Does this change lead to #1656?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/GcpProjectQueryJob.java
chanseokoh: The GCP query which may take a lot of time should be executed in a non-UI thread, so no.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
chanseokoh: It's related, but this PR is basically blocked on #1656. This will be cleaned up once #1656 is fixed.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/GcpProjectQueryJob.java
chanseokoh: It's right that it's necessary to make an explicit selection, but I don't think it's to update the UI. `projectSelector.setProjects()` has initially no selection at all.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanelTest.java
chanseokoh: With https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1653#discussion_r107927606, it will never return `null`, so we are covered.

chanseokoh:
The blocking #1656 is mixed. I merged it into this branch, and now [the problem](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1653#issuecomment-289050018) is gone. I added [one more test case](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1653/commits/10a44d6d2cfeddc563240f3ccb2d35171fa5ee4b) to cover the single account auto-selection case. @akerekes PTAL.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -451,26 +450,24 @@ private Composite createBucketSection(Composite parent) {
     return bucketComposite;
   }
 
+  @VisibleForTesting
+  Job latestGcpProjectQueryJob;  // Must be updated/accessed in the UI context.
```
akerekes: This would be better as `private` and make a getter `@VisibleForTesting`

PR #1657, opened 2017-03-24
Let data binding auto-select single account

### LINKED ISSUE #1566: 'Restore Defaults' in deploy preferences dialog doesn't restore UI components: account selector and project selector
Steps to reproduce:
1. Create a new project.
2. Open the deploy preferences, select an account, select a project, and click _OK_.
3. Re-open the deploy preferences. Verify that the previously selected account and project are loaded.
4. Click _Restore Defaults_. Verify that the account and project remain untouched. Click _OK_.
5. Re-open the deploy preferences. Observe that the account selector and the project selector have been cleared.

Expected: either
1) at step 4, _Restore Defaults_ will clear the account selector and the project selector;
or
2) at step 5, the page should load previously selected values.

I feel 1) is the correct behavior. Technically, I think at step 4, the model is cleared, but the UI components are not synchronized to reflect that.

### LINKED ISSUE #1656: Make data binding select single account instead of making AccountSelector select in deploy panel
In the deploy preferences, we auto-select an account if the account is the only available one.

The way we do is to let `AccountSelector` select it when it is instantiated (i.e., in its constructor). This means that `ProjectSelector` won't get the account selection change event through the listener, since the listener can be registered only after `AccountSelector` is created.

I think the right way is to let the data binding framework select the account, so that other binding components are properly notified and sync'ed.

Also, fixing this would probably allow a number of improvements in `AccountSelector`, such as firing selection change event only when there was a change. (Currently, the reason `ProjectSelector` has been working fine even if the data binding isn't selecting an account is that `AccountSelector` fires a lot of false selection change events. These false events, I think, is actually a bug.)

chanseokoh:
Fixes #1656, primarily. Fixes #1566 too.

Now the entity that auto-selects a single account is the deploy panel, not the `AccountSelector` itself. We already have a test in the panel if an account is auto-selected, so we are covered.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
@@ -172,12 +172,15 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
     UpdateValueStrategy modelToTarget =
         new UpdateValueStrategy().setConverter(new Converter(String.class, String.class) {
           @Override
-          public Object convert(Object expectedEmail) {
-            // Expected to be an email address, but must also ensure is a currently logged-in
-            // account
-            if (expectedEmail instanceof String
-                && accountSelector.isEmailAvailable((String) expectedEmail)) {
-              return expectedEmail;
+          public Object convert(Object savedEmail) {
+            Preconditions.checkArgument(savedEmail instanceof String);
+            // Check if the saved email is available in AccountSelector (i.e., logged in).
+            if (accountSelector.isEmailAvailable((String) savedEmail)) {
+              return savedEmail;
+            // Let data binding auto-select an account if it's the only available one.
+            // (However, if we don't require values, then don't auto-select.)
+            } else if (requireValues && accountSelector.getAccountCount() == 1) {
+              return accountSelector.getFirstEmail();
             } else {
               return null;
             }
@@ -318,9 +321,8 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
     accountLabel.setText(Messages.getString("deploy.preferences.dialog.label.selectAccount"));
     accountLabel.setToolTipText(Messages.getString("tooltip.account"));
 
-    // If we don't require values, then don't auto-select accounts
     accountSelector = new AccountSelector(this, loginService,
-        Messages.getString("deploy.preferences.dialog.accountSelector.login"), requireValues);
+        Messages.getString("deploy.preferences.dialog.accountSelector.login"));
     accountSelector.setToolTipText(Messages.getString("tooltip.account"));
     GridData accountSelectorGridData = new GridData(SWT.FILL, SWT.CENTER, true, false);
     accountSelector.setLayoutData(accountSelectorGridData);
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
@@ -174,11 +174,8 @@ public class StandardDeployPreferencesPanel extends DeployPreferencesPanel {
           @Override
           public Object convert(Object savedEmail) {
             Preconditions.checkArgument(savedEmail instanceof String);
-            // Check if the saved email is available in AccountSelector (i.e., logged in).
             if (accountSelector.isEmailAvailable((String) savedEmail)) {
               return savedEmail;
-            // Let data binding auto-select an account if it's the only available one.
-            // (However, if we don't require values, then don't auto-select.)
             } else if (requireValues && accountSelector.getAccountCount() == 1) {
               return accountSelector.getFirstEmail();
             } else {
--- a/plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountSelectorTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountSelectorTest.java
@@ -32,7 +32,6 @@ import com.google.cloud.tools.login.Account;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swtbot.swt.finder.widgets.SWTBotCombo;
 import org.junit.Before;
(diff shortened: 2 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
```diff
@@ -107,30 +95,35 @@ public String getSelectedEmail() {
     return combo.getText();
   }
 
+  /** @exception IllegalArgumentException if there is no account logged in */
```
akerekes: This should be broken into 3 lines like other javadoc one liners.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
```diff
@@ -107,30 +95,35 @@ public String getSelectedEmail() {
...
+  public String getFirstEmail() {
+    Preconditions.checkArgument(getAccountCount() > 0);
```
akerekes: Why not returning a `null`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -172,12 +172,15 @@ private void setupAccountEmailDataBinding(DataBindingContext context) {
...
-            // account
-            if (expectedEmail instanceof String
-                && accountSelector.isEmailAvailable((String) expectedEmail)) {
-              return expectedEmail;
+          public Object convert(Object savedEmail) {
+            Preconditions.checkArgument(savedEmail instanceof String);
+            // Check if the saved email is available in AccountSelector (i.e., logged in).
+            if (accountSelector.isEmailAvailable((String) savedEmail)) {
+              return savedEmail;
+            // Let data binding auto-select an account if it's the only available one.
```
akerekes: I believe this 2 line of comment is not necessary as they describe what the code does and the code is easy to understand.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -172,12 +172,15 @@ private void setupAccountEmailDataBinding(DataBindingContext context) {
...
           @Override
-          public Object convert(Object expectedEmail) {
-            // Expected to be an email address, but must also ensure is a currently logged-in
-            // account
-            if (expectedEmail instanceof String
-                && accountSelector.isEmailAvailable((String) expectedEmail)) {
-              return expectedEmail;
+          public Object convert(Object savedEmail) {
+            Preconditions.checkArgument(savedEmail instanceof String);
+            // Check if the saved email is available in AccountSelector (i.e., logged in).
```
akerekes: Not sure if this line of code is necessary.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountSelectorTest.java
```diff
@@ -396,6 +322,28 @@ public void testInitialItemOrder() {
...
+  @Test(expected = IllegalArgumentException.class)
+  public void testGetFirstEmail_noAccount() {
+    AccountSelector selector = new AccountSelector(shell, loginService, "<select this to login>");
+    selector.getFirstEmail();
+  }
+
+  @Test
+  public void testGetFirstEmail_oneAccount() {
+    when(loginService.getAccounts())
+        .thenReturn(new LinkedHashSet<>(Arrays.asList(account2)));
```
akerekes: Minor: this could work with `Collections.singleton(account2)` too.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountSelectorTest.java
```diff
@@ -396,6 +322,28 @@ public void testInitialItemOrder() {
...
+    AccountSelector selector = new AccountSelector(shell, loginService, "<select this to login>");
+    assertEquals(account2.getEmail(), selector.getFirstEmail());
+  }
+
+  @Test
+  public void testGetFirstEmail_threeAccounts() {
+    when(loginService.getAccounts())
+       .thenReturn(new LinkedHashSet<>(Arrays.asList(account3, account2, account1)));
```
akerekes: `Sets.newHashSet()`?

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
akerekes: If we keep the precondition, then it should be `checkState()` as we're not checking an argument of the method but the state of the object.

chanseokoh: Gosh, I keep slipping this. I prefer the precondition, because no one in our code calls (and should not call) this method while in a broken state.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountSelectorTest.java
chanseokoh: The javadoc of `Sets.newHashSet()` says

> Note for Java 7 and later: this method is now unnecessary and should be treated as deprecated. Instead, use the HashSet constructor directly, taking advantage of the new "diamond" syntax.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
elharo: I agree with @akerekes checkState is the right Precondition call here.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountSelectorTest.java
```diff
@@ -388,14 +314,35 @@ public void testGetAccountCount_threeAccounts() {
   @Test
   public void testInitialItemOrder() {
     when(loginService.getAccounts())
-        .thenReturn(new LinkedHashSet<>(Arrays.asList(account3, account2, account1)));
+        .thenReturn(new HashSet<>(Arrays.asList(account3, account2, account1)));
```
akerekes: with `newHashSet(E...)` you don't need the list creation

chanseokoh: Yeah, I get that, but it's trouble to use it considering https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1657#discussion_r107978531.

PR #1659, opened 2017-03-25
keep history of transformed files

elharo:
@chanseokoh @abbiefarr 

also adds test that transformed files are synced.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/Xslt.java
+++ b/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/Xslt.java
@@ -42,9 +42,9 @@ public class Xslt {
     try (InputStream in = file.getContents();
         InputStream stylesheetStream = xslt.openStream();
         InputStream resultStream = applyXslt(in, stylesheetStream)) {
-        boolean force = true;
-        boolean keepHistory = true;
-        file.setContents(resultStream, force, keepHistory, null /* monitor */);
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/Xslt.java
```diff
@@ -42,7 +42,9 @@ public static void transformInPlace(IFile file, URL xslt)
     try (InputStream in = file.getContents();
         InputStream stylesheetStream = xslt.openStream();
         InputStream resultStream = applyXslt(in, stylesheetStream)) {
-      file.setContents(resultStream, IFile.FORCE, null /* monitor */);
+        boolean force = true;
+        boolean keepHistory = true;
+        file.setContents(resultStream, force, keepHistory, null /* monitor */);
```
chanseokoh: Huh, it's a relief that we were actual using Eclipse API, so we had no sync problem.

BTW, indentation is not right.

chanseokoh: So, accordingly to the Javadoc, seems like `KEEP_HISTORY` is recommended almost always:

> Specifying <code>KEEP_HISTORY</code> is recommended except in circumstances where past states of the files are of no conceivable interest to the user.

In that regard, it'd be worth to see if there is any other `setContents()` usage in our code without the `KEEP_HISTORY` flag.

PR #1660, opened 2017-03-25
Assortment of minor fixes

chanseokoh:
Missing `@Test`s, copyright headers, minor simplifications and fixes, etc.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AbstractDeployPropertyPageForProjectsTest.java
```diff
@@ -85,7 +84,7 @@ public void testCorrectPanelIsShownForFacetedProject() throws CoreException {
   }
 
   abstract protected IProject getProject();
-  
+
```
elharo: I think we usually use `protected abstract` rather than `abstract protected`

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPageForFlexProjectTest.java
```diff
@@ -16,28 +16,22 @@
...
-import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
-import java.util.Arrays;
 import org.eclipse.core.resources.IProject;
-import org.eclipse.jst.common.project.facet.core.JavaFacet;
-import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
 import org.junit.Ignore;
-import org.junit.Rule;
 
-// until Flex facet is enabled
+// TODO: uncomment when Flex facet is enabled
```
elharo: We need to figure out a better way to handle this. We shouldn't be checking in ignored tests.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectConfigTest.java
```diff
@@ -34,35 +34,37 @@
...
+
+  @Test
   public void testProject() throws CoreException {
     try {
       IWorkspace workspace = ResourcesPlugin.getWorkspace();
       IProject project = workspace.getRoot().getProject("foobar");
+      project.create(null);
       config.setProject(project);
-      Assert.assertTrue(project.getRawLocationURI().getPath().endsWith("foobar"));
+      Assert.assertTrue(project.getLocationURI().getPath().endsWith("foobar"));
```
elharo: What's the difference between raw and cooked? Why prefer one over the other in this test?

plugins/com.google.cloud.tools.eclipse.preferences.test/src/com/google/cloud/tools/eclipse/preferences/PluginXmlTest.java
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *    http://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
```
elharo: I'm fairly picky but a single indentation space in a copyright comment exceeds even my tolerance for adding junk to the commit history.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectConfigTest.java
chanseokoh: Not sure honestly. I don't really get what this test case wants to achieve. When this test was enabled by adding `@Test`, it was failing, as `getRawLocationURI` was returning `null`. `getLocationURI` doesn't at least return `null`, and it actually makes the test pass.

chanseokoh: Not sure honestly. I don't really get what this test case wants to achieve. When this test was enabled by adding `@Test`, it was failing, as `getRawLocationURI` was returning `null`. `getLocationURI` doesn't at least return `null`, and it actually makes the test pass.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPropertyPageForFlexProjectTest.java
briandealwis: Could be called by an `assume()` check, right?

chanseokoh: Couldn't think of a good way to use `assume...()`, because the situation here is that you cannot even instantiate a test class.

I ended up adding the `.flex` bundle to `Required-Bundle` in this test bundle.

PR #1667, opened 2017-03-27
Switch to using Tycho Surefire's `hourly` setting

### LINKED ISSUE #1520: Run tests in randomized order
We should run our tests in randomized order to discover any interference between them, see [tycho-surefire-plugin docs](https://eclipse.org/tycho/sitedocs/tycho-surefire/tycho-surefire-plugin/test-mojo.html#runOrder)

briandealwis:
By default Tycho-Surefire runs tests in _filesystem order_.  This change uses `hourly` instead, which runs tests in alphabetic order, but reversing the order every other hour.

Fixes #1520


briandealwis:
Just wait another hour :-)

PR #1669, opened 2017-03-27
Provide a _debug target_ item for both 'debug' and 'run'

### LINKED ISSUE #942: Unterminatable dev_appserver.py
Even after fixing #393, we cannot run two local servers simultaneously, since subsequent servers will try to use the same datastore that the first server is using. When the second server starts, it will seem to successfully go live initially. But after several seconds, the server goes down (STOPPED state in Eclipse) showing the following error:

`OperationalError: database is locked`

Note that this database lock is expected; it is said that dev_appserver.py is not designed to run concurrently, and we won't try to fix it.

The problem is that, the python process of the servers will not terminate, while they are marked as "stopped" in Eclipse. Terminating Eclipse doesn't terminate them either. Users have to kill the processes manually.

briandealwis:
We had a few code paths where our launch may fail, such as a port already in use, which resulted in a launch could not be terminated and removed.  It seems odd, but a Platform/Debug _ILaunch_ object (the top-level elements of what's shown in the _Debug_ view) can only be _terminated_ providing it has at least one associated `IProcess` or `IDebugTarget` instance.

This change causes a 'debug target' to be immediately added to a launch.  Basically, it promotes the _server monitor_ object, that existed to (1) open the web browser on successful launch and (2) keep launch and server state changes in sync, into a debug target.  This change attaches the _Generic GCP_ icon to this target from the [GCP Icons set](https://cloud.google.com/icons/) to differentiate it from the little gears icon associated with JVMs; this requires associating a `IDebugModelPresentation` to this debug target.  The result looks like:

screen shot 2017-03-27 at 4 59 35 pm

I tried attaching the Cloud SDK process but that caused some interference with the console output.  I guess Platform/Debug tries piping the `IProcess` output into the _Console_ page and so our own reader misses half of what's returned.
 
Fixes #942


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -369,24 +377,28 @@ public void launch(ILaunchConfiguration configuration, String mode, final ILaunc
     console.clearConsole();
     console.activate();
 
-    new ServerLaunchMonitor(launch, server).engage();
+    // A launch must have at least one debug target or process, or it otherwise becomes a zombie
```
elharo: or it becomes a zombie.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkDebugTargetPresentation.java
```diff
@@ -0,0 +1,80 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkDebugTargetPresentation.java
```diff
@@ -0,0 +1,80 @@
...
+import org.eclipse.debug.ui.IValueDetailListener;
+import org.eclipse.jface.viewers.LabelProvider;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.ui.IEditorInput;
+
+/**
+ * A Platform/Debug Presentation to provide an icon for our
+ * {@link LocalAppEngineServerLaunchConfigurationDelegate.CloudSdkDebugTarget} instances.
+ */
+public class CloudSdkDebugTargetPresentation extends LabelProvider
```
elharo: This class looks testable.

briandealwis:
PTAL @elharo

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/CloudSdkDebugTargetPresentation.java
```diff
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2016 Google Inc.
```
elharo: 2017

PR #1670, opened 2017-03-27
Check App Engine application in a background job

### LINKED ISSUE #1442: Do API calls in separate thread to keep the UI responsive

chanseokoh:
Closes #1442, fixing the second part that queries existence of an App Engine application of a project.

@akerekes The fundamental mechanism is same to that of #1653.


chanseokoh:
Thinking about it, I should abandon the update job too if user changes accounts.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppEngineApplicationQueryJob.java
```diff
@@ -0,0 +1,96 @@
...
+  private final ProjectSelector projectSelector;
+  private final Predicate<Job> isLatestAppQueryJob;
+  private final String createAppLink;
+  private final Display display;
+
+  /**
+   * @param projectRepository {@link ProjectRepository#getAppEngineApplication} must be thread-safe
+   * @param isLatestQueryJob predicate that lazily determines if this job is the latest query job,
+   *     which determines if the job should update {@link ProjectSelector} or die silently. This
+   *     predicate is executed in the UI context
```
elharo: nit: period at end

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/model/GcpProject.java
```diff
@@ -70,19 +70,19 @@ public boolean equals(Object obj) {
    * @return an AppEngine object if it has been retrieved from the backend or {@code null} if it has
    * not yet been retrieved
    */
-  public AppEngine getAppEngine() {
+  public synchronized AppEngine getAppEngine() {
```
elharo: assuming only a single thread sets the value of the appEngine field, we should just mark it `volatile` instead of using `synchronized`. See:

https://stackoverflow.com/questions/9749746/what-is-the-difference-between-atomic-volatile-synchronized/37395582#37395582

If multiple threads may write this, then we should use `AtomicReference` instead.

chanseokoh: Good idea. Multiple concurrent threads may set this field, so I'd go with `AtomicReference`.

Issue #1664, opened 2017-03-26
Transitive Library Dependencies


elharo:
We need to handle the case where Library A -> Library B -> Library C. This is going to become an issue with more Google Cloud Client libraries. Right now we don't go more than one level deep.

```
      <libraryDependency id="googlecloudcore" />
```

elharo:
we now handle one level deep; will go deeper when needed

PR #1672, opened 2017-03-28
Support GCS Client Libraries

elharo:
@akerekes

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -100,7 +100,7 @@
     <library
           id="googleapiclient"
           group="clientapis"
-          name="Google API Client Library for Java"
+          name="%googleapiclient.name"
           tooltip="%googleapiclient.tooltip"
           siteUri="https://developers.google.com/api-client-library/java/" >
       <libraryFile
@@ -129,7 +129,7 @@
     <library
           id="googlecloudcore"
           group="clientapis"
-          name="Google Cloud Core Client Library for Java"
+          name="%googlecloudcore.name"
           tooltip="%googlecloudcore.tooltip"
           siteUri="https://googlecloudplatform.github.io/google-cloud-java/" >
       <libraryDependency id="googleapiclient" />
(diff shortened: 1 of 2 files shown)


chanseokoh:
This has no UI change (i.e., no GCS library in the wizards and _Add Libraries..._. Is this intended?

elharo:
Yes, the UI appears in Eclipse Add Libraries > Google API Libraries:

image

chanseokoh:
Ah, it's under a different menu item.

I see checking the Core library automatically checks the Google API library. However, if the Core library is automatically checked by the Storage library, the Google API isn't automatically checked:

selection_001

elharo:
Yes, that's #1664 and we need to get this in in order to work effectively on that.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -125,6 +125,44 @@
...
+              artifactId="joda-time"
+              groupId="joda-time"
+              version="2.9.2" />
+      </libraryFile>
+    </library>
+    
+    <library
+          id="googlecloudstorage"
+          group="clientapis"
+          name="Google Cloud Storage Client Library for Java"
```
briandealwis: Shouldn't `name` be externalized too?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -125,6 +125,44 @@
               version="1.22.0" />
       </libraryFile>
     </library>
+    
+    <library
+          id="googlecloudcore"
+          group="clientapis"
+          name="Google Cloud Core Client Library for Java"
```
briandealwis: Shouldn't `name` be externalized too?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
elharo: I think it's a proper name. Do we localize "Google Cloud Platform" for example? Let me check with the relevant team.

PR #1677, opened 2017-03-29
resolve library dependencies one level deeper

elharo:
@chanseokoh @akerekes 
This is enough to make GCS work without risking infinite recursion.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -83,6 +84,29 @@ public static Library getLibrary(String id) {
...
+    
+    return map;
+  }
+
+  // Only goes one level deeper, which is all we need for now.
+  // Does not recurse.
+  private static void resolveTransitiveDependencies(ImmutableMap<String, Library> map) {
+    for (Library library : map.values()) {
+      List<String> directDependencies = library.getLibraryDependencies();
+      List<String> transitiveDependencies = Lists.newArrayList(directDependencies);
```
chanseokoh: I avoid using these kinds of collection builders in Guava since they are not recommended anymore: https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Lists.html#newArrayList--

elharo: transitiveDependencies is mutable so that note in the Javadoc doesn't apply.

elharo:
decided not to do recursion until we need it

PR #1688, opened 2017-03-29
Ensure no overlap between source paths and WEB-INF location

### LINKED ISSUE #1647: Converted project doesn't have correct srcpath
image

The first project was created by our wizard.
The second project in this screenshot, test, was converted from a basic Java Project. Note how main.webapp is treated as a Java package.

briandealwis:
Update `AppEngineStandardFacet` install-facet to use `WebContent` if the desired `src/main/webapp` default overlaps with the project's source folders (e.g., `src`).  `WebContent` is the default location used by the WTP _Dynamic Web Facet_.

Fixes #1647


briandealwis:
@nbashirbello might need to do something similar for flex.

briandealwis:
Weird, I thought I was working against `master`, but it seems not.

chanseokoh:
I see many occurrences of `src/main/webapp` when I do a `git grep`, so I am a bit concered. Will we be okay with this change?

briandealwis:
If those locations were using `src/main/java` and `src/test/java` then this has no change.  If not, then we were in error anyways.

PR #1690, opened 2017-03-29
Enable deploying project config files along with app

### LINKED ISSUE #1676: queue.xml not included in "Deploy to App engine standard"
When we deploy a project from eclipse using "Deploy to app engine standard" only app.yaml gets deployed, queue.xml & cron.xml are not deployed. How can I add them?

chanseokoh:
Fixes #1676. Adds a checkbox to the deploy dialog that determines if project-level configurations should be "deployed" (updated) as well.

If users place `cron.xml`, `datastore-indexes.xml`, `dispatch.xml`, `dos.xml`, and/or `queue.xml` under `WEB-INF`, those files will be converted to corresponding YAML files. (`datastore-indexes.xml` converts to to `index.yaml`.) The files will be deployed together as can be seen in the `gcloud` output below, if the checkbox is selected.

```
You are about to update the following configurations:
 - <project ID>/cron  (from [<...omitted...>/tmp/1490822792021/staging/WEB-INF/appengine-generated/cron.yaml])
 - <project ID>/dos  (from [<...omitted...>/tmp/1490822792021/staging/WEB-INF/appengine-generated/dos.yaml])
 - <project ID>/queue  (from [<...omitted...>/tmp/1490822792021/staging/WEB-INF/appengine-generated/queue.yaml])
 - <project ID>/dispatch  (from [<...omitted...>/tmp/1490822792021//staging/WEB-INF/appengine-generated/dispatch.yaml])
 - <project ID>/index  (from [<...omitted...>/tmp/1490822792021/staging/WEB-INF/appengine-generated/index.yaml])
Beginning deployment of service [default]...
File upload done.
Updating service [default]...
```

A few notes:
- I used the name `ConfigDeploy` throughout the code, thinking it would eliminate any possible confusion with 'DeployConfig(uration)' (i.e., a configuration for deploy).
- For now, the default is not to deploy the config files, but we can change that, if that's desired. We can even move the checkbox into the _Avanced_ section.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployerTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+  @Before
+  public void setUp() throws IOException {
+    stagingDirectory = new Path(tempFolder.getRoot().toString());
+    tempFolder.newFile("app.yaml");
+  }
+
+  @Test
+  public void testConfigFilesFile() {
+    assertArrayEquals(
```
elharo: does order matter here? if not, individual asserts on contains would be less brittle

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployerTest.java
```diff
@@ -0,0 +1,106 @@
...
+    assertEquals(1, deployables.size());
+    assertTrue(deployables.contains(stagingDirectory.append("app.yaml").toFile()));
+  }
+
+  @Test
+  public void testComputeDeployables_noConfigFilesAndConfigDeploy() {
+    List<File> deployables =
+        AppEngineProjectDeployer.computeDeployables(stagingDirectory, true /* configDeploy */);
+    assertEquals(1, deployables.size());
+    assertTrue(deployables.contains(stagingDirectory.append("app.yaml").toFile()));
```
elharo: contains --> get(0) since there's only one

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployerTest.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Test
+  public void testComputeDeployables_configFilesExistAndNoConfigDeploy() throws IOException {
+    createFakeConfigFiles();
+
+    List<File> deployables =
+        AppEngineProjectDeployer.computeDeployables(stagingDirectory, false /* configDeploy */);
+    assertEquals(1, deployables.size());
+    assertTrue(deployables.contains(stagingDirectory.append("app.yaml").toFile()));
```
elharo: contains --> get(0) since there's only one

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -3,6 +3,7 @@ google.preferences=Google Cloud Platform-related settings
 
 # Standard deploy settings
 auto.promote=Promote the deployed version to receive all traffic
+deploy.config.files=Deploy config files together
```
elharo: "config files" is not really terminology that is used for these. We need a better term. Let me think about that.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineProjectDeployer.java
```diff
@@ -35,21 +40,43 @@
  */
 public class AppEngineProjectDeployer {
 
+  @VisibleForTesting
+  static final List<String> APP_ENGINE_CONFIG_FILES = Collections.unmodifiableList(Arrays.asList(
```
elharo: Why does this not include app.yaml? Why do we always deploy that and not deploy these? Until we enable one-by-one deploy, I think we ought to just deploy everything, no checkbox.

chanseokoh: My understanding (which might be wrong) is that apps and these configs are different in that these are GCP project-level or service-level configurations, which will affect all apps. (Not sure if flex has a different perspective.) I could imagine that people may want to iterate over apps with frequent deploys but may not want to touch project-wide settings.

But @akerekes was also suggesting to always deploy these files anyway, so I think I'll remove the checkbox.

elharo: Actually, that makes sense. Let me dig a bit deeper.

chanseokoh:
So the current checkbox label and its tooltip are like this:

selection_005

briandealwis:
Haven't yet looked at the code, but the checkbox wording could be improved with:

   _Deploy all App Engine configuration files_

Should this option only be allowed for the _default_ service?  Or should we have an additional prompt that cautions that any files found will over-ride current settings?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModel.java
```diff
@@ -29,6 +29,7 @@
   private String version;
   private boolean autoPromote;
   private boolean stopPreviousVersion;
+  private boolean configDeploy;
```
briandealwis: I think _deployConfigurationFiles_ would make more sense?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModel.java
```diff
@@ -99,6 +102,14 @@ public void setStopPreviousVersion(boolean stopPreviousVersion) {
     this.stopPreviousVersion = stopPreviousVersion;
   }
 
+  public boolean isConfigDeploy() {
```
briandealwis: `shouldDeployConfigurationFiles()`?  (and should you choose to accept this mission, perhaps rename `isAutoPromote()` and `isStopPreviousVersion()`?)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
```diff
@@ -141,6 +143,8 @@ public StandardDeployPreferencesPanel(Composite parent, IProject project,
 
     createPromoteSection();
 
+    createConfigDeploySection();
```
briandealwis: Is it worth doing this only if there are some configuration files?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
briandealwis: They're listed as _Configuration Files_ in the [docs](https://cloud.google.com/appengine/docs/standard/java/configuration-files).

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesModel.java
chanseokoh: Haven't tested, but for the data binding POJO thing to work, I think it's intentional to use this convention `isSomeValue()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/StandardDeployPreferencesPanel.java
chanseokoh: We will eventually reach the point where we'll discuss how the UI & UX would be for deploying these config files. Let's keep the idea in mind.

elharo:
I don't believe we need an extra prompt. The path of least surprise is that the new settings are used and override previous settings.

chanseokoh:
Unchecked by default. Shall we change that?

PR #1719, opened 2017-04-03
don't create two app.yaml's

### LINKED ISSUE #1682: Change the location of app.yaml to the default location when creating flex project

### LINKED ISSUE #1717: Two app.yamls
The flex new project wizard generates two app.yamls. There should only be one. 

image

elharo:
@chanseokoh fix #1717 fix #1682

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
@@ -123,8 +123,6 @@ public class CodeTemplates {
     if (isStandardProject) {
       createChildFile("appengine-web.xml", AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
           webinf, subMonitor.newChild(5), properties);
-    } else {
-      copyChildFile("app.yaml", webinf, subMonitor.newChild(5));
     }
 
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -121,16 +121,18 @@ private static IFile materialize(IProject project, AppEngineProjectConfig config
...
 
     if (isStandardProject) {
-      createChildFile("appengine-web.xml", AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
+      createChildFile("appengine-web.xml",  //$NON-NLS-1$
+          AppEngineTemplateUtility.APPENGINE_WEB_XML_TEMPLATE,
           webinf, subMonitor.newChild(5), properties);
     } else {
-      copyChildFile("app.yaml", webinf, subMonitor.newChild(5));
+      IFolder appengine = createChildFolder("appengine", main, subMonitor.newChild(5)); //$NON-NLS-1$
+      copyChildFile("app.yaml", appengine, subMonitor.newChild(5));
```
chanseokoh: Seems like we don't even need this. Someone else is creating this file already. Need to find who and where.

chanseokoh: Ah, seems like it's the facet install delegate, just like the standard one.

PR #1724, opened 2017-04-03
Minor code clean-up

chanseokoh:
Switched the argument order so that `Monitor` is the last argument.

Others are line-breaks, etc.


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -159,9 +160,9 @@ private static IFolder createFoldersForPackage(IFolder parentFolder,
   }
 
   @VisibleForTesting
-  static IFolder createChildFolder(String name, IFolder parent, SubMonitor monitor) 
+  static IFolder createChildFolder(String name, IFolder parent, SubMonitor monitor)
       throws CoreException {
-    monitor.subTask("Creating folder " + name);
+    monitor.subTask("Creating folder " + name); //$NON-NLS-1$
```
elharo: This should be NLS

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -171,12 +172,12 @@ static IFolder createChildFolder(String name, IFolder parent, SubMonitor monitor
...
-  
+
   @VisibleForTesting
-  static IFile createChildFile(String name, String template, IContainer parent, SubMonitor monitor,
-      Map<String, String> values) throws CoreException {
+  static IFile createChildFile(String name, String template, IContainer parent,
+      Map<String, String> values, SubMonitor monitor) throws CoreException {
 
-    monitor.subTask("Creating file " + name);
+    monitor.subTask("Creating file " + name); //$NON-NLS-1$
```
elharo: this too should be localized as these messages are user viisble. You don't have to do it now, but please don't suppress the warning

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -191,7 +192,7 @@ static IFile createChildFile(String name, String template, IContainer parent, Su
   @VisibleForTesting
   static void copyChildFile(String name, IContainer parent, SubMonitor monitor)
       throws CoreException {
-    monitor.subTask("Copying file " + name);
+    monitor.subTask("Copying file " + name); //$NON-NLS-1$
```
elharo: ditto

PR #1725, opened 2017-04-04
Use javax.servlet-api 3.1 for new flex project

### LINKED ISSUE #1674: Change the servlet api version for the sample Flex projects and fix at version 3.1
javax.servlet:javax.servlet-api

chanseokoh:
Fixes #1674.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
```diff
@@ -7,13 +7,15 @@ Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.newproject.test
 Bundle-Vendor: Google Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.newproject
-Require-Bundle: org.junit;bundle-version="4.12.0",
+Require-Bundle: com.google.cloud.tools.eclipse.appengine.flex,
```
elharo: Import-Package is preferred these days for OSGI reasons. Could we use that instead?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProjectTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  @Rule public TemporaryFolder tempFolder = new TemporaryFolder();
+
+  private final IProgressMonitor monitor = new NullProgressMonitor();
+  private final AppEngineProjectConfig config = new AppEngineProjectConfig();
+  private IProject project;
+
+  @Before
+  public void setUp() {
+    IWorkspace workspace = ResourcesPlugin.getWorkspace();
```
elharo: or use @Rule TestProjectCreator?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
```diff
@@ -122,27 +121,26 @@ private void addDependenciesToProject(IProject project, IProgressMonitor monitor
...
         IFile destFile = libFolder.getFile(artifactFile.getName());
         destFile.create(new FileInputStream(artifactFile), true, subMonitor.newChild(30));
       } catch (CoreException ex) {
-        logger.log(Level.WARNING, "Error downloading " +
-            libraryFile.getMavenCoordinates().toString() + " from maven", ex);
+        logger.log(Level.WARNING, "Error downloading " + //$NON-NLS-1$
+            libraryFile.getMavenCoordinates().toString() + " from maven", ex); //$NON-NLS-1$
       } catch (FileNotFoundException ex) {
-        logger.log(Level.WARNING, "Error copying over " + artifactFile.toString() + " to " +
+        logger.log(Level.WARNING, "Error copying over " + artifactFile.toString() + " to " + //$NON-NLS-1$ //$NON-NLS-2$
```
elharo: can artifactFile be null here?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProjectTest.java
chanseokoh: I consulted the Standard counterpart, and it wasn't using `TestProjectCreator`. I wondered that too at first, but the reason may be that here we are in the process of creating a new project from scratch, while `TestProjectCreator` provides a fully working project.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
chanseokoh: This is the same problem with https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1660#discussion_r108299807. This test attempts to install the flex facet, which requires the flex facet definition. However, the `.flex` bundle isn't in our target platform yet; this directive will pull in the bundle in testing.

When we eventually remove the `.flex` bundle, Eclipse will place an error marker in this file, so we won't forget removing this later.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
chanseokoh: The exception is from `new FileInputStream(artifactFile)`, so at that point `artifactFile` is not null. Also, `artifact.getFile()` never returns `null`, so we are good.

PR #1739, opened 2017-04-04
switch to devappserver1

elharo:
@chanseokoh


elharo:
stopping the server may be a problem

chanseokoh:
Seems like it works! There is a minor problem, but I think we are generally OK. For example, it still assumes that its admin port is 8000, although dev server1 doesn't have a separate port. So, we'd still see messages like below:

```
Conflicts with running server "App Engine Standard at localhost"
admin port: 8000
storage path: <default location>
```

Also, the storage path might not matter anymore, since my guess is that dev server1 creates a local datastore inside the staging directory, whereas previously, dev server2 was creating it under `/tmp` which was a contention point.

elharo:
admin port doesn't seem to make a difference. Arguably we shouldn't need to read the admin port at all. appengine-plugins-core should know what it is in each server.

chanseokoh:
- Okay, the first thing: for setting a correct "admin port" (which is just the same service port in dev appserver1), I think the right way is

```
        if (serverPort == 0) {
          serverPort = serverPortCandidate;
          // TODO: remove below when we switch back to dev appserver2.
          adminPort = serverPort;  // they are same in dev appserver1.
        }
```

at [this location](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java#L421). (If we did not plan to switch back to dev appserver2, I would have just removed the admin port field altogether.)

- But then, I noticed that we are failing to extract the service port for dev appserver1, since [our pattern matching on server output is only for dev appserver2](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java#L386).

chanseokoh:
> But then, I noticed that we are failing to extract the service port for dev appserver1, since our pattern matching on server output is only for dev appserver2.

Ah, wait. The browser was working fine, so this should already be working properly. Let me check this again.

chanseokoh:
> > But then, I noticed that we are failing to extract the service port for dev appserver1, since our pattern matching on server output is only for dev appserver2.
>
> Ah, wait. The browser was working fine, so this should already be working properly. Let me check this again.

Certainly the pattern matching should be fixed. For example, if you set the server port to 0 when doing _Run As_ > _Run on Server_ (i.e., let the server pick a random port), things will go broken. The browser was working fine when the port was different than 0, because in this case there was no port change in the launch configuration.

chanseokoh:
Finally, `appengine-plugins-core` needs some fix. Currently, stopping a server doesn't work if the server runs on a port other than 8080.

https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/370

elharo:
Re: storage path, we really haven't tested data store with either devappserver1 or devappserver2 and we really should.

elharo:
FYI, this is blocked until appengine-plugins-core 0.3.0 goes out. There are some NullPointerExceptions that won't be fixed until then.

elharo:
@chanseokoh Port 0 seems to work for me on local run. Do you still have a concern about that or do recent changes fix that?

chanseokoh:
You'll notice that you can't stop a server if run on port, e.g., 8085 or 0. I think I know why. We should set `.setAdminPort()` correctly in our `stop()` method.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -374,19 +385,27 @@ public void onStart(Process process) {
...
+    // devappserver2 patterns
+    private final Pattern moduleStartedPattern = Pattern.compile(
         "INFO .*Starting module \"(?<service>[^\"]+)\" running at: (?<url>http://.+:(?<port>[0-9]+))$");
-    private Pattern adminStartedPattern =
+    private final Pattern adminStartedPattern =
         Pattern.compile("INFO .*Starting admin server at: (?<url>http://.+:(?<port>[0-9]+))$");
 
+    // devappserver1 patterns
+    private final Pattern moduleRunningPattern = Pattern.compile(
+        "INFO: Module instance (?<service>\\w+) is running at (?<url>http://.+:(?<port>[0-9]+)/)$");
```
chanseokoh: I see `\w` is `[a-zA-Z_0-9]`, which does not include a hyphen. Not sure, but can a service name contain hyphens? Probably just safe to do `[^ ]+`. (Not tested.)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -404,20 +423,24 @@ public void onOutputLine(String line) {
...
+        int port = parseInt(portString, 0);
         if (port > 0 && (serverPortCandidate == 0 || "default".equals(serviceId))) { // $NON-NLS-1$
           serverPortCandidate = port;
         }
-      } else if ((matcher = adminStartedPattern.matcher(line)).matches()) {
+      } else if ((matcher = adminStartedPattern.matcher(line)).matches()
+          || (matcher = adminRunningPattern.matcher(line)).matches()) {
         int port = parseInt(matcher.group("port"), 0);
-        if (port > 0 && adminPort == 0) {
+        if (port > 0 && adminPort <= 0) {
```
chanseokoh: I think there was a reason we used `==` here, which I can't remember exactly. Maybe it was a conceptual distinction: -1 meaning never initialized, and 0 meaning we let the server pick a port. Maybe not a big deal, so I'm OK as long as it's working.

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviourTest.java
```diff
@@ -100,7 +101,19 @@ public void testCheckPort_portOutOfBounds_positive() throws CoreException {
...
-  private static final String[] serverOutputWithDefaultModule1 = new String[] {
+  private static final String[] devappserver1Output = new String[] {
+      "Apr 05, 2017 9:25:17 PM com.google.apphosting.utils.jetty.JettyLogger info",
+      "INFO: jetty-6.1.x",
+      "Apr 05, 2017 9:25:17 PM com.google.apphosting.utils.jetty.JettyLogger info",
+      "INFO: Started SelectChannelConnector@localhost:8080",
+      "Apr 05, 2017 9:25:17 PM com.google.appengine.tools.development.AbstractModule startup",
+      "INFO: Module instance default is running at http://localhost:8080/",
+      "Apr 05, 2017 9:25:17 PM com.google.appengine.tools.development.AbstractModule startup",
+      "INFO: The admin console is running at http://localhost:8080/_ah/admin",
```
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviourTest.java
```diff
@@ -127,18 +140,26 @@ public void testCheckPort_portOutOfBounds_positive() throws CoreException {
       "INFO     2016-11-03 21:11:21,959 admin_server.py:116] Starting admin server at: http://localhost:43679",
       "Nov 03, 2016 9:11:23 PM com.google.appengine.tools.development.SystemPropertiesManager setSystemProperties"
   };
+  
+  @Test
+  public void testExtractServerPortFromOutput_devappserver1() throws CoreException {
+    setUpServerPort(0);
+    simulateOutputParsing(devappserver1Output);
+    assertEquals(8080, serverBehavior.getServerPort());
+    assertEquals("foo", 8080, serverBehavior.getAdminPort());
```
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -345,14 +349,21 @@ private void initializeDevServer(MessageConsoleStream console) {
...
 
-    devServer = new CloudSdkAppEngineDevServer(cloudSdk);
+    devServer = new CloudSdkAppEngineDevServer1(cloudSdk);
     moduleToUrlMap.clear();
   }
+  
+  /**
+   * @return true if and only if we're using devappserver1
+   */
+  public boolean isDevAppServer1() {
```
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: I've expanded it. We certainly need to capture all legal service names here. I'm not sure to what extent we should exclude non-legal service names if they're output.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
elharo: I definitely want to set this if it's -1 at this point.

chanseokoh: Makes sense. LGTM.

PR #1742, opened 2017-04-05
Validate appengine-web.xml against schema

### LINKED ISSUE #1432: check appengine-web.xml for invalid tags
Based on https://cloud.google.com/appengine/docs/java/config/appref we could check whether the appengine-web.xml contains an invalid (mistyped?) element and give an error similar to how we give a warning about the `<application>` tag

abbiefarr:
Fix #1432 

Having trouble getting the error handler in XmlValidator#xsdValidation to report more than one error.

For example, if I have `<root>
    <illegal-tag1/>
    <illegal-tag2/>
</root>` it will report the first tag but then the validation stops.

@elharo @meltsufin

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidatorTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidatorTest.java
@@ -147,7 +147,7 @@ public class XmlValidatorTest {
   public void testXsdValidation() throws CoreException {
     String xml = "<?xml version='1.0' encoding='utf-8'?>"
         + "<appengine-web-app xmlns='http://appengine.google.com/ns/1.0'>"
-        + "<foo></foo><bar></bar>"
+        + "<foo></foo>"
         + "</appengine-web-app>";
     XmlValidator validator = new XmlValidator();
     validator.setHelper(new AppEngineWebXmlValidator());
@@ -158,7 +158,7 @@ public class XmlValidatorTest {
     validator.xsdValidation(file);
     String problemMarker = "org.eclipse.core.resources.problemmarker";
     IMarker[] markers = file.findMarkers(problemMarker, true, IResource.DEPTH_ZERO);
-    assertEquals(2, markers.length);
+    assertEquals(1, markers.length);
   }
   
   @Test
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidatorTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidatorTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.eclipse.appengine.validation;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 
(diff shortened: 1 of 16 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidator.java
```diff
@@ -89,7 +101,27 @@ void validate(IFile resource, byte[] bytes)
...
+      createSaxErrorMessage(resource, (SAXParseException) ex.getException());
+    }
+  }
+  
+  void xsdValidation(IResource resource) throws CoreException {
+    try {
+      String xsd = helper.getXsd();
+      if (xsd != null) {
+        URL xsdPath = AppEngineWebXmlValidator.class.getResource(xsd);
+        InputStream stream = xsdPath.openStream();
```
chanseokoh: You need to close the stream (preferable with the try-resources block), unless it is guaranteed that, e.g., `StreamSource` will always close it. Opening an input stream will consume a file descriptor in this case, and on systems where the number of open files are very limited (e.g., http://stackoverflow.com/questions/34588/how-do-i-change-the-number-of-open-files-limit-in-linux), OS may refuse to create any new file descriptor once it hits the limit.

Of course, the stream will be closed when it is garbage-collected, but you never know when the garbage collection will kick in. (It's basically the same principle that you shouldn't use `finalize()` almost all the time.)

I know we are a bit lenient about closing streams in tests, but that's because it is a short-lived process that will die soon in an isolated domain, and it s extremely unlikely that it will hit a limit. But for production applications, it's desirable to reclaim resources as soon as we can.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidator.java
```diff
@@ -89,7 +101,27 @@ void validate(IFile resource, byte[] bytes)
         }
       }
     } catch (SAXException ex) {
-      createSaxErrorMessage(resource, ex);
+      createSaxErrorMessage(resource, (SAXParseException) ex.getException());
```
elharo: Safer to catch SAXParseException separately rather than casting here

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/SaxErrorHandler.java
```diff
@@ -0,0 +1,66 @@
...
+  
+  @Override
+  public void warning(SAXParseException exception) throws SAXException {
+    // Do nothing
+  }
+
+  @Override
+  public void error(SAXParseException exception) throws SAXException {
+    try {  
+      XmlValidator.createSaxErrorMessage(resource, exception);
```
elharo: As long as this doesn't throw, parsing should continue. If it doesn't, maybe try the pure Xerces/SAX validation instead?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidator.java
abbiefarr: Will it catch a SAXParseException if it's wrapped in a SAXException?

elharo: No, but that's not likely to occur. SAXParseException is a subclass of SAXParseException.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidatorTest.java
```diff
@@ -131,6 +144,24 @@ public void testValidate_withBannedElements() throws IOException, CoreException
   }
   
   @Test
+  public void testXsdValidation() throws CoreException {
+    String xml = "<?xml version='1.0' encoding='utf-8'?>"
```
elharo: You don't really need the XML declaration in inline test data. It doesn't hurt but you can take it out and make the code a bit smaller.

PR #1762, opened 2017-04-07
Refactoring and code simplification

chanseokoh:
One notable change is that I removed the `credentialFile` clean-up in the `finally` block. Anything (including the credential file) in the `workDirectory` will be cleaned up by [`CleanupOldDeploysJob`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.0.2/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/CleanupOldDeploysJobTest.java), so there is no point in deleting only the credential file here.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
@@ -16,27 +16,40 @@
 
 package com.google.cloud.tools.eclipse.login;
 
-import org.junit.Assert;
-import org.junit.Test;
-
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.http.javanet.NetHttpTransport;
 import com.google.api.client.json.jackson.JacksonFactory;
 import com.google.gson.Gson;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.file.Path;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 
 public class CredentialHelperTest {
 
-  @Test
-  public void testGetJsonCredential() {
-    Credential credential = createCredential("fake_access_token", "fake_refresh_token");
-    String jsonCredential = new CredentialHelper().toJson(credential);
+  @Rule public TemporaryFolder tempFolder = new TemporaryFolder();
 
-    CredentialType credentialType = new Gson().fromJson(jsonCredential, CredentialType.class);
-    Assert.assertEquals(credentialType.client_id, Constants.getOAuthClientId());
-    Assert.assertEquals(credentialType.client_secret, Constants.getOAuthClientSecret());
-    Assert.assertEquals(credentialType.refresh_token, "fake_refresh_token");
-    Assert.assertEquals(credentialType.type, "authorized_user");
+  @Test
+  public void testToJsonFile() throws IOException {
+    Credential credential = createCredential("fake_access_token", "fake_refresh_token");
+    Path jsonFile = tempFolder.getRoot().toPath().resolve("credential-for-gcloud.json");
+    CredentialHelper.toJsonFile(credential, jsonFile);
+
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
```diff
@@ -16,27 +16,40 @@
...
 public class CredentialHelperTest {
 
+  @Rule public TemporaryFolder tempFolder = new TemporaryFolder();
+
   @Test
-  public void testGetJsonCredential() {
+  public void testToJsonFile() throws IOException {
     Credential credential = createCredential("fake_access_token", "fake_refresh_token");
-    String jsonCredential = new CredentialHelper().toJson(credential);
+    Path jsonFile = tempFolder.getRoot().toPath().resolve("credential-for-gcloud.json");
```
akerekes: `tempFolder.newFile("credential-for-gcloud.json").toPath()` looks a bit simpler, but this is also fine

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
```diff
@@ -16,27 +16,40 @@
...
 
-    CredentialType credentialType = new Gson().fromJson(jsonCredential, CredentialType.class);
-    Assert.assertEquals(credentialType.client_id, Constants.getOAuthClientId());
-    Assert.assertEquals(credentialType.client_secret, Constants.getOAuthClientSecret());
-    Assert.assertEquals(credentialType.refresh_token, "fake_refresh_token");
-    Assert.assertEquals(credentialType.type, "authorized_user");
+    try (InputStream in = new FileInputStream(jsonFile.toFile());
+        Reader reader = new InputStreamReader(in)) {
+      CredentialType credentialType = new Gson().fromJson(reader, CredentialType.class);
+      Assert.assertEquals(credentialType.client_id, Constants.getOAuthClientId());
```
akerekes: I think these `assertEquals` have their parameters reversed, should be (expected, actual).

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
chanseokoh: I verified that `newFile()` actually creates an empty file, so I'll go with `resolve()`.

akerekes: And what's the problem with that?

chanseokoh: It will then test if `toJsonFile()` actually creates a fresh new file. Using `newFile()` in the test may give the false impression that `toJsonFile()` assumes that a file must exist beforehand.

akerekes: I'm nitpicking here, but then this way it gives the false impression that the file may not exist prior to writing the credential.

chanseokoh: Yeah, that's exactly the point of this testing. Maybe you're confused. The file should not exist prior to writing the credential in the test. `toJsonFile()` should create and write a new file designated by the path `jsonFile` from a `Credential` Java object.

akerekes: But the default options of Files.write() will ensure that if the file exists, it will be truncated before writing. So in fact it doesn't matter whether it exists or not, it will be either overwritten or created.

chanseokoh: Then you're talking about a different test, e.g., `testToJsonFile_overwritesExistingFile()`. The current test is deliberately setting up an environment where there doesn't exist a file. In other words, we're not talking about whether the test will technically succeed or fail. Rather, it's about what we are testing. In this test, I want to test if it will create a new file.

Issue #1557, opened 2017-03-08
Bookshelf sample app fails to start on local server


elharo:
not immediately clear why:

![image](https://cloud.githubusercontent.com/assets/1005544/23715258/0091aa0a-03fa-11e7-8e12-e740ffd5fecf.png)

elharo:
INFO: submitting command: /Users/elharo/google-cloud-sdk/bin/dev_appserver.py /Users/elharo/workspaces/runtime-gcloud-eclipse/.metadata/.plugins/org.eclipse.wst.server.core/tmp0/bookshelf-standard-2 --host=localhost --port=8080 --admin_host=localhost --admin_port=8000 --max_module_instances=1 --automatic_restart=false

chanseokoh:
I was also perplexed by what's going on. After some time, I happened to open the _Console_ view, and I see

```
google.appengine.api.yaml_errors.EventError: Unable to assign value '[YOUR-PROJECT-ID]' to attribute 'application':
Value '[YOUR-PROJECT-ID]' for application does not match expression '^(?:(?:[a-z\d\-]{1,100}\~)?(?:(?!\-)[a-z\d\-\.]{1,100}:)?(?!-)[a-z\d\-]{0,99}[a-z\d])$'
  in "<string>", line 1, column 14:
    application: '[YOUR-PROJECT-ID]'
                 ^
```

Removing the `<application>` element fixes that problem. Running locally works, but the app gives 503. It might be that the app needs further configuration. Not sure.

```
HTTP ERROR 503

Problem accessing /. Reason:

    java.lang.IllegalStateException: Invalid storage type. Check if bookshelf.storageType property is set.
Caused by:

javax.servlet.UnavailableException: java.lang.IllegalStateException: Invalid storage type. Check if bookshelf.storageType property is set.
	at org.mortbay.jetty.servlet.ServletHolder.makeUnavailable(ServletHolder.java:415)
	at org.mortbay.jetty.servlet.ServletHolder.initServlet(ServletHolder.java:458)
	at org.mortbay.jetty.servlet.ServletHolder.doStart(ServletHolder.java:263)
	at org.mortbay.component.AbstractLifeCycle.start(AbstractLifeCycle.java:50)
	at org.mortbay.jetty.servlet.ServletHandler.initialize(ServletHandler.java:685)
	at org.mortbay.jetty.servlet.Context.startContext(Context.java:140)
[...]

PR #1785, opened 2017-04-11
web.xml 3.1 for flex

### LINKED ISSUE #1784: New flex projects has web.xml 2.5
We need to add a `web.xml` 3.1 template and have `CodeTemplate.materialize()` select the right version based on `isStandardProject`.

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.0.3C/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java#L139

chanseokoh:
Fixes #1784.

BTW, I have no idea why our current 2.5 `web.xml` template has `xmlns:web="http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"`. Examples I found on the Internet don't have it for both 2.5 and 3.1.


plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
```diff
@@ -1,9 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
+<#if version == "2.5">
 <web-app xmlns="http://java.sun.com/xml/ns/javaee"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xmlns:web="http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
```
elharo: I think you're right. We can get rid of this line.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
```diff
@@ -1,9 +1,16 @@
...
          xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd" 
          version="2.5">
+<#elseif version == "3.1">
```
elharo: This should either be two templates or we should parameterize the attribute values rather than the entire template.

briandealwis:
I know there's some authoritative site with the `xmlns:web` attribute as I copied it into the fix for #1557! But I didn't record the site.

chanseokoh:
About `xmlns:web`, I found this: https://docs.oracle.com/cd/E14571_01/web.1111/e13712/web_xml.htm#WBAPP503. I feel like I'll restore it.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
```diff
@@ -1,16 +1,15 @@
-<?xml version="1.0" encoding="utf-8"?>
 <#if version == "2.5">
```
elharo: It doesn't feel like you need an if here. Can you pass the values you want used directly?

chanseokoh: Got it. Will do.

elharo:
Please don't restore xmlns:web. It's not needed because we don't use the web:  prefix anywhere in our document. Unless it's likely to be used post-user edits (and I don't think it is) it's just irrelevant detritus.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
```diff
@@ -1,9 +1,15 @@
...
+           schema = "web-app_3_1.xsd">
+</#if>
 <?xml version="1.0" encoding="utf-8"?>
-<web-app xmlns="http://java.sun.com/xml/ns/javaee"
+<web-app xmlns="${namespace}"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xmlns:web="http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
-         xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd" 
-         version="2.5">
+         xsi:schemaLocation="${namespace} ${namespace}/${schema}"
```
elharo: ${namespace}/${schema} should be a single variable, schemaUrl

There's no rule that the namespace URL is part of the schema location.

PR #1790, opened 2017-04-13
Rename and simplify deploy-related classes

chanseokoh:
- `DeployPreferencesPanel` --> `AbstractDeployPreferencesPanel`.
- `StandardPreferencesPanel` --> `CommonPreferencesPanel`: I think the flex deploy could use this class almost as-is after a bit more refactoring.

Some simplifications:
- `DeployPreferencesPanel` was initializing `formToolKit`, but the only user of it was `StandardPreferencesPanel`, so I pushed it up.
- `DeployPreferencesPanel` didn't need to add `DisposeListener`; it's a `Composite`, so it could utilize `dispose()`.


chanseokoh:
Still renaming and moving.

chanseokoh:
With the last three commits, I further renamed and moved preferences-related classes, because all the preferences we use now are common to both standard and flex.

- `.deploy.standard/StandardDeployPreferences` --> `.deploy/DeployPreferences`
- `.deploy.standard/StandardDeployPreferencesInitializer` --> `.deploy/DeployPreferencesInitializer`
- `.deploy.standard/StandardDeployPreferencesConverter` --> `deploy/DeployPreferencesConverter`

Ready for review.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AbstractDeployPreferencesPanel.java
```diff
@@ -17,29 +17,12 @@
...
 import org.eclipse.core.databinding.DataBindingContext;
-import org.eclipse.swt.events.DisposeEvent;
-import org.eclipse.swt.events.DisposeListener;
 import org.eclipse.swt.widgets.Composite;
-import org.eclipse.ui.forms.FormColors;
-import org.eclipse.ui.forms.widgets.FormToolkit;
 
-abstract class DeployPreferencesPanel extends Composite {
-  private FormToolkit formToolkit;
+abstract class AbstractDeployPreferencesPanel extends Composite {
```
elharo: There doesn't seem to be a need to rename this class. It's not necessary to use the word "Abstract" in the name of an abstract class.

chanseokoh: Done. Merging.

PR #1791, opened 2017-04-14
Rename StandardDeployJob --> DeployJob

chanseokoh:
I'll soon make `DeployJob` an abstract class and create two new sub-classes `StandardDeployJob` and `FlexibleDeployJob`.


chanseokoh:
Merge conflicts resolved. PTAL.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
```diff
@@ -85,13 +80,11 @@
...
 
-  public StandardDeployJob(IProject project,
-                           Credential credential,
-                           IPath workDirectory,
-                           ProcessOutputLineListener stagingStdoutLineListener,
-                           ProcessOutputLineListener stderrLineListener,
-                           DefaultDeployConfiguration deployConfiguration,
-                           boolean includeOptionalConfigurationFiles) {
+  public DeployJob(IProject project, Credential credential, IPath workDirectory,
+      ProcessOutputLineListener stagingStdoutLineListener,
```
elharo: Why is this stagingStdoutLineListener but the next argument is only stderrLineListener? I suggest simply stdoutLineListener but if there's a reason these are different, Javadoc might help.

chanseokoh: Done. From the API design perspective, I think it's better to hide any `CloudSdk` related things such as `ProcessOutputLineListener` and `DefaultDeployConfiguration`, but that'd call for another opportunity for refactoring.

PR #1789, opened 2017-04-13
Tycho 1.0

briandealwis:
Update build to use Tycho 1.0.0, Maven 3.5.0 (not actually required), and Cloud SDK 150.0.0.

Tycho 1.0 was recently been released and requires Java 8 [as of 0.25](https://bugs.eclipse.org/bugs/show_bug.cgi?id=485427).  This causes some complications as we previously relied on building with Java 7 on Travis to catch any Java-8'isms.  And `tycho-compiler-plugin`'s support for using the bundle's BREE doesn't work for Neon or later as they require a Java 8 and we hit class resolution problems when compiling with a Java 7 JDK.  So this change does two things:
  1. It configures the the `tycho-compiler-plugin` to only use the BREE compilation when building with a JavaSE-1.7.
  2. It leverages `tycho-surefire-plugin`'s [default configuration of `useJDK=SYSTEM`](https://eclipse.org/tycho/sitedocs/tycho-surefire/tycho-surefire-plugin/test-mojo.html#useJDK) to run the tests using the configured Maven toolchain, which we configure to use a Java 7 runtime for Mars and Java 8 for Neon and Oxygen; I've included an example log below to show how a Java 8 project fails. 
This should catch any accidental Java 8'isms and Java 9'isms!

Tycho 1.0 caught a missing export-package in google apis, and some missing imports and an explicit Tycho version reference in `dataflow`. So I also took the opportunity to remove unnecessary dataflow configuration information that is obtained from the parent.

This change also does the following:
    - uses Maven's `-B` batch-mode to cut down on log messages
    - changes the default `tycho.toolchains` value to SYSTEM, rendering #685 obsolete, since Travis will pick up any issues
    - bumps Google Cloud SDK to 150.0.0

Fixes #1598

Log showing how converting `com.google.cloud.tools.eclipse.appengine.compat` to Java 8 causes a test failure (some lines elided):
```
[INFO] --- tycho-surefire-plugin:1.0.0:test (default-test) @ com.google.cloud.tools.eclipse.appengine.compat.test ---
[INFO] Toolchain in tycho-surefire-plugin: JDK[/usr/lib/jvm/java-7-oracle/jre]
[INFO] Expected eclipse log file: /home/travis/build/GoogleCloudPlatform/google-cloud-eclipse/plugins/com.google.cloud.tools.eclipse.appengine.compat.test/target/work/data/.metadata/.log
[INFO] Command line:
[...]

PR #1800, opened 2017-04-14
Export classes for dataflow

chanseokoh:
For me, I have to export these classes to get rid of any compilation error.


chanseokoh:
Ah, it was just that I haven't merged #1789. I just added "x-friends", as the proxy package is only for the `.dataflow.ui` bundle.

briandealwis:
Should we do the `x-friends` through a general sweep?

chanseokoh:
Maybe viable with a limited scope, but to me, it seems not worth putting much time to it. I don't really care much about `x-friends`.

elharo:
I'm not sure why you seem to need this and I don't.

chanseokoh:
Yeah, I don't need to be bothered. Closing.

Issue #1801, opened 2017-04-14
Possibly flaky test: XmlValidatorTest.testXsdValidation()


chanseokoh:
```
testXsdValidation(com.google.cloud.tools.eclipse.appengine.validation.XmlValidatorTest)  Time elapsed: 4.222 sec  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<2>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.google.cloud.tools.eclipse.appengine.validation.XmlValidatorTest.testXsdValidation(XmlValidatorTest.java:158)
```

chanseokoh:
#1803 to log more debugging messages paid off! Investigate https://travis-ci.org/GoogleCloudPlatform/google-cloud-eclipse/jobs/223288050.

chanseokoh:
Copying the log here just in case the log gets lost:

testXsdValidation(com.google.cloud.tools.eclipse.appengine.validation.XmlValidatorTest)  Time elapsed: 9.626 sec  <<< FAILURE!
[...]

PR #1803, opened 2017-04-14
Flaky test

abbiefarr:
See #1801 

I couldn't find an obvious reason why this test was flaky, but I didn't think it would hurt to have less static test data.

@elharo @meltsufin


chanseokoh:
I wonder if you need to delete `APPLICATION_MARKER` instead of `IMarker.PROBLEM` here:

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.0.3C/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidatorTest.java#L143

abbiefarr:
@chanseokoh deleting problem markers should delete application markers as well

PR #1817, opened 2017-04-17
Refactor DeployCommandHandler

chanseokoh:
Minor code refactoring. Couldn't resist doing this. Grouping code more locally: now the first part of `launchDeployJob()` is to create and show a console, and the next part is to prepare arguments (`workDirectory`, etc.) for `new DeployJob()`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/MessageConsoleUtilities.java
+++ b/plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/MessageConsoleUtilities.java
@@ -95,15 +95,11 @@ public class MessageConsoleUtilities {
     return console;
   }
 
-  public static <C extends MessageConsole> C createConsole(String name, ConsoleFactory<C> factory,
-      boolean showConsole) {
+  public static <C extends MessageConsole> C createConsole(String name, ConsoleFactory<C> factory) {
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/MessageConsoleUtilities.java
```diff
@@ -95,11 +95,15 @@ public static MessageConsole getMessageConsole(String consoleName,
...
 
-  public static <C extends MessageConsole> C createConsole(String name, ConsoleFactory<C> factory) {
+  public static <C extends MessageConsole> C createConsole(String name, ConsoleFactory<C> factory,
+      boolean showConsole) {
     ConsolePlugin plugin = ConsolePlugin.getDefault();
     IConsoleManager manager = plugin.getConsoleManager();
     C console = factory.createConsole(name);
     manager.addConsoles(new IConsole[]{console});
+    if (showConsole) {
+      manager.showConsoleView(console);
```
elharo: showConsole feels like a separate operation that perhaps doesn't belong in this method. Could we move this part back where it was?

chanseokoh: No problem. PTAL.

PR #1821, opened 2017-04-18
Implement Flex deploy execution

chanseokoh:
(This branch is based off of #1817, and it'll become slightly simpler after #1817 is merged.)

Fortunately, the only difference in std vs. flex deploy execution is basically how we stage (i.e., how to put necessary files (such as `app.yaml`) to some directory where we are ready to run `gcloud app deploy`), so we can basically reuse existing code for most other stuff.


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Delegate that takes care of App Engine environment-specific deploy behaviors for {@link
+ * DeployJob}.
+ */
+public interface DeployEnvironmentDelegate {
+
+  /**
+   * @param project Eclipse project to be deployed
+   * @param stagingDirectory directory where implementors methods should place necessary files for
```
elharo: what's "implementors" here?

chanseokoh: Should be "implementing methods".

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
```diff
@@ -31,11 +31,11 @@
 
   /**
    * @param project Eclipse project to be deployed
-   * @param stagingDirectory directory where implementors methods should place necessary files for
-   *     deployment, where this job will execute {@code gcloud app deploy}
-   * @param safeWorkDirectory directory path that implementors methods may create safely to use as
+   * @param stagingDirectory directory where implementing methods should place necessary files for
```
elharo: That's still not normal usage. "methods" is implied here. I think something like "directory where files to be deployed are placed" is enough.

chanseokoh: It may give the impression that files to be deployed are already placed, whereas the method implementation should place them. What do you think of "directory where files to be deployed are to be placed" or "should be placed"?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
```diff
@@ -0,0 +1,45 @@
...
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+
+/**
+ * Delegate that takes care of App Engine environment-specific deploy behaviors for {@link
+ * DeployJob}.
+ */
+public interface DeployEnvironmentDelegate {
```
briandealwis: The use of _Environment_ seems a bit odd.  _DeployStagingDelegate_?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
```diff
@@ -0,0 +1,45 @@
...
+   * @param stagingDirectory directory where implementing methods should place necessary files for
+   *     deployment, where {@link DeployJob} will execute {@code gcloud app deploy}
+   * @param safeWorkDirectory directory path that implementing methods may create safely to use as
+   *     a temporary work directory during staging
+   * @param cloudSdk {@link CloudSdk} that implementing methods may utilize
+   */
+  IStatus stage(IProject project, IPath stagingDirectory,
+      IPath safeWorkDirectory, CloudSdk cloudSdk, IProgressMonitor monitor) throws CoreException;
+
+  IPath getOptionalConfigurationFilesDirectory();
```
briandealwis: This method would benefit from some javadoc as it's not clear what this method is for without reading the implementations.  Given the name, I thought it would have been a configuration *parameter* provided the staging delegate (e.g., _please place any config files found **here**_), but it instead seems to be a location for the delegate to return any config files that were found and rewritten?  It should say that it should only be called after successful stage.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
```diff
@@ -168,17 +169,19 @@ private IStatus saveCredential(Path destination) {
...
     RecordProcessError stagingExitListener = new RecordProcessError();
     CloudSdk cloudSdk = getCloudSdk(credentialFile, stagingStdoutLineListener, stagingExitListener);
 
     try {
-      getJobManager().beginRule(project, progress);
-      WarPublisher.publishExploded(project, explodedWarDirectory, progress.newChild(40));
-      DeployStaging.stageStandard(explodedWarDirectory, stagingDirectory, cloudSdk,
-          progress.newChild(60));
-      return stagingExitListener.getExitStatus();
+      getJobManager().beginRule(project, null /* not worth a monitor */);
```
briandealwis: The monitor is helpful in case the job is paused pending some other job that has a lock on the project.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
```diff
@@ -98,15 +99,17 @@ public DeployJob(IProject project, Credential credential, IPath workDirectory,
       ProcessOutputLineListener stagingStdoutLineListener,
       ProcessOutputLineListener stderrLineListener,
       DefaultDeployConfiguration deployConfiguration,
-      boolean includeOptionalConfigurationFiles) {
-    super(Messages.getString("deploy.standard.runnable.name")); //$NON-NLS-1$
+      boolean includeOptionalConfigurationFiles,
+      DeployEnvironmentDelegate environmentDelegate) {
```
briandealwis: Would it make sense to move some of these arguments as getters and setters (e.g., includeOptionalConfigurationFiles).

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
chanseokoh: Ah, that makes sense.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
chanseokoh: I spent a lot of time on the name actually. It was initially `StageDelegate`, but then the method `getOptionalConfigurationFilesDirectory` became out of place, because the optional configuration files directory has nothing to do with staging for flex.

chanseokoh: By the way, it's called "standard environment" or "flexible environment" officially, so that's why "Environment" there. Name suggestions are welcome.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
chanseokoh: I'll put some Javadoc. FYI,

- Standard: `appcfg.sh` will convert XML config files into YAML files (`queue.xml` --> `queue.yaml`) and place them under `<staging-directory>/WEB-INF/appengine-generated`.
- Flexible: has nothing to do with staging, and doesn't have to be called after `stage()`. It will return a path in a source repo, for example, "src/main/appengine".

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
chanseokoh: I've been thinking about refactoring too: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1824

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployEnvironmentDelegate.java
chanseokoh: @briandealwis suggests `StagingDelegate`, and I think it'll work.

chanseokoh:
I renamed `DeployEnvironmentDelegate` to `StagingDelegate`. I also renamed `DeployStaging` to `CloudSdkStagingHelper` that is technically independent of this PR, as it becomes confusing with similar names. I actually think `CloudSdkStagingHelper` describes the class more accurately.

PR #1822, opened 2017-04-18
Minor code formatting

chanseokoh:
I find this formatting more pleasant. Agreed?


plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/AppEngineDeployOutputTest.java
```diff
@@ -20,78 +20,67 @@
...
+        "      \"traffic_split\": null,\n" +
+        "      \"version\": null\n" +
+        "    }\n" +
+        "  ]\n" +
+        "}\n";
 
-    try {
-      AppEngineDeployOutput.parse(jsonOutput);
-      Assert.fail("Failure to throw exception when parsing deploy output with more that one version entry");
-    } catch (JsonParseException e) {
```
elharo: In these cases it can be useful to assert something about the exception such as the message

chanseokoh: I recently found a different way to verify exceptions (there are many ways indeed), so I just tried it.

PR #1838, opened 2017-04-20
avoid internal class

elharo:
@briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
```diff
@@ -134,10 +135,15 @@ protected void launch(IServer server, String launchMode, SubMonitor progress)
     // IServer#start() that prompts the user *as the server continues to launch*.
     // ServerUIPlugin.saveEditors() respects the "Save editors before starting the server"
```
chanseokoh: Comment needs to be update together.

briandealwis:
I lean to -1 to this change: `ServerUIPlugin#saveEditors()` also respects the _Servers > Launching > Save dirty editors_ preference.  What if we propose to WTP to expose `saveEditors()` via `ServerUICore` or `ServerUIUtil`?

screen shot 2017-04-20 at 9 44 08 am

Issue #1830, opened 2017-04-19
service name off by one when reading name out of wizard


elharo:
E.g. try to deploy a service "silver" and what's actually deployed is "silve".

Deploy "123" and what's deployed is "123" and so forth. 

Possibly only in projects created with standard wizard. Problem happens before deployment in service element written to appengine-web.xml:

```
  <service>silver12</service>
```

elharo:
Problem is in the wizard dialog. As soon as you click **Finish** the last letter in the Service field is cut off.  If you watch closely, you can see it happen.

![image](https://cloud.githubusercontent.com/assets/1005544/25194693/46c7a666-2509-11e7-921f-5994d6c108e5.png)

elharo:
From the debugger this is happening very early, when we call

```
  public String getServiceName() {
    String text = serviceNameField.getText();
    return text;
  }
```
or sooner. Almost looks like an Eclipse bug.

briandealwis:
I've been poking at this too.  I'm flummoxed.
  - **good news**: this only seems to happen in the debugger — I can't reproduce with CT4E 1.0.3.201704111618 in my IDE
  - **it continues to happen when I switch the order of creation with the package field**
  - unrelated to the page-validator/modify listener (continues if I don't install the Text's modify listener)
  - unrelated to the use of `Text#setMessage()`
  - only happens if the service field has focus
  - it's triggered by the call within the wizard to `getContainer().run()`; if I skip that it's ok
     - tracing through `getContainer().run()`, past `ModalContext#block()` and into `Display#readAndDispatch()`, the change happens on the first call at:

        NSEvent event = application.nextEventMatchingMask(0, null, OS.NSDefaultRunLoopMode, true);

briandealwis:
If I retrieve the value in `validateLocalFields()`, it's ok.  Incidentally, turns out that we're not actually performing service name validation.

briandealwis:
I've been able to reproduce with a small standalone example outside of the debugger.  I've [filed a bug against SWT](https://bugs.eclipse.org/bugs/show_bug.cgi?id=515541).

elharo:
FWIW I initially found this outside the debugger in Oxygen M6 with 1.0.3 installed.

briandealwis:
I've since reproduced in my install too.  I suspect I must have changed the focus to another field when testing.

briandealwis:
Removed high-priority and closing since we have workarounds.

PR #1840, opened 2017-04-20
Validate service name field in AppEngineWizardPage

briandealwis:
Use appengine-plugins-core's `ServiceNameValidator` which unfortunately doesn't provide a helpful message to describe the validation error.  Any suggestions to shorten the service-name description message?

Works around #1830

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineWizardPage.java
@@ -135,7 +135,7 @@ public abstract class AppEngineWizardPage extends WizardNewProjectCreationPage {
     IStatus packageStatus = JavaPackageValidator.validate(packageName);
     if (!packageStatus.isOK()) {
       String message = Messages.getString("illegal.package.name",  //$NON-NLS-1$
-          packageStatus.getMessage());
+          packageName, packageStatus.getMessage());
       setErrorMessage(message);
       return false;
     }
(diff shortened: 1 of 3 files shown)


briandealwis:
screen shot 2017-04-20 at 12 31 46 pm

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/messages.properties
```diff
@@ -6,6 +6,7 @@ create.app.engine.standard.project=Create a new Eclipse project for App Engine s
 project.creation.failed=Failed to create project
 java.package=Java package:
 illegal.package.name=Illegal package name: {0}
+illegal.service.name=A service name may contain up to 63 numbers, letters, and hyphens, with no leading or trailing hyphens.
```
elharo: I wonder if the illegal service name should be included in the message?

briandealwis: The `{0}` in the `illegal.package.name` is used to show the validation error message produced by our `JavaPackageValidator`; the `ServiceNameValidator` just returns a true/false.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineWizardPage.java
```diff
@@ -139,6 +141,15 @@ private boolean validateLocalFields() {
       return false;
     }
 
+    String serviceName = serviceNameField.getText();
```
chanseokoh: Just so you know, I believe `.getText()` never returns null.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/messages.properties
elharo: We might want to change that, but in the meantime I still think as a user I'd appreciate seeing the illegal name I typed in this message. It would help me realize exactly which field is wrong.

briandealwis:
Ok, revised:

screen shot 2017-04-21 at 5 23 03 pm

PR #1842, opened 2017-04-20
Make AnalyticsOptIn use its own scoped pref store

chanseokoh:
Make Analytics preferences use its own scoped preference store path: `configuration://com.google.cloud.tools.eclipse.usagetracker`. It doesn't make much sense to me that it uses the bundle default store (which is by default instance-scoped if I hadn't overridden.)

I've removed the `Activator` and the awkward method override, which were the results of the past when I was completely new to Eclipse.

One thing I noticed is that the initializer is called only when the store path matches the bundle name (`com.google.cloud.tools.eclipse.preferences`), so there seems no way to do `setDefault()` on `configuration://com.google.cloud.tools.eclipse.usagetracker`.


briandealwis:
Would it make sense to move all of the Analytics stuff out of .preferences into .usagetracker?

plugins/com.google.cloud.tools.eclipse.preferences/plugin.xml
```diff
@@ -14,18 +14,12 @@
       </page>
    </extension>
    <extension
```
briandealwis: Would it make sense to move this to .usagetracker?

chanseokoh: You mean including everything Analytics-related, such as `AnalyticsOptInArea.java` and `AnalyticsPreferences.java` in this bundle, right?

briandealwis: Oh right you were fetching the value directly via ConfigurationScope.  It makes more sense if we were using the IPreferenceService, which will look up a set of scopes (default: instance, configuration, then default scope).

briandealwis: > You mean including everything Analytics-related, such as AnalyticsOptInArea.java and AnalyticsPreferences.java in this bundle, right?

Yeah.

chanseokoh:
Let me check if I can use an initializer after moving code.

chanseokoh:
Yeah, it's only by `AnalyticsPingManager`, and it felt like adding more boilerplate code with little benefit.

PR #1844, opened 2017-04-21
Remove hidden "default" deploy preferences 

### LINKED ISSUE #1846: DeployPreferencesModel might not be necessary
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1844#pullrequestreview-34061135

It sits between the data binding (UI stuff in the deploy dialog) and the persisted user preferences, so it may be possible to remove it. However, I'm not so sure if we can do this easily, because my understanding is that, we modify the model values whenever a user changes UI selections for example, so we do need to delay persisting until a user presses _OK_, _Apply_, or _Deploy_. That sounds like we nevertheless need some dirty buffer object anyway.

chanseokoh:
I've been thinking that saving a hidden user preference node to reset values in a plain Java bean is a bit hacky (and also over-engineering).

(Note: deploy preferences are project-scoped that are created when a project exists, so it's not possible to do `setDefault()` in an initializer. That's probably the reason for the hidden instance-scoped preference hack.)

Fixes #1846.


chanseokoh:
I do remember you said a very long time ago about the doubt on the model, and I also thought probably we could get rid of it. Since that'll involve some investigation and tinkering that I don't want to dig into now, I'll create a new issue.

chanseokoh:
Much better after fixing #1846.

Tests will fail. I need to fix them.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployPreferencesTest.java
```diff
@@ -16,60 +16,151 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
-import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
 import org.eclipse.core.resources.IProject;
-import org.junit.Rule;
+import org.junit.Before;
 import org.junit.Test;
 
 public class DeployPreferencesTest {
```
briandealwis: The other thing that would be useful here is to test that (1) the `IEclipsePreferences` haven't changed, and (2) they change after a `save()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployPreferencesTest.java
```diff
@@ -16,60 +16,151 @@
...
+  public void testSetIncludeOptionalConfigurationFiles() {
+    assertTrue(preferences.isIncludeOptionalConfigurationFiles());
+    preferences.setIncludeOptionalConfigurationFiles(false);
+    assertFalse(preferences.isIncludeOptionalConfigurationFiles());
+    preferences.setIncludeOptionalConfigurationFiles(true);
+    assertTrue(preferences.isIncludeOptionalConfigurationFiles());
   }
 
+  @Test
+  public void testResetToDefault() {
```
briandealwis: Does `resetToDefault()` imply a `save()`?  I'm thinking no, but I haven't thought about it much… but be sure to document whatever you decide.

If yes, could check for the existence of the `IEclipsePreferences` node.

If no, `DeployPreferences` could record a value to indicate that `save()` should remove the node.

chanseokoh: It doesn't imply `save()`, because `resetToDefaults()` is for when you press the _Reset Defaults_ button, and that just resets the intermediate model.

> If no, `DeployPreferences` could record a value to indicate that save() should remove the node.

Sorry, I didn't get it. Why remove the preference node?

chanseokoh: Ah, I got what you mean, but do you think we really need to remove the node? I don't see a great harm to persist default values.

briandealwis: It seems easier than `remove()`ing the individual fields

chanseokoh: I am not doing `remove()`. Just resetting local fields. Note that `preferences` is not `IEclipsePreferences`.

PR #1845, opened 2017-04-21
fix assorted warnings

elharo:
@chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/convert/AppEngineStandardProjectConvertJob.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/convert/AppEngineStandardProjectConvertJob.java
@@ -20,8 +20,6 @@ import com.google.cloud.tools.eclipse.appengine.compat.GpeMigrator;
 import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
 import com.google.cloud.tools.eclipse.appengine.facets.Messages;
 import com.google.cloud.tools.eclipse.util.status.StatusUtil;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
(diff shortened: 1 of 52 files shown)


plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/status/StatusUtilTest.java
```diff
@@ -61,13 +61,15 @@ public void testError_withInstanceAndException() {
...
     IStatus error = StatusUtil.error((Object) new DeleteAllVisitor().getClass(), "test error msg");
     verifyStatus(error);
   }
 
   @Test
   public void testError_withClassAsObjectAndException() {
     Throwable exception = new Exception();
+    @SuppressWarnings("cast")
     IStatus error =
         StatusUtil.error((Object) new DeleteAllVisitor().getClass(), "test error msg", exception);
```
briandealwis: I this these tests are from when we had separate `error(Class,…)` methods.  They can probably just go away.  Or at least use `DeleteAllVisitor.class`?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/convert/AppEngineStandardProjectConvertJob.java
```diff
@@ -48,14 +43,7 @@ protected IStatus run(IProgressMonitor monitor) {
 
     // Updating project before installing App Engine facet to avoid
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155.
-    try {
-      GpeMigrator.removeObsoleteGpeRemnants(facetedProject, subMonitor.newChild(20));
-    } catch (CoreException ex) {
-      // Failed to remove GPE remnants; live with it.
-      subMonitor.setWorkRemaining(20);
-      subMonitor.worked(20);
-      logger.log(Level.WARNING, "Error while removing GPE remnants", ex);
```
chanseokoh: We need to catch this and don't error out, because it will always throws the exception, e.g., if the project is not a GPE project (e.g., a plain Java project).

Because we swallow the exception, the logger is for us to get notified if the conversion fails for whatever reason.

chanseokoh: Hmm... seems like I'm wrong.

PR #1851, opened 2017-04-24
mark some methods static

### LINKED ISSUE #1768: Disable next button on dataflow archetype selection page
Only the finish button should be enabled on this page. Clicking the next button does not do anything here but it's not grayed out.

image

Or maybe the Set Default Cloud Dataflow Run Options dialog isn't checked in yet and there should be a next page?

elharo:
@tgroh a bit of cleanup


elharo:
You're right. Not sure why I didn't see that page before. At one point the next button wasn't working for me, but it's not clear why. In any case, it seems to be working now.

PR #1847, opened 2017-04-21
unit test proof of concept

elharo:
@briandealwis @chanseokoh Anyway to remove/exclude the JUnit dependency? It looks like https://resheim.net/2014/12/unit-testing-eclipse-rcp-applications.html may not have run these from the command line.

PR #1855, opened 2017-04-25
Add support for the App Engine Java 8 runtime

### LINKED ISSUE #1057: Support Java 8 runtime standard projects
https://github.com/GoogleCloudPlatform/google-cloud-intellij/issues/946

If Eclipse is doing any validation/warning based off of Java 8 usage in Java standard, it should make sure that it does not complain for cases where the user has configured
appengine-web.xml:
`<runtime>java8</runtime>`

briandealwis:
Brings in support for the new App Engine Java 8 runtime (#1057):

- Switches the _App Engine Standard_ facet to using facet groups to specify compatible Servlet API and Java Runtimes as these facet groups can be extended/added to by other bundles
- Creates new `com.google.cloud.tools.eclipse.appengine.standard.java8` and `.test` bundle that hold the Java 8 functionality.  All other changes leverage the facets framework by testing whether the _App Engine Standard v1_ facet-version is compatible with some proposed facet-version change via `IProjectFacetVersion#conflictsWith(IProjectFacetVersion)`.  No feature flag required!
    - the `ServletClasspathProvider` uses the `jst.web` version strings to decide, but we could instead bundle the Servlet API and JSP libraries into a single per-Servlet API library bundle and then just use the jst.web version string directly.
- Adds listeners and builders to translate between Java 1.8 facet and `<runtime>java8</runtime>` changes to `appengine-web.xml`.
    - I wasn't able to figure out the right XSTL magic to properly indent the `<runtime>` element when added; adding `<xsl:output indent="yes">` made some difference, but it wasn't enough on its own.
- Fixed our Server Runtime definition in `.localserver` as the module-types are J2EE modules not facets (i.e., `java` is not a module-type).

Requires no changes to deployment.

Fixes #1057

# Side Effects

# Facet Group Side-Effects
When adding an AES facet to an existing project, the help content shown uses the group labels, which is of the form:

    {{facet}} {{version}} requires any member of {{group-label}}} group.

as you can see below.

facet group labels


briandealwis:
Oops, there's one outstanding item: applying _Convert to App Engine Standard Project_ to a Java 8 project resets the support to Java 7.

chanseokoh:
If I create a plain Java 8 project and do the conversion, the project becomes a Java 7 project with Servlet 2.5. What do you think is happening?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -113,6 +115,8 @@ public void testCheckFacetCompatibility_java1_6FacetIsIncompatible() throws Core
 
   @Test
   public void testCheckFacetCompatibility_java1_8FacetIsIncompatible() throws CoreException {
+    Assume.assumeTrue("AppEngine Standard Java 8 runtime support is not present",
```
elharo: why wouldn't it be?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
```diff
@@ -132,6 +136,8 @@ public void testCheckFacetCompatibility_web2_4FacetIsIncompatible() throws CoreE
 
   @Test
   public void testCheckFacetCompatibility_web3_0FacetIsIncompatible() throws CoreException {
+    Assume.assumeTrue("AppEngine Standard Java 8 runtime support is not present",
```
elharo: ditto. Can we make the test unconditional?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -64,23 +79,27 @@ public ServletClasspathProvider() {
...
   }
 
-  private IClasspathEntry[] doResolveClasspathContainer() {
+  private IClasspathEntry[] doResolveClasspathContainer(IProjectFacetVersion dynamicWebVersion) {
+
+    String servletApiId;
+    String jspApiId;
+    if (WebFacetUtils.WEB_31.equals(dynamicWebVersion)
+        || WebFacetUtils.WEB_30.equals(dynamicWebVersion)) {
+      servletApiId = "servlet-api-3.1";
```
elharo: not 3.0 in the second case?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,14 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Google Cloud Tools for App Engine Standard Java 8 Tests
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.standard.java8.test
+Fragment-Host: com.google.cloud.tools.eclipse.appengine.standard.java8
+Bundle-Version: 0.0.1.qualifier
+Bundle-Vendor: Google, Inc.
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
```
elharo: not Java 8?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,14 @@
...
+Import-Package: com.google.cloud.tools.appengine;version="0.2.11",
```
elharo: I think we're at 0.3.1 on this

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/.project
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
elharo: I'm surprised this required a new package and bundle

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,18 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Java 8 development support for App Engine standard
+Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.standard.java8;singleton:=true
+Bundle-Version: 0.1.0.qualifier
+Bundle-Vendor: Google, Inc.
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
```
elharo: 8?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineWebBuilder.java
```diff
@@ -0,0 +1,85 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -0,0 +1,117 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -0,0 +1,117 @@
...
+import org.eclipse.jst.common.project.facet.core.JavaFacet;
+import org.eclipse.wst.common.project.facet.core.IFacetedProject;
+import org.eclipse.wst.common.project.facet.core.events.IFacetedProjectEvent;
+import org.eclipse.wst.common.project.facet.core.events.IFacetedProjectEvent.Type;
+import org.eclipse.wst.common.project.facet.core.events.IFacetedProjectListener;
+import org.eclipse.wst.common.project.facet.core.events.IProjectFacetActionEvent;
+
+/**
+ * Handle facet install and uninstalls, version changes, and generally anything else, and add/remove
+ * the our {@code appengine-web.xml} builder and its {@code <runtime>java8</runtime>} element.
```
elharo: the our --> the

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -0,0 +1,117 @@
...
+  private void addAppEngineWebBuilder(IProject project) {
+    try {
+      IProjectDescription desc = project.getDescription();
+      ICommand[] commands = desc.getBuildSpec();
+      for (int i = 0; i < commands.length; i++) {
+        if (AppEngineWebBuilder.BUILDER_ID.equals(commands[i].getBuilderName())) {
+          return;
+        }
+      }
+      ICommand[] nc = new ICommand[commands.length + 1];
```
elharo: what's nc? No abbreviation per google style

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -0,0 +1,117 @@
...
+   * Add our {@code appengine-web.xml} builder that monitors for changes to the {@code <runtime>}
+   * element.
+   */
+  private void removeAppEngineWebBuilder(IProject project) {
+    try {
+      IProjectDescription desc = project.getDescription();
+      ICommand[] commands = desc.getBuildSpec();
+      for (int i = 0; i < commands.length; i++) {
+        if (AppEngineWebBuilder.BUILDER_ID.equals(commands[i].getBuilderName())) {
+          ICommand[] nc = new ICommand[commands.length - 1];
```
elharo: ditto

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+  }
+
+  /**
+   * Add our {@code appengine-web.xml} builder that monitors for changes to the {@code <runtime>}
+   * element.
+   */
+  private void removeAppEngineWebBuilder(IProject project) {
+    try {
+      IProjectDescription desc = project.getDescription();
```
elharo: desc --> description

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
briandealwis: The Java 1.8 facet is not incompatible when the .standard.java8 bundle is available.  Tycho will run with the minimal bundles required, and so the test should pass.  But the test will likely fail when run in Eclipse since _JUnit Plug-in Tests_ defaults to running with all bundles.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
briandealwis: We could set up a `<library>` clause with the Servlet 3.0 and JSP 2._whatever_, but since 3.1 is backward compatible it didn't seem worth the while?

If we were going to change this, I think it would be better to create `<library>` entries for the specific supported `jst.web` versions.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
briandealwis: This is the Eclipse runtime.  So not until we abandon Mars.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/.project
briandealwis: It didn't, but it forced leveraging the facets constraint framework for testing compatibility (which is a good thing) rather than introducing a separate feature flag and littering `isJava8()`-style checks everywhere.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
briandealwis: The package versions are still at 0.2.11!  I wonder if the maven-bundle-plugin compares signature changes if the previous jar is found?

briandealwis:
> @chanseokoh: If I create a plain Java 8 project and do the conversion, the project becomes a Java 7 project with Servlet 2.5. What do you think is happening?

Should be fixed with 1559dc5.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -10,12 +10,39 @@
...
     </project-facet-version>
   </extension>
+   <extension
+         point="org.eclipse.wst.common.project.facet.core.groups">
+      <group
+            id="com.google.cloud.tools.eclipse.appengine.facets.standard.constraints.servlet">
+         <label>
+            supported Dynamic Web Module APIs
+         </label>
+         <include
```
elharo: Could be combined to a single line

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -10,12 +10,39 @@
...
+               facet="jst.web"
+               versions="2.5">
+         </include>
+      </group>
+      <group
+            id="com.google.cloud.tools.eclipse.appengine.facets.standard.constraints.jre">
+         <label>
+            supported Java Runtimes
+         </label>
+         <include
```
elharo: Could be combined to a single line

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -10,12 +10,39 @@
...
+            supported Dynamic Web Module APIs
+         </label>
+         <include
+               facet="jst.web"
+               versions="2.5">
+         </include>
+      </group>
+      <group
+            id="com.google.cloud.tools.eclipse.appengine.facets.standard.constraints.jre">
+         <label>
```
elharo: Could be combined to a single line

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
```diff
@@ -8,15 +8,23 @@
...
             name="%apptoolsRuntimeTypeName"
+            supportsManualCreation="false"
             vendor="%runtimeTypeVendor"
             version="1">
+         <!-- J2EE Web Modules -->
          <moduleType
                types="jst.web"
                versions="2.5">
          </moduleType>
+         <!-- Static Web Projects -->
```
elharo: This addition is good. Would it make sense to pull it out as a separate, quicker PR?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
```diff
@@ -0,0 +1,107 @@
...
+               types="jst.web"
+               versions="3.0,3.1">
+         </moduleType>
+      </runtimeModuleType>
+   </extension>
+
+   <extension
+         point="org.eclipse.wst.common.project.facet.core.groups">
+      <members
+            group="com.google.cloud.tools.eclipse.appengine.facets.standard.constraints.servlet">
```
elharo: So if I follow you, these two constraints groups are being contributed to by two different plugin.xml files. It's not redefining the group here. Is that right?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
```diff
@@ -0,0 +1,107 @@
...
+              groupId="javax.servlet"
+              artifactId="javax.servlet-api"
+              version="3.1.0" />
+      </libraryFile>
+    </library>
+    <library
+          id="jsp-api-2.3"
+          group="servlet"
+          name="Java Server Pages API 2.3"
+          siteUri="http://www.oracle.com/technetwork/java/javaee/jsp/index.html"
```
elharo: https://jsp.java.net/ looks like a better choice

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <groupId>com.google.cloud.tools.eclipse</groupId>
+    <artifactId>trunk</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
+    <relativePath>../../</relativePath>
+  </parent>
+  <artifactId>com.google.cloud.tools.eclipse.appengine.standard.java8</artifactId>
+  <version>0.1.0-SNAPSHOT</version>
+  <packaging>eclipse-plugin</packaging>
+
+  <dependencies>
```
elharo: I guess this is related to making `mvn test` work. Does this need to be here or can it be pushed up to the parent pom.xml?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Test the appengine-web.xml XSLT transforms
+ */
+public class AppEngineXsltTransformTests {
+
+  @Test
+  public void testAddBare() throws IOException, TransformerException {
+    apply("/xslt/addJava8Runtime.xsl",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?><appengine-web-app xmlns=\"http://appengine.google.com/ns/1.0\"/>",
```
elharo: You don't need the XML declarations assuming this is test input data, and taking them out makes this somewhat simpler. 

You do need it in the output string if you're just doing string comparisons. String comparisons on transformed XML are flaky since there are lot of details such as quotes and empty-element tags that can change without modifying the infoset. A JDK upgrade could break these tests. The simplest way to make this robust is to parse the output and look for the details you want such as a runtime element.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
```diff
@@ -0,0 +1,136 @@
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?><appengine-web-app xmlns=\"http://appengine.google.com/ns/1.0\">"
+            + "<runtime>java8</runtime></appengine-web-app>");
+  }
+
+  @Test
+  public void testAddExistingRuntime() throws IOException, TransformerException {
+    apply("/xslt/addJava8Runtime.xsl",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?><appengine-web-app xmlns=\"http://appengine.google.com/ns/1.0\"><runtime>XXX</runtime></appengine-web-app>",
```
elharo: remove XML declaration in input here and below

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
```diff
@@ -0,0 +1,136 @@
...
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?><appengine-web-app xmlns=\"http://appengine.google.com/ns/1.0\">"
+            + "<runtime>java8</runtime></appengine-web-app>");
+  }
+
+  @Test
+  public void testAddPreservesOrder() throws IOException, TransformerException {
+    apply("/xslt/addJava8Runtime.xsl",
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?><appengine-web-app xmlns=\"http://appengine.google.com/ns/1.0\">"
```
elharo: remove XML declaration

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
```diff
@@ -0,0 +1,136 @@
...
+    URL xslPath = AppEngineXsltTransformTests.class.getResource(templateFile);
+    if (xslPath == null) {
+      File xslFile = new File("." + templateFile);
+      assertTrue(xslFile.exists());
+      xslPath = xslFile.toURI().toURL();
+    }
+    InputStream documentStream =
+        new ByteArrayInputStream(inputString.getBytes(StandardCharsets.UTF_8));
+    InputStream transformed = Xslt.applyXslt(documentStream, xslPath.openStream());
+    String result =
```
elharo: Here we need to parse the output rather than doing a string comparison.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineWebBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.jst.common.project.facet.core.JavaFacet;
+import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
+import org.eclipse.wst.common.project.facet.core.IFacetedProject;
+import org.eclipse.wst.common.project.facet.core.IFacetedProject.Action;
+import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
+
+/**
+ * Builder to monitor user changes relating to the {@code <runtime>java8</runtime>} element in
```
elharo: Neat. I didn't know this was feasible.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineWebBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+    return null;
+  }
+
+  private void checkRuntimeElement(IFacetedProject project, IFile appEngineWebDescriptor,
+      IProgressMonitor monitor) {
+    try (InputStream input = appEngineWebDescriptor.getContents()) {
+      boolean isJava8 = AppEngineDescriptor.parse(input).isJava8();
+      if (project.hasProjectFacet(JavaFacet.VERSION_1_8) != isJava8) {
```
elharo: boolean != boolean 
is a bit surprising. I wonder if it clarify to refactor as
if 
else if
else

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -0,0 +1,134 @@
...
+      projectDescription.setBuildSpec(newCommands);
+      project.setDescription(projectDescription, null);
+      logger.finer(project + ": added AppEngineWebBuilder");
+    } catch (CoreException ex) {
+      logger.log(Level.SEVERE, "Unable to add builder for " + project, ex);
+    }
+  }
+
+  /**
+   * Add our {@code appengine-web.xml} builder that monitors for changes to the {@code <runtime>}
```
elharo: Update doc comment (remove, not add)

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
briandealwis: Any suggestions?  This approach, which is imperfect as it doesn't support indentation, at least verifies we haven't inadvertently added or removed any items.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
briandealwis: Right

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
elharo: You need to feed the string returned into a parser, then query the resulting DOM object to see if what you want/don't want is/isn't present. Some of Abbie's tests do this already, I think.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
briandealwis: Ok.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
briandealwis: That site only provides 2.1 :-(

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/pom.xml
briandealwis: I'll improve this as part of #1847

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineWebBuilder.java
briandealwis: Will clarify

chanseokoh:
A couple of things I noticed:
- If I convert a plain Java8 project, the `appengine-web.xml` created doesn't have `<runtime>java8</runtime>`.
- If I create a Java7 project, manually add the Dyn Web facet 3.1, and try to convert it, we will correctly show an info dialog, but the message doesn't help (and even misleading, I think):

incompatible dynamic web module facet version _009

These could be done in a separate PR, as we are safe that we don't expose the Java8 bundle yet, right?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
elharo: so what is this testing? that the app engine facet isn't compatible with java 8 facet, but only sometimes? This feels off to me. Why should we not simply make the app engine standard facet unconditionally compatible with Java 8?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
```diff
@@ -0,0 +1,165 @@
...
+      return builder.parse(transformed);
+    } catch (IOException | TransformerException | ParserConfigurationException | SAXException ex) {
+      // these aren't interesting for the test
+      throw new RuntimeException(ex);
+    }
+
+  }
+
+  /** Return the {@code <runtime>} value, or {@code null} if no element found. */
+  private String getRuntimeElement(Document transformed) {
```
elharo: getRuntimeValue or just getRuntime since it doesn't return an element

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
briandealwis: It's testing that when our AESv1 facet claims to not support Java 8, that our convert command properly reports that it cannot convert the project.  It's no different from a java8 feature flag and having the `Assume.assumeFalse(java8)`.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
briandealwis: Opened https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/384 but it seems the package versions are fixed in 0.3.2.  Let's fix this when we next update appengine-plugins-core.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
elharo: It is a little different. With a feature flag, all code paths are tested regardless of platform. Here the test is sometimes skipped.

More importantly, this raises another question. What does the UI here look like? Are we showing a menu item for the command that does not work in some circumstances? It's just the same existing convert menu item in the context menu, though, right?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
elharo: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1868

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,18 @@
...
+Require-Bundle: com.google.cloud.tools.eclipse.appengine.localserver;bundle-version="0.1.0",
+ org.eclipse.wst.common.project.facet.core,
+ org.eclipse.core.resources,
+ org.eclipse.core.runtime
+Import-Package: com.google.cloud.tools.appengine;version="0.2.11",
+ com.google.cloud.tools.eclipse.appengine.facets,
+ com.google.cloud.tools.eclipse.util,
+ com.google.common.io;version="15.0.0",
```
elharo: 15.0.0? Should this be upgraded yet? or unspecified?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
```diff
@@ -0,0 +1,165 @@
...
+        xslPath = xslFile.toURI().toURL();
+      }
+      InputStream documentStream =
+          new ByteArrayInputStream(inputValue.getBytes(StandardCharsets.UTF_8));
+      InputStream transformed = Xslt.applyXslt(documentStream, xslPath.openStream());
+      DocumentBuilderFactory builderFactory = DocumentBuilderFactory.newInstance();
+      builderFactory.setNamespaceAware(true);
+      DocumentBuilder builder = builderFactory.newDocumentBuilder();
+      return builder.parse(transformed);
+    } catch (IOException | TransformerException | ParserConfigurationException | SAXException ex) {
```
elharo: you could just let them escape then rather than catching

briandealwis:
[Nice find](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1855#issuecomment-297469110) @chanseokoh — I wasn't testing that the `<runtime>` element was added. Turns out I had the code only add the `<runtime>` element when adding/changing the Java facet and not the AES facet.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/AppEngineStandardProjectConvertCommandHandlerTest.java
briandealwis: There are no UI changes.  Just the presence or absence of the bundle.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src-test/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineXsltTransformTests.java
briandealwis: Are you ok `throws Exception`?  Otherwise means littering the method signatures with `throws W, X, Y, Z` which seemed like noise.

briandealwis:
> @elharo: More importantly, this raises another question. What does the UI here look like? Are we showing a menu item for the command that does not work in some circumstances? It's just the same existing convert menu item in the context menu, though, right?

There are no UI changes.  We just open the constraints for the _App Engine Standard_ facet to work with Java [1.7,1.8] and the _Dynamic Web Facet_ [2.5-3.1].

PR #1860, opened 2017-04-26
Fix supported module types for .standard.runtime definition

briandealwis:
Fix `.runtime.standard` module types.  Extracted from https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/java8


chanseokoh:
I'm curious to know what was wrong and why we need this. Care to explain this?

briandealwis:
The server runtime is meant to specify the types of J2EE modules that it supports.  `java` is a facet, not a module type, whereas `jst.web` is used both as a module type AND a facet type.

PR #1869, opened 2017-04-27
Fix up flexible wizard help

### LINKED ISSUE #1795: Flex wizard should not show help for libraries to add to build path
image

It probably should have help for App Engine service.

elharo:
@chanseokoh fix #1795

Note that I can't hard wrap the entities because the line breaks would be preserved in the displayed HTML. Thus the lines are really long.


plugins/com.google.cloud.tools.eclipse.appengine.newproject/helpContexts.xml
```diff
@@ -1,22 +1,28 @@
...
+    <description>Create a new Eclipse project that can be deployed to the App Engine flexible environment.
+
+      &project;
 
-&lt;b&gt;Project name:&lt;/b&gt; An Eclipse project name.
+      &service;
 
-&lt;b&gt;Libraries to add to build path:&lt;/b&gt;
-Adds useful libraries for the App Engine flexible environment to your project build path.
      </description>
```
chanseokoh: Nit: -1 indentation.

PR #1872, opened 2017-04-27
Kokoro continuous test

akerekes:
This is the initial version of the Windows build. It runs, and fails on:
> testStageFlexible(com.google.cloud.tools.eclipse.appengine.deploy.CloudSdkStagingHelperTest)  Time elapsed: 5.797 sec  <<< FAILURE!
java.lang.AssertionError: Could not delete project: Problems encountered while deleting resources.
	at com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator.after(TestProjectCreator.java:93)

The ubuntu build is not working due to missing UI, and also the required internal configuration is not in place.

Tweaking the log access of the windows build may be necessary.


kokoro/windows/continuous.bat
```diff
@@ -0,0 +1,16 @@
...
+start /WAIT GoogleCloudSDKInstaller.exe /S /noreporting /nostartmenu /nodesktop /logtofile /D=T:\google
+call t:\google\google-cloud-sdk\bin\gcloud.cmd components copy-bundled-python>>python_path.txt && SET /p CLOUDSDK_PYTHON=<python_path.txt && DEL python_path.txt
+call t:\google\google-cloud-sdk\bin\gcloud.cmd components update --quiet
+call t:\google\google-cloud-sdk\bin\gcloud.cmd components install app-engine-java --quiet
+set GOOGLE_CLOUD_SDK_HOME=t:\google\google-cloud-sdk
+
+wget http://www-us.apache.org/dist/maven/maven-3/3.3.9/binaries/apache-maven-3.3.9-bin.zip
+unzip apache-maven-3.3.9-bin.zip
+
+apache-maven-3.3.9/bin/mvn --fail-at-end -Ptravis -Declipse.target=oxygen verify
```
chanseokoh: Here the target is Oxygen, whereas for Ubuntu, it's Mars.

kokoro/ubuntu/continuous.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+#
+# This is just an initial version based on the Travis build script, the build
+# needs UI, which haven't been tested, just copied over from .travis.yml
```
chanseokoh: If Kokoro has `/etc/init.d/xvfb` and if it works, it may work. Let's see.

kokoro/windows/continuous.bat
briandealwis: And we're using Maven 3.5 on Travis

akerekes: There was no real strategy around this, both are arbitrarily selected and can be changed.

kokoro/ubuntu/continuous.sh
akerekes: xvfb is not listed among the installed packages and I think I ran the script and saw it failing. I haven'r tested the DISPLAY redirection.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/TestProjectCreator.java
```diff
@@ -90,7 +90,7 @@ protected void after() {
       try {
         project.delete(true, null);
       } catch (CoreException ex) {
-        fail("Could not delete project");
+        fail("Could not delete project: " + ex.getMessage());
```
briandealwis: Suggestion to replace these with:

    throw new AssertionError("Could not delete project", ex);

That should show the sub-exception too.

elharo:
Trial-and-error  is not necessary. Kokoro can build of a branch, so we don't need to check into master until it works.

PR #1875, opened 2017-04-28
Hook in Java 8 support for the build

briandealwis:
- ensure XSL templates are being packaged into the bundle
  - rename test classes so are picked up by surefire
  - specify maven-compiler-plugin source levels
  - remove test source from main source

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/build.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/build.properties
@@ -1,7 +1,7 @@
-source.. = src/,\
-           src-test/
+source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
                .,\
-               plugin.xml
+               plugin.xml,\
+               xslt/
 additional.bundles = org.eclipse.equinox.common, org.junit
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/build.properties
```diff
@@ -1,7 +1,7 @@
-source.. = src/,\
-           src-test/
```
chanseokoh: This will prevent `AppEngineXsltTransformTest.java` from being included in the final build, right? Also what's the criterion to put the test class in the main bundle while put other classes in the `.test` bundle?

Also I noticed `output.. = bin/` instead of `output.. = target/classes` in both the main bundle and the `.test` bundle, and unlike other bundles, the `target` folder doesn't have OGSi stuff. (Strangely, I don't see the `bin` folder though.) Is this normal?

Lastly, the following files which are not checked into the repo are auto-generated when I import the `java8` bundles.
```
com.google.cloud.tools.eclipse.appengine.standard.java8.test/.settings/org.eclipse.core.resources.prefs
com.google.cloud.tools.eclipse.appengine.standard.java8.test/.settings/org.eclipse.m2e.core.prefs
```

briandealwis: The `src-test` folder is pulled in via the `.classpath`.  It's not ideal as it's lost with a _Maven > Update Project…_.  Just writing up a comment about this on on #1847.

chanseokoh:
We need to put the bundles into jacoco too.

briandealwis:
ok, will fix the above and submit a new PR

PR #1878, opened 2017-04-28
Fix up some last dangling issues

briandealwis:
Some nits found by @chanseokoh in #1875

- fix up and add missing config files in .appengine.standard.java8* bundles and make robust to _Maven > Update Projects_
- update jacoco build dependencies
  - automate more in gendependencies.sh script
  - fix up whitespace in .datacore.ui* poms

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/build.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/build.properties
@@ -1,5 +1,7 @@
-source.. = src/
-output.. = bin/
+source.. = src/,\
+           src-test/
+src.excludes = src-test/
+output.. = target/classes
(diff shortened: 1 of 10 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/build.properties
```diff
@@ -1,5 +1,7 @@
-source.. = src/
-output.. = bin/
+source.. = src/,\
+           src-test/
+src.excludes = src-test/
+output.. = target/classes
```
chanseokoh: Probably fix `ouput...` in the `.test` bundle too?

briandealwis: sigh.

PR #1894, opened 2017-05-02
remove expected exception per google style

elharo:
@chanseokoh


chanseokoh:
There are compilation errors.

PR #1905, opened 2017-05-03
Refactor test code

chanseokoh:
Moved common code to traverse a `Composite` for finding a certain control to the test util bundle.


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/CompositeUtil.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2016 Google Inc.
```
elharo: 2017

PR #1908, opened 2017-05-04
Clean up code

chanseokoh:
Forgot cleaning up this: adding a flex facet takes care of creating `src/main/appengine/app.yaml`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/pom.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/pom.xml
@@ -29,6 +29,12 @@
                 <id>org.eclipse.jst.web_ui.feature</id>
                 <versionRange>0.0.0</versionRange>
               </requirement>
+              <!-- TODO: remove this once the flex facet is fully available. -->
+              <requirement>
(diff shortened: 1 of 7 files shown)


chanseokoh:
Will take care of the test failures.

chanseokoh:
Ah, now I remember why I installed the AE Standard facet and manually created `src/main/appengine/app.yaml` instead of installing the Flex facet: the facet definition is in the `.flex` bundle, which we haven't added to our target platform, so we can't use it yet.

But now I think I can manually pull the `.flex` bundle through `<extraRequirements>` in `pom.xml`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/pom.xml
```diff
@@ -29,6 +29,12 @@
                 <id>org.eclipse.jst.web_ui.feature</id>
                 <versionRange>0.0.0</versionRange>
               </requirement>
+              <!-- TODO: remove this once the flex facet is fully available. -->
```
elharo: This needs a specific issue so we don't lose track of it.

PR #1911, opened 2017-05-04
Ignore downloaded jar dependencies

briandealwis:
Missed from original Java8 PR.


chanseokoh:
Sorry, I don't get it. The directory has `mockito-core-1.10.19.jar` and `objenesis-2.2.jar` that we need to check in. And if there is any `.jar` that doesn't need to be there, you can just delete them.

briandealwis:
Oops, I need to delete those copies.  The dependencies will now be copied in at build-time; it makes updating dependencies easier (just update `/pom.xml` and go, vs having the old bins hanging around in `.test.dependencies/lib`).

briandealwis:
Ah, I was mistakenly thinking of the changes pending in PR #1847.

PR #1914, opened 2017-05-05
check for duplicate project

### LINKED ISSUE #1906: Create project with same name in maven wizard
The maven wizard lets you attempt to create a project with the same name as one that already exists in the workspace as long as you choose a new location. It then fails silently unless you look in the error log. 

image

elharo:
@chanseokoh fix #1906, maybe.

I'm not sure this is the right approach. Other options include:

1. Add a separate Project name field.
2. add a "-2", "-3" suffix to the eclipse project name if the artifact doubles up. 

I'm not sure it shoudl be impossible to create two eclipse projects in a workspace that have the same Maven artifact ID. They might have different group Ids, for example.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
@@ -323,7 +323,7 @@ public class MavenAppEngineStandardWizardPage extends WizardPage {
   }
 
   /**
-   * Check that we won't overwrite an existing location. Expects a valid Maven Artifact ID.
+   * Check that we won't overwrite an existing location. Expects a valid Maven artifact ID.
    */
   private boolean validateGeneratedProjectLocation() {
     String artifactId = getArtifactId();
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -352,6 +352,9 @@ private boolean validateMavenSettings() {
     } else if (!MavenCoordinatesValidator.validateArtifactId(artifactId)) {
       setErrorMessage(Messages.getString("ILLEGAL_ARTIFACT_ID", artifactId)); //$NON-NLS-1$
       return false;
+    } else if (ResourcesPlugin.getWorkspace().getRoot().getProject(artifactId).exists()) {
+      setErrorMessage(artifactId + " project already exists in this workspace"); //$NON-NLS-1$
```
chanseokoh: How about double quoting the `artifactId` in the UI message? Add should add this in `message.properties`.

chanseokoh: Oh, another one: our UI doesn't hint that the artifact ID will become the project name, so saying "<some name> project already exists" may be confusing, as it might imply that a "project" (as an abstract notion) already exists, while it is actually the project _name_ that already exists.

elharo: done

PR #1920, opened 2017-05-07
Make TestProjectCreator wait for project only when adding GAE standard facet

chanseokoh:
Might be just a tiny difference, but I hope this will help reduce the running time.

The changes in `setAppEngineServiceId` are simple variable renaming and replacing `StringBuilder` with `String`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/TestProjectCreator.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/TestProjectCreator.java
@@ -18,6 +18,7 @@ package com.google.cloud.tools.eclipse.test.util.project;
 
 import static org.junit.Assert.assertTrue;
 
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
 import com.google.cloud.tools.eclipse.appengine.facets.FacetUtil;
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/TestProjectCreator.java
```diff
@@ -162,8 +163,10 @@ private void addFacets() throws CoreException {
     }
     facetUtil.install(null);
 
-    // App Engine runtime is added via a Job, so wait.
-    ProjectUtils.waitForProjects(getProject());
+    if (facetedProject.hasProjectFacet(AppEngineStandardFacet.FACET)) {
```
elharo: This isn't needed for flex?

chanseokoh: Not needed for flex. For standard, we add the App Engine runtime (which doesn't exist for flex) via a job, so that is why we should wait here.

elharo:
Is this ready to be checked in?

chanseokoh:
It's ready, but I'm hesitant ATM to merge it. This shouldn't make any change to running tests if every test is written correctly, but I'm worried if it will increase flaskiness due to some poorly guarded tests. I'm basically waiting until our tests on Kokoro become very stable.

PR #1931, opened 2017-05-09
connect java8 support in build

elharo:
@briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/pom.xml
```diff
@@ -11,56 +11,4 @@
   <artifactId>com.google.cloud.tools.eclipse.appengine.standard.java8</artifactId>
   <version>0.1.0-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
-
-  <dependencies>
```
briandealwis: This prevents the in-bundle unit tests from running.  Wouldn't this be better as part of #1847?

pom.xml
```diff
@@ -52,14 +52,14 @@
     <module>plugins/com.google.cloud.tools.eclipse.appengine.localserver.test</module>
     <module>plugins/com.google.cloud.tools.eclipse.appengine.newproject</module>
     <module>plugins/com.google.cloud.tools.eclipse.appengine.newproject.test</module>
+    <module>plugins/com.google.cloud.tools.eclipse.appengine.standard.java8</module>
```
briandealwis: Not against this movement, but was it actually required?

elharo: probably not. At first I didn't see the existing modules because they were out of order.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/pom.xml
briandealwis: It's nearly done.  It was done, but there are now conflicts and kokoro seems to be failing.  Will revisit it soon.

PR #1933, opened 2017-05-09
Validate app.yaml runtime / prevent custom runtime

chanseokoh:
Checks if the `runtime:` value in `app.yaml` is `java`. If it's `custom`, shows a special error message that we don't support  custom runtime deploy yet. If `app.yaml` is malformed, it blocks deploying. ~~If there's an I/O error, I decided not to block deploying.~~


elharo:
Why not block deploying on an I/O error? It's the same situation: user wants to deploy files A, B, and C. We can't read file B. Therefore we should not surprise the user by only deploying half of what they asked for.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -26,10 +26,10 @@ error.projectId.missing=Project ID must be set for deploy.
 error.app.yaml.invalid.name=File name is not app.yaml: {0}
 error.app.yaml.not.a.file=Not a file: {0}
 error.app.yaml.non.existing=app.yaml does not exist.
-error.app.yaml.custom.runtime="runtime: custom" specified in app.yaml is not yet supported by \
- Cloud Tools for Eclipse.
-error.app.yaml.not.java.runtime=Runtime specified in app.yaml is not "java".
+error.app.yaml.custom.runtime="runtime: custom" is not yet supported by Cloud Tools for Eclipse.
+error.app.yaml.not.java.runtime=Runtime specified in app.yaml is not "java": {0}
```
elharo: Maybe

 runtime: {0} in app.yaml is not recognized.

PR #1935, opened 2017-05-10
use Files.new instead of new File...Stream

### LINKED ISSUE #1898: Prefer "Files.new{Input|Output}Stream()" over "new File{Input|Output}Stream()"
Excessive `new FileInputStream()` may contribute to a long GC pause: https://www.cloudbees.com/blog/fileinputstream-fileoutputstream-considered-harmful

There are several places where we use `new FileInputStream()`.

elharo:
@chanseokoh fix #1898

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
@@ -23,13 +23,11 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import java.io.ByteArrayInputStream;
 import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.OutputStreamWriter;
 import java.io.Reader;
+import java.io.Writer;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.logging.Logger;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
@@ -81,8 +79,7 @@ public class LibraryClasspathContainerSerializer {
       logger.warning("Container state file cannot be created, save failed");
       return;
     }
-    try (OutputStreamWriter out =
-        new OutputStreamWriter(new FileOutputStream(stateFile), StandardCharsets.UTF_8)) {
+    try (Writer out = Files.newBufferedWriter(stateFile.toPath(), StandardCharsets.UTF_8)) {
       SerializableLibraryClasspathContainer serializableContainer =
           new SerializableLibraryClasspathContainer(container,
               binaryArtifactBaseLocationProvider.getBaseLocation(),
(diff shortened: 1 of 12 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -81,8 +79,7 @@ public void saveContainer(IJavaProject javaProject, LibraryClasspathContainer co
       logger.warning("Container state file cannot be created, save failed");
       return;
     }
-    try (OutputStreamWriter out =
-        new OutputStreamWriter(new FileOutputStream(stateFile), StandardCharsets.UTF_8)) {
+    try (Writer out = Files.newBufferedWriter(stateFile.toPath(), StandardCharsets.UTF_8)) {
```
chanseokoh: Ah, this is convenient.

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
```diff
@@ -43,7 +43,8 @@ public void testToJsonFile() throws IOException {
     Path jsonFile = tempFolder.getRoot().toPath().resolve("credential-for-gcloud.json");
     CredentialHelper.toJsonFile(credential, jsonFile);
 
-    try (InputStream in = new FileInputStream(jsonFile.toFile());
+    try (InputStream in = Files.newInputStream(jsonFile);
+        // todo needs character set
```
chanseokoh: You can use `Files.newBufferedReader()` and give a charset, and the code will be further simplified as well.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/FileDownloader.java
```diff
@@ -86,8 +85,8 @@ public IPath download(URL url, IProgressMonitor monitor) throws IOException {
     connection.setReadTimeout(DEFAULT_READ_TIMEOUT_MS);
     connection.setRequestProperty("User-Agent", CloudToolsInfo.USER_AGENT);
     try (InputStream inputStream = new BufferedInputStream(connection.getInputStream());
-         OutputStream outputStream =
-             new BufferedOutputStream(new FileOutputStream(downloadedFile))) {
+         OutputStream outputStream = 
+             new BufferedOutputStream(Files.newOutputStream(downloadedFile.toPath()))) {
```
chanseokoh: I think the original code has some minor problem. I read somewhere that these streams should not be chained in a try-with-resources block to ensure closing in case it fails in the middle of the chain. That is, it should be

```java
try (...
    OutputStream out = new Files.newOutputStream(...);
    BufferedOutputStream bufferedOut = new BufferedOutputStream(out)) {
```

plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/CredentialHelperTest.java
elharo: This is a design flaw in the GSON API. It should accept an InputStream, not a Reader, and autodetect the specific Unicode encoding. We can work around this, but it would pull this PR way off on a tangent. For now since this is a test, I'd prefer to leave the behavior intact.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/FileDownloader.java
elharo: That doesn't sound right to me. The byte code should be the same either way. Do you have a reference?

Also note that the buffered stream constructors do not throw exceptions.

PR #1947, opened 2017-05-15
don't confirm sign out

### LINKED ISSUE #1946: Don't confirm sign out
This is a needless confirmation. Signing out is not an irreversible or destructive operation. The user can sign in again.

image

See http://cafe.elharo.com/ui/dontconfirm/

elharo:
@chanseokoh fix #1946


plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/GoogleLoginService.java
```diff
@@ -100,13 +100,11 @@ public GoogleLoginService() {}
...
     this(new GoogleLoginState(Constants.getOAuthClientId(), Constants.getOAuthClientSecret(),
-                              OAUTH_SCOPES, dataStore, uiFacade, loggerFacade),
-         dataStore, uiFacade, loggerFacade);
+                              OAUTH_SCOPES, dataStore, uiFacade, loggerFacade), uiFacade);
   }
 
   @VisibleForTesting
-  GoogleLoginService(GoogleLoginState loginState,
-      OAuthDataStore dataStore, LoginServiceUi uiFacade, LoggerFacade loggerFacade) {
+  GoogleLoginService(GoogleLoginState loginState, LoginServiceUi uiFacade) {
```
chanseokoh: Looks like we can also remove `uiFacade`, which in turn allows us to remove the field `loginServiceUi` from this class as well.

PR #1953, opened 2017-05-15
Minor code clean up

chanseokoh:


plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorXsltTest.java
```diff
@@ -60,14 +59,11 @@
   public void testApplyXslt()
       throws IOException, ParserConfigurationException, SAXException, TransformerException {
     try (InputStream xmlStream = stringToInputStream(WTP_METADATA_XML);
-        InputStream stylesheetStream = stringToInputStream(STYLESHEET)) {
-
+        InputStream stylesheetStream = stringToInputStream(STYLESHEET);
+        InputStream inputStream = Xslt.applyXslt(xmlStream, stylesheetStream)) {
       DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
```
elharo: there's a bug here. We need to call setNamespaceAware(true) on the factory.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorXsltTest.java
```diff
@@ -84,6 +85,14 @@ public void testWtpMetadataStylesheet_removesGpeRuntimeAndFacets()
     }
   }
 
+  private static Document parseXml(InputStream xml)
+      throws ParserConfigurationException, SAXException, IOException {
+    DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
```
elharo: This could shared as a static final field. No need to create a new factory for each request. It's designed to be thread-safe.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorXsltTest.java
```diff
@@ -56,38 +56,40 @@
       + "  </xsl:template>"
       + "</xsl:stylesheet>";
 
+  private static final Path wtpMetadataXslPath =
+      Paths.get("../com.google.cloud.tools.eclipse.appengine.compat/xslt/wtpMetadata.xsl");
+
+  private static DocumentBuilderFactory documentBuilderFactory;
+  static {
+    documentBuilderFactory = DocumentBuilderFactory.newInstance();
```
elharo: you can initialize in the same line that declares

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigratorXsltTest.java
```diff
@@ -56,38 +56,40 @@
...
+    documentBuilderFactory.setNamespaceAware(true);
```
elharo: @BeforeClass might work here. Not sure if it's preferable or not.

Issue #1941, opened 2017-05-12
Can't update Java facet to 1.8 in Maven project


elharo:
It immediately reverts to 1.7.

elharo:
Manually adding `<runtime>java8</runtime>` to appengine-web.xml makes this work, but this is not automatic as it is with a non-Maven project.

elharo:
Given that Java 8 support is still in whitelist beta, I'm inclined to declare this a non-release blocker.

elharo:
From the debugger it looks like maybe the runtime is added, but then the Java 1.8 facet add event is immediately followed by another event changing the facet back to 1.7.

elharo:
Hypothesis: when the facet is updated some job checks appengine-web.xml, sees there's no <runtime>java8</runtime> and resets the facet to Java 1.7.

elharo:
The problem is here:

```
  private void checkRuntimeElement(IFacetedProject project, IFile appEngineWebDescriptor,
      IProgressMonitor monitor) {
    try (InputStream input = appEngineWebDescriptor.getContents()) {
      boolean hasJava8Runtime = AppEngineDescriptor.parse(input).isJava8();
      boolean hasJava8Facet = project.hasProjectFacet(JavaFacet.VERSION_1_8);
      // if not the same, then we update the facet to match the appengine-web.xml
      if (hasJava8Facet != hasJava8Runtime) {
        Set<Action> updates = new HashSet<>();
        if (hasJava8Runtime) {
          updates.add(new Action(Action.Type.VERSION_CHANGE, JavaFacet.VERSION_1_8, null));
        } else {
          updates.add(new Action(Action.Type.VERSION_CHANGE, JavaFacet.VERSION_1_7, null));
        }
        logger.fine(getProject() + ": changing facets: " + updates);
        project.modify(updates, monitor);
      }
    } catch (SAXException ex) {
      // Parsing failed due to malformed XML; just don't check the value now.
    } catch (CoreException | IOException ex) {
      logger.log(Level.SEVERE, getProject() + ": error updating facets", ex);
    }
  }
```

comment out the line `updates.add(new Action(Action.Type.VERSION_CHANGE, JavaFacet.VERSION_1_7, null));` and this works. Not clear why this is only a problem for maven.

elharo:
Just noticed that adding the Java 1.8 facet to a maven project does *not* update appengine-web.xml with a `runtime` element. That may be the root cause here.

[...]

PR #1956, opened 2017-05-16
don't downgrade maven to 1.7

elharo:
@chanseokoh workaround #1941.

This may be temporary. I'm looking for a better fix that doesn't prevent downgrades.


plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineWebBuilder.java
```diff
@@ -86,7 +88,9 @@ private void checkRuntimeElement(IFacetedProject project, IFile appEngineWebDesc
         Set<Action> updates = new HashSet<>();
         if (hasJava8Runtime) {
           updates.add(new Action(Action.Type.VERSION_CHANGE, JavaFacet.VERSION_1_8, null));
-        } else {
+        } else if (!MavenUtils.hasMavenNature(project.getProject())) {
+          // see https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1941
+          // still not totally clear why this works for standard projects and not maven
```
chanseokoh: I'm not sure about this. This also disables a legitimate use case where a user manually removes `<runtime>java8</runtime>`. In this case, I think we need to downgrade.

chanseokoh: Chatted with @elharo. It's downright weird when a user's action to upgrade the facet version is reverted immediately against their will, so this is at least reasonable improvement.

PR #1957, opened 2017-05-16
don't run stylesheets on new project creation

elharo:


plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
```diff
@@ -65,10 +65,12 @@ public void handleEvent(IFacetedProjectEvent event) {
       logger.warning(project + ": cannot find appengine-web.xml");
         return;
       }
-    if (project.hasProjectFacet(JavaFacet.VERSION_1_8)) {
-      AppEngineDescriptorTransform.addJava8Runtime(descriptor);
-    } else {
-      AppEngineDescriptorTransform.removeJava8Runtime(descriptor);
+    if (event.getType() == Type.POST_VERSION_CHANGE) {
```
briandealwis: This only captures a user changing the java version, but won't capture a new-install.

briandealwis:
What the _stylesheet_?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/FacetChangeListener.java
elharo: yes, that's the intention

elharo: prior to this change, all of this fires on new project creation

chanseokoh: @briandealwis's point is that this change affects not only new project creation but also manual facet addition by users.

elharo: Anyone see a way to avoid this running ohn new projects but enable it on manual facet addition?

elharo: what if appengine-web.xml already has `<runtime>java8</runtime>` before they add the facet?

briandealwis: Now that you mention it, I do recall there being a facet-discovery mechanism via the `org.eclipse.wst.common.project.facet.core.detectors` extension mechanism.  We should have such a detector that looks for `runtime=java8`.

chanseokoh: > what if appengine-web.xml already has <runtime>java8</runtime> before they add the facet?

Shouldn't it just work? When whatever facet is added, it examines the current state and does the right thing. That is, it checks if the Java facet is installed and if it does, it either adds or removes the `<runtime>` element depending on if it is a Java 8 facet or not. And the XSLT stylesheet doesn't seem like it will duplicate `<runtime>`. (However, I haven't tested all of these.)

chanseokoh:
Was there a issue for this PR? Can't remember if we had a real problem.

In #2079, this class is being removed while `AppEngineStandardFacetChangeListener` is newly introduced in place of it. I think the new class will run the stylesheet as before, but probably it won't do harm.

chanseokoh:
And yes, in #2079, I do see that `AppEngineDescriptorTransform.removeJava8Runtime(descriptor)` is called which will run the stylesheet when creating a new project, which is unnecessary but harmless.

elharo:
There is a real issue for this PR. Converting a GPE project may change appengine-web.xml without an explicit user request. This prevents that.

chanseokoh:
Looks like it will be taken care of as part of #2079, because in that case the conversion will fail with an error message instead of adding `<runtime>java8</runtime>`.

elharo:
I don't why the conversion should fail. That may be a big in 2079

chanseokoh:
I did some testing with #2079, and below is the current state of it regarding converting existing projects:

| No. | project JRE (Java facet version) | WTP facet version | `java8` in `appengine-web.xml`? | App Engine facet version after conversion |
| --- | --- | --- | ---- | --- |
| 1 | JavaSE-1.7 (Java 7)  | 2.5  | no | JRE7 |
| 2 | JavaSE-1.7 (Java 7)  | 2.5  | yes | JRE8 (Java facet remains at 7) |
| 3 | JavaSE-1.7 (Java 7)  | 3.0  | no | fail |
| 4 | JavaSE-1.7 (Java 7)  | 3.0  | yes | fail (might be a bug) |
| 5 | JavaSE-1.8 (Java 8)  | 2.5 | no | fail |
| 6 | JavaSE-1.8 (Java 8)  | 2.5 | yes | fail (sounds like a bug) |
| 7 | JavaSE-1.8 (Java 8)  | 3.0 | no | fail |
| 8 | JavaSE-1.8 (Java 8)  | 3.0 | yes | fail (sounds like a bug) |

Failing in some cases could be argued: should we add `java8` to `appengine-web.xml` instead to make it succeed?

However, the fact that conversion always fails when the Java facet is 8 even if `appengine-web.xml` has `<runtime>java8</runtime>` seems like a bug. @briandealwis thoughts?

briandealwis:
Thank you for systematic testing.  I can only assume that the "_java8 in appengine-web.xml?_" column is an input condition? (I've labelled them as such.)  So these are testing the case "convert an existing project that happen to have an `appengine-web.xml`", right?

Right now the `installAppEngineFacet()` logic is (well, supposed to be):
  1. try to detect as an AES 7 or 8 project from the `appengine-web.xml` using the facet detectors
  2. if no AES facet, then:
     1. check if AES 7 works with no conflict (i.e., with existing facets) and use it
     2. check if there's another AES facet that works and use it
     3. if no Java facet, add highest satisfying Java facet version (will take into account AES version)
     4. if no DWP facet, add highest satisfying DWP facet version (will take into account AES version)

So @chanseokoh's tests should all be handled by Step 1 (the detectors), whereas the `ConversionTests` test Step 2.

chanseokoh:
>  I can only assume that the "java8 in appengine-web.xml?" column is an input condition? (I've labelled them as such.) So these are testing the case "convert an existing project that happen to have an appengine-web.xml", right?

Correct.

> So @chanseokoh's tests should all be handled by Step 1 (the detectors),

So you're saying the Step 1, the detector, is not handling the case 4, 6, and 8 properly, as it should have added AES8, right?

briandealwis:
> @chanseokoh: So you're saying the Step 1, the detector, is not handling the case 4, 6, and 8 properly, as it should have added AES8, right?

Just so I'm clear, your test approach was:

  1. create a project, set the facets, and ensure there is no AES facet
  2. then use _Properties > Configure > Convert to App Engine Standard Project_

And _"fail"_ means there was a failure dialog?  Or were there other modes of failure?

The `AppEngineStandardProjectConvertCommandHandler` has some hard-coded verification logic that I need to relax.

chanseokoh:
For example, for the case 8

1. Create a plain Java project.
2. Add the Java 8 facet, the WTP 3.0 facet, and the AES 8 facet. _Apply_.
3. Remove the AES 8 facet. _Apply_.
4. _Configure_ > _Convert to App Engine Standard Project_

Then you'll see

incompatible java facet version _037

briandealwis:
Ok, thanks. If I comment out the `AppEngineStandardProjectConvertCommandHandler` verification and make the conversion job a user job (`setUser(true)`) then any non-OK status returns are surfaced to the user.  For example, 

screen shot 2017-06-23 at 11 03 44 am

Not the friendliest error message.

chanseokoh:
It seems like it's trying to install JRE7 in this case (as well as the case 4 and 6). I wonder why, because at least the case 2 correctly installs JRE8.

briandealwis:
Oops, that case had no `<runtime>` element (so should be JRE7) and DWP 3.0.  So the error is correct.

Ok, I can change the `checkFacetCompatibility()` to instead check if the settings conflict with any of the _App Engine Standard Facet_ versions and provide a nicer message.

briandealwis:
Closing as fixed by #2079.

PR #1968, opened 2017-05-18
Clean up code

### LINKED ISSUE #1974: Investigate why AppEngineStandardProjectWizard.validateDependencies returns IStatus that is not being used
The intention seems to be that [`AppEngineStandardProjectWizard.validateDependencies()` returns `IStatus` that contains a UI message in case of error](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.1.0B/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/standard/AppEngineStandardProjectWizard.java#L66).

However, [it ends up just checking if the returned `IStatus` is OK or not](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.1.0B/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java#L90). The UI message is never utilized. Perhaps [we should call `StatusUtil.setErrorStatus()` as we currently do a bit below the code](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.1.0B/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java#L118).

chanseokoh:
- We've gutted out data binding from the wizard; no reason to consider `Realm` in tests.
- ~~Renamed `Coordinates` to a singular form.~~
- `groupIdField.getText().trim()` is precisely `getGroupId()`.

Fixes #1974 too.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.maven/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenAppEngineStandardWizardPage.java
```diff
@@ -180,34 +180,34 @@ public void widgetSelected(SelectionEvent event) {
     });
   }
 
-  /** Create UI for specifying desired Maven Coordinates */
-  private void createMavenCoordinatesArea(Composite container, ModifyListener pageValidator) {
-    Group mavenCoordinatesGroup = new Group(container, SWT.NONE);
-    mavenCoordinatesGroup.setText(Messages.getString("MAVEN_PROJECT_COORDINATES")); //$NON-NLS-1$
-    GridDataFactory.defaultsFor(mavenCoordinatesGroup).span(2, 1).applyTo(mavenCoordinatesGroup);
-    GridLayoutFactory.swtDefaults().numColumns(2).applyTo(mavenCoordinatesGroup);
+  /** Create UI for specifying desired Maven coordinate */
```
elharo: Standard usage in the Maven docs is plural. A groupId would be a singular coordinate but a groups ID, artifact ID, and version are three coordinates.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
```diff
@@ -86,8 +85,7 @@ public boolean performFinish() {
 
     boolean fork = true;
     boolean cancelable = true;
-    IStatus status = validateDependencies(fork, cancelable);
```
elharo: I don't see much benefit to inlining this. It's eaier to debug if you don't.

chanseokoh: Yeah, this one can be left unchanged.

chanseokoh: I actually noticed a new issue (#1974). Fixing it here.

PR #1969, opened 2017-05-18
Remove @VisibleForTesting by manually searching for controls in tests

chanseokoh:
Removes `@VisibleForTesting` on UI widgets, and extends a test util `CompositeUtil` to search for them.


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/CompositeUtil.java
```diff
@@ -45,4 +46,35 @@ public static Control findControl(Composite composite, Predicate<Control> predic
     }
     return null;
   }
+
+  /**
+   * Finds a {@link Control} of type {@code type} that comes after a {@link Label} with a text
+   * {@code label}.
+   */
+  @SuppressWarnings("unchecked")
+  public static <T> T findControlAfterLabel(Composite composite, Class<T> type, String label) {
```
elharo: How are SWT widgest ordered?

chanseokoh: Seems like there exists the notion of "drawing order" or "z-order" in SWT, AWT, etc. The `getChildren()` Javadoc says
```
/**
 * Returns a (possibly empty) array containing the receiver's children.
 * Children are returned in the order that they are drawn.  The topmost
 * control appears at the beginning of the array.  Subsequent controls
 * draw beneath this control and appear later in the array.
 * ...
 * @see Control#moveAbove
 * @see Control#moveBelow
```
It also mentions `moveAbove()` and `moveBelow()` that can explicitly manipulate that order. Basically, I see that the order is that of adding controls by default, by which you control positioning components:
https://stackoverflow.com/a/9254219/1701388
https://stackoverflow.com/a/2780964/1701388

Issue #1897, opened 2017-05-02
Flaky test on Kokoro Windows: DebugNativeAppEngineStandardProjectTest.testDebugLaunch()


chanseokoh:
I think this is not the first time I saw this on Kokoro Windows. There is a `file not found` error, and I wonder if it is related to the Windows-specific environment.

UPDATE: I _think_ the preference loading exception is not related.
```
!ENTRY org.eclipse.core.resources 4 4 2017-05-02 08:00:54.943
!MESSAGE Exception loading preferences from: /testapp/.settings/org.eclipse.jdt.core.prefs.
!STACK 1
org.eclipse.core.runtime.CoreException: File not found: T:\src\github\google-cloud-eclipse\plugins\com.google.cloud.tools.eclipse.integration.appengine\target\work\data\testapp\.settings\org.eclipse.jdt.core.prefs.
	at org.eclipse.core.internal.filesystem.Policy.error(Policy.java:45)
	at org.eclipse.core.internal.filesystem.local.LocalFile.openInputStream(LocalFile.java:385)
	at org.eclipse.core.internal.localstore.FileSystemResourceManager.read(FileSystemResourceManager.java:849)
	at org.eclipse.core.internal.resources.File.getContents(File.java:277)
	at org.eclipse.core.internal.resources.ProjectPreferences.load(ProjectPreferences.java:507)
	at org.eclipse.core.internal.resources.ProjectPreferences.load(ProjectPreferences.java:492)
	at org.eclipse.core.internal.preferences.EclipsePreferences.create(EclipsePreferences.java:370)
	at org.eclipse.core.internal.preferences.EclipsePreferences.internalNode(EclipsePreferences.java:623)
	at org.eclipse.core.internal.preferences.EclipsePreferences.node(EclipsePreferences.java:766)
	at org.eclipse.core.resources.ProjectScope.getNode(ProjectScope.java:69)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaEditor$EclipsePreferencesAdapter.getNode(JavaEditor.java:363)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaEditor$EclipsePreferencesAdapter.contains(JavaEditor.java:383)
	at org.eclipse.ui.texteditor.ChainedPreferenceStore.getVisibleStore(ChainedPreferenceStore.java:463)
[...]

PR #1529, opened 2017-03-06
Add support for switching to recommended perspective

briandealwis:
Makes the J2EE perspective the recommended perspective (the _final_ perspective) for our New Project wizards, but marks the J2EE and Web perspectives as being acceptable (the _preferred_ perspectives).

No explicit test for this, but I had to change the integration tests to switch to the J2EE perspective as the would-you-like-to-switch-perspective dialog interfered.

PR #1975, opened 2017-05-22
Log debug info for #1897

chanseokoh:
For #1897. `SWTBotTreeItem.toString()` will do the job.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/DebugNativeAppEngineStandardProjectTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/DebugNativeAppEngineStandardProjectTest.java
@@ -104,6 +104,13 @@ public class DebugNativeAppEngineStandardProjectTest extends BaseProjectTest {
     SwtBotTreeUtilities.waitUntilTreeHasItems(bot, launchTree);
     SWTBotTreeItem[] allItems = launchTree.getAllItems();
     SwtBotTreeUtilities.waitUntilTreeHasText(bot, allItems[0]);
+
+    // Log debug info for https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1897
+    if (!allItems[0].getText().contains("App Engine Standard at localhost")) {
+      for (SWTBotTreeItem treeItem : allItems) {
+        System.out.println("    [GitHub #1897 debug info] " + treeItem);
+      }
+    }
     assertTrue("No App Engine launch found",
         allItems[0].getText().contains("App Engine Standard at localhost"));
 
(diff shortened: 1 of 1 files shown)


chanseokoh:
This is funny. The Kokoro failure of this PR triggered #1897 (very opportunistic), but there was no debug info logged by this PR. This doesn't make sense, so I'm a bit perplexed.

chanseokoh:
Ah, okay, the log was quite a bit more above the assertion failure message:
```
INFO: submitting command: C:\Program Files\Java\jdk1.8.0_131\bin\java.exe -Xdebug -Xrunjdwp:transport=dt_socket,server=n,suspend=y,quiet=y,address=52119 -javaagent:C:\Program Files (x86)\Google\Cloud SDK\google-cloud-sdk\platform\google_appengine\google\appengine\tools\java\lib\agent\appengine-agent.jar -Dappengine.sdk.root=C:\Program Files (x86)\Google\Cloud SDK\google-cloud-sdk\platform\google_appengine\google\appengine\tools\java -cp C:\Program Files (x86)\Google\Cloud SDK\google-cloud-sdk\platform\google_appengine\google\appengine\tools\java\lib\appengine-tools-api.jar com.google.appengine.tools.development.DevAppServerMain --address=localhost --port=8080 --allow_remote_shutdown --disable_update_check T:\src\github\google-cloud-eclipse\plugins\com.google.cloud.tools.eclipse.integration.appengine\target\work\data\.metadata\.plugins\org.eclipse.wst.server.core\tmp0\testapp
    [GitHub #1897 debug info] TreeItem with text {}
12:45:19.861 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'OK' and with style 'SWT.PUSH')
```
There certainly exists a tree item, but it's text is empty. Maybe a race condition: we wait until there becomes one tree item, but we check its text too early before the UI thread sets a proper text.

briandealwis:
There may have been an execution error with launching devappserver.  As part of trying to diagnose the very strange test problems I encountered in #1529, I had [added some support to the tests to dump the console output from launching the devappserver](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1529/files#diff-389b5df3ae52f2bf1d8d52e9bb75eda9R134).

(I never did get to the bottom of #1529)

chanseokoh:
@briandealwis That's very interesting.

Before that, what seems weird though is that the test calls `SwtBotTreeUtilities.waitUntilTreeHasText(bot, allItems[0])`, which is supposed to wait until `allItems[0]` has a non-empty text. However, it end up seeing an empty text, so I think the method is either not functioning properly or timing out.

briandealwis:
> or timing out

As might happen if the [launched JVM never actually ran for some reason](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1529#issuecomment-285582559)…

briandealwis:
I wonder if `ProjectUtils#waitForProjects()` shouldn't instead forcibly trigger a rebuild via `IWorkspace#build()` rather than wait for the auto build job.

chanseokoh:
Hopefully #1897 is fixed by #2047. Closing.

Issue #1716, opened 2017-04-03
Maven flex support


elharo:
The flex new project wizard should optionally generate a pom.xml and set up the project to use the maven builder.

chanseokoh:
Just noticed that a single checkbox to create `pom.xml` isn't enough; `pom.xml` requires `<groupId>` and `<artifactId>`.

chanseokoh:
Made some progress, but I noticed that merely adding a `pom.xml` doesn't seem enough. For example, we currently embed a servlet-api library JAR under `lib/` as a native flex project, but for a Maven-based project, we need to do some extra stuff around adding/resolving dependencies in the Maven way. That said, I don't fully understand the code and the mechanisms that we do for the standard Maven project, so it'll take some time to look into how they work and what should be minimally done.

Progress summary:
- Added a Maven coordinates UI to the flex wizard. (Code reused after refactored from the std Maven wizard.) The UI is fully working and enabled/disabled based on the _Create as Maven Project_ check box.
- The wizard adds `pom.xml` with proper Maven coordinates user input. (`pom.xml` does need some validation before we ship it.)
- Added util methods to add a Maven nature.

chanseokoh:
Also need to set a proper output directory for `.class` files (i.e., default Java output folder) when using Maven. For now, the current output folder for flex projects is `<project>/build/classes`. Per Maven convention, this should be `<project>/target/<artifact ID>-<version>/WEB-INF/classes`.

PR #1979, opened 2017-05-23
Factor out Maven coordinates UI widget from the Maven App Engine Standard project wizard

chanseokoh:
Factoring out the Maven coordinates part to reuse it when implementing #1716.

It follows the same pattern with the libraries group UI (`LibrarySelectorGroup`) being used in the same wizard. The constructor `MavenCoordinatesUi` creates all the controls.


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenCoordinatesUi.java
```diff
@@ -0,0 +1,175 @@
...
+    versionLabel.setEnabled(checked);
+    versionField.setEnabled(checked);
+  }
+
+  /**
+   * @return {@link IStatus#OK} if there was no validation problem or the UI is disabled; otherwise
+   *     a status describing a validation problem (with a non-OK status)
+   */
+  public IStatus validateMavenSettings() {
+    boolean uiDisabled = asMavenProjectButton != null && !asMavenProjectButton.getSelection();
```
elharo: can you inline this?

PR #1986, opened 2017-05-24
add wizard title

### LINKED ISSUE #1981: Flex project wizard does not have title
selection_017

The Standard wizard has one:
selection_018

elharo:
@chanseokoh fix #1981

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/AppEngineFlexProjectWizardTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/AppEngineFlexProjectWizardTest.java
@@ -25,7 +25,7 @@ public class AppEngineFlexProjectWizardTest {
 
   @Test
   public void testWindowTitle() {
-    Assert.assertEquals("New App Engine flexibole environment Project", wizard.getWindowTitle());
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/messages.properties
```diff
@@ -1,4 +1,5 @@
 new.app.engine.standard.project=New App Engine Standard Project
+new.app.engine.flex.project=New App Engine Flex Project
```
chanseokoh: I think we should not abbreviate flex.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/AppEngineFlexProjectWizardTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AppEngineFlexProjectWizardTest {
+
+  private AppEngineFlexProjectWizard wizard = new AppEngineFlexProjectWizard();
+
+  @Test
+  public void testWindowTitle() {
+    Assert.assertEquals("New App Engine flexibole environment Project", wizard.getWindowTitle());
```
chanseokoh: Looks like this test will fail.

PR #1987, opened 2017-05-24
handle staging failure

elharo:
@chanseokoh


chanseokoh:
I'm curious. Did you hit some issue? How and what was it? Can you provide some context?

elharo:
No, I just happened to be looking at this code and it struck me as very strange that staging, which involves a lot of I/O, always succeeded. Digging deeper into appengine-plugins-core I realized this doesn't always succeed but we missed that because AppEngineException (and OperationCanceledException) are runtime exceptions.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/flex/FlexStagingDelegate.java
```diff
@@ -36,6 +39,7 @@ public FlexStagingDelegate(IPath appEngineDirectory) {
     this.appEngineDirectory = appEngineDirectory;
   }
 
+  // todo why is the cloudSdk argument not used here?
```
chanseokoh: Reason: the standard staging shells out to Cloud SDK whereas appengine-plugins-core does all copying directly via Java code. So, the interface being overridden is like a least common super set for both std and flex, which ends up having the Cloud SDK arg. There may be a way to improve it, but I don't have an immediate idea now.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/flex/FlexStagingDelegateTest.java
```diff
@@ -53,18 +54,17 @@ public void setUp() {
...
     assertTrue(stagingDirectory.append("app-to-deploy.war").toFile().exists());
     assertTrue(stagingDirectory.append("app.yaml").toFile().exists());
+    assertTrue(status.isOK());
   }
 
   @Test
-  public void testGetOptionalConfigurationFilesDirectory() throws CoreException {
+  public void testGetOptionalConfigurationFilesDirectory() {
     StagingDelegate delegate = new FlexStagingDelegate(appEngineDirectory);
-    delegate.stage(project, stagingDirectory, safeWorkDirectory, null /* cloudSdk */,
```
chanseokoh: Great.

chanseokoh:
What do you think of catching `AppEngineException` [here](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.1.0B/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java#L186) instead of catching it inside `StagingDelegate`? (The `OperationCanceledException` is already being caught.)

Otherwise, it turns out `StandardStagingDelegate` has the same problem of always returning `IStatus.OK`, so I think that also needs to be changed to be consistent. (appengine-plugins-core's standard staging can throw `AppEngineException` too like flex staging.)

elharo:
IMHO, a method should either return a status or throw an exception, not both. That is, Status.ERROR is an alternative to an exception. 

The catch of OperationCanceledException where you indicate it is a mistake because it turns a cancellation into an error. Cancellation is neither success nor failure. It's a third state that we need to handle so that Eclipse will report it properly. In other words, if the user chooses to cancel Eclipse should not pop up the same dialog it does if staging encounters a full file system.

chanseokoh:
Sounds about right. Could you fix the `StandardStagingDelegate.stage()` too similarly? appengine-plugins-core throws `AppEngineException` there.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/flex/FlexStagingDelegate.java
```diff
@@ -38,16 +42,27 @@ public FlexStagingDelegate(IPath appEngineDirectory) {
...
-    return Status.OK_STATUS;
+    try {
+      WarPublisher.publishWar(project, war, subMonitor.newChild(40));
+      CloudSdkStagingHelper.stageFlexible(appEngineDirectory, war, stagingDirectory,
+          subMonitor.newChild(60));
+      return Status.OK_STATUS;
+    } catch (AppEngineException ex) {
+      return StatusUtil.error(this, Messages.getString("deploy.job.staging.failed"), ex);
+    } catch (CoreException ex) {
+      return StatusUtil.error(this, "war publishing failed", ex);
```
chanseokoh: `Messages.getString("war.publishing.failed");`

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/flex/FlexStagingDelegate.java
```diff
@@ -38,16 +42,27 @@ public FlexStagingDelegate(IPath appEngineDirectory) {
...
   @Override
   public IStatus stage(IProject project, IPath stagingDirectory, IPath safeWorkDirectory,
-      CloudSdk cloudSdk, IProgressMonitor monitor) throws CoreException {
+      CloudSdk cloudSdk, IProgressMonitor monitor) {
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
 
-    stagingDirectory.toFile().mkdirs();
+    boolean result = stagingDirectory.toFile().mkdirs();
+    if (!result) {
+      return StatusUtil.error(this, "Could not create staging directory " + stagingDirectory);
```
chanseokoh: Localize this too?

PR #2004, opened 2017-06-05
"Files.new{Input|Output}Stream" instead of "new File{Input|Output}Stream"

chanseokoh:
There have been missing places for #1898.


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -197,14 +196,11 @@ static IFolder createChildFolder(String name, IFolder parent, IProgressMonitor m
   @VisibleForTesting
   static IFile createChildFile(String name, String template, IContainer parent,
       Map<String, String> values, IProgressMonitor monitor) throws CoreException {
-
     monitor.subTask("Creating file " + name);
 
     IFile child = parent.getFile(new Path(name));
     if (!child.exists()) {
-      child.create(new ByteArrayInputStream(new byte[0]), true /* force */, monitor);
```
elharo: I guess this line isn't needed?

chanseokoh: Yes, turns out this is not necessary.

PR #2006, opened 2017-06-05
appengine-plugins-core 0.3.5

elharo:


eclipse/ide-target-platform/category.xml
```diff
@@ -34,5 +34,5 @@
    <bundle id="com.fasterxml.jackson.core.jackson-core" version="0.0.0"/>
    <bundle id="ch.qos.logback.slf4j" version="0.0.0"/>
    <bundle id="org.slf4j.log4j" version="0.0.0"/>
-   <bundle id="org.yaml.snakeyaml" version="1.17.0"/>
+   <bundle id="org.yaml.snakeyaml" version="1.17"/>
```
chanseokoh: These version changes are not necessary as they are an OGSi version rather than a Maven artifact version. However, this would still work (I know patch versions are not mandatory here at least), so we're good.

PR #2007, opened 2017-06-05
Send AnalyticsPing for flex project wizard

### LINKED ISSUE #1671: Add analytics ping for creating a new flex project

chanseokoh:
Fixes #1671.

I also cleaned up the code, factoring in #1716.
- Now it's only the Wizard class that sends pings. (Previously, the code was scattered around too much and confusing.)
   1. `AppEngineProjectWizard` that is common to both std and flex.
   1. `MavenArchetypeProjectWizard` is another wizard, but this wizard is expected to be removed when we have a native-Mative-unified wizard.
- I changed to send a "completion" ping only after successfully creating a project. Previously, it was sending when a user pressed _Finish_.
- Some refactoring.


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
```diff
@@ -100,6 +100,16 @@ public boolean performFinish() {
 
     config.setAppEngineLibraries(page.getSelectedLibraries());
 
+    // TODO: we won't need this "instanceof" check once we implement a unified wizard:
+    // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1326
+    if (page instanceof AppEngineFlexWizardPage) {
```
chanseokoh: FYI, this is a refactoring that moves the code from the flex subclass.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
```diff
@@ -40,7 +40,7 @@
 
 public abstract class AppEngineProjectWizard extends Wizard implements INewWizard {
 
-  private AppEngineWizardPage page = null;
+  protected AppEngineWizardPage page = null;
```
elharo: See Effective Java 2E, p. 69. This is borderline and may affect thread safety. Making this private with a final setter and getter would make me a lot less nervous.

chanseokoh: While I was fixing this, I noticed that we explicitly disable the _Finish_ button on the Cloud SDK validation errors pages, so it can never get to `performFinish()`. I'll leave this field `private`.

PR #2027, opened 2017-06-08
Fix release version

chanseokoh:


briandealwis:
Hmm, the category still isn't appearing.  Investigating.

elharo:
Where/what exactly are you not seeing? 

image

image

chanseokoh:
@briandealwis try re-installing. I see 1.2.0.

briandealwis:
Huh, in my local build, the category IU isn't copied across into the target repository.  That is:
```
$ unzip -p gcp-repo/target/repository/content.jar | grep "<unit.*id='com.google.cloud.tools.eclipse.category'"
```
is empty **unless** I change the `requires.1.range` to a tight bound.

I don't know why it's working in the main builds.

Anyways, I have a patch that will pull the version from the pom.xml so that we don't have to modify the p2.inf for future releases.

briandealwis:
Ah, it works in the release build as `release/release_after_kokoro.sh` re-publishes the signed features and bundles, and then re-publishes the `metadata.product` (which pulls in the `metadata.p2.inf`).

elharo:
So, no need to hold this release, right?

briandealwis:
Good to go 👍

PR #2032, opened 2017-06-08
Add WTP library container when installing App Engine standard

### LINKED ISSUE #2028: Our project wizards don't add one of WTP classpath containers
See https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2018#issuecomment-307149093

chanseokoh:
Fixes #2028.

This takes care of both standard and flex.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
@@ -132,7 +132,7 @@ public class FacetUtil {
     IDataModel webModel = DataModelFactory.createDataModel(new WebFacetInstallDataModelProvider());
     webModel.setBooleanProperty(IJ2EEModuleFacetInstallDataModelProperties.ADD_TO_EAR, false);
     webModel.setBooleanProperty(IJ2EEFacetInstallDataModelProperties.GENERATE_DD, true);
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
```diff
@@ -132,7 +132,7 @@ public FacetUtil addWebFacetToBatch(IProjectFacetVersion webFacet) {
     IDataModel webModel = DataModelFactory.createDataModel(new WebFacetInstallDataModelProvider());
     webModel.setBooleanProperty(IJ2EEModuleFacetInstallDataModelProperties.ADD_TO_EAR, false);
     webModel.setBooleanProperty(IJ2EEFacetInstallDataModelProperties.GENERATE_DD, true);
-    webModel.setBooleanProperty(IWebFacetInstallDataModelProperties.INSTALL_WEB_LIBRARY, false);
+    webModel.setBooleanProperty(IWebFacetInstallDataModelProperties.INSTALL_WEB_LIBRARY, true);
```
briandealwis: Is it worth setting `INSTALL_WEB_LIBRARY` too? (I think it defaults to true.)
Or if `INSTALL_WEB_LIBRARY` is true by default, should we remove it here?

(I'm thinking being more explicit is useful?)

chanseokoh: Hmm, good point. I will see if I can remove it.

PR #2041, opened 2017-06-09
line break after XML declaration

### LINKED ISSUE #1940: appengine-web.xml doesn't have line break after XML declaration
when creating a new project

```
<?xml version="1.0" encoding="UTF-8"?><appengine-web-app xmlns="http://appengine.google.com/ns/1.0">

  <threadsafe>true</threadsafe>
```

should be 

```
<appengine-web-app xmlns="http://appengine.google.com/ns/1.0">

  <threadsafe>true</threadsafe>
```

elharo:
@chanseokoh fix #1940

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/xslt/addJava8Runtime.xsl
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/xslt/addJava8Runtime.xsl
@@ -7,7 +7,6 @@
   xmlns:appengine="http://appengine.google.com/ns/1.0"
   xmlns="http://appengine.google.com/ns/1.0">
 
-
   <xsl:template match="/">
   <xsl:text>
 </xsl:text>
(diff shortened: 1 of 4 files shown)


chanseokoh:
On Kokoro, the new test `assertTrue(contents.contains("?>\n"));` fails:

```
testApply(com.google.cloud.tools.eclipse.appengine.validation.XsltSourceQuickFixTest)  Time elapsed: 5.407 sec  <<< FAILURE!
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.google.cloud.tools.eclipse.appengine.validation.XsltSourceQuickFixTest.testApply(XsltSourceQuickFixTest.java:81)
```

I don't think this is due to flakiness, because the assertion right above it (`assertFalse(contents.contains("application"));`) has never failed. It may have something to do with the newline character `\n` on Windows.

chanseokoh:
Maybe the easiest solution is to change the test to this:

`assertFalse(contents.contains("?><"));`

PR #2043, opened 2017-06-09
Fix gcp-repo/metadata.p2.inf so that category is published in normal Maven builds

briandealwis:
Our CT4E category doesn't appear in the `gcp-repo` p2 repository when built from Maven.  It does appear in our release builds as we explicitly re-publish artifacts as part of the release.  

This fixes our `metadata.p2.inf` to specify an exact range to pull in the category definition.  It also changes some naming to distinguish between the category IU and the capabilities: 
- Use `$version$` to use the version of the `.product`: no need to change the `.p2.inf` file for releases
- Append `.definition` to the capability name, used to pull in the category IU from the product IU, to avoid confusion with the actual category IU.  This capability is required as Tycho's [`assemble-repository` plugin only traverses the _modules_ actually defined in this project](https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html), and it doesn't track the additional IUs defined in the `.p2.inf`.
- Change the category capability to always be version 1.0.0
- Change the requirement to be an exact version

(`assemble-repository`'s docs](https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html) note that it only follows IUs that are included, and 
> Dependencies with a strict version range, i.e. a range which only matches exactly one version of an artifact, are also considered as inclusions.
And perhaps the `[1.0.0,1.2.0]` range is considered open-ended, vs the `[1.0.0,1.0.0]` which is an exact specification.

Identified as part of #2027

### DIFF
--- a/gcp-repo/metadata.p2.inf
+++ b/gcp-repo/metadata.p2.inf
@@ -8,13 +8,17 @@
 # Follows the approach shown in
 # http://help.eclipse.org/neon/topic/org.eclipse.platform.doc.isv/guide/p2_customizing_metadata.html?cp=2_0_20_1_2#categoryGen
 
+# Pull in the IU that provides a definition; the definition version
+# is never shown in the UI and doesn't matter
 requires.1.namespace=org.eclipse.equinox.p2.iu
-requires.1.name=com.google.cloud.tools.eclipse.category
-requires.1.range=[1.0.0,1.2.0]
+requires.1.name=com.google.cloud.tools.eclipse.category.definition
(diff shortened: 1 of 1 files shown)


chanseokoh:
Can you update the first step of the release instructions?

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/wiki/Cloud-Tools-for-Eclipse-Release-Instructions

Issue #2044, opened 2017-06-09
It is possible to install Dynamic Web Module facet 3.0 with Java 7 and App Engine standard facet


chanseokoh:
Steps to reproduce:

1. Create a plain Java project with JRE 1.7.
2. Go to the _Project Facets_ project properties page. Enable facets.
3. Check App Engine Standard facet with Dynamic Web Module facet 3.0. Click _Apply_.

chanseokoh:
Related: #2034.

chanseokoh:
I will put `@Ignore` to the tests mentioned in #2049. The annotations must be deleted when this is fixed.

briandealwis:
Fixed by #2079.

Issue #2048, opened 2017-06-09
Kokoro flaky build failure: testGetProject()


chanseokoh:
This is the first time I saw this failure, so I think this happens pretty rarely.

```
testGetProject(com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidatorTest)  Time elapsed: 0.01 sec  <<< ERROR!
org.eclipse.core.internal.resources.ResourceException: Problems encountered while deleting resources.
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:773)
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:730)
	at com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidatorTest.testGetProject(XmlSourceValidatorTest.java:197)
```

chanseokoh:
Blocked by #2049.

chanseokoh:
Interestingly, just saw another Kokoro failure with this.

```
testGetFile(com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidatorTest)  Time elapsed: 0.016 sec  <<< ERROR!
org.eclipse.core.internal.resources.ResourceException: Problems encountered while deleting resources.
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:773)
	at org.eclipse.core.internal.resources.Resource.delete(Resource.java:730)
	at com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidatorTest.testGetFile(XmlSourceValidatorTest.java:175)

testGetProject(com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidatorTest)  Time elapsed: 0.016 sec  <<< ERROR!
org.eclipse.core.internal.resources.ResourceException: Resource '/test0.7604933568050114/testdata.xml' already exists.
	at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:314)
	at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:301)
	at org.eclipse.core.internal.resources.File.create(File.java:107)
	at com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidatorTest.testGetProject(XmlSourceValidatorTest.java:184)
```

PR #2050, opened 2017-06-10
Load .java8 bundle in .validation.test / ignore failing tests blocked by another bug

### LINKED ISSUE #2049: False positives in tests: XmlSourceValidatorTest.testValidate() and WebXmlValidatorTest.testValidateJavaServlet()
As it has been, the test succeeds on Travis and locally. However, when running inside Eclipse as a JUnit plug-in test, it fails. It may have something to do with the fact that `mvn verify` doesn't load all bundles that are available when running tests in Eclipse.

chanseokoh:
Fixes #2049.

In order for these two tests in `.validation.test` to work properly, we need to load the `.java8` bundle, because the `.java8` bundle additionally defines constraints between Servlet versions and Java versions.

However, because the bug in defining the constraints in `.java8` (#2044), loading the `.java8` bundle will make these tests correctly fail. Until #2044 is fixed, we need to `@Ignore` them. I made a comment in #2044 to remove these `@Ignore` annotations when fixing #2044.


chanseokoh:
@braindealwis may have some restructuring idea to fix #2044. @braindealwis what do you think? This is blocking #2048.

briandealwis:
I was holding off on this, hoping that I could make it obsolete with #2044 and friends. As that will take longer…

That said: wouldn't it be better to use `Assume` in the test rather than ignore the tests completely?

chanseokoh:
Not immediately clear how we could express the `Assume`. The end state of these tests seems to depend highly on the outcome of #2044, as these tests are asking "is it a conflict to specify 3.1 in `web.xml` if the project has an App Engine Standard facet"?

briandealwis:
Something like:
```java
    Assume.assumeTrue("Cannot be run with Java 8 support",
        AppEngineStandardFacet.FACET_VERSION.conflictsWith(JavaFacet.VERSION_1_8));
```

and

```java
    Assume.assumeTrue("Cannot be run with Java 8 support",
        AppEngineStandardFacet.FACET_VERSION.conflictsWith(WebFacetUtils.WEB_31));
```

chanseokoh:
Changed to use `Assume()`. However, when fixing #2044, these tests should be fixed together so that they run meaningful tests.

PR #2058, opened 2017-06-13
Fix Mars Eclipse target platform

chanseokoh:


chanseokoh:
:sigh: `[INFO] ide-target-platform.repo ........................... FAILURE [  1.722 s]`

I'm guessing the AERI 2.0.5 brings in guava 21.0, which is banned per our configuration.

```
[INFO] Building ide-target-platform.repo 0.1.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- tycho-packaging-plugin:1.0.0:build-qualifier-aggregator (default-build-qualifier-aggregator) @ ide-target-platform.repo ---
[INFO] The project's OSGi version is 0.1.0.201706131207
[INFO] 
[INFO] --- maven-enforcer-plugin:1.4.1:enforce (default) @ ide-target-platform.repo ---
[WARNING] Rule 1: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
Found Banned Dependency: p2.eclipse-plugin:com.google.guava:jar:21.0.0.v20170206-1425
Use 'mvn dependency:tree' to locate the source of the banned dependencies.
```

chanseokoh:
> LGTM. For future reference. I would like to understand the root cause. The binaries from version 2.0.4 got removed from the Orbit site? Is this likely to happen to us again in the future?

I don't fully understand it either. @briandealwis would be able to explain it.

briandealwis:
For Neon and Oxygen, we fetch AERI from the central release site (http://download.eclipse.org/releases/neon).  But AERI 2.x wasn't shipped with Mars, so we have to fetch it directly from one of their repositories.  Unfortunately they only maintain three repositories: (i) the latest from HEAD, (ii) the latest milestone, and (iii) the stable release.

briandealwis:
Where's Guava 21 coming from!?

chanseokoh:
I am tempted to disable AERI for now.

briandealwis:
I can't think of any other option, short of only building our AERI support (which we're not even using) for Neon and Oxygen.

chanseokoh:
Okay, let's think about what we should do with AERI later. ATM, we need to fix this broken build ASAP.

PR #2063, opened 2017-06-14
Make the current flex facet require WTP 

### LINKED ISSUE #2034: NPE when deploying App Engine flex project not configured with Dynamic Web Module facet
Currently, you can add the AE flex facet to a plain Java project without adding the Dynamic Web Module facet. This seems to cause an NPE when trying to deploy such a project.

This is also distantly related to https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2019#issuecomment-306873454.

chanseokoh:
Fixes #2034. As discussed in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2034#issuecomment-307829079, we will make the current flex facet require WTP.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
@@ -18,7 +18,11 @@ package com.google.cloud.tools.eclipse.appengine.facets;
 
 import static org.mockito.Mockito.when;
 
-import com.google.cloud.tools.eclipse.appengine.facets.AppEngineFlexFacet;
+import java.util.Arrays;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.jst.common.project.facet.core.JavaFacet;
+import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
+import org.eclipse.wst.common.project.facet.core.IConstraint;
 import org.eclipse.wst.common.project.facet.core.IFacetedProject;
 import org.eclipse.wst.common.project.facet.core.IProjectFacet;
 import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
```diff
@@ -65,4 +69,53 @@ public void testFacetLabel() {
...
+  public void testInstallConstraints_notOkWithNoServlet() {
+    IConstraint constraint = AppEngineFlexFacet.FACET_VERSION.getConstraint();
+    IStatus result = constraint.check(Arrays.asList(JavaFacet.VERSION_1_8));
+    Assert.assertFalse(result.isOK());
+  }
+
+  @Test
+  public void testInstallConstraints_notOkWithJava7Servlet31() {
+    IConstraint constraint = AppEngineFlexFacet.FACET_VERSION.getConstraint();
+    IStatus result = constraint.check(Arrays.asList(JavaFacet.VERSION_1_7, WebFacetUtils.WEB_31));
```
briandealwis: Why is this case invalid?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -113,15 +113,33 @@
...
+      <label>%flexFacetName</label>
+      <description>%flexFacetDescription</description>
     </project-facet>
     <project-facet-version facet="com.google.cloud.tools.eclipse.appengine.facets.flex" version="1">
-       <constraint>
+      <constraint>
+        <and>
           <conflicts facet="com.google.cloud.tools.eclipse.appengine.facets.standard"/>
-        </constraint>
+          <requires group="com.google.cloud.tools.eclipse.appengine.facets.flex.constraints.servlet"/>
```
briandealwis: I'm regretting this approach for the Java 8 support.  Although groups are extensible, they don't allow specifying additional constraints: it's just "at least one from this group".

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
chanseokoh: My understanding is that we don't have a Java 7 runtime: https://cloud.google.com/appengine/docs/flexible/java/dev-jetty9. The Java 8 / Jetty 9.3 Runtime for WARs, and Java 8 Runtime for JARs.

chanseokoh:
For the Java 8 / Jetty 9.3 Runtime (provided for WARs), I think these are the right constraints? Of course, the flex env is really about running as a container, and your own custom runtime can be anything you can think of, but for Java WARs which we support by the current flex facet design, we use this particular runtime. We should soon think about how we would support the Java 8 Runtime (for JARs) or a custom runtime.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
chanseokoh: So, what do you think will be better? Using `<require facet=... version=...>` constraints directly without the indirect group constraints?

elharo:
shouldn't an app compiled for Java 7 run on the Java 8 runtime?

chanseokoh:
> shouldn't an app compiled for Java 7 run on the Java 8 runtime?

Ah, I get it.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexFacetTest.java
```diff
@@ -73,21 +73,21 @@ public void testFacetLabel() {
   @Test
   public void testInstallConstraints_okWithJava7Servlet25() {
     IConstraint constraint = AppEngineFlexFacet.FACET_VERSION.getConstraint();
-    IStatus result = constraint.check(Arrays.asList(JavaFacet.VERSION_1_8, WebFacetUtils.WEB_25));
```
briandealwis: Oops, missed these :blush:

Issue #1997, opened 2017-06-01
Classes in test source folder being packaged into WAR


chanseokoh:
1. Create a new _native_ HelloWorld project.
2. Move `HelloWorld.java` to `src/test/java`. Optionally delete `src/main/java` to be sure.
3. Run locally or deploy.

Note that the `HelloWorld` servlet is accessible.

chanseokoh:
Not verified, but probably this won't happen to Maven standard projects; the output folder of `src/test/java` in the Maven projects are set to `target/test-classes` and the deployment assembly setting does not pick up that directory:

![selection_020](https://cloud.githubusercontent.com/assets/10523105/26699761/e4cfa89e-46e8-11e7-82d6-f90152b99e00.png)

![selection_021](https://cloud.githubusercontent.com/assets/10523105/26699769/ec59fbfa-46e8-11e7-845c-5f6b9217b3b9.png)

chanseokoh:
Two things should be done:
1. Remove the `/src/test/java` entry from the _Web Deployment Assembly_ sources.
1. Set the output folder of `src/test/java` to `build/test-classes` (for non-Maven projects) or `target/test-classes` (for Maven projects).

chanseokoh:
Re-opening as sometimes I still see `/src/test/java` being assembled:

![selection_030](https://user-images.githubusercontent.com/10523105/26944615-361a4de4-4c57-11e7-9c49-024dbaa54759.png)

If this happens, it usually happens when I create a project for the first time in a hosted debug run, where the wizards take some time while loading things for the first time.

This also seems to result in occasional test failure. From https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2033#issuecomment-307185425:

```
Tests run: 9, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 84.646 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProjectTest

testMostImportantFile(com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProjectTest)  Time elapsed: 31.902 sec

testAppEngineLibrariesAdded(com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProjectTest)  Time elapsed: 17.613 sec

[...]

PR #2073, opened 2017-06-19
Let project tear down at the end of each test to avoid flakiness

### LINKED ISSUE #2048: Kokoro flaky build failure: testGetProject()

chanseokoh:
Fixes #2048.

I think the tests are deleting a file too fast. It might be that the file is in use by a job when they try to delete it. This PR lets each test case set up and tear down projects, rather than trying to reuse projects across tests (`@Rule` instead of `@ClassRule`). That way, we don't need to explicitly delete files; any file in a project will be teared down properly.

I also added `ThreadDumpingWatchdog` in a separate test for future use.


chanseokoh:
The Kokoro failure is an independent flaky test: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1997#issuecomment-307189699.

PR #2069, opened 2017-06-16
Use assertArrayEquals() to see reported markers on flakey test

briandealwis:

PR #2075, opened 2017-06-19
Don't do our own manual XSD validation or handle SAX errors because Eclipse does it

### LINKED ISSUE #1801: Possibly flaky test: XmlValidatorTest.testXsdValidation()

### LINKED ISSUE #2074: Our own XSD manual validation on appengine-web.xml doesn't seem necessary
While tackling #1801, I noticed that skipping the `xsdValidation()` call in `XmlValidator.validate()` still puts error problem markers if `appengine-web.xml` does not confirm to the XSD we supply. I think this is supported natively by the Eclipse XML validation framework, as we register the XSD through an extension:

```xml
  <extension point="org.eclipse.wst.xml.core.catalogContributions">
    <catalogContribution>
      <uri
          name="http://appengine.google.com/ns/1.0"
          uri="xsd/appengine-web.xsd"/>
    </catalogContribution>
  </extension>
```

chanseokoh:
Fixes #2074. Fixes #1801.

Mainly removing `XmlValidator.xsdValidation()` and code to call it.


chanseokoh:
Uh-oh. Looks like `waitForProjects()` is not enough for the marker to show up:

```
testXsdValidation_appengineWebXml(com.google.cloud.tools.eclipse.appengine.validation.XmlValidatorTest)  Time elapsed: 7.781 sec  <<< FAILURE!
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at com.google.cloud.tools.eclipse.appengine.validation.XmlValidatorTest.testXsdValidation_appengineWebXml(XmlValidatorTest.java:153)
```

I think I hit the same problem with what @briandealwis hit in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2069#issuecomment-309488235.

briandealwis:
Does this work if you run without any test bundles?  Because although we register the [appengine-web.xsd](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml#L239) in `com.google.cloud.tools.eclipse.appengine.validation`, the `web.xml` XSDs are only registered in the test-only bundle `com.google.cloud.tools.eclipse.test.dependencies`.

chanseokoh:
In Eclipse, I launched a hosted debug run after closing `.test.dependencies`, `.test.logback`, `.test.util`, `.swtbot`, `.integration.appengine`, and all the `*.test` bundles. Eclipse shows an error problem marker for an malformed `web.xml` regardless of the presence of the App Engine facet. (It doesn't create a marker in a plain Java project, but the editor still shows a red X marker in the left gutter.)

In any case, we have never done anything about XSD validation on `web.xml`, so it's unrelated. This PR just removes our own `appengine-web.xml` XSD validation.

briandealwis:
Ah ok — I quickly skimmed and saw the change on `WebXmlValidator` which threw up some alarm bells.  Will give it a whirl this afternoon.

chanseokoh:
`appengine-web.xml` XSD validation works fine as before, which is also covered by tests.

PR #2078, opened 2017-06-20
Ensure CreateAppEngineWtpProjectTest tests use unique GAVs

briandealwis:
When running `CreateAppEngineWtpProjectTest#testNoJUnit4ClasspathIfUsingMaven()` in Eclipse, I encountered an odd NPE deep in M2E code in where `org.eclipse.m2e.core.internal.project.registry.ProjectRegistryManager#getMavenProject()`, which turns a _group/artifact/version_ (GAV) into a `MavenProjectFacade`,  returned a reference to a project that no-longer existed in the workspace.  It turns out that `testNoJUnit4ClasspathIfUsingMaven()` and `testMavenNatureEnabled()`  both use the same GAV, and the `ProjectRegistryManager`'s `ProjectRegistry` instance caches GAV &rarr; `IProject` mappings, which were stale.

This change ensures that these tests in `CreateAppEngineWtpProjectTest` use unique GAVs.  It also adds some compilation-only dependencies for `.newproject.test` and `.dataflow.ui`.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/build.properties
```diff
@@ -4,3 +4,4 @@ bin.includes = META-INF/,\
                .
 javacSource=1.7
 javacTarget=1.7
+additional.bundles = org.eclipse.emf.ecore
```
chanseokoh: I wonder if we can remove this package from MANIFEST.MF then.

Some question: if I remove the package from MANIFEST.MF, I get an error marker "The type org.eclipse.emf.ecore.EObject cannot be resolved. It is indirectly referenced from required .class files". I learned that I can remove the error by adding the package to MANIFEST.MF, but I've always wondered why the quick fix in this case doesn't offer to add it to MANIFEST.MF. Is this the case where I need to put it through `additional.bundles`?

briandealwis: The error arises as the compiler needs to examine a class (usually a superclass of some type referenced via a field, method argument or return, or the defining class hierarchy) that's not available on the classpath.  These classes aren't required at runtime.  Importing the package works, but if it's not actually used at runtime then it's unnecessary.  (And reconciling packages does take some time, so if we can avoid it, then all the better.)

chanseokoh: Thanks, now things are a little clear to me. You should be able to remove `org.eclipse.emf.ecore` from MANIFEST.MF. It works for me.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/build.properties
```diff
@@ -4,4 +4,5 @@ bin.includes = plugin.xml,\
                META-INF/,\
                .,\
                icons/
+additional.bundles = org.eclipse.core.databinding
```
chanseokoh: I believe the reason is similar to https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1490#issuecomment-282935075? (I don't have any problem without this.)

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/build.properties
briandealwis: Odd: I must have rebased this change before pushing up as I shouldn't have seen any errors with your change the `MANIFEST.MF`.

Issue #2030, opened 2017-06-08
Invalid runtime 'java8' specified on converted project


elharo:
ERROR: (gcloud.app.deploy) Error Response: [400] Invalid runtime 'java8' specified. Accepted runtimes are: [go, java, java7, php, php55, python, python27]

![image](https://user-images.githubusercontent.com/1005544/26940465-c15a0b24-4c48-11e7-8a49-cf38494251ff.png)

elharo:
Possible cause: Java 8 only in installed JREs and CT4E "helpfully" adds `<runtime>java8</runtime>` to appengine-web.xml

elharo:
Reproduced this. Definitely happens on project conversion.

elharo:
It's not, but that's not the issue. This was supposed to be a Java 7 project. The bug is in our code that is being overly eager to add `<runtime>java8</runtime` to the appengine-web.xml.

briandealwis:
I think this is a user configuration error: I suspect your project had a JavaSE-1.8 execution environment. I verified that converting a JavaSE-1.7 project preserves the Java 1.7 facet and the `appengine-web.xml` is untouched.

<img width="738" alt="java-build-path" src="https://user-images.githubusercontent.com/202851/26976485-f52c11d2-4cf1-11e7-94bc-dc2389e2d119.PNG">

If the user's project is configured with JavaSE-1.8, then they can already use Java 8'isms, and so we should set the `<runtime>`.

Perhaps more to the point, if I have an existing 1.8 servlet project that I'm aiming to port to App Engine, I may get strange results trying to deploy the project.

elharo:
That is indeed what happened but that does not mean this isn't a bug. Java 7 is the expected default for App Engine. Unless the user makes a conscious decision to convert it to Java 8 we should not convert it for them. There is no reason for a user to expect that convertin a project from GPE will aklso make it a Java 8 project.

briandealwis:
We'll have to back this out once Java 8 goes GA.  My preference is to prompt the user to say something along the line of:
> Your project is Java 8, but Java 8 support for App Engine is not yet GA. Switch to Java 7?

elharo:
I'm not sure we will. At some point Java 8 becomes the default regardless of runtime element. 

We should not change the user's project without their explicit request. If they haven't taken a deliberate action to ask for Java 8 we shouldn't force it on them.

chanseokoh:
[...]

Issue #1965, opened 2017-05-17
Prevent manually changing Dynamic Web Module facet to >2.5 when using Java 7 runtime


chanseokoh:
Previously, I think there was a facet constraint (or such a thing) that prevented having, e.g., an infeasible Java 7 + Dyn Web 3.1 facet combination. However, now it is possible to manually update the Dyn Web facet to 3.1 through the _Project Facets_ preference page.

If a project is in this situation, then uninstalling the App Engine facet and reinstalling it won't bring back the App Engine Standard Runtime library. (Probably we should also prevent installing the App Engine facet too). From this observation, it might be possible that we still have some facet constraint.

briandealwis:
I've explored a few approaches.

In light of some of the Java 8 experiences to now (#1961, 1963, #1965, #2030, #2044, #2050), here are a few questions to drive evaluation:
  - What should be necessary to mark a project as a Java 8?  Do we want to prevent a project from being accidentally marked as Java 8?
      - Perhaps it should be driven *only* by the `<runtime>` element in `appengine-web.xml`, rather than the Project Facets?
  - is there an advantage to explicitly representing the different runtimes (jre7 vs jre8)?
     - are the Java 7 vs Java 8 environments truly interchangeable beyond the language differences?  We have a few places where we load the `appengine-web.xml` descriptor to check the `<runtime>`, and I wonder if we'll have more.

## Approach 0: _status quo_ with a facet detector ##

I thought we might be able to use the `org.eclipse.wst.common.project.facet.core.detectors` facet detectors to pull out the relevant facets.  I created a facet detector that looks for an `appengine-web.xml` and configures the facets as appropriate.

But this is only triggered by `IFacetedProjectWorkingCopy#detect()`, which is invoked by by the project context menu's _Configure > Convert to Faceted Form…_.  `#detect()` actually works pretty well, and we should use this as part of our _Convert to App Engine Standard Project_.

## Approach 1: _status quo_ with a facet validator ##

[...]

Issue #1963, opened 2017-05-17
Import GCP Projects wizard


elharo:
should we have a Google section here with imports for various types of App Engine, GPE, etc.projects? 

![image](https://cloud.githubusercontent.com/assets/1005544/26155975/d7281ee0-3ae2-11e7-89bf-127cfd1a08b2.png)

elharo:
Or perhaps instead, per @briandealwis "Now that you mention it, I do recall there being a facet-discovery mechanism via the org.eclipse.wst.common.project.facet.core.detectors extension mechanism. We should have such a detector that looks for runtime=java8."

PR #2079, opened 2017-06-20
Change Java 8 support to use separate App Engine Standard facet versions

### LINKED ISSUE #1961: Import Java 8 project
See what happens when we import an existing Java 8 app engine standard project (`<runtime>java8</runtime>`). Make sure the facets (Java and Dynamic Web project) are set appropriately without further action.

briandealwis:
- defines new utility functions on FacetUtil to workaround limitations of equivalents on `IFacetedProjectWorkingCopy` which only consider "fixed" (locked) facet versions
   - `getHighestSatisfyingVersion()` finds the highest-version of a facet that satisfies the existing facets defined on an `IFacetedProject` (replaces `IFacetedProjectWorkingCopy#getHighestAvailableVersion()`)
   - `conflictsWith()` checks if a facet version conflicts with existing facets on a project (replaces `IFacetedProjectWorkingCopy#isFacetAvailable()`)
- creates separate facet versions for the App Engine Standard `JRE7` and `JRE8` runtimes
   - configures `JRE7` as the default version for the App Engine Standard facet
   - this supports separate dependencies for each runtime, so we don't have a DWP 3.1 on `JRE7`
- provides _facet detectors_ for the App Engine Standard `JRE7` and `JRE8` facet versions
   - these look for `appengine-web.xml` and configure the appropriate facet versions depending on the `<runtime>` setting 
   - change `WebProjectUtil#findInWebInf()` to look at several standard locations (`src/main/webapp`, `WebContent`, `war`, and `web`) to support the detector
- changes `AppEngineStandardFacet#installAppEngineFacet()` to use the detectors first
   - otherwise tries to install the _default version_ of the facet (set to `JRE7`), if there is no conflict
   - otherwise tries to install the highest satisfying version of `AppEngineStandardFacet` (i.e., `JRE8`)
- changes the facet listener from listening to `JavaFacet` changes to `AppEngineStandardFacet` changes (i.e., runtime changes), and propagates those changes to `appengine-web.xml` `<runtime>`
- changes the `appengine-web.xml` builder to propagate `<runtime>` changes to the `AppEngineStandardFacet` *and* the `JavaFacet`
- allows changing from `AppEngineStandardFacet` `JRE7` <—> `JRE8`

Fixes #1961, #1963, #1965, #2030, #2044, #2050


chanseokoh:
About `FacetedProjectWorkingCopy`: the cause of #1155 was, IIRC, that `CovertJob` takes a snapshot of a working copy at one point without grabbing a lock, and when it commits changes, it overwrites whatever changes others may have made concurrently in between: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155#issuecomment-268337638.

In light of this, aren't we making the same mistake of not grabbing a lock to the project being modified?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StringVersionComparator.java
```diff
@@ -0,0 +1,23 @@
...
+public class StringVersionComparator implements Comparator<String> {
+
+  // Since this comparator is usually specified via plugin.xml, one would normally
+  // implement this as an IExecutableExtensionFactory that returns Ordering.natural().
+  // But the fproj framework instantiates the specified class *directly* and so we
+  // must actually implement the comparator.
+
+  @Override
+  public int compare(String o1, String o2) {
+    return Ordering.natural().compare(o1, o2);
```
chanseokoh: I think simple string comparison does work with versions, e.g., "1.9.0" vs "1.10.0".

chanseokoh: (... assuming that facet versions may use semantic versions)

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -47,15 +50,16 @@
   private static final Logger logger = Logger.getLogger(AppEngineStandardFacet.class.getName());
 
   public static final String ID = "com.google.cloud.tools.eclipse.appengine.facets.standard";
-  public static final String VERSION = "1";
 
   public static final IProjectFacet FACET = ProjectFacetsManager.getProjectFacet(ID);
-  public static final IProjectFacetVersion FACET_VERSION = FACET.getVersion(VERSION);
+  public static final IProjectFacetVersion JRE7 = FACET.getVersion("JRE7");
+  // public static final IProjectFacetVersion JRE8 = FACET.getVersion("JRE8");
```
chanseokoh: Why not having this one and removing `AppEngineStandardFacetChangeListener.APP_ENGINE_STANDARD_JRE8`?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -54,6 +55,12 @@
       type="UNINSTALL">
       <delegate class="com.google.cloud.tools.eclipse.appengine.facets.StandardFacetUninstallDelegate"/>
     </action>
+    <action
+      id="com.google.cloud.tools.eclipse.appengine.facets.standard.version.action"
+      facet="com.google.cloud.tools.eclipse.appengine.facets.standard"
```
chanseokoh: +1 indent level

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -169,19 +173,50 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
...
+      return;
+    }
+
+    Set<IProjectFacet> previousFixedFacets = fpjwc.getFixedProjectFacets();
     FacetUtil facetUtil = new FacetUtil(facetedProject);
-    facetUtil.addFacetToBatch(FACET_VERSION, null /* config */);
+    // See if the default AppEngine Standard facet is ok
+    if (!FacetUtil.conflictsWith(fpjwc, FACET.getDefaultVersion())) {
+      facetUtil.addFacetToBatch(FACET.getDefaultVersion(), null);
+      fpjwc.addProjectFacet(FACET.getDefaultVersion());
```
chanseokoh: We're using `FacetUtil`, so I think there's no point of using `fpjwc`? Moreover, there is no `commitChanges()` onward, so  I think `fpjwc` will have no effect at all.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StringVersionComparator.java
briandealwis: I'll document that this is for pure-string comparisons, like "JRE7" vs "JRE8"; a version comparator isn't necessary for numeric versions as the faceted project framework handles that out of the box.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
briandealwis: I'll add a comment here to point to `AppEngineStandardFacetChangeListener.APP_ENGINE_STANDARD_JRE8`.  I moved it as tests fail on Tycho since `.java8` isn't loaded and so the _JRE8_ version isn't found.   `AppEngineStandardFacetChangeListener` isn't an ideal class, but it seemed better than creating a brand new class just to hold `JRE8`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
briandealwis: It's necessary for the `getHighestSatisfyingVersion()`.  I'll add a comment.

briandealwis:
Prior to #1167 (fix for #1155) we installed the AES, DWP, and Java facets added as independent operations via `IFacetedProject.installProjectFacet()`.  It was the addition of the DWP facet that triggered the `ConvertJob`, which created an `IFacetedProjectWorkingCopy` as the Java facet was added, such the IFPWC was stale.  Using an IFPWC for `AppEngineStandardFacet.installAppEngineFacet()` should have the same atomic effect as using `IFacetedProject.modify()`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
briandealwis: And I plan to replace `FacetUtil.add*Batch*` with IFPWC separately.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
chanseokoh: Another option is to add it to the test `pom.xml` (or even just adding it to `MANIFEST.MF`, which I think isn't a good idea), but I guess you don't like doing so?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -169,19 +173,51 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
     if (facetedProject.hasProjectFacet(FACET)) {
       return;
     }
+
+    final IFacetedProjectWorkingCopy fpjwc = facetedProject.createWorkingCopy();
+    fpjwc.detect(subMonitor.newChild(20));
+    fpjwc.commitChanges(subMonitor.newChild(20));
```
chanseokoh: So this should be `.modify()` to avoid https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155#issuecomment-310122465 as we discussed?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacetTest.java
```diff
@@ -46,17 +46,25 @@
...
-          AppEngineStandardFacet.FACET_VERSION);
+          AppEngineStandardFacet.JRE7);
 
   @Test
   public void testStandardFacetExists() {
     Assert.assertEquals("com.google.cloud.tools.eclipse.appengine.facets.standard",
         AppEngineStandardFacet.ID);
-    Assert.assertEquals("1", AppEngineStandardFacet.VERSION);
+    Assert.assertEquals("JRE7", AppEngineStandardFacet.JRE7.getVersionString());
+    if (AppEngineStandardFacet.FACET.getVersions().size() > 1) {
```
elharo: this feels like a case for Assume, maybe with a separate test method

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -169,19 +173,51 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
     if (facetedProject.hasProjectFacet(FACET)) {
       return;
     }
+
+    final IFacetedProjectWorkingCopy fpjwc = facetedProject.createWorkingCopy();
```
elharo: workingCopy

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardJre7ProjectFacetDetector.java
```diff
@@ -0,0 +1,55 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetVersionChangeDelegate.java
```diff
@@ -0,0 +1,18 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StringVersionComparator.java
```diff
@@ -0,0 +1,24 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StringVersionComparator.java
```diff
@@ -0,0 +1,24 @@
...
+package com.google.cloud.tools.eclipse.appengine.facets;
+
+import java.util.Comparator;
+
+/**
+ * A {@link Comparator<String>} for the Faceted Project framework that compares versions that are
+ * strings, like "JRE7" vs "JRE8", using the natural string ordering. Not intended for versions that
+ * have some numeric meaning (e.g., 1.2.0): use the standard Faceted Project
```
elharo: : use --> . Use

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -38,10 +38,15 @@
  * Utility classes for processing WTP Web Projects (jst.web and jst.utility).
  */
 public class WebProjectUtil {
-  final static String DEFAULT_WEB_PATH = "src/main/webapp";
+  /** Default top-level locations for WEB_INF */
+  public static final String DEFAULT_WEB_PATH = "src/main/webapp";
+
+  /** All possible top-level locations for WEB_INF */
+  final static String[] DEFAULT_WEB_PATHS = {DEFAULT_WEB_PATH, "WebContent", "war", "web"};
```
elharo: non-private and mutable. Dangerous.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardFacetVersionChangeTests.java
```diff
@@ -0,0 +1,135 @@
...
+import org.eclipse.jst.common.project.facet.core.JavaFacet;
+import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
+import org.eclipse.wst.common.project.facet.core.IFacetedProject;
+import org.eclipse.wst.common.project.facet.core.IFacetedProject.Action;
+import org.junit.Rule;
+import org.junit.Test;
+import org.xml.sax.SAXException;
+
+/**
+ * Test changing the App Engine Standard Facet results in appropriate {@code <runtime>} changes in
```
elharo: Tets that

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardFacetVersionChangeTests.java
```diff
@@ -0,0 +1,135 @@
...
+      .withFacetVersions(JavaFacet.VERSION_1_8, WebFacetUtils.WEB_25,
+          AppEngineStandardFacetChangeListener.APP_ENGINE_STANDARD_JRE8);
+
+  @Test
+  public void testJre8FacetVersionId() {
+    assertEquals("JRE8",
+        AppEngineStandardFacetChangeListener.APP_ENGINE_STANDARD_JRE8.getVersionString());
+  }
+
+  /** Should be able to change a App Engine Standard JRE7 project to JRE8 with no other changes. */
```
elharo: an App Engine

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardJre8ProjectFacetDetector.java
```diff
@@ -0,0 +1,55 @@
+
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardJre8ProjectFacetDetector.java
```diff
@@ -0,0 +1,55 @@
...
+import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
+import org.eclipse.wst.common.project.facet.core.IFacetedProjectWorkingCopy;
+import org.eclipse.wst.common.project.facet.core.ProjectFacetDetector;
+import org.xml.sax.SAXException;
+
+public class AppEngineStandardJre8ProjectFacetDetector extends ProjectFacetDetector {
+  private static final Logger logger = Logger.getLogger(AppEngineStandardJre8ProjectFacetDetector.class.getName());
+
+  @Override
+  public void detect(IFacetedProjectWorkingCopy fpjwc, IProgressMonitor monitor)
```
elharo: workingCopy (no abbreviations per google style)

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacetTest.java
briandealwis: As there is a separate test in `AppEngineStandardFacetVersionChangeTests`, I'll remove this.  It is a bit confusing.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StringVersionComparator.java
briandealwis: Reworded.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
briandealwis: It is _final_: but the order was wrong.  This should be made private.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
chanseokoh: Ah, I see `.facet` is independent of `.java8` in that it's not aware of the JRE8 facet. Please disregard.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
chanseokoh: It's `final`, but anyone can actually replace elements of the array, so it is not immutable.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
briandealwis: Hmm…  this codepath is the same as that used by WTP's [_Configure > Convert to Faceted Project_](http://git.eclipse.org/c/webtools-common/webtools.common.git/tree/plugins/org.eclipse.wst.common.project.facet.ui/src/org/eclipse/wst/common/project/facet/ui/internal/ConvertProjectToFacetedFormRunnable.java?h=master#n81).  In the absence of some known conflicts, I'm thinking we _should_ use it, and use it more.  More to point, I should change the detectors to also install the AES runtimes too,  and we should use it to install the runtimes in our new project wizards.

I don't think we should get too caught up with the `ConvertJob` fiasco.  Yes it was painful to debug, but in large part because we weren't installing facets atomically, and we were auto-installing the runtime.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -42,7 +43,8 @@
   public static final String DEFAULT_WEB_PATH = "src/main/webapp";
 
   /** All possible top-level locations for WEB_INF */
-  private static final String[] DEFAULT_WEB_PATHS = {DEFAULT_WEB_PATH, "WebContent", "war", "web"};
+  private static final List<String> DEFAULT_WEB_PATHS =
```
elharo: This coould be declared as ImmutableList<String> for clarity.

briandealwis:
Discovered last night that Maven-related tests are breaking.  They were passing previously.  The `NewMavenBasedAppEngineProjectWizardTest.testHelloWorld` created project is being created as a JRE8 project.  Digging into it.

briandealwis:
Found a bug in the JRE7 detector (:blush:).
I also missed that installing the DWP facet needs a configuration object.

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -7,39 +7,40 @@
     <project-facet id="com.google.cloud.tools.eclipse.appengine.facets.standard">
        <label>%standardFacetName</label>
        <description>%standardFacetDescription</description>
+       <version-comparator
+             class="com.google.cloud.tools.eclipse.appengine.facets.StringVersionComparator">
+       </version-comparator>
+       <default-version
```
elharo: this could be a one-line empty element tag

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -7,39 +7,40 @@
...
-         <include
-               facet="java"
-               versions="1.7">
-         </include>
-      </group>
-   </extension>
+  
+  <extension
+        point="org.eclipse.wst.common.project.facet.core.aliases">
+     <!--
```
elharo: Is this comment useful to future developers?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -169,19 +173,51 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
...
+    final IFacetedProjectWorkingCopy workingCopy = facetedProject.createWorkingCopy();
+    workingCopy.detect(subMonitor.newChild(20));
+    workingCopy.commitChanges(subMonitor.newChild(20));
+
+    if (facetedProject.hasProjectFacet(FACET)) {
+      // success!
+      return;
+    }
+
+    // we continue to update fpjwc to use FacetUtil.getHighestSatisfyingVersion()
```
elharo: workingCopy

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetVersionChangeDelegate.java
elharo: ping

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
```diff
@@ -130,15 +142,32 @@ public FacetUtil addWebFacetToBatch(IProjectFacetVersion webFacet) {
...
+   * the {@code WEB-INF/web.xml}.
+   */
+  public static IDataModel createWebFacetDataModel(IContainer webappFolder) {
+    return createWebFacetDataModel(webappFolder.getProjectRelativePath().toString());
+  }
+
+  /**
+   * Create a Dynamic Web Facet configuration.
+   */
+  public static IDataModel createWebFacetDataModel(String webInfFolder) {
```
chanseokoh: I think this is `webAppFolder`, not `webInfFolder`. (Also we should capitalize `App`.)

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardJre7ProjectFacetDetector.java
```diff
@@ -0,0 +1,81 @@
...
+import org.xml.sax.SAXException;
+
+public class AppEngineStandardJre7ProjectFacetDetector extends ProjectFacetDetector {
+  private static final Logger logger =
+      Logger.getLogger(AppEngineStandardJre7ProjectFacetDetector.class.getName());
+
+  @Override
+  public void detect(IFacetedProjectWorkingCopy workingCopy, IProgressMonitor monitor)
+      throws CoreException {
+    SubMonitor progress = SubMonitor.convert(monitor, 10);
```
chanseokoh: This is unused.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -169,19 +173,51 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
     if (facetedProject.hasProjectFacet(FACET)) {
       return;
     }
+
+    final IFacetedProjectWorkingCopy workingCopy = facetedProject.createWorkingCopy();
```
chanseokoh: Doesn't need to be `final`.

chanseokoh:
I see a little problem:

1. Create a Hello-World native project.
1. Go to the facet property page. Change the WTP version to 3.0. (You'll notice you also need to change the App Engine facet version to JRE8.)
1. Open `appengine-web.xml`. (Note `<runtime>java8`.)
1. Delete `<runtime>java8</runtime>`.


Although the hosted debug run doesn't pop up any error, but in the main Eclipse console log, you'll see

```
Jun 22, 2017 5:20:27 PM com.google.cloud.tools.eclipse.appengine.standard.java8.AppEngineWebBuilder checkRuntimeElement
SEVERE: P/A: error updating facets
org.eclipse.core.runtime.CoreException: One or more constraints have not been satisfied.
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProject.modifyInternal(FacetedProject.java:363)
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProject.mergeChangesInternal(FacetedProject.java:1181)
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProject.access$2(FacetedProject.java:1117)
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProject$1.run(FacetedProject.java:324)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2241)
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProject.modify(FacetedProject.java:339)
	at com.google.cloud.tools.eclipse.appengine.standard.java8.AppEngineWebBuilder.checkRuntimeElement(AppEngineWebBuilder.java:108)
	at com.google.cloud.tools.eclipse.appengine.standard.java8.AppEngineWebBuilder.build(AppEngineWebBuilder.java:72)
	at org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:734)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:205)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:245)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:300)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:303)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:359)
	at org.eclipse.core.internal.events.BuildManager.build(BuildManager.java:382)
	at org.eclipse.core.internal.events.AutoBuildJob.doBuild(AutoBuildJob.java:144)
	at org.eclipse.core.internal.events.AutoBuildJob.run(AutoBuildJob.java:235)
Contains: App Engine Java Standard Environment JRE7 requires Dynamic Web Module 2.5.
```

chanseokoh:
Another problem:

1. Create a Hello-World native project.
1. Go to the facet property page. Change the WTP version to 3.1. (You'll notice you also need to change the App Engine facet version to `JRE8`.) Click _Apply_.
1. Try to downgrade the WTP version, e.g., to 3.0 or 2.5. You see a message that you can't. No explanation as to why. Trying to downgrading the App Engine facet version to `JRE7` together doesn't help.

chanseokoh:
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/1957#issuecomment-310524821

briandealwis:
> [@elharo](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2079#pullrequestreview-45824114): what do you think about simply JAVA_7 and JAVA_8 or maybe JAVA7 and JAVA8 instead of JRE7 and JRE8? It would be closer to how the runtimes are named.

I originally had it as `java7` and `java8` and found it looked weird when `java8` actually supported Java 1.7.

> [@chanseokoh](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2079#issuecomment-310506465): 2. You'll notice you also need to change the App Engine facet version to JRE8.)
> 3. Open appengine-web.xml. (Note <runtime>java8</runtime>.)

Thanks for testing!  You explicitly changed the AES facet to JRE8 ("I want to use the AES Java 8 Runtime"), so the `<runtime>java8</runtime>` element is correct:

   changes to the AES facet version &larr;&rarr; `appengine-web.xml`'s `<runtime>` element

This behaviour is different from previously in that it was the _Java Facet_ (not the AES facet) that was tied to the `<runtime>` element.

When converting a project (#1957), the `<runtime>` element should drive the choice of the AES, Java, and DWP facet versions.

But this does show that when we remove the `<runtime>` element, we need to downgrade the Java and DWP facet versions.

> [@chanseokoh](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2079#issuecomment-310508516): Try to downgrade the WTP version, e.g., to 3.0 or 2.5. You see a message that you can't. No explanation as to why. Trying to downgrade the App Engine facet version to JRE7 together doesn't help.

That's independent of CT4E: it's because the DWP facet (aka `jst.web`) [doesn't allow downgrading versions from 3.0 or 3.1](http://git.eclipse.org/c/jeetools/webtools.javaee.git/tree/plugins/org.eclipse.jst.j2ee/plugin.xml?h=master#n707).  My guess is because of annotations?

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
briandealwis: This is auto-generated

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
briandealwis: I want to warn future developers from deleting this alias as _version=1_ is in the wild.  I'll try again:
> CT4E had been deployed with a single version of the App Engine Standard facet, which was subsequently replaced with two facet versions with different IDs.

chanseokoh:
> You explicitly changed the AES facet to JRE8 ("I want to use the AES Java 8 Runtime"), so the <runtime>java8</runtime> element is correct:

Absolutely. This is working as intended. The bug is when trying to delete `<runtime>java8</runtime>`, as you noted:

> But this does show that when we remove the <runtime> element, we need to downgrade the Java and DWP facet versions.

Currently, it logs an exception, which we need to fix in this PR. A problem I see is that, we may be able to downgrade the Java version, but as you noted below, DWP doesn't allow downgrading the version. What could be done then?

> That's independent of CT4E: it's because the DWP facet (aka jst.web) doesn't allow downgrading versions from 3.0 or 3.1.

Ah, I see.

briandealwis:
Possible approaches:
  - hand-code the downgrade (exploring this right now) — hoping that cascading errors surface as problems (e.g., _`web.xml` specifies incompatible version_ from validation)
  - create a problem marker on `appengine-web.xml` saying something like _"Unable to downgrade to 'java7' runtime changes: {{summary}}"_
  - add an `<action>` for `jst.web` that to allow downgrading from 3.0/3.1 to 2.5 (eek!)

briandealwis:
Big whack o' changes to bring the conversion world into what we discussed, which can be summed up as: _always respect the `<runtime>` setting in the `appengine-web.xml`_.

To handle removing the `<runtime>java8</runtime>` element, which should force a project into Java 7 and Servlet 2.5 (aka Dynamic Web Facet), I explored a couple of options but finally settled on the simplest: adding support to WTP to change from Servlet 3.x to 2.5.

I've added tests to test each of the additional configurations.  The final summary is:

 Has appengine-web.xml?|`<runtime>`| Java facet version | Dynamic Web Project facet version | App Engine facet version after conversion (RESULT)
--- | --- | --- | --- | ---
YES | none| NONE| NONE |JRE7, Java&rarr;1.7, DWP&rarr;2.5
YES | `java8`| NONE| NONE |JRE8, Java&rarr;1.8, DWP&rarr;3.1
no |  | Java 7| 2.5 |JRE7 (creates appengine-web.xml)
no |  |Java 7| 3 |JRE8 (creates appengine-web.xml; Java facet remains at 1.7)
no |  |Java 8| 2.5 |JRE8 (creates appengine-web.xml; DWP remains at 2.5)
no |  |Java 8| 3 |JRE8 (creates appengine-web.xml)
yes| none| Java 7| 2.5| JRE7
yes| none| Java 7| 3| JRE7, DWP&rarr;2.5
yes| none| Java 8| 2.5| JRE7, Java&rarr;1.7
yes| none| Java 8| 3| JRE7, DWP&rarr;2.5, Java&rarr;1.7
yes| `java8`| Java 7| 2.5| JRE8, Java&rarr;1.8
yes| `java8`| Java 7| 3| JRE8, Java&rarr;1.8
yes| `java8`| Java 8| 2.5| JRE8 (DWP facet remains at 2.5)
yes| `java8`| Java 8| 3| JRE8

chanseokoh:
I looked into one test failure, and it looks like there is a concurrency issue:

```
Tests run: 13, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 74.774 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProjectTest
testAppEngineLibrariesAdded(com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProjectTest)  Time elapsed: 6.553 sec  <<< ERROR!
java.lang.IllegalArgumentException: null
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProjectWorkingCopy.setPrimaryRuntime(FacetedProjectWorkingCopy.java:1339)
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProjectWorkingCopy.revertChanges(FacetedProjectWorkingCopy.java:2095)
	at org.eclipse.wst.common.project.facet.core.internal.FacetedProjectWorkingCopy.commitChanges(FacetedProjectWorkingCopy.java:2035)
	at com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet.installAppEngineFacet(AppEngineStandardFacet.java:180)
	at com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProject.addAppEngineFacet(CreateAppEngineStandardWtpProject.java:50)
	at com.google.cloud.tools.eclipse.appengine.newproject.CreateAppEngineWtpProject.execute(CreateAppEngineWtpProject.java:121)
	at com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProjectTest.testAppEngineLibrariesAdded(CreateAppEngineStandardWtpProjectTest.java:71)
```

In my Eclipse-hosted JUnit runs, this test failure can happen in any test cases in `CreateAppEngineStandardWtpProjectTest` that (indirectly) call `installAppEngineFacet()`. The probably of this failure happening was very high in my runs.

`FacetedProjectWorkingCopy.commitChanges()` calls `revertChanges()`, which does the following:

```java
                    setTargetedRuntimes( this.project.getTargetedRuntimes() );
                    setPrimaryRuntime( this.project.getPrimaryRuntime() );
```
For our tests, `project.getTargetedRuntimes()` and `project.getPrimaryRuntime()` should return the same runtime "App Engine Standard Runtime". It appears that `FacedProjectWorkingCopy` frequently retrieves a lot of information directly from `project`, and the above code snippet is one such example. I added debug code to see what the values are at each point when the failure occurs:

```java
                    System.out.println(">>> WC's targetedRuntimes: " + targetedRuntimes);
                    setTargetedRuntimes( this.project.getTargetedRuntimes() );
                    System.out.println(">>> WC's targetedRuntimes: " + targetedRuntimes);
                    System.out.println(">>> Project's getPrimaryRuntime(): " + this.project.getPrimaryRuntime());
```

The result is that sometimes the working copy's `targetedRuntimes` is set to `[App Engine Standard Runtime]`, which comes from `project.getTargetedRuntimes()` as expected, but `project.getPrimaryRuntime()` may return `null`.
```
>>> WC's targetedRuntimes(): []
>>> WC's targetedRuntimes(): [App Engine Standard Runtime]
>>> Project's getPrimaryRuntime(): null
```

I think this is happening because we have a job that modifies `project` concurrently: our App Engine Runtime installation job. Our installation job calls `addTargetedRuntimes()` first and then calls `setTargetedRuntime()`.

chanseokoh:
> I think this is happening because we have a job that modifies project concurrently: our App Engine Runtime installation job. Our installation job calls addTargetedRuntimes() first and then calls setTargetedRuntime().

I'll back out on this.

briandealwis:
I'm not sure why it's calling `revertChanges()`.  Will add some diagnostics around the `commitChanges()`.

chanseokoh:
It's the normal control flow that it calls `revertChanges()`. (A comment says it reverts changes made to the working copy after the commit, so that callers can reuse the working copy again with the initial state.)

chanseokoh:
> It's the normal control flow that it calls revertChanges(). (A comment says it reverts changes made to the working copy after the commit, so that callers can reuse the working copy again with the initial state.)

... which may raise another problem. Didn't you assume that the changes made to the working copy won't be reverted after `commitChanges()` in `installAppEngineFacet()`?

briandealwis:
Oh, I missed this: `AppEngineStandardRuntimeChangeListener` calls `AppEngineStandardFacet#installAppEngineFacet()`.  Previously the runtimes were set by a job spawned by StandardFacetInstallDelegate (i.e., after the App Engine Standard facet had been added).  The new code uses the new facet detectors that configure a runtime on the `IFacetedProjectWorkingCopy` (which is desirable for the _Configure > Convert to Faceted Project_ workflow), and committing the changes with `IFPWC#commitChanges()` sets the runtimes first before applying facet changes.  So we get double-changes happening in parallel:

```
Thread [main] (Suspended)	
	owns: RunnableLock  (id=167)	
	AppEngineStandardFacet.installAppEngineFacet(IFacetedProject, boolean, IProgressMonitor) line: 184	
	CreateAppEngineStandardWtpProject.addAppEngineFacet(IProject, IProgressMonitor) line: 50	
	CreateAppEngineStandardWtpProject(CreateAppEngineWtpProject).execute(IProgressMonitor) line: 121	
	CreateAppEngineStandardWtpProjectTest.testAppEngineLibrariesAdded() line: 71	
	NativeMethodAccessorImpl.invoke0(Method, Object, Object[]) line: not available [native method]	
	NativeMethodAccessorImpl.invoke(Object, Object[]) line: 62	
        […]

Thread [Worker-1] (Suspended (breakpoint at line 183 in AppEngineStandardFacet))	
	AppEngineStandardFacet.installAppEngineFacet(IFacetedProject, boolean, IProgressMonitor) line: 183	
	AppEngineStandardRuntimeChangeListener$InstallAppEngineFacetJob.run(IProgressMonitor) line: 63	
	Worker.run() line: 55	
```

briandealwis:
Our jobs like the `InstallAppEngineFacetJob` should acquire the appropriate scheduling rule.  And although the various `IPF`/`IPFWC` acquire project locks, it would be helpful to acquire the locks around the execution work like `CreateAppEngineWtpProject#execute()` to ensure the jobs are serialized appropriately.

For now I have CreateAppEngineWtpProject acquire the workspace lock as it's required for `FacetedProjectManager#create()` when updating the project description.  I wonder if we can relax this afterwards.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -118,17 +118,44 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
+    // at
+    // org.eclipse.wst.common.project.facet.core.internal.FacetedProjectFrameworkImpl.create(FacetedProjectFrameworkImpl.java:721)
+    // at
+    // org.eclipse.wst.common.project.facet.core.ProjectFacetsManager.create(ProjectFacetsManager.java:375)
+    // at
+    // com.google.cloud.tools.eclipse.appengine.newproject.standard.CreateAppEngineStandardWtpProject.addAppEngineFacet(CreateAppEngineStandardWtpProject.java:48)
+    // at
+    // com.google.cloud.tools.eclipse.appengine.newproject.CreateAppEngineWtpProject.execute(CreateAppEngineWtpProject.java:126)
+
+    Job.getJobManager().beginRule(newProject.getWorkspace().getRoot(), subMonitor.newChild(1));
```
chanseokoh: We already hold a workspace lock as an `WorkspaceModifyOperation`, so I think this is unnecessary?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
```diff
@@ -75,6 +75,7 @@ private AppEngineRuntimeInstallJob(IFacetedProject facetedProject) {
       super(Messages.getString("appengine.install.runtime.to.project", // $NON-NLS$
           facetedProject.getProject().getName()));
       this.facetedProject = facetedProject;
+      setRule(facetedProject.getProject());
```
chanseokoh: I'm concerned about possible deadlock: this job waits for the first `ConvertJob`, and when this job is being created (or scheduled), the first `ConvertJob` is already scheduled, which may be in the sleeping state. If we grab a lock, the `ConvertJob` may not begin.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
briandealwis: Good point.  I should have checked.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java
briandealwis: I see what you mean 😭  I think the `ConvertJob` will begin — but it may become 'paused' in the middle when it tries to `commitChanges()` and acquire a workspace lock.  That said, the `NonSystemJobSuspender` will suspend this job, won't it?

chanseokoh: Um... wasn't talking about that exactly (I was wrong), but I think you got another point. The suspender will suspend the second `ConvertJob`, but it won't suspend the first `ConvertJob`. We have this comment in `AppEngineFacetInstallDelegate`:

```java
    installJob.schedule();
    // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1155
    // The first ConvertJob has already been scheduled (which installs JSDT facet), and
    // this is to suspend the second ConvertJob temporarily.
    NonSystemJobSuspender.suspendFutureJobs();
  }
```
The first `ConvertJob` is the one that installs the JSDT facet (you'll see the comment in the code as well), and it does `commitChanges()`. (I don't think it tries to acquire a workspace lock, but I think it'll be blocked on the lock to modify project facets anyway.) The `AppEngineRuntimeInstallJob.waitUntilJsdtIsFixedFacet()` waits for the first `ConvertJob` to finish, but if we're holding a project lock, I think we'll get into a deadlock. (Not tested though.)

I may be wrong, however. Cannot be confident until we do some actual testing.

chanseokoh: Hmm... interesting. I *think* we have a deadlock situation like below (I think the job "Configuring for JavaScript" is the first `ConvertJob`, because we resume the suspender at the end of "Install App Engine runtimes"), but the deadlock is resolved after some time automatically:

selection_002

Not sure if this is because the Job manager has an ability to resolve deadlocks. But then, I don't see any log about it, so this may not be a deadlock situation at all.

chanseokoh: Ah, the deadlock resolution is because I added a time limit when waiting for the first `ConvertJob` with `MAX_JSDT_CHECK_RETRIES`:

```java
        IProjectFacet jsdtFacet = ProjectFacetsManager.getProjectFacet(JSDT_FACET_ID);
        for (int times = 0;
            times < MAX_JSDT_CHECK_RETRIES && !facetedProject.isFixedProjectFacet(jsdtFacet);
            times++) {
 ```

After 10 seconds, it just gives up. So, this is a deadlock that is always reproducible.

Comment (chanseokoh) [inline plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java]: I think there should be a way to apply a scheduling rule to some portion of code of a `Job`, rather than applying it to the entire job? If we grab a lock after `waitUntilJsdtIsFixedFacet()`, at least we can get away with this deadlock.

Comment (briandealwis) [inline plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/StandardFacetInstallDelegate.java]: > The first ConvertJob is the one that installs the JSDT facet, and it does commitChanges(). (I don't think it tries to acquire a workspace lock, but I think it'll be blocked on the lock for modifying project facets anyway.)

`FacetedProjectWorkingCopy#commitChanges()` calls `FacetedProject#mergeChanges()` which does [acquire the workspace lock](http://git.eclipse.org/c/webtools-common/webtools.common.git/tree/plugins/org.eclipse.wst.common.project.facet.core/src/org/eclipse/wst/common/project/facet/core/internal/FacetedProject.java?h=master#n1109)

> I think there should be a way to apply a scheduling rule to some portion of code of a Job, rather than applying it to the entire duration of the job? If we grab a lock after waitUntilJsdtIsFixedFacet(), at least we can get away with this deadlock.

Yeah, we should be able to wrap `AppEngineStandardFacet#installAppEngineRuntime()`.  In fact, maybe it's best to put these inside `installAppEngineRuntime()` and `installAppEngineFacet()`.

File: plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardFacetChangeListener.java [hunk 26]
```diff
@@ -49,26 +55,42 @@ public void handleEvent(IFacetedProjectEvent event) {
...
+          logger.fine(project + ": removing <runtime>java8</runtime> from appengine-web.xml");
+          AppEngineDescriptorTransform.removeJava8Runtime(descriptor);
+        }
       }
-    if (project.hasProjectFacet(JavaFacet.VERSION_1_8)) {
-      AppEngineDescriptorTransform.addJava8Runtime(descriptor);
-    } else {
-      AppEngineDescriptorTransform.removeJava8Runtime(descriptor);
+    } catch (SAXException | IOException | CoreException ex) {
+
```
chanseokoh: Log something or rethrow?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardFacetChangeListener.java
```diff
@@ -49,26 +55,42 @@ public void handleEvent(IFacetedProjectEvent event) {
...
+    }
+  }
+
+  private boolean isJava8(IFile descriptor) throws IOException, CoreException, SAXException {
```
chanseokoh: `static`

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardJre7ProjectFacetDetector.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        progress.worked(1);
+      }
+
+      if (!workingCopy.hasProjectFacet(WebFacetUtils.WEB_25)) {
+        if (workingCopy.hasProjectFacet(WebFacetUtils.WEB_FACET)) {
+          logger.fine(projectName + ": changing to Dynamic Web 2.5 facet");
+          workingCopy.changeProjectFacetVersion(WebFacetUtils.WEB_25);
+        } else {
+          logger.fine(projectName + ": setting Dynamic Web 2.5 facet");
```
briandealwis: I thought they were pretty fine looking myself too ;-)

Fine isn't shown by default.  These really are for debugging.  Would you prefer INFO, or FINER?

chanseokoh: I wasn't concerned about the logging level. Rather it was of an aesthetic reason and about the code taking up screen real estate. I don't really care, so leave them if you want.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -176,32 +176,43 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
 
     try {
       if (facetedProject.hasProjectFacet(FACET)) {
+        // nothing to do, move along
         return;
       }
 
+      /*
+       * So the project does not have the App Engine Standard facet. We first try our AES JRE7 and
+       * JRE8 detectors: these detectors look for appengine-web.xml and either add or change the
```
chanseokoh: Q: does the JRE7 detector runs before the JRE8 detector (or the other way around)? Or, is the order random? Will the order matter?

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
```diff
@@ -108,29 +120,35 @@ private static boolean isGpeClasspath(IClasspathEntry entry) {
...
   @VisibleForTesting
-  static void removeGpeNature(IProject project) {
+  static boolean removeGpeNature(IProject project) {
     try {
+      boolean hadNature = NatureUtils.hasNature(project, GPE_GAE_NATURE_ID);
       NatureUtils.removeNature(project, GPE_GAE_NATURE_ID);
+      return hadNature;
     } catch (CoreException ex) {
       logger.log(Level.WARNING, "Failed to remove GPE nature.", ex);
+      return true;
```
briandealwis: I'm going to back out these changes to `GpeMigrator` method return-type as I didn't end up using them.

briandealwis: I'll surface the `CoreException` upwards: it's only thrown if the project doesn't exist, isn't open, or has some other major issue. In which case there's not much we can do: the project itself is unusable.

plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/GpeMigrator.java
```diff
@@ -108,29 +120,35 @@ private static boolean isGpeClasspath(IClasspathEntry entry) {
...
     // To remove the facets, we will directly modify the WTP facet metadata file (using XSLT):
     // .settings/org.eclipse.wst.common.project.facet.core.xml
     IFile metadataFile = facetedProject.getProject().getFile(FACETS_METADATA_FILE);
     if (!metadataFile.exists()) {
-      return;
+      return false;
     }
-
+    // must make a copy
+    Set<IProjectFacetVersion> originalFacets = new HashSet<>(facetedProject.getProjectFacets());
```
briandealwis: They're not big, we have to compute the results anyways, and it helped when debugging.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
```diff
@@ -120,6 +131,7 @@ public FacetUtil addWebFacetToBatch(IProjectFacetVersion webFacet) {
       return this;
     }
 
+    // FIXME: should this try to find web.xml first?
```
briandealwis: Cursory examination, I didn't really understand the intention of this code.  Will add a comment.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
```diff
@@ -130,15 +142,32 @@ public FacetUtil addWebFacetToBatch(IProjectFacetVersion webFacet) {
...
+   * the {@code WEB-INF/web.xml}.
+   */
+  public static IDataModel createWebFacetDataModel(IContainer webAppFolder) {
+    return createWebFacetDataModel(webAppFolder.getProjectRelativePath().toString());
+  }
+
+  /**
+   * Create a Dynamic Web Facet configuration.
+   */
+  public static IDataModel createWebFacetDataModel(String webAppFolder) {
```
briandealwis: I used it in the new JRE7 and JRE8 detectors

chanseokoh: I know, but this one particular can be `private`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -56,12 +63,18 @@ public static IFolder getWebInfDirectory(IProject project) {
...
+    for (String possibleWebInfContainer : DEFAULT_WEB_PATHS) {
+      // simplify mocking: get the location as two parts and check for null despite that getFolder()
+      // should be @NonNull
+      IFolder defaultLocation = project.getFolder(possibleWebInfContainer);
+      if (defaultLocation != null && defaultLocation.exists()) {
+        defaultLocation = defaultLocation.getFolder(WEB_INF);
+        if (defaultLocation != null && defaultLocation.exists()) {
+          return defaultLocation;
+        }
+      }
```
briandealwis: Hmm, kinda: the detectors uses WebProjectUtil#findInWebInf() to find the `appengine-web.xml` file, which uses these lookup locations.  The AppEngine Java SDK demos have the `appengine-web.xml` in `war/`, not `src/main/webapp`.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineWebBuilderTest.java
```diff
@@ -50,52 +51,142 @@ public void testNoBuilder() throws CoreException {
...
+        testProject.getFacetedProject().getProjectFacetVersion(AppEngineStandardFacet.FACET));
+    assertEquals("removing <runtime>java8</runtime> should change java to 1.7",
+        JavaFacet.VERSION_1_7,
+        testProject.getFacetedProject().getProjectFacetVersion(JavaFacet.FACET));
+    assertEquals(WebFacetUtils.WEB_25,
+        testProject.getFacetedProject().getProjectFacetVersion(WebFacetUtils.WEB_FACET));
+  }
+
+  /**
+   * Project adding App Engine Standard facet should have builder, and should downgrade the Dynamic
```
briandealwis: Good point.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/src/com/google/cloud/tools/eclipse/appengine/standard/java8/AppEngineStandardFacetChangeListener.java
briandealwis: oops

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
briandealwis: Order doesn't matter: they both parse the content and then check if the `<runtime>` element matches their requirement, and return if not.

briandealwis:
In adecdba008da79ad22adb6e8c225d40236c94e77 I removed the `catch(CoreException)` in `installAppEngineFacet()`: by allowing DWP 3.x to be downgraded to 2.5, there should be no known conflicts introduced by our detectors, and so any conflicts that do arise are important.  (Allowing the downgrade also really simplified this code — `installAppEngineFacet()` was becoming pretty hairy trying to determine the conflicts and whether they were known or not).

briandealwis:
In 55221544cf9c61ac9add5417a0e658f1e2ed037e, I change the `AppEngineWebBuilder`, which monitors for changes to the `appengine-web.xml` to cause additions of `<runtime>java8</runtime>` to always upgrade to Java 8 and DWP 3.1.

briandealwis:
Hmm, hit a deadlock on the Neon build, and I've been able to reproduce it locally on Mars.

The cast of characters:
  - _Worker-2_ is `InstallAppEngineFacetJob`, a job spawned by the installation of an App Engine Runtime in `AppEngineStandardRuntimeChangeListener`.  It has acquired the workspace root's lock, but is blocked on the `IFacetedProject` as the IFP is being modified… by _ModalContext_.
  - _ModalContext_ is executing `CreateAppEngineStandardWtpProject`, and is in the middle of `FacetedProject#mergeChanges()` which acquired the IFP's modification lock; it originally has the workspace root, but somehow calls into the `DependencyGraph` which yields the workspace root's lock, thus allowing _Worker-0_ to start running.  So it's blocked waiting on _Worker-0_ to reacquire the workspace root lock.

Observations: 
  - _Worker-0_'s `InstallAppEngineFacetJob` never normally runs as it is triggered by the addition of an App Engine Runtime, which is itself usually triggered by the addition of the App Engine facet.
     - I had to change this job to be scheduled to run in 200ms as `AppEngineStandardFacet#installAppEngineFacet()` now results in the App Engine Runtime being installed _before_ the AES facet and its dependent facets (part of `IFacetedProjectWorkingCopy#commitChanges()`).
- Although it looks to be  _DependencyGraph_-related, our previous `IDependencyGraph.INSTANCE.preUpdate()`/`.postUpdate()` workaround doesn't work.
- The immediate problem is that `DependencyGraphImpl#isUpdateNecessary()` is returning true, as the DependencyGraph must be updated for the new project (that we're in the middle of configuring).
- This actual deadlock is happening as the Dynamic Web Project install delegate (`WebFacetInstallDelegate`) is adding the WTP EAR Libraries container.  

So two potential fixes:
  - Force update of the DependencyGraph by explicitly calling `DependencyGraphImpl#waitForUpdates()`
  - We don't support EARs, so 🤞 we should be able to disable the EAR library processing and maybe prevent it?

[deadlock.txt](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/files/1113876/deadlock.txt)

briandealwis:
Looks like disabling EAR libraries is sufficient.  I don't see any negative implications, since AppEngine doesn't support EARs any more.

briandealwis:
Although the don't-add-EAR-libraries works (4 rebuild on Travis haven't reproduced it), I'm not that keen on this solution. The root cause of this deadlock is two-fold:
  1. we've introduced workspace locks in fear of potential deadlock, but they're being released by WTP's `IDependencyGraph`
  2. it's actually our own spawned jobs to install the AES facets and runtimes that are causing the problems 

Thinking about (2), we spawn a job from our `StandardFacetInstallDelegate` to ensure we install the AES runtime. And conversely, we spawn a job from our  `AppEngineRuntimeChangeListener` (IIRC) to ensure we have the AES facet (and`appengine-web.xml`). these calls currently operate in a state of ignorance of how they've been triggered. in this particular case, the runtime-listener is unaware that it was triggered as part of an AES facet install, so that spawning a facet install is unnecessary. 

We could try to work out a way to pass information about the current operation in a way that works between threads. For example `AppEngineStandardFacet.installAppEngineFacet()` could put the IFPWC in a map so that `AppEngineStandardFacet.installAppEngineRuntime()`  could see that there is a pending operation. But it still has the possibility of the runtime install happening at the same time as the facet install.

Another possibility is to change `AppEngineStandardFacet.installAppEngineFacet()` to  manually apply the IFPWC changes, as determined by the detectors, rather than use `commitChanges()': first applying the facet changes, then applying the rutile changes.  This would mimic what was done previously, and ensures that the `AppEngineRuntimeChangeListener` sees the facet is available.  But it still has the possibility of the runtime install happening at the same time as the facet install.

The last alternative is to use a separate lock to synchronise our `AppEngineStandardFacet` installation tasks on the IFPWC, like an  Eclipse Jobs `ILock`, that is independent of the workspace lock. I think this last alternative is the safest in that it ensures or changes are serialised.

briandealwis:
> The last alternative is to use a separate lock to synchronise our AppEngineStandardFacet installation tasks on the IFPWC, like an Eclipse Jobs ILock, that is independent of the workspace lock. I think this last alternative is the safest in that it ensures or changes are serialised.

This is the approach I've gone with.  I'll trigger a couple of rebuilds.

chanseokoh:
Personally, I wouldn't mind using a single static lock instead of caching fine-grained locks at the project level. Adding AES facets to multiple projects would be a very rare event.

briandealwis:
@chanseokoh gave his ok to latest changes.

PR #2082, opened 2017-06-22
turn on dataflow

elharo:
@tgroh @briandealwis I'd like to soft launch this (no fanfare) early next week.


tgroh:
What's left to do to make this soft-launch happen?

elharo:
We need to get login working at a minimum. I'd also like to fix some of the recent bugs @chanseokoh has found.

briandealwis:
Do we want to be able to install Dataflow separately from the rest of CT4E?  Should we have a separate feature that pulls in the minimal core?

(Or do we wait and see if people ask.)

elharo:
Login should work now.

PR #2081, opened 2017-06-21
Add "src/test/java" with test-only output folder before adding WTP facet

chanseokoh:
Fixes #1997 in this way: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1997#issuecomment-309914517.

@briandealwis this will cause a conflict with #2079, so beware.

PR #2084, opened 2017-06-23
Fix fragile "src/test/java" removal from web deploy assembly

### LINKED ISSUE #1997: Classes in test source folder being packaged into WAR

chanseokoh:
Fixes #1997 by this way: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1997#issuecomment-310707483. I think this is much nicer than the previous approach #2081.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProjectTest.java
```diff
@@ -143,22 +141,17 @@ private boolean containsSegment(IPath path, String segment) {
...
-  private void assertNoTestClassesInDeploymentAssembly() {
-    StructureEdit core = StructureEdit.getStructureEditForRead(project);
-    try {
-      WorkbenchComponent component = core.getComponent();
-      assertEquals(0, component.findResourcesBySourcePath(new Path("/src/test/java"), 0).length);
-      assertEquals(1, component.findResourcesBySourcePath(new Path("/src/main/java"), 0).length);
-    } finally {
-      core.dispose();
+    while (!creator.deployAssemblyEntryRemoveJob.jobCompleted) {
+      Thread.sleep(50);
```
elharo: could this spin forever?

briandealwis: It'll be killed by Tycho's test timeout.  But could use a `Stopwatch` from Guava to keep track.

I think a comment here re: NonSystemJobSuspender would be helpful.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/DeployAssemblyEntryRemoveJob.java
```diff
@@ -0,0 +1,77 @@
...
+ * Removes entries from WTP's deployment assembly whose source path matches a given path.
+ */
+// Intended to be used by our wizards during project creation. See
+// https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1997#issuecomment-310707483
+class DeployAssemblyEntryRemoveJob extends Job {
+
+  private final IProject project;
+  private final IPath sourcePath;
+
+  // Cannot use "join()", as "NonSystemJobSuspender" makes it return immediately without running it.
```
briandealwis: Oh interesting.  What a tangled web we weave…  I wonder if we shouldn't make NonSystemJobSuspender specific to ConvertJob.

But this comment seems to belong elsewhere (it's actually for the users of this job).  And instead of this boolean, callers can use `getState() == Job.NONE` to see if it's been executed.

An alternative is to make `DeployAssemblyEntryRemoveJob` to be a system job — at least for the `CreateAppEngineWtpProject` case.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/DeployAssemblyEntryRemoveJob.java
```diff
@@ -0,0 +1,77 @@
...
+  public DeployAssemblyEntryRemoveJob(IProject project, IPath sourcePath) {
+    super(Messages.getString("deploy.assembly.test.source.remove.job")); //$NON-NLS-1$
+    this.project = project;
+    this.sourcePath = sourcePath;
+  }
+
+  @Override
+  protected IStatus run(IProgressMonitor monitor) {
+    try {
+      Job.getJobManager().join(
```
briandealwis: Can just use `getJobManager()`. This `join()` is worth a comment or pointer as to why it's necessary.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProjectTest.java
chanseokoh: It is guaranteed to terminate unless some jobs hangs with a deadlock (which shouldn't happen), but I see no harm in waiting for a set time. I'll make the job a system job so that it's not affected by `NonSystemJobSuspendeder`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/DeployAssemblyEntryRemoveJob.java
chanseokoh: > I wonder if we shouldn't make NonSystemJobSuspender specific to ConvertJob.

I am down with this idea, and frankly, wanted to do so back then. At the time we introduced the job suspended, I believe we were concerned that Eclipse code may change in the future, so we decided to devise a future-proof method. But knowing that the `ConvertJob` bug was fixed in Oxygen, I regret we should have just waited for the `ConvertJob` (with a checking condition like `instanceof ConvertJob`).

> An alternative is to make DeployAssemblyEntryRemoveJob to be a system job — at least for the CreateAppEngineWtpProject case.

Making it a system job seems a good idea. I will make a comment inside `CreateAppEngineWtpProject` about it too.

chanseokoh: > And instead of this boolean, callers can use getState() == Job.NONE to see if it's been executed.

This doesn't seem to work, because jobs (attempted to be) scheduled after the `NonSystemJobSuspender` starts are in the NONE state. That is, if the suspender is in play, the world will never see any non-system job, until the suspender resumes.

BTW, the field is here only to enable tests. In production, as long as the job runs eventually, we're good. However in tests, I need to wait until this job completes, which cannot be done with `.join()` or `getState()`.

Issue #2076, opened 2017-06-19
ProjectUtils.waitForProjects() not enough to wait for markers to appear/disappear, leading to flaky tests


chanseokoh:
We have a few places where we test our validation puts problem markers properly, e.g.,

- Tests in `XmlValidatorTest` such as `testXsdValidation_appengineWebXml()` and `testValidate_badXml()`
- [`XsltSourceQuickFixTest.testApply()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.2.0B/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XsltSourceQuickFixTest.java#L86)
- `ToServlet25SourceQuickFixTest.testConvertServlet()`

We call `ProjectUtils.waitForProjects()` to give some time for the markers to be updated. However, it seems like we need to wait further, as the tests occasionally fail: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2075#issuecomment-309565629 and https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2069#issuecomment-309488235.

This issue does not necessarily mean we want to overload `waitForProjects()` further to also cover marker updates. We could put some code to wait for markers in the tests themselves.

briandealwis:
```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 25.888 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.validation.XsltSourceQuickFixTest
testApply(com.google.cloud.tools.eclipse.appengine.validation.XsltSourceQuickFixTest)  Time elapsed: 25.687 sec  <<< FAILURE!
java.lang.AssertionError: [/test0.06296464251836487/appengine-web.xml:2: Project ID should be specified at deploy time]
	at org.junit.Assert.fail(Assert.java:88)
	at com.google.cloud.tools.eclipse.test.util.ArrayAssertions.assertIsEmpty(ArrayAssertions.java:58)
	at com.google.cloud.tools.eclipse.test.util.ArrayAssertions.assertIsEmpty(ArrayAssertions.java:36)
	at com.google.cloud.tools.eclipse.appengine.validation.XsltSourceQuickFixTest.testApply(XsltSourceQuickFixTest.java:92)
```
It's like the old markers are being removed independently?

briandealwis:
[...]

PR #2099, opened 2017-06-30
Check on-disk vs in-editor representations

briandealwis:
Add a test to `XsltSourceQuickFixTest` to check the on-disk representation.  The XML editor seems to maintain a separate model from the viewer document, and my suspicion is that #2076 is caused as this XML model is out-of-sync with the viewer document, and the save is using the stale model.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XsltSourceQuickFixTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XsltSourceQuickFixTest.java
@@ -26,6 +26,11 @@ import com.google.cloud.tools.eclipse.test.util.project.ProjectUtils;
 import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
 import com.google.cloud.tools.eclipse.ui.util.WorkbenchUtil;
 import com.google.common.base.Function;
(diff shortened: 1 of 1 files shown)


chanseokoh:
No harm in adding the check, but this doesn't apply to the `testXsdValidation_appengineWebXml()` in #2076, so I am skeptical. `testXsdValidation_appengineWebXml()` doesn't involve any viewer:

```java
  @Test
  public void testXsdValidation_appengineWebXml() throws CoreException {
    String xml = "<appengine-web-app xmlns='http://appengine.google.com/ns/1.0'>"
        + "<foo></foo>"
        + "</appengine-web-app>";
    IProject project = appEngineStandardProjectCreator.getProject();
    IFile file = project.getFile("WebContent/WEB-INF/appengine-web.xml");
    file.setContents(
        new ByteArrayInputStream(xml.getBytes(StandardCharsets.UTF_8)), true, false, null);
    ProjectUtils.waitForProjects(project);   // Wait until Eclipse puts an error marker.

    String problemMarker = "org.eclipse.core.resources.problemmarker";
    IMarker[] markers = file.findMarkers(problemMarker, true, IResource.DEPTH_ZERO);
    assertEquals(1, markers.length);
    assertTrue(markers[0].getAttribute(IMarker.MESSAGE).toString().contains(
        "Invalid content was found starting with element 'foo'."));
  }
```

As shown above, having an on-disk content that would trigger markers should be enough to create markers.

briandealwis:
Good point.  I'll try digging a bit deeper into the XML editor code at some point.

PR #2112, opened 2017-07-06
Suspend only ConvertJobs instead of all non-system jobs

chanseokoh:
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2076#issuecomment-313492723
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2084#discussion_r123837495

Given that the `ConvertJob` issue (#1155) was fixed in Oxygen, it seems safe now to suspend only `ConvertJob` instead of all non-system jobs. What do you think @briandealwis?


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/ConvertJobSuspender.java
```diff
@@ -83,7 +88,7 @@ static void resumeInternal() {
   }
 
   static void suspendJob(Job job, long scheduleDelay) {
-    if (job.isSystem()) {
+    if (!"Configuring for JavaScript".equals(job.getName())) {
```
chanseokoh: Good idea.

PR #2120, opened 2017-07-08
Clear bucket validation error message after validation success / remove unnecessary code

### LINKED ISSUE #2117: SelectFirstMatchingPrefixListener.OnCompleteListener doesn't seem to work as intended
According to [its Javadoc](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.2.0B/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/util/SelectFirstMatchingPrefixListener.java), listeners should be notified only when the combo is auto-completed, but at the current state, [listeners are always notified whenever a user types anything into the combo text area](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.2.0B/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/util/SelectFirstMatchingPrefixListener.java#L86). I don't think this is intended, as this listener provides no value. (It's just implementing a duplicate of `ModifyListener`.)

### LINKED ISSUE #2119: Bucket validation error message in Dataflow wizard should be cleared when the error gets cleared
For example, when typing `gs://`, the previous error message that `gs:/` is problematic stays there.

selection_013

The only reason this wasn't a big problem was that, (due to the bug #2117) whenever anything is typed into the field and if it is not an invalid bucket name, another job that actually tries to access the bucket is scheduled and will eventually overwrite any lingering message. However, that is going to change after fixing #2117, so this will become a real problem. The fix is very easy though: clear the message if a bucket input becomes valid.

chanseokoh:
Fixes #2119. Fixes #2117.

At the same time, refactored `butcketNameOk()` so that it just returns validation results and let callers update UI.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
@@ -233,7 +233,7 @@ public class RunOptionsDefaultsComponent {
       return;
     }
 
-    IStatus status = bucketNameOk();
+    IStatus status = bucketNameStatus();
     if (!status.isOK()) {
       messageTarget.setError(status.getMessage());
       return;
@@ -362,7 +362,7 @@ public class RunOptionsDefaultsComponent {
     return bucketName;
   }
 
-  private IStatus bucketNameOk() {
+  private IStatus bucketNameStatus() {
     return bucketNameValidator.validate(trimBucketName());
   }
 
@@ -370,7 +370,7 @@ public class RunOptionsDefaultsComponent {
 
     @Override
     public void modifyText(ModifyEvent event) {
-      boolean enabled = !trimBucketName().isEmpty() && bucketNameOk().isOK();
+      boolean enabled = !trimBucketName().isEmpty() && bucketNameStatus().isOK();
       createButton.setEnabled(enabled);
     }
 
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/util/SelectFirstMatchingPrefixListener.java
```diff
@@ -54,60 +47,33 @@ public void setContents(SortedSet<String> contents) {
...
-      // Don't fire on this modification of the combo
-      targetCombo.removeModifyListener(this);
-      targetCombo.select(comboIndex);
-
-      // Select the text that was added
-      targetCombo.setSelection(new Point(prefix.length(), firstMatch.length()));
 
-      // Fire on later user modifications of the combo
-      targetCombo.addModifyListener(this);
-    } finally {
```
chanseokoh: Here, I just removed the `try-finally` block and de-indented code, but apparently, the GitHub diff doesn't highlight this in an effectively.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -346,28 +351,26 @@ public void run() {
...
-      messageTarget.setError(status.getMessage());
-      setPageComplete(false);
-    }
-    boolean enabled = status.isOK() && !bucketName.isEmpty();
-    return enabled;
+    return bucketName;
   }
-  
+
+  private IStatus bucketNameOk() {
```
tgroh: `bucketNameStatus`?

PR #2124, opened 2017-07-09
Proof of concept

elharo:
uses Eclipse core expressions library to add stackdriver monitoring library only to non-standard projects.


plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -202,6 +205,51 @@ unspecified, the javadoc will be downloaded from the artifact location if availa
       </xs:complexType>
    </xs:element>
 
+   <xs:element name="enablement">
+      <xs:annotation>
+         <xs:documentation>
```
briandealwis: There's a schema to define the enablement language.
`<include schemaLocation="schema://org.eclipse.core.expressions/schema/expressionLanguage.exsd"/>
`

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -202,6 +205,51 @@ unspecified, the javadoc will be downloaded from the artifact location if availa
...
+            An Eclipse core expressions enablement element.
+            See http://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fcore%2Fexpressions%2Fpackage-summary.html
+            This schema is very restricted and needs to be expanded.
```
briandealwis: This text should define what is the default variable.  I assume it's an `org.eclipse.core.resources.IProject`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -24,6 +24,7 @@
 import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.RegistryFactory;
 
+
```
briandealwis: This seems an odd change.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -145,4 +149,16 @@ void setGroup(String group) {
...
+
+  /**
+   * @param expression an expression to be evaluated to determine whether
+   *     this library is enabled in context
+   */
+  void setEnablement(Expression expression) {
+    this.expression = expression;
+  }
+
+  public Expression getEnabled() {
```
briandealwis: `getEnablement`?

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/LibrarySelectorGroup.java
```diff
@@ -74,18 +87,55 @@ private void createContents(Composite parentContainer) {
...
 
+  private static boolean enabled(Library library) {
+    Expression expression = library.getEnabled();
+    if (expression == null) {
+      return true; 
+    }
+
+    // fill in the default variable with the currently selected project
+    List<IProject> defaultVariable = new ArrayList<>();
+    // todo is this the best way to grab the selection? If so move to a utility package?
```
briandealwis: Pass the project into the method.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
elharo: What is the schema:// protocol? How is it resolved?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/LibrarySelectorGroup.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
briandealwis: It's a PDE extension, I guess.  But the above is what's done to reference the expression language in various extension points like [`org.eclipse.ui.handlers`](http://git.eclipse.org/c/platform/eclipse.platform.ui.git/tree/bundles/org.eclipse.ui/schema/handlers.exsd?h=master#n30).

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/LibrarySelectorGroup.java
```diff
@@ -73,19 +86,62 @@ private void createContents(Composite parentContainer) {
...
+  private static IProject getSelectedProject() {
+    // todo is this the best way to grab the selection? If so move to a utility package? 
+    IWorkbench workbench = PlatformUI.getWorkbench();
+    IWorkbenchWindow activeWorkbenchWindow = workbench.getActiveWorkbenchWindow();
+    IWorkbenchPage activePage = activeWorkbenchWindow.getActivePage();
+    ISelection selection = activePage.getSelection();
+    if (selection instanceof IStructuredSelection) {
+      IStructuredSelection structuredSelection = (IStructuredSelection) selection;
+      Object element = structuredSelection.getFirstElement();
+      if (element instanceof IResource) {
```
chanseokoh: This can be simplified further. Take a look at `ProjectFromSelectionHelper.getProject()`.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/LibrarySelectorGroup.java
```diff
@@ -73,19 +86,62 @@ private void createContents(Composite parentContainer) {
...
+    try {
+      EvaluationResult result = expression.evaluate(context);
+      return result == EvaluationResult.TRUE;
+    } catch (CoreException ex) {
+      // default to including the library
+      return true;
+    }
+  }
+
+  private static IProject getSelectedProject() {
```
chanseokoh: Ah, good point about the new project wizard.

elharo: It would be easy to simply not check constraints for the new project case. However, it's harder and will require some more thought to handle those cases we do want to check. It may be that the new project case should have hard-coded methods that find the relevant unfiltered lists of projects.

elharo:
The example expression used here is

```
  <extension point="org.eclipse.core.expressions.definitions">
    <definition id="com.google.cloud.tools.eclipse.appengine.onlyInStandardProjects">
      <adapt type="org.eclipse.core.resources.IProject">
        <test
            property="org.eclipse.wst.common.project.facet.core.projectFacet"
            value="com.google.cloud.tools.eclipse.appengine.facets.standard" />
      </adapt>
    </definition>
  </extension>
```

This requires an IProject. But I wonder if we could configure an expression definition to support multiple types? E.g. if the default variable is an IProject use that, otherwise use some other class?

elharo:
I've now verified that it is possible to configure expressions such that they work both with and without an IProject. That is, we can still use an IProject as the default variable when one is present and use something else (likely custom) when it isn't.

chanseokoh:
Need to merge with master (#2182 necessary) to make Travis work.

elharo:
I'm going to close this. Looks like we're going down another path.

Issue #435, opened 2016-07-22
Define tab group for App Engine Standard Launch configuration type


elharo:
![image](https://cloud.githubusercontent.com/assets/1005544/17070067/ff47b1f2-5027-11e6-8c00-6c20c26e1f3a.png)

discovered as part of #429

elharo:
Even an empty tag group would be a good start.

elharo:
I suspect we hook it up in localserver/plugin.xml here:

    <extension point="org.eclipse.debug.ui.launchConfigurationTypeImages">    
      <launchConfigurationTypeImage
           configTypeID="com.google.cloud.tools.eclipse.AppToolsLaunchConfigurationType"   
           icon="platform:/plugin/com.google.cloud.tools.eclipse.appengine.ui/icons/gcp-16x16.png"   
           id="com.google.cloud.tools.eclipse.ExternalLaunchConfigurationTypeImage">   
      </launchConfigurationTypeImage>   
    </extension>

elharo:
FYI, going forward this may be a good place to put port configuration. We need to think about that.

PR #2125, opened 2017-07-10
Set auto-restart based on launch mode, and max-modules=1  in debug mode

### LINKED ISSUE #1428: Changes to Java files are not published to local devserver (devappserver2)
If the user changes the source code while the local devappserver is running, the changes are noticed by the runtime (i.e. the Servers view indicates that the application should be republished). But running publish from the context menu of the server does not actually refresh the code, although the Servers view indicates that the application is not in sync with the sources.

### LINKED ISSUE #2111: Changes to Java files are not published to local devserver (devappserver1)
- Cloud Tools for Eclipse Version: 1.2.0.201706082053
- OS: Windows 10 10.0
- Java Version: 1.8.0_73

Last update made it so that publishing locally doesn't update all the changed code. 
Working on a web application and if i make changes to the front end, so the basic JSP files, HTML/CSS, all that updates fine on publish locally.
But when trying to publish edited code in the backend..(Servlets,Models,Database), it doesn't work. Only when i restart the server do the changes register (when editing backend code).. 

Before the update, everything was working smoothly and didn't have this problem. Looked into the settings but didn't see anything that could be causing this to happen. Forgive me if i'm asking a noob question and forgot to flip a switch somewhere but pretty sure this is a bug.

briandealwis:
Change `LocalAppEngineServerLaunchConfigurationDelegate` to:
  - set _automatic-restart=true_ when launching in run mode, and _false_ in debug mode
  - only set _max-modules=1_ when launching in debug mode (only relevant for devappserver2; devappserver1 only has a single module)

Makes devappserver 1 or 2 selectable by setting `com.google.cloud.tools.eclipse.appengine.devappserver` property to `1` (default) or `2`.

Requires https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/435
Fixes #1428 
Fixes #2111

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
@@ -94,8 +94,7 @@ import org.eclipse.wst.server.core.ServerUtil;
 public class LocalAppEngineServerLaunchConfigurationDelegate
     extends AbstractJavaLaunchConfigurationDelegate {
 
-  static final boolean DEV_APPSERVER2 =
-      "2".equals(System.getProperty("com.google.cloud.tools.eclipse.appengine.devappserver", "1"));
+  static final boolean DEV_APPSERVER2 = false;
   
   private static final Logger logger =
       Logger.getLogger(LocalAppEngineServerLaunchConfigurationDelegate.class.getName());
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
@@ -94,7 +94,8 @@ import org.eclipse.wst.server.core.ServerUtil;
 public class LocalAppEngineServerLaunchConfigurationDelegate
     extends AbstractJavaLaunchConfigurationDelegate {
 
-  private static final boolean DEV_APPSERVER2 = false;
+  static final boolean DEV_APPSERVER2 =
+      "2".equals(System.getProperty("com.google.cloud.tools.eclipse.appengine.devappserver", "1"));
   
   private static final Logger logger =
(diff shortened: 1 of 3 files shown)


chanseokoh:
This will affect only dev_appserver2. Can we say this fixes #2111, as we currently use dev_appserver1?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -94,7 +94,8 @@
 public class LocalAppEngineServerLaunchConfigurationDelegate
     extends AbstractJavaLaunchConfigurationDelegate {
 
-  private static final boolean DEV_APPSERVER2 = false;
+  static final boolean DEV_APPSERVER2 =
```
elharo: This is my concern. I'm not sure this should be configurable via a system property, or other user preference.

briandealwis:
Amended title and description, and description for #1428 and #2111.  Unfortunately devappserver1 doesn't expose any knobs to change auto-restart behaviour.

ludoch:
The devappserver1  knob to change auto-restart behaviour is to pass a sys property 

-Dappengine.fullscan.seconds=XX
where XX is short (1 or 2...)

to the devappserver1 JVM process.

ludoch:
For example of Gradle usage, see https://github.com/GoogleCloudPlatform/gradle-appengine-plugin/issues/97

briandealwis:
Updated patch to properly set _automatic-restart_ based on launch mode; with [appengine-plugins-core #435](https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/435) will work with both devappserver 1 and 2.  _max-modules_ doesn't apply to devappserver1.

(Oddly enough, _automatic-restart=true_ doesn't seem to work on devappserver2 at the moment.)

briandealwis:
PTAL @elharo

briandealwis:
This patch fixes both devappserver 1 and 2.  We will need to cut a new release of appengine-plugins-core though.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
elharo: Ping. I'm still not sold on the idea of condtionally enabling devappserver2 at this point int time. IMHO, we shoudl stick to devapopserver1 until devappserver2 is ready.

PR #2133, opened 2017-07-11
Utility to read version from Maven Central

elharo:
@tgroh @chanseokoh Initially I tried to just refactor DataflowArtifactRetriever but teasing out the dataflow dependencies from that class got way too complex, so I decided to build this in stages.


plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import org.apache.maven.artifact.versioning.ArtifactVersion;
+import org.apache.maven.artifact.versioning.InvalidVersionSpecificationException;
+import org.apache.maven.artifact.versioning.VersionRange;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ArtifactRetrieverTest {
```
tgroh: These tests will all perform remote calls, yes?

It is worth at minimum commenting that this is the case, and noting that in the case of an outage of the remote repositories the tests are expected to fail.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -0,0 +1,196 @@
...
+
+  private final LoadingCache<String, ArtifactVersion> latestVersion =
+      CacheBuilder.newBuilder()
+          .refreshAfterWrite(4, TimeUnit.HOURS)
+          .build(
+              new CacheLoader<String, ArtifactVersion>() {
+
+                @Override
+                public ArtifactVersion load(String coordinates) throws Exception {
+                  Document document = getMetadataDocument(coordinates);
```
tgroh: Is it worth caching the document as well?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -0,0 +1,196 @@
...
+      if (versionRange.containsVersion(version)) {
+        return version;
+      }
+    }
+    return null;
+  }
+
+  private Document getMetadataDocument(String coordinates) throws CoreException {
+    String[] x = keyToId(coordinates);
+    String groupId = x[0];
```
tgroh: Is it worth having an internal wrapper class with two fields that has real names for groupId/artifactId?

If you don't think so, then document the form that `coordinates` is permitted to be in wherever it is consumed (at least for the first time, i.e. with this style of key the cache should specify the requirements of its key)

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowMavenCoordinates.java
```diff
@@ -0,0 +1,33 @@
...
+public class DataflowMavenCoordinates {
+
+  static final String DATAFLOW_GROUP_ID = "com.google.cloud.dataflow";
+  static final String DATAFLOW_SDK_ARTIFACT = "google-cloud-dataflow-java-sdk-all";
+  /**
+   * Versions which are known to have been released.
+   */
+  static final NavigableSet<ArtifactVersion> KNOWN_VERSIONS =
+      ImmutableSortedSet.<ArtifactVersion>of(
+          new DefaultArtifactVersion("1.9.0"), new DefaultArtifactVersion("2.0.0-beta1"));}
```
elharo: done

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
elharo: Only the GetGuava tests, not the others

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: I guess we do load it twice. Rather caching the document, we could just update latest and all available versions at the same time.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: done

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -0,0 +1,198 @@
...
+public class ArtifactRetriever {
+  
+  private static final Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+  
+  @VisibleForTesting
+  static URL getMetadataUrl(String groupId, String artifactId) {
+    String groupPath = groupId.replace('.', '/');
+    try {
+      return new URL(
+          "https://repo1.maven.org/maven2/" + groupPath + "/" + artifactId + "/maven-metadata.xml");
```
briandealwis: Can you make the base URL replaceable?  This code will fail behind firewalls or who run with Maven proxies.  We should be able to pull this info from m2e.

(I don't think this should be static)

plugins/com.google.cloud.tools.eclipse.util/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Export-Package: com.google.cloud.tools.eclipse.util,
  com.google.cloud.tools.eclipse.util.status
 Import-Package: com.google.common.annotations;version="[20.0.0,21.0.0)",
  com.google.common.base;version="[20.0.0,21.0.0)",
+ com.google.common.cache;version="20.0.0",
```
chanseokoh: Should be "[20.0.0,21.0.0)" like others.

elharo: done

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: I was thinking we should push this code into appengine-plugins-core since IntelliJ will need it too. Thus I don't want to depend on m2e. Is there another way you would suggest setting the repo?

briandealwis: Pass it in the constructor, defaulting to `https://repo1.maven.org/maven2` (and perhaps use `repo.maven.apache.org`?).

CT4E would then use m2e to find the URL for `central`.  `MavenPlugin.getInstance().getRepositoryRegistry().getRepositories(bitMask)` returns a list of repositories, and bit mask would likely be just `IRepositoryRegistry.SCOPE_SETTINGS`.  We'd iterate through looking for a mirror of _central_, which has a repository URL.

elharo: Possible, though that would make it a non-singleton.

elharo: also, why repo.maven.apache.org instead of repo1.maven.org?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -0,0 +1,217 @@
...
+    return getLatestIncrementalVersion(idToKey(groupId, artifactId), range);
+  }
+
+  /**
+   * Returns the latest version of the specified artifact in the version range, or null if there is
+   * no such version.
+   * 
+   * @param coordinates Maven coordinates in the form groupId:artifactId
+   */
+  private ArtifactVersion getLatestIncrementalVersion(String coordinates, VersionRange range) {
```
briandealwis: What does _incremental_ mean?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
briandealwis: It's the network fetch that dominates here, right?  So this seems pretty important to address.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
tgroh: Maven version numbers are in the form of `Maj.Min.Inc`, so this is effectively the latest version within the provided version range.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: done

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
briandealwis: Could avoid this if `ArtifactRetriever` had a configurable URL-retrieval (`Function<URL,InputStream>`?).  It would also provide a hook in case we need to do something relating to network proxy (viz our `GoogleApiFactory` and `ProxyFactory` classes).

(Thinking back to our suspicion of some Travis failures being from the XML Schema validation hitting network issues trying to download schemas.)

elharo: That's more complexity than I'm willing to introduce just for testing.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowMavenCoordinates.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class DataflowMavenCoordinates {
+
+  static final String DATAFLOW_GROUP_ID = "com.google.cloud.dataflow";
+  static final String DATAFLOW_SDK_ARTIFACT = "google-cloud-dataflow-java-sdk-all";
+
+  /** Versions which are known to have been released. */
+  static final NavigableSet<ArtifactVersion> KNOWN_VERSIONS =
+      ImmutableSortedSet.<ArtifactVersion>of(
+          new DefaultArtifactVersion("1.9.0"), new DefaultArtifactVersion("2.0.0-beta1"));
```
tgroh: This can be updated to "2.0.0"

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
briandealwis: This will need to be resolved eventually, but it can wait.

PR #2159, opened 2017-07-17
Use Guava Preconditions

chanseokoh:
Also converted `Require-Bundle` to `Import-Package`.


plugins/com.google.cloud.tools.eclipse.dataflow.core/META-INF/MANIFEST.MF
```diff
@@ -37,9 +48,7 @@ Import-Package: com.google.cloud.tools.eclipse.util,
  org.eclipse.m2e.core.repository,
  org.osgi.framework,
  org.osgi.service.prefs
-Require-Bundle: com.google.cloud.tools.eclipse.googleapis,
- com.google.guava;bundle-version="[20.0.0,21.0.0)",
- org.eclipse.m2e.maven.runtime,
+Require-Bundle: org.eclipse.m2e.maven.runtime,
```
briandealwis: Could this not be turned into `Import-Package`s too?

chanseokoh: I tried this the other day, but I was not able to make it work.

chanseokoh:
`com.google.api.client.util` is currently in `.googleapis` and `.projectutils`. I actually tried removing them, and there was no compilation error. However, I decided to back it out, as I remembered for `.googleapis` removing some imports in `MANIFEST.MF` don't cause compilation error but are nonetheless required, which otherwise would result in NoClassDefFound at runtime.

PR #2164, opened 2017-07-18
Make AESv8 require Java 8

### LINKED ISSUE #2161: Launch failure with AESv8 runtime but Java 1.7 facet
We current allow the AESv8 facet to support the Java 7 facet, effectively treating the Java facet as a language support level.  #1928 ensures that we use a project-compatible JVM to launch the project; if the Java 7 facet is used, then we may attempt to launch with a Java 1.7 JVM if one is present.

But when the project bears a `<runtime>java8</runtime>`, `appengine-plugins-core` launches the devappserver with `-Duse_jetty9_runtime=true`.  Although Jetty 9.2 supported Java 7, 9.3 and beyond [now require Java 8](https://www.eclipse.org/jetty/documentation/current/what-jetty-version.html).

The AESv8 facet should only support the Java 8 facet.

briandealwis:
Force the _App Engine Standard JRE8_ facet to require the Java 8 facet.  This was a bit more involved than I had anticipated as we want to choose the best version of the AES facet given the current settings, and avoid upgrading the Java version only if required.

Fixes #2161


briandealwis:
PTAL

PR #2167, opened 2017-07-19
Initialize ILaunchConfiguration properly when using project-specific or workspace-wide default values

### LINKED ISSUE #2166: PipelineArgumentsTab should properly initialize argumentValues of PipelineLaunchConfiguration
```java
  @Override
  public void performApply(ILaunchConfigurationWorkingCopy configuration) {
   ...
    Map<String, String> overallArgValues = new HashMap<>(launchConfiguration.getArgumentValues());
    if (!defaultOptionsComponent.isUseDefaultOptions()) {
      overallArgValues.putAll(defaultOptionsComponent.getValues());
    }
    ...
    launchConfiguration.setArgumentValues(overallArgValues);
    ...
  }
```

The `overallArgValues` is supposed to hold a GCP project ID, GCS staging bucket, and soon-to-be-introduced Google account email. `defaultOptionsComponent` is the UI part that contains these values. However, as can be seen above, `overallArgValues` is initialized only when `isUseDefaultOptions` is true (when a user decides to use values defined in the current launch configuration which will override project-wide or workspace-wide values). I think we should initialize the `overallArgValues` too when `isUseDefaultOptions` is false, with the values from a project or a workspace.

chanseokoh:
Fixes #2166.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
@@ -399,7 +399,6 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
           IJavaLaunchConfigurationConstants.ATTR_PROJECT_NAME, "");
       return getProject(projectValue);
     } catch (CoreException ex) {
-      // BTW, I don't think this can really happen.
       DataflowUiPlugin.logError(ex, "Cannot read project value from launch configuration.");
       return null;
     }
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -382,10 +393,25 @@ private PipelineOptionsHierarchy getPipelineOptionsHierarchy(
     return pipelineOptionsHierarchyFactory.global(monitor);
   }
 
+  private IProject getProject(ILaunchConfiguration configuration) {
+    try {
+      String projectValue = configuration.getAttribute(
+          IJavaLaunchConfigurationConstants.ATTR_PROJECT_NAME, "");
+      return getProject(projectValue);
+    } catch (CoreException ex) {
+      // BTW, I don't think this can really happen.
```
elharo: Javadoc of getAttribute says otherwise.

chanseokoh: Yeah, maybe this can happen when having trouble reading a value from a launch config saved on disk.

chanseokoh:
Closing as I backed out of #2166.

PR #2173, opened 2017-07-19
Load latest, unspecified library version

elharo:
@chanseokoh @briandealwis 

@tgroh Is dataflow ever likely to need to load a non-release version?


chanseokoh:
Probably a good idea to merge with master to pull in #2182, which has a Travis fix.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -134,8 +137,14 @@ private static MavenCoordinates getMavenCoordinates(IConfigurationElement[] chil
     if (!Strings.isNullOrEmpty(repository)) {
       mavenCoordinates.setRepository(repository);
     }
-    
+
+    // Only look up latest version if version isn't specified in file.
     String version = mavenCoordinatesElement.getAttribute(ATTRIBUTE_NAME_VERSION);
+    if (Strings.isNullOrEmpty(version) || "LATEST".equals(version)) {
+      // todo don't load beta versions
+      version = retriever.getLatestArtifactVersion(groupId, artifactId).toString();
```
chanseokoh: The (original) Javadoc of `getLatestArtifactVersion` says it can potentially return `null`:

> Returns the latest published artifact version, or null if there is no such version.

However it is not clear to me how there can be no published version at all, other than providing wrong coordinates (in which case, the `pom.xml` is just wrong?)

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -193,6 +169,16 @@ private ArtifactVersion getLatestIncrementalVersion(String coordinates, VersionR
     }
   }
 
+  private static ArtifactVersion getLatestReleasedVersion(
+      NavigableSet<ArtifactVersion> allVersions) {
+    for (ArtifactVersion version : allVersions.descendingSet()) {
```
chanseokoh: Just checking in. Does this sorting works correctly as below?

- "19.0" comes before "2.0".
- "1.0.0" comes before "1.0.0-SNAPSHOT".

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
```diff
@@ -30,10 +29,20 @@
   private ArtifactRetriever retriever = new ArtifactRetriever();
   
   @Test
-  public void testGetGuavaLatest() {
+  public void testGetServletLatest() {
+    // group and version IDs changed with 3.0 so this should be stable
+    ArtifactVersion servlet = retriever.getLatestArtifactVersion("javax.servlet", "servlet-api");
```
chanseokoh: About the unit tests making network connections, one option would be to think of these as integration tests and move them to the integration bundle? (I thought this can fall into the integration test category conceptually.)

Otherwise, we could create a new issue to mock this part later.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -146,30 +120,32 @@ public ArtifactRetriever() {
   }
 
   /**
-   * Returns the latest published artifact version, or null if there is no such version.
+   * Returns the latest published non-beta artifact version, or null if there is no such version.
```
chanseokoh: In this case where you consider the entire version range, I think the correct doc is something like below, unless you did not intend it.

 "The latest published non-beta artifact version or the latest published beta version if there all versions are beta. Otherwise null (i.e., wrong groupId or artifactId)." I was a bit confused before I looked into the `getLatestIncrementalVersion()` code.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -146,30 +120,32 @@ public ArtifactRetriever() {
...
    */
   public ArtifactVersion getLatestArtifactVersion(String groupId, String artifactId) {
     return getLatestIncrementalVersion(idToKey(groupId, artifactId), null);
   }
 
   /**
-   * Returns the latest published artifact version in the version range, or null if there is no such
-   * version.
+   * Returns the latest published non-beta artifact version in the version range, 
+   * or null if there is no such version.
```
chanseokoh: This can still return a beta artifact, if all versions are beta in the given range.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -146,30 +120,32 @@ public ArtifactRetriever() {
...
   public ArtifactVersion getLatestArtifactVersion(
       String groupId, String artifactId, VersionRange range) {
     return getLatestIncrementalVersion(idToKey(groupId, artifactId), range);
   }
 
   /**
-   * Returns the latest version of the specified artifact in the version range, or null if there is
-   * no such version.
+   * Returns the latest non-beta version of the specified artifact in the version range,
+   * or null if there is no such version.
```
chanseokoh: Ditto.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -134,8 +137,14 @@ private static MavenCoordinates getMavenCoordinates(IConfigurationElement[] chil
     if (!Strings.isNullOrEmpty(repository)) {
       mavenCoordinates.setRepository(repository);
     }
-    
+
+    // Only look up latest version if version isn't specified in file.
     String version = mavenCoordinatesElement.getAttribute(ATTRIBUTE_NAME_VERSION);
+    if (Strings.isNullOrEmpty(version) || "LATEST".equals(version)) {
+      // todo don't load beta versions
```
elharo: done; comment was out of date.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: There can be no non-beta version.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
elharo: good idea; done

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: It should return null if there are no non-beta versions.

chanseokoh: Actually, it should say "non-qualifier" rather than "beta".

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: In descending order yes.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
chanseokoh: So, should we prepared for NPE (for `.toString()`)?

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
chanseokoh: Remove the three network-using tests, since they were moved to the integration bundle.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
chanseokoh: I think we should still update the Javadoc, because it can return a beta version.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/ArtifactRetrieverTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: how about "release"

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
elharo: I don't see a code path that allows that to happen. Do you see one?

chanseokoh: I don't see it either. I thought it was another potentially useful util method.

chanseokoh: > I don't see it either. I thought it was another potentially useful util method.

Sorry, that was a comment at a wrong place.

If you look at `getLatestIncrementalVersion()`, it first tries to get a latest release (without a qualifier) version:

```java
      NavigableSet<ArtifactVersion> allVersions = availableVersions.get(coordinates);
      ArtifactVersion latest = getLatestReleasedVersion(allVersions);
      if (range == null || range.containsVersion(latest)) {
        return latest;
      }
```

If that fails, it falls back to include qualifier versions. 

```java
      NavigableSet<ArtifactVersion> allVersions = availableVersions.get(coordinates);
      ArtifactVersion latest = getLatestInRange(range, allVersions);
      return latest;
```

chanseokoh:
Generally looks good to me, but need to resolve conflicts.

elharo:
Merged. PTAL @chanseokoh

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactoryTest.java
```diff
@@ -16,28 +16,73 @@
...
+  
+  @Test
+  public void testCreate_useLatestVersion() throws LibraryFactoryException {
+    Mockito.when(configuration.getChildren("libraryFile")).thenReturn(libraryFiles);
+    
+    Library library = factory.create(configuration);
+    String version = library.getLibraryFiles().get(0).getMavenCoordinates().getVersion();
+    DefaultArtifactVersion artifactVersion = new DefaultArtifactVersion(version);
+    int majorVersion = artifactVersion.getMajorVersion();
+    Assert.assertTrue(majorVersion >= 22);
```
chanseokoh: Where does this number 22 come from? I can't find it.

elharo: latest version of Guava at the time of writing. (Now 23). When the configuration does not specify a specific version of the artifact, the maven repo is queried to find the latest available release version.

PR #2190, opened 2017-07-22
Rework test dependencies

briandealwis:
This was extracted from #1847, where we were looking at supporting Maven-style in-project unit tests.  As part of that effort, I started hitting package-uses conflicts as we had mixtures of Hamcrest bundles.  This change pulls our test dependencies like Mockito, Hamcrest, and JUnit into a single bundle, `com.google.cloud.tools.eclipse.test.dependencies`.  It simplifies our `MANIFEST.MF`s as we don't need mandatory attributes to pull in just the right bundles.


plugins/com.google.cloud.tools.eclipse.test.dependencies/META-INF/MANIFEST.MF
```diff
@@ -5,38 +5,99 @@ Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.dependencies;singleton:
...
+   org.junit.runner,
+   org.mockito.internal.debugging,
+   org.mockito.internal.runners,
+   org.mockito.internal.util,
+   org.junit.runner.manipulation",
+ org.mockito.stubbing;provider=google;version="1.10.19";uses:="org.mockito.invocation,org.mockito.internal.progress",
+ org.mockito.stubbing.answers;provider=google;version="1.10.19";uses:="org.mockito.internal.stubbing.answers",
+ org.mockito.verification;provider=google;version="1.10.19";uses:="org.mockito.internal.verification.api,org.mockito.internal.util",
+ org.objenesis;provider=google;version="2.2.0";uses:="org.objenesis.instantiator,org.objenesis.strategy"
+Bundle-Classpath: lib/mockito-core-1.10.19.jar,
```
chanseokoh: Do we still need these jars? You are importing them from Orbit, right?

pom.xml
```diff
@@ -336,6 +356,17 @@
...
+
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-compiler-plugin</artifactId>
+          <version>2.5.1</version>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-surefire-plugin</artifactId>
+          <version>2.12.4</version>
```
chanseokoh: Curiosity: can you explain why these are needed?

plugins/com.google.cloud.tools.eclipse.test.dependencies/pom.xml
```diff
@@ -12,7 +12,46 @@
...
+            </goals>
+            <configuration>
+              <outputDirectory>lib</outputDirectory>
+              <ignoredDependencies>
+                <!-- ignore ourselves; otherwise copies our Bundle-Classpath -->
+                <ignoreDependency>com.google.cloud.tools.eclipse:com.google.cloud.tools.eclipse.test.dependencies</ignoreDependency>
+              </ignoredDependencies>
+              <!-- includeScope defaults to test -->
+              <excludeScope>runtime</excludeScope>
+            </configuration>
```
chanseokoh: Can you explain what the net result of using this maven-dependency-plugin is? I'm trying to understand what's going on.

briandealwis: I'll add a note to the pom to explain.  Also realized that this benefits from an m2e connector for the `maven-dependency-plugin` to be installed, so I'll add a note to README.md

pom.xml
briandealwis: A semi-mistake: it was from the `unit` branch for the in-bundle unit-tests.  The `c.g.c.t.e.appengine.standard.java8` bundle includes some such unit tests and includes these plugins with a version. I'll remove those versions to pick up from here.

plugins/com.google.cloud.tools.eclipse.test.dependencies/META-INF/MANIFEST.MF
chanseokoh: OK, I think they are still needed.

briandealwis: Yes, they're needed: these are the jars copied down from Maven Central.

eclipse/ide-target-platform/category.xml
```diff
@@ -25,7 +25,6 @@
    <bundle id="org.eclipse.jetty.servlet" version="0.0.0"/>
    <bundle id="org.eclipse.jetty.server" version="0.0.0"/>
    <bundle id="org.eclipse.jetty.util" version="0.0.0"/>
-   <bundle id="org.hamcrest" version="1.1.0.v20090501071000"/>
```
briandealwis: @chanseokoh here's the problem: I removed the `org.hamcrest` 1.1 bundle from `category.xml` but failed to add the 1.3 bundles.

PR #2198, opened 2017-07-25
Set project JDK for deploy staging

### LINKED ISSUE #2195: Use project SDK when staging App Engine apps for deploy
- Eclipse version: _Oxygen Release (4.7.0)_
- Cloud Tools for Eclipse version: _1.2.0.201706082053_
- Google Cloud SDK version: _163.0.0_
- OS: _Windows 10 v1703_
- Java version: _jdk1.8.0_111_

I have a Maven App Engine standard project which contains jsp files (and this is the reason I think that the project is looking for JDK)

If I deploy the project with `appengine:deploy` works fine
If I deploy with the plugin I get
```
Unable to stage:
java.lang.RuntimeException: Cannot get the System Java Compiler. Please use a JDK, not a JRE.
	at com.google.appengine.tools.admin.Application.compileJspJavaFiles(Application.java:1201)
	at com.google.appengine.tools.admin.Application.compileJsps(Application.java:1178)
	at com.google.appengine.tools.admin.Application.populateStagingDirectory(Application.java:901)
	at com.google.appengine.tools.admin.Application.createStagingDirectory(Application.java:809)
	at com.google.appengine.tools.admin.AppAdminImpl.stageApplication(AppAdminImpl.java:539)
	at com.google.appengine.tools.admin.AppAdminImpl.stageApplicationWithDefaultResourceLimits(AppAdminImpl.java:492)
	at com.google.appengine.tools.admin.AppCfg$StagingAction.execute(AppCfg.java:2510)
	at com.google.appengine.tools.admin.AppCfg.executeAction(AppCfg.java:393)
	at com.google.appengine.tools.admin.AppCfg.<init>(AppCfg.java:216)
	at com.google.appengine.tools.admin.AppCfg.<init>(AppCfg.java:116)
	at com.google.appengine.tools.admin.AppCfg.main(AppCfg.java:112)
com.google.appengine.tools.admin.AdminException: Unable to stage app: Cannot get the System Java Compiler. Please use a JDK, not a JRE.
	at com.google.appengine.tools.admin.AppAdminImpl.stageApplication(AppAdminImpl.java:543)
	at com.google.appengine.tools.admin.AppAdminImpl.stageApplicationWithDefaultResourceLimits(AppAdminImpl.java:492)
	at com.google.appengine.tools.admin.AppCfg$StagingAction.execute(AppCfg.java:2510)
	at com.google.appengine.tools.admin.AppCfg.executeAction(AppCfg.java:393)
	at com.google.appengine.tools.admin.AppCfg.<init>(AppCfg.java:216)
	at com.google.appengine.tools.admin.AppCfg.<init>(AppCfg.java:116)
	at com.google.appengine.tools.admin.AppCfg.main(AppCfg.java:112)
Caused by: java.lang.RuntimeException: Cannot get the System Java Compiler. Please use a JDK, not a JRE.
	at com.google.appengine.tools.admin.Application.compileJspJavaFiles(Application.java:1201)
	at com.google.appengine.tools.admin.Application.compileJsps(Application.java:1178)
	at com.google.appengine.tools.admin.Application.populateStagingDirectory(Application.java:901)
	at com.google.appengine.tools.admin.Application.createStagingDirectory(Application.java:809)
	at com.google.appengine.tools.admin.AppAdminImpl.stageApplication(AppAdminImpl.java:539)
	... 6 more

```

I don't think it is a project or OS error because using directly Maven the deploy works fine

Here is my Java installation
image

Eclipse configuration
image

Project configuration
image

chanseokoh:
~~Fixes #2195.~~ (incomplete: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2195#issuecomment-318439239)

I tried to write a fancy test to see if the `CloudSdk` picks up the right JDK, but I realized that our test Eclipse instance will have only one default VM installed (which could be either Java 7 or Java 8 per Travis execution env).


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
```diff
@@ -216,9 +220,16 @@ private CloudSdk getCloudSdk(Path credentialFile,
         .exitListener(processExitListener)
         .appCommandMetricsEnvironment(CloudToolsInfo.METRICS_NAME)
         .appCommandMetricsEnvironmentVersion(CloudToolsInfo.getToolsVersion())
-        .appCommandOutputFormat("json")
-        .build();
-    return cloudSdk;
+        .appCommandOutputFormat("json");
+
+    IJavaProject javaProject = JavaCore.create(project);
```
briandealwis: I wonder if it might not be better to push this to the creator of the DeployJob in DeployCommandHandler.

chanseokoh: What's the benefit or rational? `DeployJob` can figure this out by itself, and we have one less parameter a caller has to pass in.

briandealwis: The JDK/JRE to use seems like another piece of configuration for the DeployJob,  and the DeployCommandHandler is already doing a lot of configuration.  I think it's easier to understand and predict if configuration computation like this is kept in one place.  We may find that the project JDK may not be sufficient for CloudSdk (e.g., may need Java 8).

chanseokoh: > and the DeployCommandHandler is already doing a lot of configuration.

I'm going to digress a bit, but I don't think this is really true, because what it does is merely pre-constructing a `CloudSdk`-specific class from the save project preferences. (BTW, I think is a not good design. `DeployJob` can completely hide `CloudSdk`-specific stuff.) Sure, it additionally passes a work directory, but Elliotte once said it shouldn't (he left a comment somewhere) and rather `DeployJob` should just create one inside. I agree on that. Then there's some message console arguments, but this is just a separate between the background Job and the UI command handler; we don't have other options but to pass the UI arguments. So, then it just boils down to passing the actual user deploy preferences.

We may find that the project JDK may not be sufficient for CloudSdk (e.g., may need Java 8).

Going back to the main discussion, that's a good point. Now I think we should have a separate `CloudSdk` instance, because the issue here is only for case of App Engine standard staging. And for that `CloudSdk`, I think it should always use the project-specific JDK, which shouldn't be configurable. What you pointed out is about running `CloudSdk` in general, and that could be made configurable (made sometime later, if a need arises). I'll do some refactoring.

chanseokoh: This refactoring is going to take quite some time. There are many things to clean up to share code.

chanseokoh:
@briandealwis I refactored `DeployJob` and created `CloudSdkProcessFacade` that hides the complex low-level gcloud process stuff. It's now very simple to stage or deploy using `CloudSdk` from the `DeployJob` perspective.

A project JDK is set only for staging `CloudSdk` instance.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
chanseokoh: > What you pointed out is about running CloudSdk in general, and that could be made configurable (made sometime later, if a need arises).

But then, deploy does not require Java at all. It's only for the staging.

chanseokoh:
Hmm... maybe I could move the standard staging-specific code (`CloudSdkProcessFacade.forStandardStaging()`) down to `stager`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
```diff
@@ -133,28 +112,54 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
         return saveStatus;
       }
 
-      IStatus stagingStatus = stageProject(credentialFile, stagingDirectory, progress.newChild(30));
+      synchronized (this) {
```
briandealwis: Why's this necessary to be synchronized? runInWorkspace() should be guarded.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkProcessFacade.java
```diff
@@ -0,0 +1,174 @@
...
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+
+public class CloudSdkProcessFacade {
```
briandealwis: Would definitely benefit from a class comment.  Describe the specifics of how it configures the underlying CloudSdk instance.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
```diff
@@ -133,28 +112,54 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
+        if (canceled) {
+          return Status.CANCEL_STATUS;
+        }
+        deployCloudSdkFacade = CloudSdkProcessFacade.forDeploy(credentialFile, stderrLineListener);
+        stagingCloudSdkFacade = CloudSdkProcessFacade.forStandardStaging(getProjectVm(project),
```
briandealwis: Odd to have `forStandardStaging()` when this could be for flexible.

Perhaps both of these are moved into the appropriate stager/deployer?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
```diff
@@ -208,8 +192,7 @@ private static void launchCleanupJob() {
   }
 
   private static IPath getTempDir() {
-    return Platform
-        .getStateLocation(Platform.getBundle("com.google.cloud.tools.eclipse.appengine.deploy"))
+    return Platform.getStateLocation(FrameworkUtil.getBundle(DeployJob.class))
```
briandealwis: `DeployCommandHandler.class`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
chanseokoh: This part may change a bit, but I'll explain anyway. It's guarding the section against `canceling()`. I believe `canceling()` can be executing while the job is running. It's called when a cancellation request is made:

```
	 * A hook method indicating that this job is running and {@link #cancel()}
	 * is being called for the first time.
	 * <p>
	 * Subclasses may override this method to perform additional work when
	 * a cancelation request is made.  This default implementation does nothing.
```

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/DeployJob.java
chanseokoh: I'll move these into the appropriate stager/deployer.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
chanseokoh: I noticed that previously it isn't the `.deploy.ui` but just `.deploy'. I think the first author wanted to have the directory under the non-UI bundle, which seems OK to me.

chanseokoh:
Hold on. I pushed a commit just see if tests will pass.

chanseokoh:
I expect the tests to pass. As a first step for #2195, this PR will just make the standard staging use a project JRE. The sophisticated logic to find any suitable JDK (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2195#issuecomment-318439239) beyond the project JRE will be the next PR.

@briandealwis PTAL

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
briandealwis: Worth a comment.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/StagingDelegate.java
```diff
@@ -52,4 +54,10 @@ IStatus stage(IProject project, IPath stagingDirectory,
    */
   IPath getOptionalConfigurationFilesDirectory();
 
+  /**
+   * Does not guarantee cancellation/termination. This can be called concurrently from a different
+   * thread than the thread running {@link #stage}.
+   */
+  void interrupt();
```
briandealwis: Worth making this `abstract`?  You override it in both Standard and Flexible.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -458,12 +458,12 @@ public void launch(ILaunchConfiguration configuration, String mode, final ILaunc
...
-      
+
       IJavaProject javaProject = JavaCore.create(modules[0].getProject());
       IVMInstall vmInstall = JavaRuntime.getVMInstall(javaProject);
-      
-      String javaHome = vmInstall.getInstallLocation().getAbsolutePath();
-      serverBehaviour.startDevServer(mode, devServerRunConfiguration, Paths.get(javaHome),
+
+      Path javaHome = vmInstall.getInstallLocation().toPath();
+      serverBehaviour.startDevServer(mode, devServerRunConfiguration, javaHome,
```
briandealwis: To be honest, it would probably be best passing the `IVMInstall` into `startDevServer()` (cf https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/353), though that can be done whenever that issue is tackled.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/StagingDelegate.java
chanseokoh: No need. This is an `interface`.

briandealwis: Uh weird!  I'm sure what I saw was a class with an empty method body.  Of course you're right.

PR #2201, opened 2017-07-27
Configure .test.dependencies to use specific artifacts

briandealwis:
Simplify `.test.dependencies` to copy specific artifacts rather than copying a filtered set of `test`-scoped dependencies.  This avoids having artifacts cluttering the `lib/` directory that are deliberately avoided (like `hamcrest-core-1.1`, which is pulled in from `mockito-core`, since we pull in hamcrest 1.3 via Orbit instead).


chanseokoh:
After landing this, we should do `git rm --cached /path/to/already/committed/jars`.

https://stackoverflow.com/questions/6535362/gitignore-after-commit

briandealwis:
Now landed.

PR #2209, opened 2017-07-31
Add hamcrest.core and .library to category.xml / remove hamcrest.integration from .test.dependencies

chanseokoh:
I can check that this does pull in the said bundle, but is this the best?

Another option would be to remove it from the `MANIFEST.MF` in `.test.dependencies`, because no one seems to use it for now. But for completeness, it may be a good idea to leave it there whether it's used or not.

```
Require-Bundle: org.junit;bundle-version="[4.12.0,4.12.1)";visibility:=reexport,
 org.hamcrest.core;bundle-version="[1.3.0,1.3.1)";visibility:=reexport,
 org.hamcrest.integration;bundle-version="[1.3.0,1.3.1)";visibility:=reexport,
 org.hamcrest.library;bundle-version="[1.3.0,1.3.1)";visibility:=reexport
```

### DIFF
--- a/eclipse/ide-target-platform/category.xml
+++ b/eclipse/ide-target-platform/category.xml
@@ -25,7 +25,8 @@
    <bundle id="org.eclipse.jetty.servlet" version="0.0.0"/>
    <bundle id="org.eclipse.jetty.server" version="0.0.0"/>
    <bundle id="org.eclipse.jetty.util" version="0.0.0"/>
-   <bundle id="org.hamcrest.integration" version="0.0.0"/>
+   <bundle id="org.hamcrest.core" version="1.3.0.v201303031735"/>
+   <bundle id="org.hamcrest.library" version="1.3.0.v201505072020"/>
    <bundle id="org.mortbay.jetty.server" version="6.1.26"/>
    <bundle id="org.mortbay.jetty.util" version="6.1.26"/>
    <bundle id="org.freemarker.freemarker" version="2.3.25.stable-incubating"/>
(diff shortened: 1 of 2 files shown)


briandealwis:
We [previously pulled in `org.hamcrest` 1.1 from Orbit](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2190/files#diff-2a1709698f95b9b69afeeddd27df4deaL28), which pulled in 1.3 of `org.hamcrest.core`, `org.hamcrest.integration`, and `org.hamcrest.library`.  

JUnit's `Assert.assertThat()` implementations delegates to `.integration`'s `MatcherAssert.assertThat()`.  Which is somewhat odd as `org.junit` has a `Require-Bundle: org.hamcrest.core` but I suppose most installations seem to include `org.hamcrest` 1.1, and so it all just seems to work.

So I think we should change the category.xml to explicitly include all 3 hamcrest bundles: `org.hamcrest.core`, `org.hamcrest.integration`, and `org.hamcrest.library`.

chanseokoh:
With what version? This PR uses "0.0.0".

briandealwis:
Sorry, I was updating my comment as you wrote the above.

My preference would be to explicitly include `org.hamcrest.core` and `org.hamcrest.library` 1.3, and ditch `org.hamcrest.integration` from `com.google.cloud.tools.eclipse.test.dependencies`.

chanseokoh:
FTR: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2190#pullrequestreview-53385395

PR #2221, opened 2017-08-02
Change project creation URL

### LINKED ISSUE #2023: Link in deploy dialog to create new GCP project doesn't seem to work
Not sure if we had a similar issue here or internally.

selection_027

No matter which account is selected, I am presented with the following screen on the Developer Console:

selection_028

chanseokoh:
Fixes #2023. The `authUser` part will continue to be appended as usual.


chanseokoh:
This should work. It'll take users to the App Engine application creation page after project creation.

chanseokoh:
The "authuser" parameter is dynamically added, depending on the current account selection:

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.3.0A/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java#L348

PR #2231, opened 2017-08-03
Pin springboot-appengine-standard test to use maven-jar-project 2.6

briandealwis:
m2e-wtp 1.2 (shipped with Mars) uses reflection to access certain fields in the `org.apache.maven.plugins:maven-jar-plugin`, which [broke with `maven-jar-plugin` 3.0.0](https://bugs.eclipse.org/bugs/show_bug.cgi?id=494583).  This can cause test failures in `com.google.cloud.tools.eclipse.integration.appengine.ImportMavenAppEngineStandardProjectTest`.  
```
Failed tests: 
  ImportMavenAppEngineStandardProjectTest.runImport:76 Imported Maven project has errors
/springboot-appengine-standard/pom.xml:1: org.apache.maven.archiver.MavenArchiver.getManifest(org.apache.maven.project.MavenProject, org.apache.maven.archiver.MavenArchiveConfiguration)
```

(For some reason, the test never fails on Travis.)

~~The easy fix is to pin `org.apache.maven.plugins:maven-war-plugin` to version 2.6 in `ImportMavenAppEngineStandardProjectTest`'s test project.  This test project is kept as a zip file, so the actual fix is as follows:~~
** patch removed: see below **


briandealwis:
Hold on, I just reproduced this again with the fix.

chanseokoh:
Got it. BTW, I noticed that sometimes the test passes locally. I think I was really unlucky to have 4 consecutive fails today. It's still true it fails easily though.

chanseokoh:
BTW, we don't see the detailed error message here, do we? Only the file name and the line number. I wonder `ProjectUtils.formatProblem()` is incomplete.

briandealwis:
I don't think there is a detailed message: there's a failure involving reflection.

This appears to be a transient error that disappears after a second or two; looks like there are some more jobs that we need to wait on in `waitForProjects()`.

briandealwis:
The correct patch is below (maybe time for an eye check):
```patch
--- springboot-appengine-standard/pom.xml.orig	2017-08-04 14:28:41.000000000 -0400
+++ springboot-appengine-standard/pom.xml	2017-08-04 14:34:57.000000000 -0400
@@ -108,7 +108,7 @@
 
       <plugin>
         <artifactId>maven-war-plugin</artifactId>
-        <version>3.1.0</version>
+        <version>2.6</version>
         <configuration>
           <failOnMissingWebXml>false</failOnMissingWebXml>
         </configuration>
```

briandealwis:
PTAL @chanseokoh

PR #2233, opened 2017-08-04
fix spelling

elharo:
@chanseokoh


elharo:
Yes, I think codecov has started being more aggressive about their checks.

PR #2238, opened 2017-08-04
add new dataflow wizard to toolbar button

### LINKED ISSUE #2216: Add Google Cloud Dataflow Project to Create New project submenu under the GCP toolbar icon

elharo:
@tgroh @chanseokoh fix #2216


briandealwis:
Ideally this should be done as a menuContribution from the `.dataflow.ui` bundle.  But this toolbar definition needs refactoring, so I opened #2239.

PR #2252, opened 2017-08-08
Copy Web Fragment JAR when deploying to App Engine

### LINKED ISSUE #2206: Deploy to App Engine Standard Skips spring-web jar file
(please ensure you are running the latest version of CT4E with _Help > Check for Updates_.)

- Cloud Tools for Eclipse version:  
Cloud Tools for Eclipse 1.2

- Google Cloud SDK version: _(run `gcloud version`)_
gcloud version
Google Cloud SDK 164.0.0
app-engine-go
app-engine-java 1.9.54
app-engine-python 1.9.57
bq 2.0.24
core 2017.07.25
gcloud
gsutil 4.27

- OS:
MacOS sierra 10.12

- Java version:
java -version
java version "1.8.0_141"
Java(TM) SE Runtime Environment (build 1.8.0_141-b15)
Java HotSpot(TM) 64-Bit Server VM (build 25.141-b15, mixed mode)

**What did you do?**
I ran Deploy to App Engine Standard from Eclipse

I expected my application to deploy successfully to app engine

**What did you see instead?**
My application does not completely deploy.  The Jar file spring-web-4.3.9.RELEASE.jar does not get deployed even though it is a maven dependency.  Therefore the application fails with a class not found exception.  The application runs fine in the local Eclipse environment.  I am using the Java8 runtime.

These are my Maven dependencies:

image

These are the files in the `workspace/.metadata/.plugins/com.google.cloud.tools.eclipse.appengine.deploy/tmp/1501385321020/staging/WEB-INF/lib` directory.  Notice how spring-web is missing:

```
ls -ln
total 18320
-rw-r--r--  1 501  20   106862 Jul 29 23:28 Astrolib-1.jar
-rw-r--r--  1 501  20    61829 Jul 29 23:28 commons-logging-1.2.jar
-rw-r--r--  1 501  20  2256213 Jul 29 23:28 guava-18.0.jar
-rw-r--r--  1 501  20    55784 Jul 29 23:28 jackson-annotations-2.8.0.jar
-rw-r--r--  1 501  20   282438 Jul 29 23:28 jackson-core-2.8.8.jar
-rw-r--r--  1 501  20  1238590 Jul 29 23:28 jackson-databind-2.8.8.jar
-rw-r--r--  1 501  20   414240 Jul 29 23:28 jstl-1.2.jar
-rw-r--r--  1 501  20   348450 Jul 29 23:28 objectify-5.1.13.jar
-rw-r--r--  1 501  20   380572 Jul 29 23:28 spring-aop-4.3.9.RELEASE.jar
-rw-r--r--  1 501  20   763050 Jul 29 23:28 spring-beans-4.3.9.RELEASE.jar
-rw-r--r--  1 501  20  1140303 Jul 29 23:28 spring-context-4.3.9.RELEASE.jar
-rw-r--r--  1 501  20  1121720 Jul 29 23:28 spring-core-4.3.9.RELEASE.jar
-rw-r--r--  1 501  20   263287 Jul 29 23:28 spring-expression-4.3.9.RELEASE.jar
-rw-r--r--  1 501  20   915475 Jul 29 23:28 spring-webmvc-4.3.9.RELEASE.jar
```

Any help would be very appreciated.  I'm stuck and cannot deploy this application.

chanseokoh:
Fixes #2206.

This works well, but I am not 100% certain if anything is missing or if there is a better implementation. So, putting up a PR to get early feedback.

Another problem is that compilation fails on Travis with the following error:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.0.0:compile (default-compile) on project com.google.cloud.tools.eclipse.appengine.deploy: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/GoogleCloudPlatform/google-cloud-eclipse/plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java:[92] 
[ERROR] 	IModule[] childModules = deployable.getChildModules();
[ERROR] 	                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] Access restriction: The method 'FlatComponentDeployable.getChildModules()' is not API (restriction on classpath entry '/home/travis/.m2/repository/p2/osgi/bundle/org.eclipse.wst.web/1.1.800.v201312041437/org.eclipse.wst.web-1.1.800.v201312041437.jar')
```

However, `mvn verify` works locally. Not sure why it fails on Travis.


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -63,10 +74,39 @@ private static void publish(IProject project, IPath destination, boolean explode
...
+      if ("jst.webfragment".equals(module.getModuleType().getId())
+          && module.getName().toLowerCase(Locale.US).endsWith(".jar")) {
+        ProjectModule projectModule = (ProjectModule)
+            module.loadAdapter(ProjectModule.class, monitor);
+
+        for (IModuleResource resource : projectModule.members()) {
+          IFile file = resource.getAdapter(IFile.class);
+          File file2 = resource.getAdapter(File.class);
+          if (file != null) {
+            webFragmentJars.add(new ModuleFile(file, resource.getName(), new Path("WEB-INF/lib")));
```
briandealwis: We should use the layout defined in the deployment assembly.

chanseokoh:
@briandealwis said he's looking into something in detail.

chanseokoh:
```
              // Need to set ModuleFile name to full path to get the right end-result, but may
              ...
                resources.add(new ModuleFile(..., deployPath, ...);
```

I doubt it. I tried printing out resource names (`.getName()`) of a sample web module, and what I get is like this:

```
META-INF/
  MANIFEST.MF
  maven/
    a/
      b/
        pom.properties
        pom.xml
WEB-INF/
  classes/
    a/
      HelloAppEngine.class
  appengine-web.xml
  lib/
    jstl-1.2.jar
    spring-webmvc-4.3.6.RELEASE.jar
    spring-aop-4.3.6.RELEASE.jar
    spring-beans-4.3.6.RELEASE.jar
    spring-context-4.3.6.RELEASE.jar
    spring-core-4.3.6.RELEASE.jar
    commons-logging-1.2.jar
    spring-expression-4.3.6.RELEASE.jar
  web.xml
favicon.ico
index.html
```

That is, resource names for files or folders are not full path. FYI, the code to traverse a list of resources:

```java
  private static void traverse(int d, IModuleResource[] resources) {
    for (IModuleResource resource : resources) {
      for (int i = 0; i < 2*d; i++) System.out.print(" ");
      
      if (resource instanceof IModuleFolder) {
        IModuleFolder folder = (IModuleFolder) resource;
        System.out.println(resource.getName() + "/");
        traverse(d + 1, folder.members());
      } else
        System.out.println(resource.getName());
    }
  }
```

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -63,10 +74,39 @@ private static void publish(IProject project, IPath destination, boolean explode
     J2EEFlexProjDeployable deployable =
         new J2EEFlexProjDeployable(project, ComponentCore.createComponent(project));
```
chanseokoh: That's cool. The implement is in your code above, right?

briandealwis:
I saw this in practice: previously I had the name set to `childModule.getName()` but that resulted in `spring-web-4.3.8.RELEASE.jar` being placed in the root.

chanseokoh:
> I saw this in practice: previously I had the name set to childModule.getName() but that resulted in spring-web-4.3.8.RELEASE.jar being placed in the root.

I also got into the same problem too. To me, the correct solution seems to set the `path` to be the parent folder (not the full path) of the resource. So, if I print the path together with the name (`System.out.println(resource.getModuleRelativePath() + ":" + resource.getName()`), the output becomes like this:

```
:META-INF/
  META-INF:MANIFEST.MF
  META-INF:maven/
    META-INF/maven:a/
      META-INF/maven/a:b/
        META-INF/maven/a/b:pom.properties
        META-INF/maven/a/b:pom.xml
:WEB-INF/
  WEB-INF:classes/
    WEB-INF/classes:a/
      WEB-INF/classes/a:HelloAppEngine.class
  WEB-INF:appengine-web.xml
  WEB-INF:lib/
    WEB-INF/lib:jstl-1.2.jar
    WEB-INF/lib:spring-webmvc-4.3.6.RELEASE.jar
    WEB-INF/lib:spring-aop-4.3.6.RELEASE.jar
    WEB-INF/lib:spring-beans-4.3.6.RELEASE.jar
    WEB-INF/lib:spring-context-4.3.6.RELEASE.jar
    WEB-INF/lib:spring-core-4.3.6.RELEASE.jar
    WEB-INF/lib:commons-logging-1.2.jar
    WEB-INF/lib:spring-expression-4.3.6.RELEASE.jar
  WEB-INF:web.xml
:favicon.ico
:index.html
```

briandealwis:
Ah, that's an interesting find.

chanseokoh:
`IJ2EEModule.isBinary()` (or `IWebFragmentModule.isBinary()`) has an interesting Javadoc. I think checking `isBinary()` covers our case. guaranteeing that it's a zipped module and there's only one member.

```
	 * Returns true if this is a binary (zipped) module, and
	 * false if it is expanded.
	 * 
	 * <p>If true, members() should return only a single element -
	 * the binary (jar or zip file) that contains the contents of
	 * this module. (a single IModuleResource, e.g.
	 * myejb.jar) Also, getLocation() should return the full path
	 * up to and including the binary itself. (e.g.
	 * c:\temp\myejb.jar)</p>
	 * 
	 * <p>If false, members() should return the entire contents
	 * of the module, starting at the root. There should be no
	 * preceeding directory structure. (an array of
	 * IModuleResources, e.g. index.html, WEB-INF/web.xml,
	 * ...) In this case, getLocation() should return the path to
	 * the root folder containing these resources.</p>
```

chanseokoh:
Wait... something's different. Looking into it.

chanseokoh:
Nope. It's working fine. I updated the code to use the APIs from @briandealwis's comment.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -71,12 +70,7 @@ private static void publish(IProject project, IPath destination, boolean explode
...
-    J2EEFlexProjDeployable deployable =
-        new J2EEFlexProjDeployable(project, ComponentCore.createComponent(project));
-
-    List<IModuleResource> resourcesToDeploy = Lists.newArrayList(deployable.members());
-    resourcesToDeploy.addAll(computeWebFragmentJarResources(deployable, progress));
-    IModuleResource[] resources = resourcesToDeploy.toArray(new IModuleResource[0]);
+    IModuleResource[] resources = flattenResources(project, progress);
 
     if (exploded) {
       publishHelper.publishFull(resources, destination, progress.newChild(100));
```
briandealwis: Was there a difference between `publishFull` and `publishSmart`?  Or did you just figure that since we're deploying to a new directory each time, there's little point with the `publishSmart`?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -85,28 +79,46 @@ private static void publish(IProject project, IPath destination, boolean explode
...
+      Collections.addAll(resources, delegate.members());
+
+      // now handle web fragment child modules, if they exist
+      for (IModule child : delegate.getChildModules()) {
+        ModuleDelegate childDelegate = (ModuleDelegate)
+            child.loadAdapter(ModuleDelegate.class, monitor);
+        IWebFragmentModule webFragmentModule = (IWebFragmentModule)
+            child.loadAdapter(IWebFragmentModule.class, monitor);
+        if (childDelegate == null || webFragmentModule == null || !webFragmentModule.isBinary()) {
+          continue;
```
briandealwis: I think we need to log these situations.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
chanseokoh: Good catch. There was a different. For example, `publishSmart()` was deleting existing files not known to itself, so if we put a web frag JAR into `WEB-INF/lib` first and later it tries to publish the main lib JARs into `WEB-INF` (i.e., if the web frag JAR comes first in the `IModuleResource[]` list), `publishSmart()` clears the web frag. I figured out we always create a new directory each time, so I guess there's no reason to use `publishSmart()`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -85,28 +79,46 @@ private static void publish(IProject project, IPath destination, boolean explode
...
+        }
+
+        // e.g., "WEB-INF/lib/spring-web-4.3.6.RELEASE.jar"
+        IPath zip = new Path(delegate.getPath(child));
```
briandealwis: It would be helpful to indicate that this is a destination, and not the actual zip.

chanseokoh:
Can't think of a way to have an integration test, since we can't actually deploy. I'll go for non-integration tests.

briandealwis:
You could import the project and then just run the WarPublisher.

chanseokoh:
Makes sense. I'll create a separate test project zip, since `springboot-appengine-standard.zip` seems to require `JavaSE-1.8` (`Assume.assumeTrue()` being used).

chanseokoh:
Well maybe I could still reuse `springboot-appengine-standard` if I ignore build errors.

chanseokoh:
I ended up creating a separate test project zip (`web-fragment-example.zip`), as `springboot-appengine-standard.zip` seemed excessive with too many JARs being copied.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -60,13 +74,57 @@ private static void publish(IProject project, IPath destination, boolean explode
...
+      Collections.addAll(resources, delegate.members());
+
+      // now handle web fragment child modules, if they exist
+      for (IModule child : delegate.getChildModules()) {
+        ModuleDelegate childDelegate = (ModuleDelegate)
+            child.loadAdapter(ModuleDelegate.class, monitor);
+        IWebFragmentModule webFragmentModule = (IWebFragmentModule)
+            child.loadAdapter(IWebFragmentModule.class, monitor);
+        if (childDelegate == null || webFragmentModule == null || !webFragmentModule.isBinary()) {
+          logger.log(Level.WARNING, "child modules other than web fragments are not supported:"
```
briandealwis: binary web-fragments

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/WebFragmentWarPublishTest.java
```diff
@@ -0,0 +1,79 @@
...
+  public void testPublishWar_webFragmentJarPublished() throws IOException, CoreException {
+    List<IProject> projects = ProjectUtils.importProjects(getClass(),
+        "test-projects/web-fragment-example.zip", false /* checkBuildErrors */, monitor);
+    assertEquals(1, projects.size());
+    project = projects.get(0);
+
+    IFile war = project.getFile("my-app.war");
+    WarPublisher.publishWar(project, war.getLocation(), monitor);
+
+    IFolder exploded = project.getFolder("exloded-war");
```
briandealwis: ex**p**loded (missing 'p')

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/WebFragmentWarPublishTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.After;
+import org.junit.Test;
+
+public class WebFragmentWarPublishTest {
+
+  private final IProgressMonitor monitor = new NullProgressMonitor();
+  private IProject project;
+
+  @After
+  public void tearDown() throws CoreException {
```
briandealwis: Should probably do the import with a `@BeforeClass` so as to amortize the maven resolution time.  Could even merge with the ImportMaven test with `@BeforeClass` to avoid further test time since the tests are non-destructive.

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/WebFragmentWarPublishTest.java
```diff
@@ -0,0 +1,79 @@
...
+    if (project != null) {
+      ProjectUtils.waitForProjects(project);
+      project.delete(true, null);
+    }
+  }
+
+  @Test
+  public void testPublishExploded_webFragmentJarPublished() throws IOException, CoreException {
+    List<IProject> projects = ProjectUtils.importProjects(getClass(),
+        "test-projects/web-fragment-example.zip", false /* checkBuildErrors */, monitor);
```
briandealwis: Could you change the `spring-web` version to _4.3.8.RELEASE_ to match the `spring-boot-appengine` test project?  Saves downloading two versions of the same artifact.

chanseokoh: Good idea!

plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisherTest.java
```diff
@@ -70,7 +70,6 @@ public void testWriteProjectToStageDir_cancelled() throws CoreException {
   public void testPublishExploded() throws CoreException {
     IProject project = projectCreator.getProject();
     IFolder exploded = project.getFolder("exloded-war");
```
briandealwis: This folder is missing a `p` too

chanseokoh: Gosh!

Issue #2248, opened 2017-08-08
importMavenProject() in ImportMavenAppEngineStandardProjectTest.runImport() timed out on Kokoro


chanseokoh:
This happened on Kokoro. It seems like it just took a lot of time to download a lot of artifacts when importing a project. (Sometimes the network is pretty slow or unstable on Kokoro.) Probably all we can do is to increase the timeout, unless we remove the test.

```
05:56:11.890 [Worker-7] INFO  o.e.m.c.i.e.AbstractTransferListenerAdapter - Downloaded https://repo.maven.apache.org/maven2/org/slf4j/jul-to-slf4j/1.7.25/jul-to-slf4j-1.7.25.pom
05:56:11.921 [Worker-7] INFO  o.e.m.c.i.e.AbstractTransferListenerAdapter - Downloading https://repo.maven.apache.org/maven2/org/slf4j/log4j-over-slf4j/1.7.25/log4j-over-slf4j-1.7.25.pom
05:56:12.531 [Worker-7] INFO  o.e.m.c.i.e.AbstractTransferListenerAdapter - Downloaded https://repo.maven.apache.org/maven2/org/slf4j/log4j-over-slf4j/1.7.25/log4j-over-slf4j-1.7.25.pom
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 34.797 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.integration.appengine.ImportMavenAppEngineStandardProjectTest
runImport(com.google.cloud.tools.eclipse.integration.appengine.ImportMavenAppEngineStandardProjectTest)  Time elapsed: 33.297 sec  <<< ERROR!
org.eclipse.swtbot.swt.finder.widgets.TimeoutException: Timeout after: 30000 ms.: Project does not exist! P/springboot-appengine-standard
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:522)
	at org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot.waitUntil(SWTWorkbenchBot.java:1)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:496)
	at org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot.waitUntil(SWTWorkbenchBot.java:1)
[...]

PR #2256, opened 2017-08-10
Use Maven Central GCS mirror on Kokoro

chanseokoh:
Let's see how much time this will cut on Kokoro.

### DIFF
--- a/kokoro/maven-settings.xml
+++ b/kokoro/maven-settings.xml
@@ -1,0 +1,10 @@
+<settings>
+  <mirrors>
+    <mirror>
+      <id>google-maven-central</id>
+      <name>GCS Maven Central mirror</name>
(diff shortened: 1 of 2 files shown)


chanseokoh:
That's disappointing. However, I wonder if Maven Central throttles down download speed through some network settings. In such a case, we might observe better throughput?

chanseokoh:
This doesn't seem to help: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2248#issuecomment-321655914

briandealwis:
> @patflynn: apparently this shouldn't be effective. I'm told Maven Central is already on GCP

Unless something changed recently, [it was on AWS S3](http://takari.io/2015/10/28/google-maven-central.html).  That said, `repo.maven.apache.org` and `repo1.maven.org` resolve to Fastly CDN buckets, which has direct connects with GCP.

PR #2271, opened 2017-08-14
distinguish missing and illegal input values in new dataflow project wizard

### LINKED ISSUE #1796: Start the Dataflow wizard with a prompt, not an error message.
Per https://wiki.eclipse.org/User_Interface_Guidelines#Language 5.3 this is wrong:

image

elharo:
fix #1796 @tgroh @chanseokoh


plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectValidationStatus.java
```diff
@@ -22,71 +22,92 @@
...
   /**
    * Returns if the dataflow project being validated can be created with the current arguments.
    */
   public boolean isValid() {
-    return valid;
+    return status == Status.VALID;
+  }
+  
+  /**
+   * Returns if the dataflow project being validated can be created with the current arguments.
```
chanseokoh: The Javadoc needs to be updated.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/wizard/NewDataflowProjectWizard.java
```diff
@@ -52,9 +52,9 @@ public boolean performFinish() {
     }
     try {
       getContainer().run(true, true, creator);
-    } catch (InvocationTargetException | InterruptedException e) {
+    } catch (InvocationTargetException | InterruptedException ex) {
       // TODO: handle
-      DataflowUiPlugin.logError(e, "Error encountered when trying to create project");
+      DataflowUiPlugin.logError(ex, "Error encountered when trying to create project");
```
chanseokoh: I was going to say this:
https://github.com/GoogleCloudPlatform/google-cloud-intellij/pull/1579#discussion_r130979346

I didn't really like the rule being applied everywhere blindly. How about us allowing a one-letter var for exceptions?

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectCreator.java
```diff
@@ -347,8 +347,10 @@ private DataflowProjectValidationStatus validateMavenGroupId() {
   }
 
   private DataflowProjectValidationStatus validatePackage() {
-    if (Strings.isNullOrEmpty(packageString)
-        || !JAVA_PACKAGE_REGEX.matcher(packageString).matches()) {
+    if (Strings.isNullOrEmpty(packageString)) {
+      return DataflowProjectValidationStatus.OK;
```
chanseokoh: Previously, if it's null or empty, it returned `ILLEGAL_PACKAGE`, which is an error. Now you are allowing it. Can you verify there will be no issue when creating a project without a package name?

But even if is OK, I realized a little technical problem. If you ever enter anything into the package field (which will always happen the field is auto-filled when you enter a Maven group ID), you cannot make the field empty: What happens is that, if you delete the auto-filled value in the package field, the value is automatically restored as soon as the focus is lost from the field.

Maybe we should just return `.MISSING`.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectValidationStatus.java
elharo: done

PR #2280, opened 2017-08-15
test for NewDataflowProjectWizard

elharo:
@tgroh


plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/MavenUtilsTest.java
```diff
@@ -43,5 +43,10 @@ public void testMavenNature_nonMavenProject() {
 
     Assert.assertFalse(MavenUtils.hasMavenNature(project));
   }
+  
+  @Test
+  public void testCreateRepository() throws CoreException {
+    MavenUtils.createRepository("foo", "http://www.example.com");
```
chanseokoh: It's not clear what it's testing. Change the test method name?

elharo: moved form this PR

PR #2293, opened 2017-08-17
Disable Maven Aether from resuming file downloads in tests

briandealwis:
Workaround repository corruption seen in #2284


chanseokoh:
A lot of test failures seem to be due to connection problems to http://dl-ssl.google.com/android/eclipse/.

PR #2295, opened 2017-08-18
upgrade sample HTML to XHTML 5

elharo:
@briandealwis


elharo:
same as any other application. When I started working on an app engine hackathon project, I almost immediately found myself using features that are only available in HTML 5.

PR #2299, opened 2017-08-18
begin moving to ArtifactRetriever

elharo:


plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManagerTest.java
```diff
@@ -248,22 +260,22 @@ public void hasPinnedDependencyTrackedDependency() {
...
 
-  private Dependency dataflowDependency() {
-    Dependency dep = new Dependency();
-    dep.setArtifactId("google-cloud-dataflow-java-sdk-all");
-    dep.setGroupId("com.google.cloud.dataflow");
-    return dep;
+  private static Dependency dataflowDependency() {
+    Dependency dependency = new Dependency();
+    dependency.setArtifactId("google-cloud-dataflow-java-sdk-all");
+    dependency.setGroupId("com.google.cloud.dataflow");
```
briandealwis: Could these be `DataflowMavenCoordinates.GROUP_ID` and `.ARTIFACT_ID`?

elharo: Yes, though I prefer to use string literals in tests since it gives an extra layer of protection against a mistyped constant initializer.

PR #2307, opened 2017-08-19
remove some log messages

elharo:
@briandealwis


elharo:
They got in my way when I was looking at other things. I've never bought into the "log everything because some day you might need it" philosophy. I find that when I do need to go to the logs, I'm usually wading through page after page of irrelevant logs I don't care about. I prefer more focused logs that only record legitimate, unexpected error conditions.

PR #2315, opened 2017-08-22
Fix up ClientApisLibrariesSelectorGroupTest for `cloudtranslation`

briandealwis:


briandealwis:
Closing since same fix in #2314

PR #2317, opened 2017-08-23
upgrade to endpoints v2

### LINKED ISSUE #2310: Endpoints v2 support
@lesv @patflynn "Cloud Endpoints Frameworks v1 has been deprecated and will be shut down on August 2, 2018. We recommend that you use the latest version of this feature, which is renamed to Cloud Endpoints Frameworks for App Engine. This new version supports App Engine standard environment..."

What should we do about Endpoints v2? Should we add it to the list of available libraries or replace the current endpoints v1?

It appears to run on App Engine Standard Java 7.

image

elharo:
fix #2310


elharo:
The official word is "Cloud Endpoints Frameworks v1 has been deprecated and will be shut down on August 2, 2018." so I'd prefer not to create new projects with Endpoints v1.

elharo:
There aren't a lot of dependencies because we don't yet do much with endpoints aside from add it to the classpath.

PR #2318, opened 2017-08-23
remove DataflowArtifactRetriever

elharo:
@tgroh @briandealwis


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -96,6 +96,7 @@ URL getMetadataUrl(String groupId, String artifactId) {
                 }
               });
 
+  // todo consider a factory to create or build retrievers for each repo
```
briandealwis: Or pulling the URL from a preference?

elharo: I think the reason we added the constructor was to allow multiple repos, though we don't seem to be using that yet.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectCreator.java
```diff
@@ -109,6 +112,9 @@ public String getLabel() {
       return label;
     }
 
+    /**
+     * @return the artifact ID of the archetype
```
elharo: I've filed a separate issue to rename this for clarity.

PR #2324, opened 2017-08-23
Fix enablement of Dataflow Run Options fields

### LINKED ISSUE #2265: Dataflow Run Options Create Bucket enabled before account and project are selected
Pressing the button in this case does nothing:

image

briandealwis:
- centralize validation computation into single method `validate()`
- rename VERIFY_LOCATION_DELAY_MS to NEXT_KEY_DELAY_MS
- fix FetchStagingLocationsJob and VerifyStagingLocationJob to properly cancel and restart
- ensure MessageTarget.set* are propagated to the Launch Config dialog
- expand abbreviation in UI
- get rid of ListenableFutureProxy and just use ListenableFuture

Fixes #2265


elharo:
should we mark this as fixing #2265?

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManager.java
```diff
@@ -126,25 +126,23 @@ public boolean hasTrackedDataflowDependency(IProject project) {
...
-              e);
+    Model model = getModelFromProject(project);
+    if (model != null) {
+      Dependency dependency = getDataflowDependencyFromModel(model);
+      if (dependency != null) {
+        String version = dependency.getVersion();
+        if (!Strings.isNullOrEmpty(version)) {
+          try {
+            return VersionRange.createFromVersionSpec(version);
+          } catch (InvalidVersionSpecificationException e) {
```
elharo: e --> ex

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManager.java
```diff
@@ -126,25 +126,23 @@ public boolean hasTrackedDataflowDependency(IProject project) {
...
+            String message =
+                String.format("Could not create version range from existing version %s", version);
```
elharo: seems like overkill for format when a trivial concatenation would work equally well

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/FetchStagingLocationsJob.java
```diff
@@ -32,34 +32,39 @@
 public class FetchStagingLocationsJob extends Job {
   private final GcsDataflowProjectClient gcsClient;
 
+  private final String accountEmail;
   private final String cloudProject;
   private final SettableFuture<SortedSet<String>> stagingLocations;
 
-  private FetchStagingLocationsJob(GcsDataflowProjectClient gcsClient, String cloudProject) {
+  public FetchStagingLocationsJob(GcsDataflowProjectClient gcsClient, String accountEmail,
+      String cloudProject) {
```
elharo: Is this a project ID or name or number? probably the ID

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -233,93 +330,105 @@ public void setEnabled(boolean enabled) {
...
+    return fetchStagingLocationsJob == null ? null : fetchStagingLocationsJob.getStagingLocations();
+  }
+
   /**
    * Fetch the staging locations from GCS in a background task and update the Staging Locations
    * combo.
    */
-  private void updateStagingLocations(String project) {
+  @VisibleForTesting
+  synchronized void updateStagingLocations(String project, long scheduleDelay) {
```
chanseokoh: I think this method is called exclusively by the UI thread. My preference is to not use `synchronized` if it is not needed. It makes people wonder the keyword is there for a good reason. People don't use `synchronized` for SWT listener methods. I believe the main reason for the single UI thread architecture is to avoid complex concurrency management such as using `synchronized`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -233,93 +330,105 @@ public void setEnabled(boolean enabled) {
...
 
-    IStatus status = bucketNameStatus();
-    if (!status.isOK()) {
-      messageTarget.setError(status.getMessage());
-      return;
+  /**
+   * Ensure the staging location specified in the input combo is valid.
+   */
+  @VisibleForTesting
+  synchronized void startStagingLocationCheck(long schedulingDelay) {
```
chanseokoh: This method is called only in an SWT listener and in the UI thread. I don't think `synchronized` is necessary. Please see my previous comment.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -233,93 +330,105 @@ public void setEnabled(boolean enabled) {
     stagingLocationInput.setEnabled(enabled);
   }
 
+
+  /**
+   * Return the fetched staging locations result future, or null if no job is yet scheduled.
+   */
+  private synchronized Future<SortedSet<String>> getStagingLocationsResult() {
```
chanseokoh: This method is called only in the UI thread, so `synchronized` is not necessary. Please see my comment below.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -233,93 +330,105 @@ public void setEnabled(boolean enabled) {
...
-
-    final String bucketNamePart = extractBucketNamePart();
-    if (bucketNamePart.isEmpty()) {
-      // If the bucket name is empty, we don't have anything to verify; and we don't have any
-      // interesting messaging.
-      setPageComplete(true);
-      return;
-    }
+  @VisibleForTesting
+  synchronized Future<VerifyStagingLocationResult> getVerifyStagingLocationResult() {
```
chanseokoh: This method is called only in the UI thread, so `synchronized` is not necessary. Please see my previous comment.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -153,35 +151,134 @@ public RunOptionsDefaultsComponent(Composite target, int columns, MessageTarget
...
+    Preconditions.checkNotNull(future, "verification job should be underway");
+    if (!future.isDone()) {
+      messageTarget.setInfo("Verifying staging location...");
+      createButton.setEnabled(false);
+      return;
+    }
+
+    try {
+      VerifyStagingLocationResult result = future.get();
+      if (!result.email.equals(accountSelector.getSelectedEmail())
```
chanseokoh: Previously, `VerifyStagingLocationResult.email` was needed to get the email given to the job for this check at this point. Now that the field `verifyStagingLocationJob` has the `getEmail()` field, I wonder if we want to remove `.email` from the result type. (The result is from the current `verifyStagingLocationJob`.) However, it may still be harmless to keep `.email`.

The same argument applies to the `.stagingLocation` field.

chanseokoh:
There may be some timing issue leading to a flaky test. I feel like there is an additional delay in updating the _Create_ button?

```
testEnablement_nonExistentStagingLocation(com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest)  Time elapsed: 0.248 sec  <<< FAILURE!
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest.testEnablement_nonExistentStagingLocation(RunOptionsDefaultsComponentTest.java:240)
```

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
chanseokoh: > The future needs the email and staging location to check for staleness (vs the current contents of the page).

Exactly. I was saying `verifyStagingLocationJob.getEmail()` is same as `result.email` at this point (`result` is obtained from `verifyStagingLocationJob.getVerifyResult().get()`). But as I said, I don't mind much about having `.email` anyway.

chanseokoh:
Just for future reference,

> This method is called only in the UI thread, so synchronized is not necessary.

I realized my comment can mislead people: the fact alone that this method is called only in the UI thread doesn't make `synchronized` unnecessary. (For example, it may be required if there were a non-UI thread calling another `synchronized` method.) It is the fact that _all_ `synchronized` class methods (or code sections) run in the UI thread.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -153,35 +153,136 @@ public RunOptionsDefaultsComponent(Composite target, int columns, MessageTarget
...
+        startStagingLocationCheck(NEXT_KEY_DELAY_MS);
       }
     });
     createButton.addSelectionListener(new CreateStagingLocationListener());
 
-    stagingLocationInput.addModifyListener(new EnableCreateButton());
+    stagingLocationInput.addModifyListener(new ModifyListener() {
+      @Override
+      public void modifyText(ModifyEvent e) {
+        validate();
```
chanseokoh: nit: what if we move this `validate()` into the `addModifyListener()` code a little bit above?

briandealwis:
PTAL @elharo — there have been some minor changes since you reviewed.

PR #2326, opened 2017-08-23
Handle pipeline options computation as UI-blocking work

### LINKED ISSUE #2289: Concurrency flaw in Dataflow PipelineArgumentsTab UI class
While I was working on #2232, I noticed that there is a concurrency flaw, which cannot easily be fixed unless we make a fundamental change to the overall workings of `PipelineArgumentsTab`.

The class can launch two types of async jobs: [hierarchy-update job from `updateHierarchy()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.3.0B/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java#L345) and [UI-update job from `updatePipelineOptions()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.3.0B/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java#L395).

- The hierarchy-update job calls `getPipelineOptionsHierarchy(majorVersion, progress)` asynchronously to get a new hierarchy, probably because it may take a long time.
   ```java
    Job job = new Job("Update Hierarchy") {
      @Override
      public IStatus run(IProgressMonitor progress) {
        hierarchy = getPipelineOptionsHierarchy(majorVersion, progress);
        return Status.OK_STATUS;
      }
    };
   ```
   So, `getPipelineOptionsHierarchy()` will run in a non-UI thread. However, the fundamental flaw is that the outcome of the method depends on the states of `PipelineArgumentsTab` (e.g., the `launchConfiguration` field); the states can be modified concurrently in another thread (in fact, the UI thread). It is not safe to call `getPipelineOptionsHierarchy()` asynchronously.

- The same kind of problem happens with the UI-update job too. It calls `launchConfiguration.getOptionsHierarchy(hierarchy)` in a non-UI thread, probably because it may take a long time.
   ```java
    Job job = new Job("Update Pipeline Options Form") {
      @Override
        optionsHierarchyFuture.set(launchConfiguration.getOptionsHierarchy(hierarchy));
        return Status.OK_STATUS;
      }
    };
   ```
   However, `launchConfiguration` is a stateful object, whose states can be modified concurrently in the UI thread. It is not safe to access `launchConfiguration` asynchronously.

briandealwis:
Execute pipeline computation using the modal context (`IRunnableWithContext` obtained from the `ILaunchConfigurationDialog`): this launches the computation on a background thread, but spins the UI thread (with a progress bar, if it takes longer than some threshold).  The actual pipeline options form updating is handled done with a busy-indicator.

Fixes #2289

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
@@ -35,6 +35,7 @@ import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.debug.ui.ILaunchConfigurationDialog;
 import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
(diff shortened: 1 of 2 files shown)


chanseokoh:
No concurrency after this, which I think is super nice. That said, I think retaining `Future` with the future listener mechanism becomes an over-engineering and over-complication. I verified that the code below works.

```java
        @Override
          final Map<PipelineOptionsType, Set<PipelineOptionsProperty>> types =
              launchConfiguration.getOptionsHierarchy(hierarchy);

          display.syncExec(new Runnable() {
            @Override
            public void run() {
              BusyIndicator.showWhile(display, new Runnable() {
                @Override
                public void run() {
                  pipelineOptionsForm.updateForm(launchConfiguration, types);
                  updateLaunchConfigurationDialog();
                }
              });
            }
          });
        }
      });
```

PR #2328, opened 2017-08-25
Report error from server on bucket-creation failure

### LINKED ISSUE #2313: Report staging-location creation errors
In testing #2265, I specified a bucket named `blah.blah.blah`.  No validation error is reported in the UI, but the actual bucket creation fails (but the actual reason is never surfaced).

Before attempt to create bucket:
screen shot 2017-08-21 at 10 11 42 pm

After attempt to create bucket:
screen shot 2017-08-21 at 10 16 44 pm

Looking with the debugger, the create-bucket fails with:
```
400 Bad Request
{
  "code" : 400,
  "errors" : [ {
    "domain" : "global",
    "message" : "Bucket name contains \".\", but is not under a currently recognized top-level domain.",
    "reason" : "invalid"
  } ],
  "message" : "Bucket name contains \".\", but is not under a currently recognized top-level domain."
}
```

briandealwis:
Fixes #2313 

Use Guava's `InternetDomainName#isValid()` to perform some basic validation of bucket names with ".".

Use a JFace `ControlDecoration` to report the server-side error in a hover.  
screen shot 2017-08-25 at 1 00 42 pm

### Avoiding hovers

I did try including the error text in the pref-page's message area, but the message becomes too long and the resulting text is collapsed and hard to read.

tmp-37694d8rr4v8qixxw


plugins/com.google.cloud.tools.eclipse.ui.util/META-INF/MANIFEST.MF
```diff
@@ -16,6 +16,7 @@ Import-Package: com.google.cloud.tools.eclipse.ui.util,
  com.google.cloud.tools.eclipse.util.status,
  com.google.common.annotations;version="[20.0.0,21.0.0)",
  com.google.common.base;version="[20.0.0,21.0.0)",
+ com.google.common.net;version="15.0.0",
```
elharo: why 15.0.0? If this is part of Guava it should be 20.0.0, right?

plugins/com.google.cloud.tools.eclipse.ui.util/src/com/google/cloud/tools/eclipse/ui/util/databinding/BucketNameValidator.java
```diff
@@ -26,14 +27,23 @@
...
+ * contain up to 222 characters, but each dot-separated component can be no longer than 63
+ * characters.</li>
+ * <li>Cloud Storage considers bucket names that contain dots to be domain names, and as such these
+ * bucket names must:
+ * <ul>
+ * <li>Be syntactically valid DNS names (for example, <code>bucket..example.com</code> is not valid
+ * because it contains two dots in a row)</li>
+ * <li>Has a <em>public suffix</em> (although GCS states it should ends with a TLD, due to the
+ * domain verification requirements</li>
+ * </ul>
```
elharo: </ul></li>

plugins/com.google.cloud.tools.eclipse.ui.util/META-INF/MANIFEST.MF
briandealwis: Oops!  Missed that.

PR #2333, opened 2017-08-29
share a single retriever for Maven Central

elharo:
@briandealwis since we want to cache shared artifacts


elharo:
If we need it we'll do that later. ATM we're not even using the constructor that takes a repo URL as an argument.

elharo:
Ping @briandealwis

PR #2260, opened 2017-08-11
Repeatedly wait for problem markers in testValidate_badXml()

briandealwis:
Towards fixing #2076.

Wait for `ConvertJob`, fixes #2087

PR #2337, opened 2017-08-29
Repeatedly wait for markers in XsltSourceQuickFixTest#testApply()

briandealwis:
Applies same approach in #2260, to work around issues seen in #2076.


briandealwis:
Hold off — I'll adjust these logging patches to (1) only log when unexpected and (2) call out that it's a warning.

briandealwis:
PTAL @elharo

PR #2341, opened 2017-08-30
resolve dependencies from Maven Central rather than hardcoded XML

elharo:


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
```diff
@@ -77,10 +78,12 @@
         try {
           List<ArtifactResult> artifacts =
               system.resolveDependencies(session, request).getArtifactResults();
+          List<Artifact> dependencies = new ArrayList<>();
           for (ArtifactResult result : artifacts) {
             Artifact dependency = result.getArtifact();
-            dependencies.add(dependency.getGroupId() + ":" + dependency.getArtifactId() + ":"
-                + dependency.getVersion());
+            dependencies.add(dependency);
+            /* dependencies.add(dependency.getGroupId() + ":" + dependency.getArtifactId() + ":"
```
briandealwis: Don't need this line, right?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
```diff
@@ -215,8 +215,8 @@ private SourceAttacherJob createSourceAttacherJob(IJavaProject javaProject,
     return new SourceAttacherJob(javaProject, containerPath, libraryPath, new Callable<IPath>() {
       @Override
       public IPath call() throws Exception {
-          return repositoryService.resolveSourceArtifact(libraryFile, artifact.getVersion(),
-                                                         monitor);
+        return repositoryService.resolveSourceArtifact(libraryFile, artifact.getVersion(),
```
briandealwis: We're going to need something like this for the javadoc too.  I've never looked into how M2E does its lazy source attachment, but it might be worth a look.

elharo: It looks like we're already doing that. Is there any chance Eclipse is reading Javadoc form the sources directly rather than a Javadoc bundle?

FYI, this isn't really ready for review. I just pushed it up to see what other tests it caught. I'm probably going to throw this away and start over.

briandealwis: Oops, sorry, I meant lazy javadoc if we're downloading it.  But yes, Eclipse will parse javadoc directly if source is there, so that might be enough.  Not all artifacts in MC have source though (that may be just older ones).

PR #2343, opened 2017-08-31
Equals methods

elharo:
@briandealwis Looks like we're going to need this to avoid duping transitive dependencies. 

These methods quite deliberately do not consider all fields of their classes.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
@@ -20,7 +20,8 @@ import com.google.common.base.Preconditions;
 import java.text.MessageFormat;
 
 /**
- * Describes a Maven artifact.
+ * Describes a Maven artifact. MavenCoordinates objects are considered to be equal
+ * if and only if the group ID and artifact IDs are the same.
  */
 public final class MavenCoordinates {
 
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Describes a Maven artifact.
```
briandealwis: This needs a comment calling out that `equals()` is different.

elharo: Isn't the comment on the equals method sufficient?

elharo:
This may change, but ATM the use case we have is group ID and artifact ID. We're going to need to think harder about version conflicts; but I suspect we do not want two different versions of the same artifact to show up in our container. I want HashSets to not contain two different versions and the contains method to tell me if there's already a different version of the same artifact we need to handle somehow.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
briandealwis: Nobody looks at javadoc on `equals()`.

briandealwis:
But resolving conflicts is more than just overwrite-the-last or don't-add-if-present.  Maven apparently uses the [_nearest-wins_](http://techidiocy.com/maven-dependency-version-conflict-problem-and-resolution/) approach; Eclipse converts the dependencies into a SAT problem to come up with a working recipe.  Use a separate class that encodes the information required.

elharo:
This doesn't so much resolve conflicts as alert us to them. With a version aware equals method we don't even see the conflicts.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
elharo: OK, done.

PR #2347, opened 2017-09-01
Load translate dependencies transitively

elharo:
@briandealwis For early consideration. The big question is where and when do we resolve transitive dependencies? I'm thinking this should be as late as possible, basically when we add a library to the classpath. There's no reason to hit Maven Central and download all possibly referenced jars when most projects won't even use them.


plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -189,7 +189,14 @@ WAR file of the project, under &lt;tt&gt;WEB-INF/lib&lt;/tt&gt; if the parent &l
                </xs:documentation>
             </xs:annotation>
          </xs:attribute>
-
+         <xs:attribute name="loadDependencies" type="xs:boolean" default="false">
+            <xs:annotation>
+               <xs:documentation>
+                  If the loadDependencies attribute is present and has the value true, 
+                  then all transitive dependencies of the Maven artifact should also be included.
```
briandealwis: s/included/loaded/ to be consistent?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
```diff
@@ -95,8 +94,8 @@ public String getArtifactId() {
 
   @Override
   public String toString() {
-    return MessageFormat.format("MavenCoordinates [repository={0}, {1}:{2}:{3}:{4}:{5}]",
-                                repository, groupId, artifactId, type, classifier, version);
+    return String.format("MavenCoordinates [repository={0}, {1}:{2}:{3}:{4}:{5}]", repository,
```
briandealwis: `String.format()` isn't equivalent to `MessageFormat.format()`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -318,10 +318,10 @@
           name="%cloudtranslation.name"
           tooltip="%cloudtranslation.tooltip"
           javaVersion="1.8"
-          siteUri="https://cloud.google.com/translate/docs/reference/libraries" >
-      <libraryDependency id="googlecloudcore" />
+          siteUri="https://cloud.google.com/translate/docs/reference/libraries">
       <libraryFile
-            javadocUri="https://googlecloudplatform.github.io/google-cloud-java/0.21.0/apidocs/index.html?com/google/cloud/translate/package-summary.html">    
+          loadDependencies="true"
+          javadocUri="https://googlecloudplatform.github.io/google-cloud-java/0.21.0/apidocs/index.html?com/google/cloud/translate/package-summary.html">
```
briandealwis: Since we use this to feed into Eclipse JDT for javadoc hovers, this should be just `https://googlecloudplatform.github.io/google-cloud-java/0.21.0/apidocs`. This is only used for specific types and methods, rather than a general entry point to documentation.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -99,29 +105,64 @@ static Library create(IConfigurationElement configurationElement) throws Library
...
-        libraryFile.setFilters(getFilters(libraryFileElement.getChildren()));
-        // todo do we really want these next two to be required?
-        libraryFile.setSourceUri(
-            getUri(libraryFileElement.getAttribute(ATTRIBUTE_NAME_SOURCE_URI)));
-        libraryFile.setJavadocUri(
-            getUri(libraryFileElement.getAttribute(ATTRIBUTE_NAME_JAVADOC_URI)));
-        String exportString = libraryFileElement.getAttribute(ATTRIBUTE_NAME_EXPORT);
-        if (exportString != null) {
-          libraryFile.setExport(Boolean.parseBoolean(exportString));
+        String loadDependencies = libraryFileElement.getAttribute("loadDependencies");
```
briandealwis: `loadDependencies` should be a constant too

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -99,29 +105,64 @@ static Library create(IConfigurationElement configurationElement) throws Library
...
+        if ("true".equalsIgnoreCase(loadDependencies)) {
+          try {
+            Collection<LibraryFile> files = loadTransitiveDependencies(mavenCoordinates);
+            libraryFiles.addAll(files);
+          } catch (CoreException ex) {
+            throw new LibraryFactoryException(
+                "Problem loading transitive dependencies for " + mavenCoordinates, ex);
```
briandealwis: I wonder if this should be externalized?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -99,29 +105,64 @@ static Library create(IConfigurationElement configurationElement) throws Library
...
             libraryFileElement.getChildren(ELEMENT_NAME_MAVEN_COORDINATES));
-        LibraryFile libraryFile = new LibraryFile(mavenCoordinates);
-        libraryFile.setFilters(getFilters(libraryFileElement.getChildren()));
-        // todo do we really want these next two to be required?
-        libraryFile.setSourceUri(
-            getUri(libraryFileElement.getAttribute(ATTRIBUTE_NAME_SOURCE_URI)));
-        libraryFile.setJavadocUri(
-            getUri(libraryFileElement.getAttribute(ATTRIBUTE_NAME_JAVADOC_URI)));
-        String exportString = libraryFileElement.getAttribute(ATTRIBUTE_NAME_EXPORT);
-        if (exportString != null) {
```
briandealwis: Worth considering if we should have an `exportDependencies` flag too?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
elharo: Do we use it for that though? I looked and didn't find the path where that happened. I'm not 100% sure we're using this at all.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: Maybe. I'm tempted to dename all these constants. Do you find it easier to read the constant field or the literal?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: Likely once it lands in its final location. I wonder, do we need some UI for surfacing this to the end user?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: In general, what should our default be? To export dependencies or not to export them?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
elharo: reverted

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -132,6 +132,16 @@ limitations under the License.
                </xs:documentation>
             </xs:annotation>
          </xs:attribute>
+         <xs:attribute name="includesAllDependencies" type="xs:boolean" default="true">
```
briandealwis: An idle thought, but would it be worth making this a `dependencies="ignore"` and `"include"`?  Are there other possibilities for handling dependencies like _only-immediate_ dependencies?  Or including _optional_ dependencies too?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
briandealwis: Hmm, Maven projects don't export their _Maven dependencies_:

screen shot 2017-09-01 at 12 41 47 pm

And `WEB-INF/lib` libraries are not either for WTP _Dynamic Web Projects_:

screen shot 2017-09-01 at 12 40 37 pm

So I'd say no exporting by default?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
```diff
@@ -16,9 +16,10 @@
 
 package com.google.cloud.tools.eclipse.appengine.libraries.model;
 
-import com.google.common.base.Preconditions;
 import java.text.MessageFormat;
```
briandealwis: Looks like your import rules are out of sync: Google style now just sorts them as a block with no special priority on java.* or com.google.*

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
elharo: Wouldn't surprise me. The Java classes will diverge a little further from the XML but that's fine. Let me play with that.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: Makes sense, though it is a change to the current behavior. That's not going to change which jars we deploy to the server, right?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
elharo: Yes, that's true. I don't think we've updated the external Google stylesheet for Eclipse but let me check.

Yep. Still uses the old import ordering: https://github.com/google/styleguide/issues/270

Issue #2352, opened 2017-09-05
Support launching devappserver with user-supplied credential (including CT4E login credentials, service accounts, etc) and project ID used for Veneer libraries


briandealwis:
Launching a local app in the devappserver that uses Google services (e.g., BigQuery) is difficult as we don't provide means to configure the application default credential nor the project.  There has been some discussion about [the ADC](https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/439#issuecomment-320137458). ~~#429 will at least allow~~ (UPDATED by @chanseokoh: we can do this now) setting environment variables (like `GOOGLE_APPLICATION_CREDENTIALS` or `GOOGLE_CLOUD_PROJECT`), but it would be better to have some decent UI support, especially for selecting a user credential.  Perhaps in the server definitions?

Steps to repeat:

  1. Create a new _Google App Engine Standard_ project with Java 7 with Maven support; in my testing, I used groupId=`com.example.bq`, artifactId=`bq`, and javaPackage=`com.example.bq`
  2. UPDATED by @chanseokoh: _Build Path_ > _Add Libraries..._ > _Google API Libraries_
~~Add the following to the `pom.xml`'s `<dependencies>` (as per the [BigQuery client library installation instructions](https://cloud.google.com/bigquery/docs/reference/libraries#installing_the_client_library);~~
UPDATED by @chanseokoh: App Engine API will be provided for App Engine projects
~~note that we also have to bring in the `appengine-api-1.0-sdk`, discussed further below:~~
  3. Add the following sample code to the `HelloAppEngine#doGet()`, cribbed [from the BigQuery _Using the client library_ code](https://cloud.google.com/bigquery/docs/reference/libraries#using_the_client_library):
```java
    // Instantiates a client
    BigQuery bigquery = BigQueryOptions.getDefaultInstance().getService();

    // Prepares a new dataset
    String datasetName = "my_new_dataset";
    Dataset dataset = null;
    DatasetInfo datasetInfo = DatasetInfo.newBuilder(datasetName).build();

    // Creates the dataset
    dataset = bigquery.create(datasetInfo);

    response.setContentType("text/plain");
    response.setCharacterEncoding("UTF-8");
    response.getWriter().printf("Dataset %s created.\n", dataset.getDatasetId().getDataset());
```

Launching this app will fail for two reasons:
   ~~1. Cannot find a credential for access~~
   2. Cannot find a Project ID

[...]

PR #2354, opened 2017-09-05
Add environment tab to run configuration (again) and wire to CloudSdk

### LINKED ISSUE #429: Allow easy way to provide environment variables for Cloud emulators

briandealwis:
Provides a stop-gap measure for specifying Google services.

Fixes #429 and allows specifying `GOOGLE_APPLICATION_CREDENTIALS` and `GOOGLE_CLOUD_PROJECT` as per #2352.


plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegateTest.java
```diff
@@ -294,18 +301,48 @@ public void testGenerateRunConfiguration_withProgramArgs() throws CoreException
...
+    assertEquals("bar", parsedEnvironment.get("foo"));
+    assertEquals(System.getenv("PATH"), parsedEnvironment.get("baz"));
+    verify(launchConfiguration).getAttribute(eq(ILaunchManager.ATTR_ENVIRONMENT_VARIABLES),
+        anyMapOf(String.class, String.class));
+    verify(launchConfiguration, atLeastOnce())
+        .getAttribute(eq(ILaunchManager.ATTR_APPEND_ENVIRONMENT_VARIABLES),
+        anyBoolean());
+  }
+
+  @Test(expected = CoreException.class)
```
chanseokoh: Nit: `expected` is discouraged per internal doc. Use try-catch with `fail()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -299,6 +299,22 @@ DefaultRunConfiguration generateServerRunConfiguration(ILaunchConfiguration conf
...
 
+    boolean environmentAppend =
+        configuration.getAttribute(ILaunchManager.ATTR_APPEND_ENVIRONMENT_VARIABLES, true);
+    if (!environmentAppend) {
+      throw new CoreException(StatusUtil.error(this, "'Replace environment' not yet supported"));
+    }
+    // getEnvironment() applies var substitutions, but also then joins the key-value pairs
+    String[] environment = getEnvironment(configuration);
+    if (environment != null) {
+      Map<String, String> asMap = new HashMap<String, String>();
```
chanseokoh: Nit: `new HashMap<>();`

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -299,6 +299,22 @@ DefaultRunConfiguration generateServerRunConfiguration(ILaunchConfiguration conf
       devServerRunConfiguration.setAdditionalArguments(programArguments);
     }
 
+    boolean environmentAppend =
+        configuration.getAttribute(ILaunchManager.ATTR_APPEND_ENVIRONMENT_VARIABLES, true);
+    if (!environmentAppend) {
+      throw new CoreException(StatusUtil.error(this, "'Replace environment' not yet supported"));
```
chanseokoh: Can the error message more specific to point out that the _Replace environment_ is a checkbox in the _Environment_ launch tab? Users will hit the error far later when actually executing the dev appserver, so users can get easily lost about what this means.

Ideally, the _Replace environment_ checkbox should be disabled, but I guess this is difficult to implement?

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegateTest.java
```diff
@@ -294,18 +301,48 @@ public void testGenerateRunConfiguration_withProgramArgs() throws CoreException
...
+    DefaultRunConfiguration config = new LocalAppEngineServerLaunchConfigurationDelegate()
+        .generateServerRunConfiguration(launchConfiguration, server, ILaunchManager.RUN_MODE);
+
+    Map<String, String> parsedEnvironment = config.getEnvironment();
+    assertNotNull(parsedEnvironment);
+    assertEquals("bar", parsedEnvironment.get("foo"));
+    assertEquals(System.getenv("PATH"), parsedEnvironment.get("baz"));
+    verify(launchConfiguration).getAttribute(eq(ILaunchManager.ATTR_ENVIRONMENT_VARIABLES),
+        anyMapOf(String.class, String.class));
+    verify(launchConfiguration, atLeastOnce())
```
chanseokoh: I don't see much value in these `verify` checks. We already check the values of `parsedEnvironment`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
briandealwis: Unlike some of the other tabs, it is possible but it's brittle (fragile superclass and all).  I forgot to open an issue with appengine-plugins-core to add support for append/replace (https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/446).

plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegateTest.java
briandealwis: We've been using them pretty systematically throughout.

briandealwis:
`AbstractLaunchConfigurationDelegate#getEnvironment()` checks and does the append-to-existing environment as part of its handling, which appengine-plugins-core does too.  I think we're better off deferring to appengine-plugins-core.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
chanseokoh: > Can the error message be more specific to point out that the _Replace environment_ is a checkbox in the _Environment_ launch tab?

Can we still do this? I think the current error message is too distance from the actual error source.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
```diff
@@ -299,6 +302,25 @@ DefaultRunConfiguration generateServerRunConfiguration(ILaunchConfiguration conf
...
+    }
+    // Could use getEnvironment(), but it does `append` processing and joins as key-value pairs
+    Map<String, String> environment = configuration.getAttribute(
+        ILaunchManager.ATTR_ENVIRONMENT_VARIABLES, Collections.<String, String>emptyMap());
+    if (environment != null) {
```
chanseokoh: Probably we don't need this null check, as we return the empty set when it's undefined.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerLaunchConfigurationDelegate.java
briandealwis: oops, will do

chanseokoh:
LGTM. BTW, in my Eclipse testing, the Append/Replace environment options are disabled. They are enabled in your environment?

selection_003

briandealwis:
It's disabled until a variable has been added:

screen shot 2017-09-11 at 3 18 35 pm

screen shot 2017-09-11 at 3 17 58 pm

briandealwis:
@chanseokoh can you tick the approve box when you have a moment?

PR #2362, opened 2017-09-06
Mini Project Selector

### LINKED ISSUE #2217: the project field in the dataflow wizard should use the project selector used for app engine
image

briandealwis:
Creates a combo-based project selector, used in Dataflow's Runtime Defaults options.
Fixes #2217

This is functionally complete, though in using it I realize that it needs some way to indicate that a network check is in progress as populating the projects list is noticeable.

Right now the mini selector must created explicitly by name. I wondered about changing `ProjectSelector` to be a facade / factory with `create{Small,Large}Selector()` to create a particular sized selector that would respond in a similar way.  Not sure if this makes sense though since the MiniSelector can only ever select a single project.

screen shot 2017-09-06 at 4 05 20 pm


plugins/com.google.cloud.tools.eclipse.ui.util.test/src/com/google/cloud/tools/eclipse/ui/util/DisplayExecutorTest.java
```diff
@@ -14,10 +14,11 @@
  * the License.
  */
 
-package com.google.cloud.tools.eclipse.dataflow.ui.util;
+package com.google.cloud.tools.eclipse.ui.util;
 
 import static org.mockito.Mockito.verify;
 
+import com.google.cloud.tools.eclipse.ui.util.DisplayExecutor;
```
chanseokoh: Unnecessary addition.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/messages.properties
```diff
@@ -1,2 +1,3 @@
 projectselector.header.name=Name
 projectselector.header.id=ID
+resolving.project.id=Resolving project ''{0}''
```
chanseokoh: This doesn't seem to be used.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectsProvider.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+  }
+
+  @Override
+  public Object[] getElements(Object inputElement) {
+    if (fetchProjectsJob != null) {
+      Future<GcpProject[]> projectsFuture = fetchProjectsJob.getProjects();
+      if (projectsFuture != null && projectsFuture.isDone()) {
+        try {
+          return projectsFuture.get();
```
chanseokoh: One caveat: if you launch two jobs by switching from a credential A to a credential B quickly, you'll have two jobs running in parallel. It is possible that the earlier job (for the credential A) completes after the latest job (for the credential B). The result is that the account selector is on the credential B while the project selector lists projects from the credential A.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/MiniSelector.java
```diff
@@ -0,0 +1,156 @@
...
+  public void setProject(GcpProject project) {
+    setProject(project.getId());
+  }
+
+  /**
+   * Set the currently selected project by ID; return the project or {@code null} if no project by
+   * that ID.
+   */
+  public void setProject(final String projectId) {
+    projectsProvider.resolve(projectId, new ProjectsProvider.Callback<GcpProject>() {
```
chanseokoh: `resolve()` is designed to execute the callback in its `displayExecutor`, so maybe it is an duplicate effort to run it in the `displayExecutor` of `MiniSelector` here?

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponentTest.java
```diff
@@ -255,43 +298,82 @@ public void testEnablement_nonExistentStagingLocation()
...
+    component.setCloudProjectText("project");
+    spinEventsUntilResolvedProject();
     component.setStagingLocationText("gs://bucket/object");
     verify(messageTarget, never()).setError(anyString());
   }
+
+  /**
+   * Spin the display loop while the waitCondition is true or we timeout.
+   * 
+   * @param waitCondition
```
elharo: fill in or delete

briandealwis:
Fixes #2217

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/MiniSelector.java
briandealwis: Oops, that's a mistake: `ProjectProvider` should be dispatching on its active thread.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectsProvider.java
briandealwis: The first job will be cancelled before the second job is scheduled. `checkCancelled()` should prevent the future from being set if either the job was explicitly cancelled or should the provider's credential changed (though now that I look at the code, the future should be cancelled too, and it should also support external cancelling the future).

Actually looking at the code for `FetchProjectsJob`, I should extract this out to a common base class for use with other implementations as there's a few gotchas.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/messages.properties
briandealwis: ➡️ 🗑

plugins/com.google.cloud.tools.eclipse.ui.util.test/src/com/google/cloud/tools/eclipse/ui/util/DisplayExecutorTest.java
briandealwis: ➡️ 🗑

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectsProvider.java
chanseokoh: If you think about it, `checkCancelled()` cannot guard it; `Job.cancel()` may be called after the last location of `checkCancelled()` but before setting the future. In other words, `cancel()` does not guarantee cancellation, and the cancelled job may run to completion, triggering the listener. If timing is bad, the listener of the earlier job may execute later than the second job.

As you said, you also need to cancel the future, but since all `Future.cancel()` does is to set the cancelled flag (or additionally the thread interrupted flag), it could suffer from the same problem if not done correctly. Also note that `Future.cancel()` will make `Future.isDone()` return true.

briandealwis: `checkCancelled()` also treats the `ProjectProvider`'s credential having changed as indicating cancellation.

Good point re: isDone() under cancelled;

briandealwis:
This change separates out the mechanics of background network job checking into two new classes:
  - `FuturisticJob<T>`, an abstract class that separates reporting of the _computation results_ (via a future of type `T`) from the _job results_ (i.e., we could fetch a list of buckets associated with a project, which fails with a 401 due to insufficient permission, but the job result would be `Status.OK_STATUS`).  Subclasses need to override `compute()` to perform the actual computation and either returns a result or throws an exception, and `isStale()` to check if this job is still current.
  - `PluggableJob<T>`, which extends `FuturisticJob` and allows `compute()` to be provided as a `Callable<T>` and `isStale()` as an optional `Predicate<Job>`

`FuturisticJob` also adds a new method `abandon()` as the Eclipse Jobs' `Job#cancel()` doesn't do anything unless the job is running; if the job is complete, then `cancel()` is a no-op.  Whereas we want to be able to _abandon_ jobs that are no longer current: for example, a list-projects job should be abandoned if the user selects a different account.  Unfortunately `cancel()` is `final`, so we instead add a new `abandon()` method.

In writing code to use the futures, I realized I was writing repetitive code to check if the job/future was still current, etc.  So I added `onSuccess()` and `onError()` listeners to the job, whose callbacks are only called providing the job is still current.

Issue #2211, opened 2017-08-01
Category version in install binary has ".qualifier"


chanseokoh:
We always had the `.qualifier` part in our code, so this seems to be a post-effect of #2043.

![selection_031](https://user-images.githubusercontent.com/10523105/28839650-51a6f072-76c2-11e7-82f1-349a485dc99a.png)

![selection_032](https://user-images.githubusercontent.com/10523105/28839657-54834ea8-76c2-11e7-9877-59ae23042233.png)

briandealwis:
Ah it's because of our [post-signing product re-publishing step](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/wiki/Release-Details#steps-covered-by-releaserelease_after_kokorosh):
> 6. Run the Eclipse p2 Product Publisher to add our custom p2 metadata to the repository.

which manually invokes the p2 _product publisher_  and thus doesn't benefit from the _.qualifier_ expansion performed by Tycho.

Given that this is a release version, and the category IU is never actually downloaded to the user (and thus doesn't need to be unique), we should just update the [release instructions](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/wiki/Cloud-Tools-for-Eclipse-Release-Instructions) to drop the _.qualifier` when updating the .product version.

This issue can go away whenever we're able to do signing as part of the build process.

briandealwis:
Updated release doc.

PR #2363, opened 2017-09-07
Version 1.3.1

elharo:
@chanseokoh

### DIFF
--- a/gcp-repo/metadata.product
+++ b/gcp-repo/metadata.product
@@ -9,7 +9,7 @@
   .p2.inf).  p2 generates unique qualifiers for categories, so it would be
   best for our non-release builds to have unique versions too.
 -->
-<product uid="com.google.cloud.tools.eclipse.dist" version="1.3.1" useFeatures="true" includeLaunchers="false">
+<product uid="com.google.cloud.tools.eclipse.dist" version="1.3.1.qualifier" useFeatures="true" includeLaunchers="false">
 
    <configIni use="default">
    </configIni>
--- a/gcp-repo/metadata.product
+++ b/gcp-repo/metadata.product
@@ -9,7 +9,7 @@
   .p2.inf).  p2 generates unique qualifiers for categories, so it would be
   best for our non-release builds to have unique versions too.
 -->
-<product uid="com.google.cloud.tools.eclipse.dist" version="1.3.0.qualifier" useFeatures="true" includeLaunchers="false">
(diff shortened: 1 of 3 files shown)


gcp-repo/metadata.product
```diff
@@ -9,7 +9,7 @@
   .p2.inf).  p2 generates unique qualifiers for categories, so it would be
   best for our non-release builds to have unique versions too.
 -->
-<product uid="com.google.cloud.tools.eclipse.dist" version="1.3.0.qualifier" useFeatures="true" includeLaunchers="false">
+<product uid="com.google.cloud.tools.eclipse.dist" version="1.3.1" useFeatures="true" includeLaunchers="false">
```
chanseokoh: I think the release instructions are not up to date. I remember @briandealwis and I decided that we retain `.qualifier` here and drop it temporarily when actually building binaries in a local tag branch (which is indeed not ideal). I'll update the release instructions.

chanseokoh: We should retain `.qualifier` in this PR. I have updated the release instructions accordingly:

> 5. Checkout the tag: `git checkout <tag>`.
> 6. Use `git status` to verify there are no local changes. Make sure you see "nothing to commit, working directory clean".
> 7. Drop the `.qualifier` in `gcp-repo/metadata.product` (e.g., `version="1.3.0.qualifier"` to `version="1.3.0"`) and use the exact release version ([#2211](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2211)). Do not commit. `git status` should show that `gcp-repo/metadata.product` is the only file that changed.

@briandealwis we are good with this, aren't we?

briandealwis: That's what we did before.  Perhaps we should change it slightly and:
  - increment `1.3.1.qualifier`
  - commit & push to _master_
  -  tag 1.3.1
  - drop the `.qualifier` and do the actual build

Otherwise `master` stays at `1.3.0.qualifier` and we'll need to fix that up separately.

chanseokoh: That's exactly what I am saying. This PR will change the version to `1.3.1.qualifier`.

briandealwis:
The rationale for dropping the .qualifier from the product was: since this is a marketing version, this version only ever appears in the category metadata that's shown in the Install Wizard, and we will never re-release a version (we won't re-release 1.3.1, we'll just do 1.3.2), it made sense to have the product version reflect the marketing version.  We leave the .qualifier for our normal builds in case we ever aggregate builds (e.g., _nightlies_).

PR #2364, opened 2017-09-07
export library files by default

elharo:
@chanseokoh This restores the behavior of the previous release.


chanseokoh:
Seems like the Dataflow team just released a sample archetype 1.9.1, up from 1.9.0. That's why the tests are failing. Not seeing 2.1.0 yet, but given that 1.9.1 is out, I think it's possible that they roll out 2.1.0 out soon.

chanseokoh:
@tgroh is it expected that the 2.1.0 versions of the Starter and Example Dataflow sample archetypes are released soon? We need to fix tests based on sample versions.

elharo:
The tests shouldn't depend on the specific released versions of the archetypes.

PR #2375, opened 2017-09-12
Make RunOptionsDefaultsComponent able to be partially filled

briandealwis:
- introduce new `partial` argument on `RunOptionsDefaultsComponent` constructors: if `true` then the component should be able to be partially filled, and be complete providing the field values are valid
- rearrange the ordering of checks in `RunOptionsDefaultsComponent#validate()` as the _fetch staging locations for project P_ serves as a proxy for project validity
- adds new tests to verify that partial support is handled; required properly mocking project validation checks; switched to Mockito's _doXXX.when(obj).yyy()_ syntax as some mocks were classes and wouldn't handle mock-overrides properly

With this patch, I can create new dataflow projects without specifying some or all of the run-options defaults.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
@@ -86,6 +86,11 @@ public class RunOptionsDefaultsComponent {
   private final MessageTarget messageTarget;
   private final Composite target;
   
+  /**
+   * If true, then this component is allowed to be partially-complete.
+   */
+  private final boolean partial;
+
   private final AccountSelector accountSelector;
   private final Text projectInput;
   private final Combo stagingLocationInput;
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -86,39 +86,47 @@
...
   public RunOptionsDefaultsComponent(Composite target, int columns, MessageTarget messageTarget,
       DataflowPreferences preferences) {
-    this(target, columns, messageTarget, preferences, null);
+    this(target, columns, messageTarget, preferences, null, false);
   }
 
   public RunOptionsDefaultsComponent(Composite target, int columns, MessageTarget messageTarget,
-      DataflowPreferences preferences, WizardPage page) {
-    this(target, columns, messageTarget, preferences, page,
+      DataflowPreferences preferences, WizardPage page, boolean partial) {
```
elharo: "partial" isn't clear to me. Maybe allowIncomplete or requireComplete with sense reversed?

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -206,40 +214,31 @@ public void modifyText(ModifyEvent event) {
   }
 
   private void validate() {
+    // we set pageComplete to the value of `partial` if the fields are valid
```
elharo: Is this true? Do we ever set page complete to true directly? That is, is partial is false when do we call setPageComplete(true)?

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
briandealwis: ok: `allowIncomplete`

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
briandealwis: We do [set it `true` directly](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/4d930745128bb6f80024587e2c59aa69ce3e72ff/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java#L303)

Issue #2105, opened 2017-07-05
Add "Deploy to App Engine Flexible..." context menu for Maven-based flex JAR projects


chanseokoh:
The context menu item is enabled when a project has the Maven nature and the App Engine flex JAR facet (#2104). The deploy dialog will look like the current standard or flex WAR deploy dialog and work the same way.

We can use M2E to package a JAR for deployment, and the JAR should be what `mvn package` (or `clean package`?) would produce. The JAR packing starts once a user clicks on the _Deploy_ button. If M2E fails to generate a JAR this way, the deploy process will terminate with an error. We won’t attempt any sophisticated checking on the JAR beyond checking if the Maven builder has produced it.

chanseokoh:
Fixed by #2389. Closing.

PR #2380, opened 2017-09-13
JAR-packaging deploy stager for Maven projects

chanseokoh:
Towards #2105. To be merged to `master`. (Nothing is wired up.)

Introduces a JAR stager class that packages a JAR. It runs `mvn package` through the Maven launch configuration mechanism. A temporary (not saved) and private (not visible to users) launch config is created and launched. 

A simple Spring Boot app works, and this PR actually adds the app as a test case.


chanseokoh:
This can be merged to master directly.

PR #2389, opened 2017-09-14
Add and wire up deploy context menu for Flex JAR Maven projects

### LINKED ISSUE #2105: Add "Deploy to App Engine Flexible..." context menu for Maven-based flex JAR projects

chanseokoh:
Fixes #2105.

_Merging to the feature branch `gae-flex-jar-feature`_. After the merge, the feature branch becomes complete and can be merged to `master` to enable deploying Maven JAR projects.

- The (existing) flex deploy context menu is enabled if it has the flex JAR facet and has the Maven nature (in addition to current enabled state for the flex WAR project).
- The command handler determines if a project is a WAR or JAR Maven project and uses an appropriate deploy artifact stager.

Again, this will complete the (initial) implementation of deploying a flex JAR Maven project.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
@@ -129,7 +129,8 @@ public abstract class DeployCommandHandler extends AbstractHandler {
     return severity != IMarker.SEVERITY_ERROR;
   }
 
-  private void launchDeployJob(IProject project, Credential credential) throws IOException {
+  private void launchDeployJob(IProject project, Credential credential)
+      throws IOException, CoreException {
     AnalyticsPingManager.getInstance().sendPing(
         AnalyticsEvents.APP_ENGINE_DEPLOY, AnalyticsEvents.APP_ENGINE_DEPLOY_STANDARD, null);
 
@@ -166,7 +167,7 @@ public abstract class DeployCommandHandler extends AbstractHandler {
     deploy.schedule();
   }
 
-  protected abstract StagingDelegate getStagingDelegate(IProject project);
+  protected abstract StagingDelegate getStagingDelegate(IProject project) throws CoreException;
 
   private static String getConsoleName(String projectId) {
     Date now = new Date();
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployCommandHandler.java
@@ -18,16 +18,23 @@ package com.google.cloud.tools.eclipse.appengine.deploy.ui.flexible;
 
 import com.google.cloud.tools.eclipse.appengine.deploy.StagingDelegate;
 import com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexDeployPreferences;
+import com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexJarMavenProjectStagingDelegate;
 import com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexStagingDelegate;
 import com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployCommandHandler;
 import com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployPreferencesDialog;
 import com.google.cloud.tools.eclipse.appengine.deploy.ui.Messages;
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineFlexJarFacet;
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineFlexWarFacet;
 import com.google.cloud.tools.eclipse.googleapis.IGoogleApiFactory;
 import com.google.cloud.tools.eclipse.login.IGoogleLoginService;
+import com.google.cloud.tools.eclipse.util.MavenUtils;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.swt.widgets.Shell;
+import org.eclipse.wst.common.project.facet.core.IFacetedProject;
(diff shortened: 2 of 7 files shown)


chanseokoh:
I did an end-to-end test of importing a [sample Spring Boot Maven project](https://github.com/spring-guides/gs-rest-service/tree/master/complete) and deploying to App Engine flex. The deploy works fine, and the app works.

chanseokoh:
Looks like there are other tests that I need to fix.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
```diff
@@ -166,7 +167,7 @@ public void done(IJobChangeEvent event) {
     deploy.schedule();
   }
 
-  protected abstract StagingDelegate getStagingDelegate(IProject project);
+  protected abstract StagingDelegate getStagingDelegate(IProject project) throws CoreException;
```
elharo: I'm surprised this didn't already throw an exception. Were we missing a runtime exception somewhere?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployCommandHandler.java
```diff
@@ -39,14 +46,22 @@ protected DeployPreferencesDialog newDeployPreferencesDialog(Shell shell, IProje
   }
 
   @Override
-  protected StagingDelegate getStagingDelegate(IProject project) {
+  protected StagingDelegate getStagingDelegate(IProject project) throws CoreException {
     String appYamlPath = new FlexDeployPreferences(project).getAppYamlPath();
     IFile appYaml = project.getFile(appYamlPath);
     if (!appYaml.exists()) {
```
elharo: aha! This should not be a runtime exception since it indicates a problem outside the code itself. Separate bug but we should fix it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployCommandHandler.java
```diff
@@ -39,14 +46,22 @@ protected DeployPreferencesDialog newDeployPreferencesDialog(Shell shell, IProje
...
       throw new IllegalStateException("Invalid app.yaml path from project preferences.");
     }
     IPath appEngineDirectory = appYaml.getParent().getLocation();
-    return new FlexStagingDelegate(appEngineDirectory);
+
+    IFacetedProject facetedProject = ProjectFacetsManager.create(project);
+    if (AppEngineFlexWarFacet.hasFacet(facetedProject)) {
+      return new FlexStagingDelegate(appEngineDirectory);
+    }
+    if (AppEngineFlexJarFacet.hasFacet(facetedProject) && MavenUtils.hasMavenNature(project)) {
```
elharo: else if?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployCommandHandler.java
```diff
@@ -39,14 +46,22 @@ protected DeployPreferencesDialog newDeployPreferencesDialog(Shell shell, IProje
...
+      return new FlexJarMavenProjectStagingDelegate(appEngineDirectory);
```
elharo: What if it has the Jar facet but is not a Maven project?

chanseokoh: It cannot happen because the deploy command is not enabled in that case (for now), but if it can ever happen, it will throw the runtime exception below.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
chanseokoh: `CoreException` is a checked exception that was not needed before. It is now thrown from `ProjectFacetsManager.create()` (e.g., when a project is closed) introduced in this PR, so I decided to propagate.

elharo: I think it was needed. We were just throwing the wrong (unchecked) exception instead.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployCommandHandler.java
```diff
@@ -39,14 +46,22 @@ protected DeployPreferencesDialog newDeployPreferencesDialog(Shell shell, IProje
...
-    return new FlexStagingDelegate(appEngineDirectory);
+
+    IFacetedProject facetedProject = ProjectFacetsManager.create(project);
+    if (AppEngineFlexWarFacet.hasFacet(facetedProject)) {
+      return new FlexStagingDelegate(appEngineDirectory);
+    } else if (AppEngineFlexJarFacet.hasFacet(facetedProject)
+        && MavenUtils.hasMavenNature(project)) {
+      return new FlexJarMavenProjectStagingDelegate(appEngineDirectory);
+    }
+    throw new IllegalStateException("BUG: command enabled for non-flex or non-Maven-flex projects");
```
elharo: This should be two different exception messages then, one for no flex facet at all and the other for a flex jar facet without a maven nature.

PR #2395, opened 2017-09-15
Fix flakey MiniProjectSelector tests

briandealwis:
Spin the display event loop until a project has been selected.  The selection-change event could be called multiple times.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.eclipse.projectselector;
 
+import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.instanceOf;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -38,10 +39,16 @@ import com.google.cloud.tools.eclipse.test.util.ui.ShellTestResource;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.ISelectionChangedListener;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.viewers.SelectionChangedEvent;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swtbot.swt.finder.SWTBot;
+import org.eclipse.swtbot.swt.finder.waits.DefaultCondition;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -54,6 +61,13 @@ public class MiniSelectorTest {
   public IGoogleApiFactory apiFactory;
   @Rule
   public ShellTestResource shellResource = new ShellTestResource();
+  
+  private SWTBot bot;
+
+  @Before
+  public void setUp() {
+    bot = new SWTBot(shellResource.getShell());
+  }
 
   @Test
   public void testNoCredential() {
@@ -72,7 +86,7 @@ public class MiniSelectorTest {
     assertNotNull(selector.getSelection());
     assertTrue(selector.getSelection().isEmpty());
     selector.setProject("foo.id");
-    spinEvents(); // setProject() waits for the project list to be returned
+    waitUntilResolvedProject(selector); // waits for the project list to be returned
 
     assertFalse(selector.getSelection().isEmpty());
     assertNotNull(selector.getProject());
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
```diff
@@ -89,31 +101,45 @@ public void testListenerFired() {
...
       @Override
       public void selectionChanged(SelectionChangedEvent event) {
-        assertThat(event.getSelection(), instanceOf(IStructuredSelection.class));
-        IStructuredSelection selection = (IStructuredSelection) event.getSelection();
-        assertEquals(1, selection.size());
-        assertThat(selection.getFirstElement(), instanceOf(GcpProject.class));
-        GcpProject gcpProject = (GcpProject) selection.getFirstElement();
-        calledAndCorrect[0] =
-            "foo".equals(gcpProject.getName()) && "foo.id".equals(gcpProject.getId());
+        lastSelection.set(event.getSelection());
```
chanseokoh: This should be happening only in the UI thread. Do we need `AtomicReference`?

plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
```diff
@@ -89,31 +101,45 @@ public void testListenerFired() {
...
+    assertEquals("foo.id", gcpProject.getId());
   }
 
-  private void spinEvents() {
-    while (Display.getCurrent().readAndDispatch());
+  private void waitUntilResolvedProject(final MiniSelector selector) {
+    bot.waitUntil(new DefaultCondition() {
+      @Override
+      public boolean test() throws Exception {
+        if (Display.getCurrent() != null) {
```
chanseokoh: This seems weird. We are in the UI thread, we already created `MiniSelector` and went through a lot of SWT methods at this point. This can never be null. I think we can remove it.

plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
```diff
@@ -89,31 +101,45 @@ public void testListenerFired() {
...
+          // seems surprising that this is required?
```
chanseokoh: I looked at the `waitUtil()` code, and it just repeatedly calls `Thread.sleep()`. It were fine if the tests were running in a non-UI thread like our existing integration test bundle does, but since we are in the UI thread, this is certainly necessary.

briandealwis: But it also means that the SWT event loop isn't processed when this is the UI thread.

plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
briandealwis: The nice thing with using SWTBot is that it doesn't matter whether the test was run in the UI thread or a different thread (tycho-surefire `<useUIThread>`).

plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
briandealwis: It's just meant for a convenient reference holder.  `SoftReference` and `WeakReference`  may lose the reference.  `final ISelection[] lastSelection = new ISelection[1];`.  I'll change this.

plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
chanseokoh: > But it also means that the SWT event loop isn't processed when this is the UI thread.

No idea if the SWT event loop will be executing in another thread, but what is certain is that `waitUtil()` is just a loop with `Thread.sleep()` and `ICondition.test()`.

```java
	public void waitUntil(ICondition condition, ...) ... {
		while (true) {
			...
				if (condition.test()) return;
			...
			sleep(interval);  // This is just `Thread.sleep()`.
		}
	}
```

chanseokoh:
Let's get this in. It's happening a lot.

plugins/com.google.cloud.tools.eclipse.projectselector.test/src/com/google/cloud/tools/eclipse/projectselector/MiniSelectorTest.java
briandealwis: Ah sorry, my comment was meant as a warning for future readers as to why this was required.  I'll make it clear.

Issue #1558, opened 2017-03-08
Build path contains duplicate entries in Java Project


elharo:
1. Create a Java project (not a GCP project)
2. Add all 3 App Engine Libraries to project
3. Get this error dialog:

![image](https://cloud.githubusercontent.com/assets/1005544/23719119/138043ca-0408-11e7-9c4e-e762d5303301.png)

Despite the error, the entries appear to have been added anyway.

chanseokoh:
This happens also on App Engine projects too, and I think I know why:

1. `BuildPath.addLibraries()` explicitly sets the new libraries by calling `javaProject.setRawClasspath()`.
2. We make `AppEngineLibrariesPage.getNewContainers()` return the libraries that we just added (`return added;`). However, interestingly, the underlying framework will add the libraries we return, which we already added. That's why we get the error message. (I find that this behavior isn't clear from its Javadoc.)

So, we could either 1) don't explicitly add the libraries in `BuildPath.addLibraries()`; or 2) simply return `null` from `AppEngineLibrariesPage.getNewContainers()`.

The option 1) seems to break new project wizards, so we may want to go with 2).

elharo:
I now have this firing 100% of the time at head so we need to fix this before release. I'm nervous about making getNewContainers() return null since that violates the interface contract. This may need a little refactoring so that we only call javaProject.setRawClasspath() from the new project wizards.

chanseokoh:
+1 for refactoring.

PR #2409, opened 2017-09-20
Use one master container

### LINKED ISSUE #2381: Library export conflict build error when adding Objectify and Cloud Core Client Library together
The same error as in #2369, but with Objectify and Cloud Core Client Library on App Engine. This must have been like this (not a regression).

`Invalid classpath publish/export dependency /home/chanseok/.m2/repository/com/google/guava/guava/20.0/guava-20.0.jar. The project contains another dependency with the same archive name.`

elharo:
fix #2381 @chanseokoh @briandealwis


briandealwis:
Re: the jars in `plugins/com.google.cloud.tools.eclipse.googlesapis/lib`: you added them prior to merging in the .gitignore change from #2401.
```
*   6e277be Merge branch 'master' into container [Elliotte Harold, 2 days ago]
|\  
| * e8e61c1 Clean up plugins/com.google.cloud.tools.eclipse.googleapis configuration (#2401) [Brian de Alwis, 3 days ago]
* | eaf09c9 wip [Elliotte Harold, 3 days ago]
|/  
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -92,6 +96,7 @@ public void setSelection(IClasspathEntry containerEntry) {
 
   @Override
   public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
+    // todo can we use the currentEntries to tick the checkboxes in the library selector group?
```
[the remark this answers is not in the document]
elharo: moved

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -115,7 +115,7 @@ void setSiteUri(URI siteUri) {
   /**
    * @param libraryFiles artifacts associated with this library, cannot be <code>null</code>
    */
-  synchronized void setLibraryFiles(List<LibraryFile> libraryFiles) {
+  public synchronized void setLibraryFiles(Collection<LibraryFile> libraryFiles) {
```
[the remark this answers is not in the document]
elharo: This came because I was using a set to remove the duplicates. Should we sort the set into a list? Perhaps alphabetically by jar name? That would certainly make the classpath a little easier to scan in the package explorer.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
```diff
@@ -92,7 +92,7 @@ limitations under the License.
             <xs:choice minOccurs="0" maxOccurs="unbounded">
                <xs:element ref="libraryDependency" />
             </xs:choice>
-            <xs:choice minOccurs="1" maxOccurs="unbounded">
+            <xs:choice minOccurs="0" maxOccurs="unbounded">
```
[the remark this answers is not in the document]
elharo: yes. When first loaded the master library has no jars, and none are listed in plugin.xml which this schema inspects directly.

chanseokoh:
Maybe this is an opportunity to update our container prefix and the `Library` Javadoc to remove App Engine, since going forward, we'll have this brand new, single uber container.

```java
/**
 * A library that can be added to App Engine projects, e.g. App Engine Endpoints library.
 *
 */
public final class Library {
  public static final String CONTAINER_PATH_PREFIX =
      "com.google.cloud.tools.eclipse.appengine.libraries";
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -101,23 +106,33 @@ public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
...
       } else {
+        Set<LibraryFile> masterFiles = new HashSet<>();
         for (Library library : libraries) {
           if (!library.isResolved()) {
             library.resolveDependencies();
           }
+          masterFiles.addAll(library.getLibraryFiles());
         }
+
+        Library masterLibrary = CloudLibraries.getMasterLibrary();
```
chanseokoh: I think renaming the local var to `masterContainer` or `masterLibraryContainer` improves readability.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -101,23 +106,33 @@ public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
...
+        masterFiles.addAll(masterLibrary.getLibraryFiles());
+        masterLibrary.setLibraryFiles(masterFiles);
+        ArrayList<Library> masterLibraries = new ArrayList<>();
+        masterLibraries.add(masterLibrary);
```
chanseokoh: - `List<Library> ... = Collections.singletonList(masterContainer);` reads it better to me.
- The var name `masterLibraries` makes me a bit confused, as it sounds like there are multiple masters.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -101,23 +106,33 @@ public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
...
+        
+        // todo This takes a long time. Use a real progress monitor
         IClasspathEntry[] added =
-            BuildPath.listAdditionalLibraries(project, libraries, new NullProgressMonitor());
+            BuildPath.addNativeLibraries(project, masterLibraries, new NullProgressMonitor());
```
chanseokoh: We should keep `list..` here, otherwise we will hit #1558 we fixed a long time ago. #1558 doesn't happen for App Engine projects created by our wizards, but this happens when you try to add libraries on a plain Java project.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -118,17 +125,37 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
+      }
 
-    BuildPath.addLibraries(newProject, config.getAppEngineLibraries(), subMonitor.newChild(2));
+      Library masterLibrary = CloudLibraries.getMasterLibrary();
+      // new project so no existing files in master yet
+      masterLibrary.setLibraryFiles(masterFiles);
+      
+      // todo pointless; refactor so addLibraries takes a Collection<LibraryFile>
+      ArrayList<Library> masterLibraries = new ArrayList<>();
+      masterLibraries.add(masterLibrary);
```
chanseokoh: Ditto. (Using `Collections.singletonList()` and about the `masterLibraries` name.)

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
chanseokoh: However, I noticed one more problem even if we kept the `list..()`: it won't update the container the second time a user tries to add new libraries. This is because `runContainerResolverJob()` is not called in `BuildPath.prepareLibraries()` if `addToClasspath` is `false`. It is the `runContainerResolverJob()` method that can actually update the newly set library files.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
elharo: Passing ImmutableLists to methods that don't specifically take an ImmutableList in their signature makes me nervous.

I've refactored so masterLibraries no longer appears here.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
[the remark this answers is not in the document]
elharo: reverted to list, sorted roughly by maven coordinates

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
elharo: fixed by refactoring and simplifying the API

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
elharo: Restored list, modulo some refactoring. Still need to figure out what to do about the runContainerResolverJob()

elharo: Now runs the container resolver job after listing

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
```diff
@@ -78,4 +78,30 @@ public boolean isExport() {
   void setExport(boolean export) {
     this.export = export;
   }
+  
+  @Override
+  public String toString() {
+    return this.mavenCoordinates.toStringCoordinates();
```
chanseokoh: `this` not needed in these methods.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
```diff
@@ -78,4 +78,30 @@ public boolean isExport() {
...
+        .compareTo(other.mavenCoordinates.toStringCoordinates());
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (other == null || !(other instanceof LibraryFile)) {
+      return false;
+    }
+    return this.mavenCoordinates.toStringCoordinates()
+        .equals((((LibraryFile) other).mavenCoordinates.toStringCoordinates()));
```
chanseokoh: Smallish: maybe `mavenCoordinates.compareTo((LibraryFile) other) == 0`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
chanseokoh: It may be safer to have `toStringCoordinates()` in this class (maybe then with a name `getMavenCoordinatesString()`), because changes in the `MavenCooridnates` class may affect the results of these methods? Not a big deal though.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -101,23 +106,29 @@ public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
...
             library.resolveDependencies();
           }
+          masterFiles.addAll(library.getLibraryFiles());
         }
-        IClasspathEntry[] added =
-            BuildPath.listAdditionalLibraries(project, libraries, new NullProgressMonitor());
+
+        Library masterLibrary = CloudLibraries.getMasterLibrary();
+        masterFiles.addAll(masterLibrary.getLibraryFiles());
+        masterLibrary.setLibraryFiles(new ArrayList<LibraryFile>(masterFiles));
```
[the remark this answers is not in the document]
elharo: That sounds right. Possibly we don't need to get this from an extension point at all.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -8,6 +8,14 @@
        whether statically or dynamically --> 
   <extension point="com.google.cloud.tools.eclipse.appengine.libraries"
              id="com.google.cloud.tools.eclipse.appengine.libraries.defaultapis">
+             
+    <library 
+          id="master-container"
+          group="master"
+          name="Google APIs"
```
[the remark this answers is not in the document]
elharo: This also contains Objectify and Endpoints. It's not just the google-cloud-java APIs

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -79,55 +69,47 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
+    List<IClasspathEntry> newEntries = readEntries(javaProject, library);
+    subMonitor.worked(1);
+    
+    ClasspathUtil.addClasspathEntries(javaProject.getProject(), newEntries, subMonitor);
+    runContainerResolverJob(javaProject);
+    
+    return newEntries.toArray(new IClasspathEntry[0]);
+  }
 
+  private static List<IClasspathEntry> readEntries(IJavaProject javaProject, Library library)
```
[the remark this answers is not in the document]
elharo: I'm not sure I follow. Are you suggesting a rename?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
elharo: this removed

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
elharo: I've always though equals comes first. Guess it's just that I started learning Java before Comparable was a thing. :-)

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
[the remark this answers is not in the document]
chanseokoh: Yeah, was suggesting a rename.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -65,28 +72,47 @@
...
+   * Returns the library with the specified ID, or null if not found
    */
   public static Library getLibrary(String id) {
-    // todo this method is only called by LibraryClasspathContainerResolverService and tests.
-    // Perhaps here or in LibraryClasspathContainerResolverService is where we should 
-    // resolve the transitive dependencies
     return libraries.get(id);
   }
+
+  private static final LoadingCache<IJavaProject, Library> masterLibraries =
```
chanseokoh: Caching `Library` doesn't seem the right direction. The cache is gone when Eclipse restarts, which means that you have no info about which projects have which libraries. Steps to reproduce a problem:

1. Create a fresh project and add Endpoints and Objectify. Verify they are both added.
1. _File_ > _Restart_ (or restart Eclipse by any means)
1. Add App Engine API to the same project.

The project will remove Objects and Endpoints added earlier.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/xsd/com.google.cloud.tools.eclipse.appengine.libraries.xsd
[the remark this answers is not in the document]
elharo: reverted since master library has been removed from plugin.xml

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: Makes sense. Let me think about this. I suspect we need a ClasspathContainerInitializer that preloads the cache. Or some other way of retrieving the container from the IJavaProject.

elharo: I'm having trouble figuring out how to read the specific jars from the master container. There may be some massive hacks that can make this work, but it feels like there should be a more straightforward way to list out the contents of the master-container once I have a reference to it.

briandealwis: The previous approach took the Library ID from the container path.  I think you'll need to explicitly record the selected Library IDs either into the serialized JSON, or in a separate file alongside.

elharo: Any idea how we might  record the selected Library IDs into the serialized JSON? I can't find any API or documentation for this file. There must be some code in Eclipse somewhere that's writing this stuff, probably in some internal package. 

Alternatively, if there were an API for reading the actual jar paths out of an IClasspathEntry that represents a container, that would let us make this work.

elharo:
PTAL @briandealwis @chanseokoh

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -118,17 +125,33 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
       addJunit4ToClasspath(newProject, subMonitor.newChild(2));
-    }
+      IJavaProject javaProject = JavaCore.create(newProject);
+      
+      List<Library> libraries = config.getAppEngineLibraries();
+      // todo duplicates some code in CloudLibrariesPage
+      SortedSet<LibraryFile> masterFiles = new TreeSet<>();
+      for (Library library : libraries) {
+        if (!library.isResolved()) {
+          library.resolveDependencies();
```
briandealwis: This call should take an `IProgressMonitor`.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -118,17 +125,33 @@ public void execute(IProgressMonitor monitor) throws InvocationTargetException,
...
+      enableMavenNature(newProject, subMonitor.newChild(4));
+      BuildPath.addMavenLibraries(newProject, config.getAppEngineLibraries(), subMonitor.newChild(5));
     } else {
       addJunit4ToClasspath(newProject, subMonitor.newChild(2));
-    }
+      IJavaProject javaProject = JavaCore.create(newProject);
+      
+      List<Library> libraries = config.getAppEngineLibraries();
+      // todo duplicates some code in CloudLibrariesPage
+      SortedSet<LibraryFile> masterFiles = new TreeSet<>();
```
briandealwis: This code seems like it should be hidden in BuildPath#addNativeLibrary()`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
```diff
@@ -113,16 +113,18 @@ public IStatus resolveContainer(IJavaProject javaProject, IPath containerPath,
     
     try {
       String libraryId = containerPath.segment(1);
-      Library library = CloudLibraries.getLibrary(libraryId);
+      Library library = null;
+      if (CloudLibraries.MASTER_CONTAINER_ID.equals(libraryId)) {
```
briandealwis: I believe this code should be something like:
```
Library library = CloudLibraries.getLibrary(javaProject, libraryId);
```
`getLibrary()` will check and return a library definition for `libraryId`, and if not will attempt to load a serialized library definition from a file generated from that `libraryId`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
briandealwis: It's our responsibility to read/write the jar list.  The Maven container transforms the pom.xml into a list of jars; the PDE container transforms the Require-Bundles and Import-Packages into a list of jars.

So we need a single method, `CloudLibraries.load(IJavaProject, String containerId)`.  If a `Library` exists with ID `containerId`, then return it, otherwise look for a serialized library descriptor for the `IJavaProject` (e.g., in `.settings/com.google.cloud.tools.eclipse.xxx/<containerId>.json`) and create a `Library` from that previously-serialized descriptor.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -65,28 +72,47 @@
...
+      CacheBuilder.newBuilder().build(new CacheLoader<IJavaProject, Library>() {
```
chanseokoh: Still trying to understand what the cache does. If we read/write JSON and can figure out which jars the master container has, why do we need a cache? (Certainly adding new APIs is an infrequent event, and even if it were frequent, such an event can hardly be a performance bottleneck.)

elharo: There are multiple master libraries, one per project. The cache is just a map from JavaProject to master library with some stronger thread-safety guarantees and simpler initialization than we would get from a HashMap.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
```diff
@@ -23,6 +23,7 @@ Import-Package: com.google.cloud.tools.eclipse.appengine.ui,
  com.google.cloud.tools.eclipse.util.status,
  com.google.common.annotations;version="[20.0.0,21.0.0)",
  com.google.common.base;version="[20.0.0,21.0.0)",
+ com.google.common.cache;version="20.0.0",
```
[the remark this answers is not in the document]
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: Wee could do that, yes. However it means we add the initialization logic, including the check as to whether a master library for the project already exists, in all three places. 
And we'd still need to store at least  a map of JavaProjects --> Master Libraries, and all three locations would have to update this map. It seemed simpler to me to have it in a single method call.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
elharo: Exactly.

briandealwis:
I hit NPEs if I try to import an existing project (File > Import > General > Existing Project)
```
!ENTRY org.eclipse.core.resources 4 2 2017-09-26 16:51:39.616
!MESSAGE Problems occurred when invoking code from plug-in: "org.eclipse.core.resources".
!STACK 0
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:770)
	at com.google.cloud.tools.eclipse.appengine.libraries.model.Library.setLibraryFiles(Library.java:119)
	at com.google.cloud.tools.eclipse.appengine.libraries.persistence.SerializableLibraryClasspathContainer.toLibraryClasspathContainer(SerializableLibraryClasspathContainer.java:58)
	at com.google.cloud.tools.eclipse.appengine.libraries.persistence.LibraryClasspathContainerSerializer.loadContainer(LibraryClasspathContainerSerializer.java:104)
	at com.google.cloud.tools.eclipse.appengine.libraries.LibraryClasspathContainerInitializer.initialize(LibraryClasspathContainerInitializer.java:77)
	at org.eclipse.jdt.internal.core.JavaModelManager.initializeContainer(JavaModelManager.java:2868)
	at org.eclipse.jdt.internal.core.JavaModelManager$10.run(JavaModelManager.java:2774)
	at org.eclipse.jdt.internal.core.JavaModelManager.initializeAllContainers(JavaModelManager.java:2812)
	at org.eclipse.jdt.internal.core.JavaModelManager.getClasspathContainer(JavaModelManager.java:1895)
	at org.eclipse.jdt.core.JavaCore.getClasspathContainer(JavaCore.java:3298)
	at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:2693)
	at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:2857)
	at org.eclipse.jdt.internal.core.JavaProject.getResolvedClasspath(JavaProject.java:1962)
	at org.eclipse.jdt.internal.core.DeltaProcessingState.getRootInfos(DeltaProcessingState.java:311)
	at org.eclipse.jdt.internal.core.DeltaProcessingState.initializeRoots(DeltaProcessingState.java:256)
	at org.eclipse.jdt.internal.core.DeltaProcessor.processResourceDelta(DeltaProcessor.java:1858)
	at org.eclipse.jdt.internal.core.DeltaProcessor.resourceChanged(DeltaProcessor.java:2059)
	at org.eclipse.jdt.internal.core.DeltaProcessingState.resourceChanged(DeltaProcessingState.java:477)
	at org.eclipse.core.internal.events.NotificationManager$1.run(NotificationManager.java:299)
	at org.eclipse.core.internal.events.NotificationManager.notify(NotificationManager.java:289)
	at org.eclipse.core.internal.events.NotificationManager.broadcastChanges(NotificationManager.java:152)
	at org.eclipse.core.internal.resources.Workspace.broadcastPostChange(Workspace.java:373)
	at org.eclipse.core.internal.resources.Workspace.endOperation(Workspace.java:1470)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2250)
	at org.eclipse.ui.actions.WorkspaceModifyOperation.run(WorkspaceModifyOperation.java:130)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:119)
```

elharo:
That's interesting. Let me see if I can reproduce.

chanseokoh:
The cache is keyed by `IJavaProject`, but it turns out it is effectively keyed by the project name. So, when you delete a project and create a project with the same name, the new project picks up the master container info of the deleted project. To illustrate,

1. Create a new project.
1. Add Endpoints.
1. Delete the project. (Tick _Delete project contents on disk_ so that you delete the project folder as well.)
1. Create a new project with the same name.
1. Add Objectify.

The new project will have both Endpoints and Objectify.

chanseokoh:
But this is more like a corner case.

chanseokoh:
BTW, I found a way to get `IClasspathEntry`s from our container, although I'm not sure if this is an appropriate usage:

```java
      JavaModelManager manager = JavaModelManager.getJavaModelManager();

      // Master Container ID
      IPath containerPath = new Path("com.google.cloud.tools.eclipse.appengine.libraries/master-container");
      IClasspathContainer existingCointainer = manager.containerGet(javaProject, containerPath);
      System.out.println("Container: " + existingCointainer);

      // I noticed it can return this empty container when it's being initialized.
      if (existingCointainer == JavaModelManager.CONTAINER_INITIALIZATION_IN_PROGRESS) {
        System.out.println("Container initialization in progress. Did you just create it? Try later.");
      } else
        for (IClasspathEntry entry : existingCointainer.getClasspathEntries())
          System.out.println(entry);
```

Then it outputs the libraries I added to the master container:

```

<...>/.m2/repository/com/google/appengine/appengine-api-1.0-sdk/1.9.56/appengine-api-1.0-sdk-1.9.56.jar
    [CPE_LIBRARY][K_BINARY][isExported:true]
    [AccessRuleSet ...
        [classpath entry: <...>/.m2/repository/com/google/appengine/appengine-api-1.0-sdk/1.9.56/appengine-api-1.0-sdk-1.9.56.jar]]
    [attributes:com.google.cloud.tools.eclipse.appengine.libraries.repository=central,
                com.google.cloud.tools.eclipse.appengine.libraries.groupid=com.google.appengine,   
                com.google.cloud.tools.eclipse.appengine.libraries.artifactId=appengine-api-1.0-sdk,
                com.google.cloud.tools.eclipse.appengine.libraries.type=jar,
                com.google.cloud.tools.eclipse.appengine.libraries.version=1.9.56,
                org.eclipse.jst.component.dependency=/WEB-INF/lib,
                javadoc_location=https://cloud.google.com/appengine/docs/java/javadoc/]
<...>/.m2/repository/com/google/guava/guava/20.0/guava-20.0.jar
    [CPE_LIBRARY][K_BINARY]
    [sourcePath:<...>/.m2/repository/com/google/guava/guava/20.0/guava-20.0-sources.jar]
    [isExported:true]
    [attributes:com.google.cloud.tools.eclipse.appengine.libraries.repository=central,
                com.google.cloud.tools.eclipse.appengine.libraries.groupid=com.google.guava,
                com.google.cloud.tools.eclipse.appengine.libraries.artifactId=guava,
                com.google.cloud.tools.eclipse.appengine.libraries.type=jar,   
                com.google.cloud.tools.eclipse.appengine.libraries.version=20.0,
                org.eclipse.jst.component.dependency=/WEB-INF/lib,
                javadoc_location=https://google.github.io/guava/releases/20.0/api/docs/]
<...>/.m2/repository/com/googlecode/objectify/objectify/5.1.21/objectify-5.1.21.jar
    [CPE_LIBRARY][K_BINARY]
    [sourcePath:<...>/.m2/repository/com/googlecode/objectify/objectify/5.1.21/objectify-5.1.21-sources.jar]
    [isExported:true]
    [attributes:com.google.cloud.tools.eclipse.appengine.libraries.repository=central,
                com.google.cloud.tools.eclipse.appengine.libraries.groupid=com.googlecode.objectify,
                com.google.cloud.tools.eclipse.appengine.libraries.artifactId=objectify,
                com.google.cloud.tools.eclipse.appengine.libraries.type=jar,
                com.google.cloud.tools.eclipse.appengine.libraries.version=5.1.21,
                org.eclipse.jst.component.dependency=/WEB-INF/lib,
                javadoc_location=https://www.javadoc.io/doc/com.googlecode.objectify/objectify/5.1.21]
```

briandealwis:
> Was this a project that had a master container or an older one that didn't?

This was a AESv7 Hello World project created with 1.3.1.  So no master container.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -65,28 +72,47 @@
...
+
+        @Override
+        public Library load(IJavaProject project) throws JavaModelException {
```
chanseokoh: No one throws `JavaModelException`. If we remove it, we can just call `.getUnchekced()` in `getMasterLibrary()` below. It would be good to leave a comment here saying "Do not throw a checked exception, as cache clients call getUnchecked()", if we remove it.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
```diff
@@ -27,17 +30,21 @@
...
 
   public LibraryClasspathContainer(IPath path, String description,
-      List<IClasspathEntry> classpathEntries) {
+      List<IClasspathEntry> classpathEntries, List<LibraryFile> libraryFiles) {
     Preconditions.checkNotNull(path, "path is null");
     Preconditions.checkNotNull(description, "description is null");
     Preconditions.checkArgument(!description.isEmpty(), "description is empty");
     Preconditions.checkNotNull(classpathEntries, "classpathEntries is null");
+    Preconditions.checkNotNull(libraryFiles, "libraryFiles is null");
+    // Preconditions.checkArgument(classpathEntries.size() == libraryFiles.size());
```
chanseokoh: Remove it?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -79,53 +73,66 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
+   * Returns the entries added to the classpath.
    */
-  static IClasspathEntry[] addLibraries(
-      IJavaProject javaProject, List<Library> libraries, IProgressMonitor monitor)
-          throws JavaModelException, CoreException {
-
-    // todo or we could grab the transitive dependency graph here if needed
-    // at this point we have the Java project so we could even filter
-    // entries already present in other libraries
+  public static IClasspathEntry[] addNativeLibrary(IJavaProject javaProject,
```
chanseokoh: I've been thinking about the interface for a while. Going forward, if we are going to make `BuildPath` not general in that it always and only adds a single master library, it'd be much clearer and less confusing if we rename this to `addNativeMasterLibrary` (or `addNativeMasterContainer` or `addNativeMasterLibraryContainer`). The Javadoc would also signify that this method returns either 1) a singleton `IClasspathEntry` array, whose only element is the master container; or 2) an empty array if the container already exists in the project.

In any case, I don't see anywhere its return value is used. We could make it `void`.

Perhaps we could create a new issue for this renaming and updating Javadocs.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -79,53 +73,66 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
     SubMonitor subMonitor = SubMonitor.convert(monitor,
         Messages.getString("adding.app.engine.libraries"), //$NON-NLS-1$
-        libraries.size() + 1); // + 1 because we pass the submonitor along below
-
-    List<IClasspathEntry> rawClasspath = Lists.newArrayList(javaProject.getRawClasspath());
+        2);
+    
+    Library masterLibrary = collectLibraryFiles(javaProject, libraries);
+    
+    List<IClasspathEntry> newEntries = computeEntries(javaProject, masterLibrary);
```
chanseokoh: As with the `addNativeLibrary` naming: if we make this class only for the master library, `computeEntries` always return either 1) an empty list if `javaProject` already has the master container; or 2) a singleton list containing the master container. So, it would be much less confusing to do

```java
   if (javaProject doesn't have the master container) {
       ClasspathUtil.addClasspathEntries(..., newEntries /* a one-size array containing the master container */, ...);
   }
   runContainerResolverJob();
}
```
As I said, we could do this in a separate PR.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -79,53 +73,66 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
-    }
-
-    return newEntries.toArray(new IClasspathEntry[0]);
+    Library masterLibrary = CloudLibraries.getMasterLibrary(javaProject);
+    masterFiles.addAll(masterLibrary.getLibraryFiles());
+    masterLibrary.setLibraryFiles(new ArrayList<LibraryFile>(masterFiles));
+    return masterLibrary;
   }
 
+  private static List<IClasspathEntry> computeEntries(IJavaProject javaProject, Library library)
```
chanseokoh: If we make this class only for the master library, this method becomes not so useful, because all it does is to return either 1) an empty list if `javaProject` already has the master container; or 2) a singleton list containing the master container. It would be clearer to rename this and add a Javadoc. As I said, all this refactoring can be done in another PR.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -79,53 +73,66 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
+  
   /**
-   * @return the entries to be added to the classpath. Does not add them to the classpath.
+   * Returns the entries to be added to the classpath. Does not add them to the classpath.
    */
-  public static IClasspathEntry[] listAdditionalLibraries(
-      IJavaProject javaProject, List<Library> libraries, IProgressMonitor monitor)
-          throws JavaModelException, CoreException {
-    return prepareLibraries(javaProject, libraries, monitor, false);
+  public static IClasspathEntry[] listNativeLibrary(IJavaProject javaProject, Library library)
```
chanseokoh: Similarly, this really means `listNativeMasterLibrary()`, returning either a singleton array or an empty list.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
chanseokoh: If we reach that point, I prefer the name `addNativeMasterContainer()` and rename the parameter `libraries` to `librariesToInclude`. Currently, it is kind of confusing in that the method sounds like it is adding a single "library" (`addNativeLibrary` - which is really the master container) but it is accepting multiple "libraries" (the parameter).

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/SerializableLibraryClasspathContainer.java
```diff
@@ -17,38 +17,56 @@
...
+  LibraryClasspathContainer toLibraryClasspathContainer(IJavaProject javaProject, IPath baseDirectory,
       IPath sourceBaseDirectory) {
     List<IClasspathEntry> classpathEntries = new ArrayList<>();
+
     for (SerializableClasspathEntry entry : entries) {
       classpathEntries.add(entry.toClasspathEntry(baseDirectory, sourceBaseDirectory));
     }
-    return new LibraryClasspathContainer(new Path(path), description, classpathEntries);
+    
+    Library masterLibrary = CloudLibraries.getMasterLibrary(javaProject);
```
briandealwis: Why is this general-purpose deserializer adding what could be arbitrary libraryFiles to the project's master library?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
elharo: Yes, at this point the return value is only used in tests so this probably should be void. I'm not sure how much those tests are testing now.
 addNativeLibraries, perhaps since it can add more than one. "ToInclude" doesn't seem necessary.
Let's not overthink this one as long as it works for now. I expect a lot more changes around all this code as we move to purely transitive dependencies, and any cleanup work we do here is likely to get reworked anyway.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
elharo: Again, let's wait to see where we land once we load the dependencies from Maven, and don't have library dependencies.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
elharo: again clearer once there's only a master container left

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
elharo: This one, I think, operates on the argument, not the master library.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: good catch, done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/SerializableLibraryClasspathContainer.java
elharo: It's actually not. I tried that but it hit some of the same issues as earlier. Instead it adds an empty list when it reads a legacy JSON (no libraryFiles element) so other parts of the code don't blow up when they get a null. This seems to be sufficient to avoid the problems on import you found yesterday.

elharo:
@briandealwis What do you suggest instead of a master library cache? I suppose we could try to rewrite getMasterLibrary to find an existing library in the project but no, we need the library here, not the container.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
chanseokoh: That's true. And at the current state, the argument `library` is always `masterLibrary`, so I'm also implying we could remove the argument going forward. However, let's not overthink it.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
chanseokoh: Missed this one. Should be `"[20.0.0,21.0.0)"`

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
```diff
@@ -23,7 +23,9 @@ Import-Package: com.google.cloud.tools.eclipse.appengine.ui,
...
  com.google.common.collect;version="[20.0.0,21.0.0)",
+ com.google.common.util.concurrent;version="15.0.0",
```
chanseokoh: Should be `"[20.0.0,21.0.0)"`

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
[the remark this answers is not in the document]
chanseokoh: Still not fixed.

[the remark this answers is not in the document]
elharo: oops, fixed something else. Now this is fixed.

briandealwis:
Oh man, what a tar baby.  Previously we created containers for each Library definition, where each container path embedded the Library ID, and each Library definition was assumed to be persistent (i.e., loaded from `plugin.xml`).  Given a Library ID, we could identify the list of jars. As a speed up, we also serialize out a list of the container's constituent jars into `<project>/.settings/com.google.cloud.tools.eclipse.appengine.libraries/<libraryId>.container`.  But we computed this cached list of jars when loading the classpath container, not when creating the classpath container.  Our `BuildPath` code creates the containers (with the ID), and JDT's container initialization process then computes the jar list and serializes it out.  Subsequent restarts are fast as the container initialization just uses that computed list.

With this new master library, we don't actually have a persistent library definition, so we were having to use the `CloudLibraries` master-libraries cache to keep ahold of the master library definition so that the subsequent container initializer can get the list of jar files to be serialized out.  This is different from before, since that list was computed from the Library ID encoded in the container path.

I've pushed up a separate branch to [`containers-no-cache`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/containers-no-cache): it removes the `CloudLibraries` master-libraries cache, and causes `BuildPath#addNativeLibraries()` to also serialize out the list of Library IDs.  It changes the LibraryClasspathContainerResolverService to use that list of serialized Library IDs when it computes the serialized list.

I think the serialized list of Library IDs could instead be saved as a IClasspathAttribute on the container, but I didn't feel like affecting more interfaces to pass along classpath attributes.

@elharo I've dismissed my -1 review.  Avoiding the CloudLibraries shared variable feels like a mild improvement.  But if you'd prefer the current approach, then ok.

chanseokoh:
> As a speed up, we also serialize out a list of the container's constituent jars

Q. Is the (de)serialization only for speeding up, or required to do something (if so, what)?

elharo:
Yes, this is a tar baby. I'm hopeful we can cut some of this out as we move away from preregistered library definitions.

briandealwis:
> Q. Is the (de)serialization only for the speed-up, or required to do something (if so, what)?

Providing we write out the selected libraries (either as I've done with a separate file, or — better IMHO — as IClasspathAttributes) it is just a speedup. The serialised file lists the resolved files relative to the user's maven repository, and pointers to the source artifacts too.  Those can all be recomputed  given the selected library IDs. 

Without the definitions, we need to either record or find sufficient information *somewhere* to/repopulate the container.  For example PDE uses the `MANIFEST.MF`

elharo:
IClasspathAttributes would probably be the best. Unfortunately we still need to read the old JSON files for compatibility.

PR #2411, opened 2017-09-21
Refactor a few deploy-related classes

chanseokoh:
No new functionality. Refactoring code a bit to prepare for adding the existing-WAR/JAR deploy dialog. (Also renaming wrong vars.)

The dialog works fine.


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployPreferencesPanel.java
```diff
@@ -58,33 +67,50 @@ protected void createCenterArea() {
         includeOptionalConfigurationFilesButton, "includeOptionalConfigurationFiles");
   }
 
-  private void createAppYamlSection() {
+  /**
+   * Helper method to create a row of a {@link Label}, a {@link Text} for a file path input, and
+   * a "browse" {@link Button} to open a {@link FileDialog}. It also does data binding between
+   * the {@link Text} field and the given {@code modelPropertyName}.
+   */
+  protected Text createBrowseFileRow(String modelPropertyName, String labelText, String Tooltip,
```
briandealwis: Would it make sense to provide the databinding property rather than the `modelPropertyName`?  Isolate assumptions that the model is a POJO?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppYamlValidator.java
```diff
@@ -44,9 +47,13 @@
   private final IPath basePath;
   private final IObservableValue appYamlPath;
 
-  public AppYamlValidator(IPath basePath, IObservableValue appYamlPath) {
+  public AppYamlValidator(IPath basePath, Text fileField) {
```
briandealwis: I think you should try to keep this validator SWT-agnostic.  More to the point, we may have the file field being computed in a different way.

(TBH, the basePath could also be an observable value, though we'd just use a `WritableValue`)

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/DeployArtifactValidator.java
```diff
@@ -17,22 +17,30 @@
...
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.Text;
 
 public class DeployArtifactValidator extends FixedMultiValidator {
 
   private final IPath basePath;
   private final IObservableValue deployArtifactPath;
 
-  public DeployArtifactValidator(IPath basePath, IObservableValue deployArtifactPath) {
+  public DeployArtifactValidator(IPath basePath, Text fileField) {
```
briandealwis: Ditto re: keeping it SWT agnostic.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppYamlValidator.java
chanseokoh: I'll make both constructors `public`. That way, the validator is SWT-agnostic by its nature, and the SWT-constructor is just a convenience method.

I think keeping `basePath` as a constant makes understanding code easy. However, I might find some usefulness in the near future, so let me think about it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexDeployPreferencesPanel.java
chanseokoh: Thanks, I could think of better refactoring, which is consistent with an existing pattern.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppYamlValidator.java
chanseokoh: I'll keep `basePath` as `IPath`. Making it a databinding observable value (which implies a dynamically changing value) adds cognitive complexity whose feature is not used at all.

chanseokoh:
Ready for review.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java
```diff
@@ -223,12 +224,23 @@ private void setupProjectSelectorDataBinding() {
...
-  private void setupTextFieldDataBinding(Control control, String modelPropertyName,
+  protected void setupTextFieldDataBinding(Text text, String modelPropertyName,
+      ValidationStatusProvider validator) {
+    ISWTObservableValue textValue = WidgetProperties.text(SWT.Modify).observe(text);
+    IObservableValue modelValue = PojoProperties.value(modelPropertyName).observe(model);
+
+    bindingContext.bindValue(textValue, modelValue);
+    if (requireValues) {
+      bindingContext.addValidationStatusProvider(validator);
+    }
```
briandealwis: This doesn't seem right: if !requiresValues, then we won't validate the field contents (e.g., we could have an invalid bucket name).

chanseokoh: That is the point of `requireValues` that was introduced a very long time ago. The purpose is to allow the properties page (as opposed to the deploy dialog) to save invalid values (e.g., non-existing app.yaml).

BTW, this is the original code. I did not change it.

briandealwis: Weird.  We should either rename it (`mustBeValidated`?) or, what I would have thought was the original intent, add support for these to be optional but valid fields?

chanseokoh: `mustBeValidate` sounds good. These are not about the required-optional notion.

briandealwis: Where did this code come from?  I don't see it in `AppEngineDeployPreferencesPanel`.  The code I do see still does validate the fields, whereas this will only add the provided validator if `requireValues == true`.

briandealwis: Oh [found it](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2411/files#diff-57c80aef5b529e610d70cdde8d3056f1L84).  Hmm, the code isn't equivalent: the previous code in `FlexDeployPreferencesPanel` was for a specific file (the app.yaml), and the validator was checking the file contents.  But this is now a utility method, and the validator may be doing other validation that is required.

chanseokoh: I'll tell you shortly.

BTW, I just realized this branch got a wrong merge, so I need to fix it first.

briandealwis: > These are not about the required-optional notion.

That doesn't look to be the case in `AppEngineDeployPreferencesPanel`.

chanseokoh: This will clear your doubt: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.3.1C/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java#L571

I do know the whole history of `requireValues`. It was renamed once. The value is true for the deploy dialog, and false for the property page. In the property page, we allow saving invalid values.

chanseokoh:
My code changes got disappeared. Something got messed up.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java
chanseokoh: Oh, I see what you're saying. Sorry, my bad. So you're thinking that validation should always be done no matter what. I was thinking skipping the validation based on the following test:

```java
  @Test
  public void testAppYamlPathValidation_noValidationIfMustBeValidatedIsFalse() {
    FlexDeployPreferencesPanel panel = createPanel(false /* mustBeValidated */);

    Text appYamlField = findAppYamlField(panel);
    appYamlField.setText("non/existing/app.yaml");
    assertNull(getAppYamlPathValidationStatus(panel));
  }
```

I was just thinking that the property page shouldn't run the validator to make it accept invalid values, but probably we should be able to make the validator just return INFO rather than ERROR. I agree your intent should be the way.

PR #2412, opened 2017-09-21
Simplify project-list update and integrate ProjectsProvider into MiniSelector

briandealwis:
Move the project-list updating from being performed in the IContentProvider (`ProjectProvider`) into the `MiniSelector` to simplify coordinating the populating of the combo and restoring/making project selection.

PR #2414, opened 2017-09-22
reduce visibility

elharo:
@chanseokoh


briandealwis:
`/push` is failing as the branch is missing the commit from #2412.

PR #2421, opened 2017-09-25
Handle validation errors as INFO instead of skipping it

### LINKED ISSUE #2419: FlexDeployPreferencesPanel interprets requireValues in a wrong way
Currently, the panel interprets `requireValues` as whether to skip validation. The current intent was to allow saving invalid values in the deploy properties page, but this is less than ideal, because we can actually do both validation and allow saving an invalid value.

Basically, `AppYamlValidator` should work like `ProjectSelectionValidator` so that, for an invalid value, it returns an INFO status rather than an ERROR status. So the following test should be changed to `..._noErrorForInvalidAppYamlIfRequireValueIsFalse`.

```java
  @Test
  public void testAppYamlPathValidation_noValidationIfRequireValuesIsFalse() {
    FlexDeployPreferencesPanel panel = createPanel(false /* requireValues */);

    Text appYamlField = findAppYamlField(panel);
    appYamlField.setText("non/existing/app.yaml");
    assertNull(getAppYamlPathValidationStatus(panel));
  }
```

`DeployArtifactValidator` should work in the same manner too.

chanseokoh:
Fixes #2419.

@briandealwis Heaps-up; I'm yet to add tests and haven't checked if all tests are passing. This early PR open is to see if we are aligned.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanelTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanelTest.java
@@ -107,7 +107,7 @@ public class AppEngineDeployPreferencesPanelTest {
   @Test
   public void testAutoSelectSingleAccount() {
     when(loginService.getAccounts()).thenReturn(oneAccountSet);
-    deployPanel = createPanel(true /* requireValues */);
+    deployPanel = createPanel(false /* validationErrorAsInfo */);
     assertThat(deployPanel.getSelectedCredential(), is(credential));
 
     // verify not in error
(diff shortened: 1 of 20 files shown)


chanseokoh:
Hmm... this still may not be consistent with how the deploy bucket is being validated. I will have to look further.

chanseokoh:
On second thought, the bucket validation and the project version validation can always be enforced and don't need treating errors as INFO, as the validation on them is about simple syntax checks.

chanseokoh:
Abandoing as we decided to close #2419.

chanseokoh:
> -1: the issue is really separating validity from completeness. The requireValues was about completeness. But we still want the different fields to be valid if provided.

FTR, we do want to ignore validity for some fields if it is the property page case (as opposed to the deploy dialog), to prevent users from being stuck in the page. (For example, if users accidentally navigated to the property page while the saved `app.yaml` value points to a non-existing file, they cannot close the page or move to another property page until they can luckily find another `app.yaml` in their system with an acceptable content and save that path as a temporary value.)

The `requireValues` was originally introduced to fix this problem: if it is false, users can leave the required fields (the account selector and the project selector) unselected, and users do not get stuck in the property page in case users are not logged in or do not have any GCP project. So, at that point, we could argue `requireValues` implied completeness. However, at the current state, the meaning of `requireValues` clearly went beyond the notion of completeness, such that it is for the specific purpose of fixing the said UX problem: if it is false, it disables validation on app.yaml, so that users do not get stuck in the UI.

However, do note that some fields are always optional+validated (e.g., the project version field and the stage bucket field) even in the property page, completely irrelevant to the `requireValues` setting. This is OK because they are free from the said UX problem: the validation on these fields is a simple syntax check, so users have no way to save/load an invalid value. Even if somehow an invalid value could be loaded, it is trivial to satisfy the syntax check, or as they are optional fields, users could just make them empty.

PR #2423, opened 2017-09-26
Fix NPEs and IAEs when creating new Dataflow launch configuration

### LINKED ISSUE #2410: Cannot create/open a Dataflow run/debug configuration: NPE and IllegalArgumentException
To reproduce, just try to create a Dataflow launch config.

The first NPE is due to `RunOptionsDefaultsComponent.getProject()` returning `null` (because initially no project is selected in `MiniSelector`). We could at least make it return an empty string instead, but looking at `updateDefaultableInputValues()`, I'm not sure what it means if we are not setting `customValues` synchronously.

```
	at com.google.cloud.tools.eclipse.dataflow.ui.launcher.DefaultedPipelineOptionsComponent.updateDefaultableInputValues(DefaultedPipelineOptionsComponent.java:136)
	at com.google.cloud.tools.eclipse.dataflow.ui.launcher.DefaultedPipelineOptionsComponent.setPreferences(DefaultedPipelineOptionsComponent.java:105)
	at com.google.cloud.tools.eclipse.dataflow.ui.launcher.PipelineArgumentsTab.initializeFrom(PipelineArgumentsTab.java:316)
	at org.eclipse.debug.ui.AbstractLaunchConfigurationTabGroup.initializeFrom(AbstractLaunchConfigurationTabGroup.java:90)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationTabGroupWrapper.initializeFrom(LaunchConfigurationTabGroupWrapper.java:201)
```

The second exception seems to be that creating a bogus `Event` instance to simulate calling `ModifyListener.modifyText()` doesn't work (`listener.modifyText(new ModifyEvent(new Event()))`):

```
java.lang.IllegalArgumentException: null source
	at java.util.EventObject.<init>(EventObject.java:56)
	at org.eclipse.swt.internal.SWTEventObject.<init>(SWTEventObject.java:38)
	at org.eclipse.swt.events.TypedEvent.<init>(TypedEvent.java:71)
	at org.eclipse.swt.events.ModifyEvent.<init>(ModifyEvent.java:35)
	at com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponent$5.selectionChanged(RunOptionsDefaultsComponent.java:413)
	at org.eclipse.jface.viewers.StructuredViewer$3.run(StructuredViewer.java:877)
	at org.eclipse.ui.internal.JFaceUtil$1.run(JFaceUtil.java:50)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:173)
	at org.eclipse.jface.viewers.StructuredViewer.firePostSelectionChanged(StructuredViewer.java:874)
	at org.eclipse.jface.viewers.StructuredViewer.setSelection(StructuredViewer.java:1734)
	at org.eclipse.jface.viewers.Viewer.setSelection(Viewer.java:388)
	at com.google.cloud.tools.eclipse.projectselector.MiniSelector$2.accept(MiniSelector.java:170)
	at com.google.cloud.tools.eclipse.projectselector.MiniSelector$2.accept(MiniSelector.java:1)
	at com.google.cloud.tools.eclipse.projectselector.ProjectsProvider$2.accept(ProjectsProvider.java:111)
	at com.google.cloud.tools.eclipse.projectselector.ProjectsProvider$2.accept(ProjectsProvider.java:1)
	at com.google.cloud.tools.eclipse.util.jobs.FuturisticJob$1.run(FuturisticJob.java:110)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:35)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:135)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:3794)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3433)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:827)
	at org.eclipse.jface.window.Window.open(Window.java:803)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationsDialog.open(LaunchConfigurationsDialog.java:1154)
```

briandealwis:
Fixes #2410

Add `getProjectId()` helpers that return an `"" ` if no project is selected.  I don't much like this, but it keeps consistency with the helpers to return selected emails and staging locations.

Create tests for `DefaultedPipelineOptionsComponent` and fixed up class behaviour to be more consistent.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponentTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponentTest.java
@@ -1,0 +1,243 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.eclipse.dataflow.ui.launcher;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+
+import com.google.cloud.tools.eclipse.dataflow.core.preferences.DataflowPreferences;
+import com.google.cloud.tools.eclipse.dataflow.ui.page.MessageTarget;
+import com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponent;
+import com.google.cloud.tools.eclipse.test.util.ui.ShellTestResource;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.stubbing.Answer;
+
+/**
+ * DefaultedPipelineOptionsComponent is responsible for saving and restoring custom values when the
+ * user toggles the <em>use defaults</em> button.
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultedPipelineOptionsComponentTest {
+  @Rule
+  public ShellTestResource shellCreator = new ShellTestResource();
+  @Mock
+  MessageTarget messageTarget;
+
+  RunOptionsDefaultsComponent defaultOptions;
+  String accountEmail = "";
+  String projectId = "";
+  String stagingLocation = "";
+
+  @Before
+  public void setUp() {
+    // This is easier than subclassing RunOptionsDefaultComponent
+    Answer<String> answerAccountEmail = new Answer<String>() {
+      @Override
+      public String answer(InvocationOnMock invocation) throws Throwable {
+        return accountEmail;
+      }
+    };
+    Answer<Void> recordAccountEmail = new Answer<Void>() {
+      @Override
+      public Void answer(InvocationOnMock invocation) throws Throwable {
+        accountEmail = invocation.getArgumentAt(0, String.class);
+        return null;
+      }
+    };
+
+    Answer<String> answerProjectId = new Answer<String>() {
+      @Override
+      public String answer(InvocationOnMock invocation) throws Throwable {
+        return projectId;
+      }
+    };
+    Answer<Void> recordProjectId = new Answer<Void>() {
+      @Override
+      public Void answer(InvocationOnMock invocation) throws Throwable {
+        projectId = invocation.getArgumentAt(0, String.class);
+        return null;
+      }
+    };
+    Answer<String> answerStagingLocation = new Answer<String>() {
+      @Override
+      public String answer(InvocationOnMock invocation) throws Throwable {
+        return stagingLocation;
+      }
+    };
+    Answer<Void> recordStagingLocation = new Answer<Void>() {
+      @Override
+      public Void answer(InvocationOnMock invocation) throws Throwable {
+        stagingLocation = invocation.getArgumentAt(0, String.class);
+        return null;
+      }
+    };
+
+    defaultOptions = mock(RunOptionsDefaultsComponent.class);
+    doAnswer(answerAccountEmail).when(defaultOptions).getAccountEmail();
+    doAnswer(recordAccountEmail).when(defaultOptions).selectAccount(anyString());
+    doThrow(IllegalStateException.class).when(defaultOptions).getProject();
+    doAnswer(answerProjectId).when(defaultOptions).getProjectId();
+    doAnswer(recordProjectId).when(defaultOptions).setCloudProjectText(anyString());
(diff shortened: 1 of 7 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponentTest.java
```diff
@@ -0,0 +1,243 @@
...
+/**
+ * DefaultedPipelineOptionsComponent is responsible for saving and restoring custom values when the
+ * user toggles the <em>use defaults</em> button.
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultedPipelineOptionsComponentTest {
+  @Rule
+  public ShellTestResource shellCreator = new ShellTestResource();
+  @Mock
+  MessageTarget messageTarget;
```
chanseokoh: This and the fields below can be `private`. I think some can be `final` too.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
```diff
@@ -52,6 +51,13 @@
 
   public DefaultedPipelineOptionsComponent(Composite parent, Object layoutData,
       MessageTarget messageTarget, DataflowPreferences preferences) {
+    this(parent, layoutData, messageTarget, preferences, null);
+  }
+
+  @VisibleForTesting
+  public DefaultedPipelineOptionsComponent(Composite parent, Object layoutData,
```
chanseokoh: Perhaps `public` can be removed.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
```diff
@@ -52,6 +51,13 @@
 
   public DefaultedPipelineOptionsComponent(Composite parent, Object layoutData,
       MessageTarget messageTarget, DataflowPreferences preferences) {
+    this(parent, layoutData, messageTarget, preferences, null);
```
chanseokoh: What do you think of passing `new RunOptionsDefaultsComponent()` instead of `null`? That could simplify the `?` ternay-operator below.

chanseokoh:
Did some testing and found a pretty minor problem that may or may not be related to this change, but this PR makes things work at least, so I'll create a separate issue.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
briandealwis: I agree it's awkward.  Unfortunately cannot as it's parented using the `defaultsGroup` which is created inside the constructor.  I had explored creating an interface `RunOptionsContainer` to represent the `RunOptionsDefaultsComponent` API, so that it could be more easily mocked, but it didn't much help.

chanseokoh: Ah, gotcha.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
```diff
@@ -131,36 +142,41 @@ public boolean isUseDefaultOptions() {
...
-      // TODO: Give this a separate input
-      customValues.put(
-          DataflowPreferences.GCP_TEMP_LOCATION_PROPERTY, defaultOptions.getStagingLocation());
-      String defaultAccountEmail = preferences.getDefaultAccountEmail();
-      defaultOptions.selectAccount(Strings.nullToEmpty(defaultAccountEmail));
-      String defaultProject = preferences.getDefaultProject();
-      defaultOptions.setCloudProjectText(Strings.nullToEmpty(defaultProject));
-      String defaultStagingLocation = preferences.getDefaultStagingLocation();
-      defaultOptions.setStagingLocationText(Strings.nullToEmpty(defaultStagingLocation));
+      saveCustomValues();
```
chanseokoh: I think I just learned that this logic (the original logic) is weird, does not make much sense, is useless, but luckily OK because correct custom values are always loaded to the UI by `setCustomValues()`. That is, if you delete the method `saveCustomValues()`, I believe it will continue to work fine. What I think is wrong with the logic is that, if the _use defaults_ button is checked, it means the UI (`defaultOptions`) is currently showing the default (preferences) values, so it will save these preferences values to `customValues`.

However, let's not bother with it here. It's the problem of the original logic.

briandealwis: Ah, I think you're seeing odd behaviour due to the call to `updateDefaultableInputValues()` in the constructor.  Looking back at the code, I added this as I was trying to mock the `RunOptionsDefaultsComponent` and I wasn't configuring it to use the defaults from the preferences unlike the real `RunOptionsDefaultsComponent`.  I'll fix that up.

briandealwis:
PTAL @chanseokoh

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
chanseokoh: I was talking about the nonsensical behavior (I think) that is irrelevant of this PR. `customValues` is to store values entered when the _use defaults_ button is unchecked, isn't it? I was saying the code is storing values of the (project or workspace) defaults into `customValues`. That is not right, isn't it?

PR #2433, opened 2017-09-27
add translate API; switch most libraries to automatic dependency resolution

elharo:
@chanseokoh @briandealwis

Issue #2400, opened 2017-09-18
Properties for Classpath container don't make sense


elharo:
1. Create new project and include some libraries such as Objectify.
2. Select Objectify in Project Explorer
3. Context Menu > Properties

![image](https://user-images.githubusercontent.com/1005544/30540748-7c2a76b2-9c46-11e7-80a9-e83221f76ff7.png)

elharo:
For comparison here are the properties for the JUnit container:

![image](https://user-images.githubusercontent.com/1005544/30540945-2c2d15f6-9c47-11e7-9d3e-92ac99a80cca.png)

elharo:
can hit this in other places too:

![image](https://user-images.githubusercontent.com/1005544/30548004-f348e748-9c5e-11e7-80e1-0d5dee7dfda1.png)

elharo:
This may make a little more sense when we switch to a single container. We will need to check what is currently selected and set the checkboxes accordingly.

elharo:
To check what is currently selected and set the checkboxes accordingly we need to save the list of libraries. I think we can do this as extra classpath attributes on our master container. 

In `CloudLibrariesPage` this is passed to the `initialize` method as `currentEntries`:

```
  @Override
  public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
    this.project = project;
  }
```

More specifically it is the current entry with the path `com.google.cloud.tools.eclipse.appengine.libraries/master-container`.

When we add/remove a library to this path we should add/remove an `IClasspathAttribute` that notes the library ID. 

A classpath attribute defines a name/value pair that can be persisted with a classpath entry. Such an attribute can be created using the factory method `JavaCore.newClasspathAttribute(String name, String value)`.

briandealwis:
The list of libraries is now persisted by #2436: see `LibraryClasspathContainerSerializer#loadLibraryIds()`. That list would make sense to be added as a classpath attribute.

chanseokoh:
This makes sense now, as we now have a single master container.

PR #2436, opened 2017-09-28
Record selected libraries for master container

briandealwis:
Removes the _master library_ cache in `CloudLibraries` and instead causes `BuildPath` to generate sufficient metadata to regenerate the container contents.  Currently the metadata is generated to a file in `<project>/.settings/com.google.cloud.tools.eclipse.appengine.libraries/master-libraries.container`.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -135,7 +162,7 @@ public static Library collectLibraryFiles(IJavaProject javaProject, List<Library
     return newEntries.toArray(new IClasspathEntry[0]);
   }
 
-  private static IClasspathEntry makeClasspathEntry(Library library) throws CoreException {
+  private static IClasspathEntry makeClasspathEntry(final Library library) throws CoreException {
```
elharo: why final?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
```diff
@@ -39,6 +38,7 @@ public LibraryClasspathContainer(IPath path, String description,
     Preconditions.checkArgument(!description.isEmpty(), "description is empty");
     Preconditions.checkNotNull(classpathEntries, "classpathEntries is null");
     Preconditions.checkNotNull(libraryFiles, "libraryFiles is null");
+    // Preconditions.checkArgument(classpathEntries.size() == libraryFiles.size());
```
elharo: delete

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainerResolverJob.java
```diff
@@ -40,6 +44,11 @@ public LibraryClasspathContainerResolverJob(IJavaProject javaProject) {
 
   @Override
   protected IStatus run(IProgressMonitor monitor) {
+    // may have been deleted before this job runs
+    if (!javaProject.exists()) {
```
elharo: This addition looks important in itself. Is that so?

briandealwis: It was to remedy some NPEs occurring during test tear-down.  BuildPath starts a library resolver job (this job) to generate the serialized container cache file, but the test may be finished and the project deleted by the time this job starts to execute.

briandealwis:
I'll re-apply after #2433 and will re-try encoding the library IDs as a IClasspathAttribute.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -110,6 +117,39 @@ public LibraryClasspathContainer loadContainer(IJavaProject javaProject, IPath c
...
+      out.write(gson.toJson(libraryIds.toArray()));
+    }
+  }
+
+  public List<String> loadLibraryIds(IJavaProject javaProject, IPath containerPath)
+      throws IOException, CoreException {
+    File stateFile = getContainerStateFile(javaProject, LIBRARY_LIST_PATH, false);
+    if (stateFile == null) {
+      logger.warning("No library-id state file found: " + stateFile);
+      return null;
```
elharo: Would an empty list be safer here?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -110,6 +117,39 @@ public LibraryClasspathContainer loadContainer(IJavaProject javaProject, IPath c
...
+    }
+    try (Reader reader = Files.newBufferedReader(stateFile.toPath(), StandardCharsets.UTF_8)) {
+      JsonArray array = gson.fromJson(reader, JsonArray.class);
+      if (array == null) {
+        return null;
```
elharo: ditto

briandealwis:
It still reads in the old format: those containers specify an explicit library ID.

> will re-try encoding the library IDs as a IClasspathAttribute

And that doesn't work as the attributes are on the container, and the initializer is to create such a container because it doesn't already exist.  So I cleaned up how the container state files are referenced.

briandealwis:
Squashed into a single commit and pushed.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -101,19 +115,21 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
       masterFiles.addAll(library.getLibraryFiles());
     }
 
-    Library masterLibrary = CloudLibraries.getMasterLibrary(javaProject);
-
-    masterFiles.addAll(masterLibrary.getLibraryFiles());
+    Library masterLibrary = new Library(CloudLibraries.MASTER_CONTAINER_ID);
+    masterLibrary.setName("Google APIs"); //$NON-NLS-1$
+    masterLibrary.setLibraryDependencies(dependentIds);
+    masterFiles.addAll(masterLibrary.getLibraryFiles()); // does this actually do anything?
```
elharo: no, not here since masterLibrary was just created. Delete this.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/SerializableLibraryClasspathContainer.java
```diff
@@ -59,7 +58,8 @@ LibraryClasspathContainer toLibraryClasspathContainer(IJavaProject javaProject,
       classpathEntries.add(entry.toClasspathEntry(baseDirectory, sourceBaseDirectory));
     }
     
-    Library masterLibrary = CloudLibraries.getMasterLibrary(javaProject);
+    Library masterLibrary = new Library(CloudLibraries.MASTER_CONTAINER_ID);
+    masterLibrary.setName("Google APIs"); //$NON-NLS-1$
     if (libraryFiles == null) { // we deserialized an old version
       libraryFiles = new ArrayList<>();
```
chanseokoh: This field was added in the previous PR (#2409, [code](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2409/files#diff-22cc39f47cb2fa69d82c706bd981d445R63)) and is used to do `masterLibrary.setLibraryFiles()` just below. Do we still need this?

elharo: Yes, I think we do. GSON can have set this field to null.

chanseokoh: Sorry, I was asking if we can remove the field entirely, if it is no longer necessary because we now record library IDs?

briandealwis: We do still use the field to record the actual jars (e.g., from the extension points).

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -46,10 +53,11 @@
  */
 @Creatable
 public class LibraryClasspathContainerSerializer {
-
   private static final Logger logger =
       Logger.getLogger(LibraryClasspathContainerSerializer.class.getName());
 
+  private static final String MASTER_LIBRARY_LIST_FILE_ID = "master-libraries";
```
briandealwis: I wonder if we should name this something like `_libraries`, and rename it something like `CONTAINER_LIBRARY_LIST_FILE_ID`.  Thinking toward adding support #2400 where we currently have the App Engine container too.

Issue #2416, opened 2017-09-22
Consider safer pattern to retrieve deploy preferences during deploy


chanseokoh:
The `DeployPreferences()` class is for saving/loading (App Engine standard) deploy parameters (such as a GCP project, a Google account, etc) to/from user preferences. As long as we have an `IProject` ref, we can conveniently retrieve such deploy preferences anywhere at any point by `new DeployPreferences(IProject)`. We do call this constructor to get values during deploy in a couple places. We don't have any problem currently as we do things right, but the current implementation has a potential risk:

- At the language level, `DeployPreferences` is mutable, you might get a different value later if someone changes the value somehow (which isn't currently the case) in the process of deploying.
- Although this isn't currently the case, the implementation of `DeployPreferences` may change in the future in a way that getting values always returns most-up-to-date preferences values (e.g., users updated the preferences while a deploy is in progress) rather than the values at the time when the instance is constructed by `new DeployPreferences(IProject)`.

So, for future safety, I think it is better to explicitly take an immutable snapshot of the preferences before launching a deploy job and passing the snapshot.

chanseokoh:
> ... returns most-up-to-date preferences values (e.g., users updated the preferences while a deploy is in progress) rather than the values at the time when the instance is constructed by new DeployPreferences(IProject).

Oh, this can really happen theoretically because `FlexExistingDeployArtifactStagingDelegate.getDeployArtifact()` does it in a concurrent job (the deploy job).

```
  protected IPath getDeployArtifact(IProject project, IPath safeWorkDirectory,
      IProgressMonitor monitor) throws CoreException {
    String artifactPath = new FlexExistingArtifactDeployPreferences().getDeployArtifactPath();
    return new Path(artifactPath);
  }
```

Issue #2103, opened 2017-07-05
Consider adding context menu for WAR/JAR flex deploy


chanseokoh:
After #2012, consider adding context menu items "Deploy JAR to App Engine Flexible..." and "Deploy WAR to App Engine Flexible..." when a user selects a JAR or a WAR.

PR #2439, opened 2017-09-29
Add and enable global WAR/JAR deploy menu

chanseokoh:
The menu is always enabled.

selection_007

Can't say I'm 100% satisfied with the current code structure, but I decided to go for reasonably minimal code change, as otherwise refactoring starts to propagate to too many files.

Did some actual testing:
- Absolute JAR path and relative app.yaml path, outside the workspace:
selection_005

- Relative JAR path and absolute app.yaml, inside the workspace:
selection_008

I also tested if the existing standard, flex WAR, and flex JAR (Maven projects) deployments work well.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
@@ -8,3 +8,7 @@ deployStandardMenuLabel=Deploy to App Engine Standard...
 deployFlexCommandDescription=Uploads the project to Google App Engine flexible environment.
 deployFlexCommandName=Deploy to App Engine Flexible
 deployFlexMenuLabel=Deploy to App Engine Flexible...
+
+deployFlexExistingWarJarCommandDescription=Uploads existing WAR/JAR to Google App Engine \
+ flexible environment.
+deployFlexExistingWarJarCommandName=Deploy Existing WAR/JAR to App Engine Flexible
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexExistingArtifactDeployCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexExistingArtifactDeployCommandHandler.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.eclipse.appengine.deploy.ui.flexible;
+
+import com.google.cloud.tools.eclipse.appengine.deploy.StagingDelegate;
+import com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexExistingArtifactDeployPreferences;
+import com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexExistingDeployArtifactStagingDelegate;
+import com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployCommandHandler;
+import com.google.cloud.tools.eclipse.appengine.deploy.ui.DeployPreferencesDialog;
+import com.google.cloud.tools.eclipse.appengine.deploy.ui.Messages;
+import com.google.cloud.tools.eclipse.googleapis.IGoogleApiFactory;
+import com.google.cloud.tools.eclipse.login.IGoogleLoginService;
+import com.google.cloud.tools.eclipse.util.status.StatusUtil;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.swt.widgets.Shell;
+
+public class FlexExistingArtifactDeployCommandHandler extends DeployCommandHandler {
+
+  @Override
+  protected DeployPreferencesDialog newDeployPreferencesDialog(Shell shell, IProject project,
+      IGoogleLoginService loginService, IGoogleApiFactory googleApiFactory) {
+    String title = Messages.getString("deploy.preferences.dialog.title.flexible");
+    return new FlexExistingArtifactDeployPreferencesDialog(shell, title, loginService,
+        googleApiFactory);
+  }
+
+  @Override
+  protected StagingDelegate getStagingDelegate(IProject project) throws CoreException {
+    String appYamlPath = new FlexExistingArtifactDeployPreferences().getAppYamlPath();
+    IFile appYaml = resolveFileAgainstWorkspace(appYamlPath);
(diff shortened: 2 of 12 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
```diff
@@ -8,3 +8,7 @@ deployStandardMenuLabel=Deploy to App Engine Standard...
 deployFlexCommandDescription=Uploads the project to Google App Engine flexible environment.
 deployFlexCommandName=Deploy to App Engine Flexible
 deployFlexMenuLabel=Deploy to App Engine Flexible...
+
+deployFlexExistingWarJarCommandDescription=Uploads existing WAR/JAR to Google App Engine \
+ flexible environment.
+deployFlexExistingWarJarCommandName=Deploy Existing WAR/JAR to App Engine Flexible
```
elharo: and this is the menu item text, right? delete "Existing", perhaps change to "Deploy Existing WAR/JAR file App Engine Flexible"

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
```diff
@@ -8,3 +8,7 @@ deployStandardMenuLabel=Deploy to App Engine Standard...
 deployFlexCommandDescription=Uploads the project to Google App Engine flexible environment.
 deployFlexCommandName=Deploy to App Engine Flexible
 deployFlexMenuLabel=Deploy to App Engine Flexible...
+
+deployFlexExistingWarJarCommandDescription=Uploads existing WAR/JAR to Google App Engine \
```
elharo: Is this a tooltip? If so, "Uploads pre-existing WAR or JAR file to..."

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
chanseokoh: Your last suggestion reads weird to me. ("... file App Engine Flexible"). You mean "Deploy Existing App Engine Flexible WAR/JAR File"?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexExistingArtifactDeployCommandHandler.java
```diff
@@ -0,0 +1,83 @@
...
+  protected DeployPreferencesDialog newDeployPreferencesDialog(Shell shell, IProject project,
+      IGoogleLoginService loginService, IGoogleApiFactory googleApiFactory) {
+    String title = Messages.getString("deploy.preferences.dialog.title.flexible");
+    return new FlexExistingArtifactDeployPreferencesDialog(shell, title, loginService,
+        googleApiFactory);
+  }
+
+  @Override
+  protected StagingDelegate getStagingDelegate(IProject project) throws CoreException {
+    String appYamlPath = new FlexExistingArtifactDeployPreferences().getAppYamlPath();
```
briandealwis: This repeated use of `new FlexExistingArtifactDeployPreferences()` seems unusual — it makes sense for the global dialog, but I think it would be good to extract a method that is passed in the artifact and app.yaml.  

I also think this handler can do double-duty, both deploying an existing artifact that's currently selected (e.g., right-clicking on a jar/war and optionally app.yaml in the _Project Explorer_), as well as to prompt for the file.

Given that you don't use the IProject argument, and from looking at DeployCommandHandler, I wonder if you might not be better off  not subclassing?

chanseokoh: > I also think this handler can do double-duty, both deploying an existing artifact that's currently selected (e.g., right-clicking on a jar/war and optionally app.yaml in the Project Explorer), as well as to prompt for the file.

We'll get to that point eventually (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2103). Just not now.

Regarding the other refactoring suggestions, I did expect such comments, and I'll say this again:

> Can't say I'm 100% satisfied with the current code structure, but I decided to go for reasonably minimal code change, as otherwise refactoring starts to propagate to too many files.

To add more, I spent quite a lot of time on what would be the ideal code, and all the refactoring attempts ended up touching a lot of files and a huge PR where the majority of the diff is a distraction from what should be reviewed due to refactoring. So, if it isn't a tangible and small suggestion that can be done in this PR, I'd prefer put off big high-level restructuring ideas.

chanseokoh: Also for the `new FlexExistingArtifactDeployPreferences()`, we already have an issue: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2416

chanseokoh:
PTAL

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
```diff
@@ -8,3 +8,7 @@ deployStandardMenuLabel=Deploy to App Engine Standard...
 deployFlexCommandDescription=Uploads the project to Google App Engine flexible environment.
 deployFlexCommandName=Deploy to App Engine Flexible
 deployFlexMenuLabel=Deploy to App Engine Flexible...
+
+deployFlexExistingWarJarCommandDescription=Uploads pre-existing WAR or JAR file to Google App \
```
elharo: the Google App...

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexExistingArtifactDeployCommandHandler.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  protected DeployPreferences getDeployPreferences(IProject project) {
+    return new FlexExistingArtifactDeployPreferences();
+  }
+
+  @Override
+  protected IProject getSelectedProject(ExecutionEvent event)
+      throws ExecutionException, CoreException {
+    return null;
```
elharo: Should this be throw new UnsupportedOperationException?
Either way, this does make me wonder if the abstractions are correct. I.e. should DeployCommandHandler have a getProject method when not all deploy commands have projects?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
chanseokoh: The message continues in the next line. It's a description, so it doesn't need to end with three dots either.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/flexible/FlexExistingArtifactDeployCommandHandler.java
chanseokoh: Yeah, I also see the abstractions are somewhat weird with the introduction of the global dialog, which doesn't required a project. That's a gripe here, but unfortunately, the existing codebase assumed there is always a project to be deployed, and getting away from it with the global dialog requires massive refactoring. I tried some ideas, but it has been tricky when also considering #2416. Let me file a bug for having better abstractions.

BTW, we need to return `null` for now.

chanseokoh: Tech debt: #2449

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/plugin.properties
chanseokoh: Ah, you mean adding "the".

PR #2440, opened 2017-09-29
Publish project dependency JARs

### LINKED ISSUE #2426: Staging fails to publish external and other project dependencies
- Cloud Tools for Eclipse version: 1.3.1.201709121426
- Google Cloud SDK version: 172.0.1
- OS:OSX 10.12.6 
- Java version:1.7.0_60

**What did you do?**
As i am trying to upgrade from the old google appengine plugin to google cloud tools, i try to deploy a webproject that has a dependency on another (plain java) project in eclipse using the "deploy to appengine standard" button

deployment

**What did you see instead?**
/var/folders/yg/lqfxnjzx2hx54z1609vl5fb80000gn/T/1506502634400-0/org/apache/jsp/WEB_002dINF/jsp/contact_002dselection/page_002dtop_jsp.java:15: error: package nl.anp.aps.entities does not exist
import nl.anp.aps.entities.User;
                          ^
6 errors
Unable to stage app: Failed to compile the generated JSP java files.

Note: 
- the missing imports are from the dependend java project
- deployment with the previous "google plugin for eclipse 4.4/4.5/4.6" worked fine. 
- mvn appengine:deploy also works ok
- i have no spaces in my gcloud sdk path

chanseokoh:
Fixes #2426. However, for native projects, users still need to configure the deploy assembly (which is actually easy).

Description:
Previously, we were publishing web fragment child module only. This PR now publishes all J2EE child modules and project dependency child modules.


plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ChildModuleWarPublishTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.junit.AfterClass;
+import org.junit.Test;
+
+public abstract class ChildModuleWarPublishTest {
```
chanseokoh: Not a new class actually. Code extracted from `WebFragmentWarPublishTest` below to reuse the base.

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ProjectDependencyPublishTest.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Arrays;
+import java.util.List;
+import org.eclipse.core.runtime.CoreException;
+import org.junit.BeforeClass;
+
+public class ProjectDependencyPublishTest extends ChildModuleWarPublishTest {
+
+  @BeforeClass
+  public static void setUp() throws IOException, CoreException {
+    loadTestProjectZip("test-projects/project-dep-publish-example.zip", "dep_publish");
```
chanseokoh: I have zipped up three projects: A) `dep_publish`; B) `simple-dep`; and C) `simple-dep-b`. A requires B, and B requires C. (There is a chain.)

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -31,13 +31,14 @@
...
 import org.eclipse.wst.server.core.IModule;
 import org.eclipse.wst.server.core.ServerUtil;
 import org.eclipse.wst.server.core.model.IModuleResource;
 import org.eclipse.wst.server.core.model.ModuleDelegate;
 import org.eclipse.wst.server.core.util.ModuleFile;
-import org.eclipse.wst.server.core.util.PublishHelper;
+import org.eclipse.wst.server.core.util.PublishUtil;
 
 /**
  * Writes a WAR file of a project, or the exploded contents of it to a destination directory.
```
elharo: Looks like it just writes the exploded contents, not the WAR file?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -51,40 +52,40 @@
...
-    publish(project, destination, true /* exploded */, monitor);
+  public static void publishExploded(IProject project, IPath destination,
+      IPath safeWorkDirectory, IProgressMonitor monitor) throws CoreException {
+    publish(project, destination, safeWorkDirectory, true /* exploded */, monitor);
   }
 
-  public static void publishWar(IProject project, IPath destination, IProgressMonitor monitor)
-      throws CoreException {
-    publish(project, destination, false /* exploded */, monitor);
+  public static void publishWar(IProject project, IPath destination, IPath safeWorkDirectory,
```
elharo: This writes the WAR file.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -51,40 +52,40 @@
...
+    Preconditions.checkNotNull(safeWorkDirectory, "safeWorkDirectory is null"); //$NON-NLS-1$
 
     SubMonitor progress = SubMonitor.convert(monitor, 100);
     progress.setTaskName(Messages.getString("task.name.publish.war"));
 
-    PublishHelper publishHelper = new PublishHelper(null);
-    IModuleResource[] resources = flattenResources(project, progress);
+    IModuleResource[] resources = flattenResources(project, safeWorkDirectory, progress);
 
     if (exploded) {
```
elharo: I think you can refactor to avoid the if and reduce the cyclomatic complexity. Return resources from this method and call publishFull or publishExploded from publishWar or publishExploded. Then the exploded argument and the if go away.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -101,27 +102,38 @@ private static void publish(IProject project, IPath destination, boolean explode
...
+        String zipName = destination.lastSegment();  // to be created
+        IPath zipParent = destination.removeLastSegments(1);
+
+        boolean alreadyZip = j2eeModule != null && j2eeModule.isBinary()
+            || utilityModule != null && utilityModule.isBinary();
+        if (alreadyZip) {
+          // per "isBinary()" Javadoc, "members()" should have a single resource.
+          IModuleResource zipResource = childDelegate.members()[0];
+
+          File javaIoFile = zipResource.getAdapter(File.class);
```
elharo: This seems weird. Why would either of these calls work once and not the next time? Or why would one fail and not the other?

chanseokoh: It's the Eclipse API that forces us to check both. `org.eclipse.wst.server.core.util.ModuleFile` with either the Java `File` or the Eclipse `IFile`:
```java
public class ModuleFile implements IModuleFile {
	private IFile file;
	private File file2;
        ...
	public ModuleFile(IFile file, String name, IPath path) { ...
          this.file = file;
          ... }

	public ModuleFile(File file, String name, IPath path) { ...
          this.file2 = file;
          ... }
```
We should be prepared to handle either use case, hence trying both.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
chanseokoh: > Return resources from this method and call publishFull or publishExploded from publishWar or publishExploded.

Will do so.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
chanseokoh: This is not wrong, since the Javadoc is for the class, which has methods to create both a WAR file or an exploded directory.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
chanseokoh: Turns out this method does so little without the `if` statement and it just became `flatternResources` itself. Done.

chanseokoh:
PTAL

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
elharo: Where is IModuleResource.getAdapter documented? I can't find it.

PR #2442, opened 2017-09-30
remove supportsManualCreation="false"

elharo:
@briandealwis @chanseokoh still not sure if this should be added to the server type

Meanwhile this fix lets us show up in the list of runtimes:

image

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/plugin.xml
@@ -8,7 +8,6 @@
             description="%runtimeTypeDescription"
             id="com.google.cloud.tools.eclipse.appengine.standard.runtime"
             name="%apptoolsRuntimeTypeName"
-            supportsManualCreation="false"
             vendor="%runtimeTypeVendor"
             version="1">
          <!-- J2EE Web Modules -->
@@ -28,6 +27,7 @@
          </moduleType>
       </runtimeType>
    </extension>
+ 
    <extension
          point="org.eclipse.wst.server.core.serverTypes">
       <serverType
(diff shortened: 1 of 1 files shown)


briandealwis:
The downside to this is that it allows creating multiple runtimes which isn't actually useful.

I *think* the better approach might be to use the `org.eclipse.ui.startup` extension point to launch a task that creates the runtime.

elharo:
My concern was that supportsManualCreation doesn't seem to be an actual attribute. It isn't documented here, for example:

https://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.servertools.doc.isv%2Fhtml%2Fextpts_reference%2Fservertools%2Forg_eclipse_wst_server_core_runtimeTypes.html

Where do you find it?

elharo:
W.r.t org.eclipse.ui.startup, that would always create a runtime as long as CT4E is installed, right? Is that something we want to do? How heavyweight is this?

elharo:
OK, found supportsManualCreation in the schema at https://git.eclipse.org/c/servertools/webtools.servertools.git/tree/plugins/org.eclipse.wst.server.core/schema/runtimeTypes.exsd so although poorly documented it's not a complete fiction. I can understand not supporting manual creation, but we do need to be able to create a runtime on import if it doesn't already exist.

elharo:
This is a little ugly but maybe `LibraryClasspathContainerSerializer.loadContainer()` could to launch the task that creates the runtime if needed?

briandealwis:
The runtime is pretty lightweight from what I've seen: it's basically just a serialized identifiers.  You can see your runtime definitions in `.metadata/.plugins/org.eclipse.core.runtime/.settings/org.eclipse.wst.server.core.prefs`.  Our definition is:
```
<runtime
    id="App Engine Standard Runtime" name="App Engine Standard Runtime"
    runtime-type-id="com.google.cloud.tools.eclipse.appengine.standard.runtime"
    timestamp="0"/>
```

PR #2445, opened 2017-10-03
Load google-cloud-java libraries from appengine-plugins-core metadata

elharo:
image


elharo:
Looks like we need Require-Bundle instead. No idea why.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
```diff
@@ -27,7 +27,7 @@ Import-Package: com.google.cloud.tools.appengine.cloudsdk.serialization;version=
  com.google.common.cache;version="[20.0.0,21.0.0)",
  com.google.common.collect;version="[20.0.0,21.0.0)",
  com.google.gson;version="[2.7,3.0)",
- javax.json;version="1.1",
+ javax.json;version="1.0.4",
```
briandealwis: org.glassfish.javax.json 1.0.4 exports its packages as version 1.0, not 1.0.4:
```
Bundle-SymbolicName: org.glassfish.javax.json
Bundle-Version: 1.0.4
Export-Package: javax.json;uses:="javax.json.stream,javax.json.spi";version="1.0",
  javax.json.stream;uses:="javax.json";version="1.0",
  javax.json.spi;uses:="javax.json,javax.json.stream";version="1.0"
```

chanseokoh:
No need to check if it's Java 7 or 8? App Engine standard or not?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -64,6 +87,63 @@ public static Library getLibrary(String id) {
...
+          for (JsonObject client : clients.toArray(new JsonObject[0])) {
+            JsonString language = client.getJsonString("language");
+            if (language != null && "java".equals(language.getString())) {
+              String toolTip = client.getString("infotip");
+              library.setToolTip(toolTip);
+              JsonObject coordinates = client.getJsonObject("mavenCoordinates");
+              String groupId = coordinates.getString("groupId");
+              String artifactId = coordinates.getString("artifactId");
+              ArtifactVersion version =
+                  ArtifactRetriever.DEFAULT.getLatestArtifactVersion(groupId, artifactId);
```
chanseokoh: This always tries to use the latest version, which will almost all the time succeed, instead of the versions curated in `libraries.json`. Is this intended?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -64,6 +87,63 @@ public static Library getLibrary(String id) {
     return libraries.get(id);
   }
 
+  private final static Supplier<List<Library>> clientApis = Suppliers.memoize(new Supplier<List<Library>>() {
+    public List<Library> get() {
+      Bundle bundle = FrameworkUtil.getBundle(CloudSdk.class);
+      URL url = bundle.getResource("/com/google/cloud/tools/libraries/libraries.json");
```
briandealwis: An alternative URL is `platform:/plugin/com.google.cloud.tools.appengine/com/google/cloud/tools/libraries/libraries.json`

(I wonder if we should rename that bundle to `.cloudsdk`?)

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
briandealwis: Getting the latest artifact version is potentially quite expensive, and we're going to do this on every library.  This doesn't feel like the right place.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -64,6 +87,63 @@ public static Library getLibrary(String id) {
...
+      
+      try (InputStream in = url.openStream()) {
+        JsonReaderFactory factory = Json.createReaderFactory(null);
+        JsonReader reader = factory.createReader(in);
```
chanseokoh: Turns out we can by-pass the factory and call the static method `Json.createReader(in)` directly.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/META-INF/MANIFEST.MF
```diff
@@ -27,6 +27,7 @@ Import-Package: com.google.cloud.tools.appengine.cloudsdk.serialization;version=
  com.google.common.cache;version="[20.0.0,21.0.0)",
  com.google.common.collect;version="[20.0.0,21.0.0)",
  com.google.gson;version="[2.7,3.0)",
+ javax.json,
```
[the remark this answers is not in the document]
elharo: There is a 1.1 version that doesn't work for us. Added

 javax.json;version="[1.0,1.1)",

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: Does that URL work without  importing the bundle?

Renaming the bundle sounds like a good idea longterm. We've been talking about renaming appengine-plugins-core anyway. If we do this, we should do it as part of a broader rename to avoid confusion from mismatches.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -70,6 +93,63 @@ public static Library getLibrary(String id) {
     return libraries.get(id);
   }
 
+  private final static Supplier<List<Library>> clientApis = Suppliers.memoize(new Supplier<List<Library>>() {
+    public List<Library> get() {
```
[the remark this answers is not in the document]
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: neat. done.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: Yes, it's intended. I'm assuming the metadata will go out of date.

elharo: Where would you suggest doing it? It's only grabbed once and only requested when the first time the user opens the relevant panel. We could add a progress monitor.

elharo: This could be a 30 page conference paper (which I've been threatening to submit for years now; maybe one of these days I'll actually write it); but tldr; GSON is, like most data binding products, brittle and unreliable. We don't want to serialize and unserialize objects. We want to query a data store for specific information we care about. 

Yes, this would work better if we had an actual query language such as XPath or if Java were more weakly typed, but GSON doesn't address either of those issues.

elharo:
It's compiled with Java 8.

--
Elliotte Rusty Harold

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
briandealwis: We'd want it done on demand, no?  The user doesn't need the version number at selection time (if I'm looking for Cloud Datastore, then I won't be interested that Cloud Translate is version 1.22.3453).

briandealwis:
> It's compiled with Java 8.

So we can use `[1,2)` in the MANIFEST.MF, but should make sure our target platform and the 3rdparty.feature pull in 1.0.

elharo:
I suppose we could use [1,2) in the MANIFEST.MF, but what's the reason to do that? More specific version numbers appear safer to me.

I see your point on later loading of the versions. Let me see if I can figure out a reasonable approach for that.

chanseokoh:
NPE when adding Video Surveillance, Data Loss Prevention, or Cloud Spanner. There might be other APIs that cause this.

```
	at com.google.cloud.tools.eclipse.appengine.libraries.model.MavenCoordinates.updateVersion(MavenCoordinates.java:216)
	at com.google.cloud.tools.eclipse.appengine.libraries.model.Library.resolveDependencies(Library.java:184)
	at com.google.cloud.tools.eclipse.appengine.libraries.BuildPath.collectLibraryFiles(BuildPath.java:122)
	at com.google.cloud.tools.eclipse.appengine.libraries.ui.CloudLibrariesPage.getNewContainers(CloudLibrariesPage.java:111)
	at org.eclipse.jdt.internal.ui.wizards.buildpaths.ClasspathContainerWizard.performFinish(ClasspathContainerWizard.java:97)
	at org.eclipse.jdt.internal.ui.wizards.buildpaths.newsourcepage.AddLibraryToBuildpathAction$1.performFinish(AddLibraryToBuildpathAction.java:98)
	at org.eclipse.jface.wizard.WizardDialog.finishPressed(WizardDialog.java:799)
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
```diff
@@ -193,6 +201,22 @@ public Builder setClassifier(String classifier) {
       return this;
     }
     
+  }
+
+  /**
+   * Check Maven Central to find the latest release version of this artifact.
+   * This check is made at most once. Subsequent checks are no-ops.
+   */
+  public void updateVersion() {
```
chanseokoh: `MavenCoordinates` was a lightweight, strictly immutable class, so I don't like adding an update method that much, in addition to the fact that his method makes a long-running network call. I prefer leaving this class as a stateless constant representation. To update a version, you'd want to utilize `.toBuilder()`, which was intended for updating things. That said, if you add a new field, you also need to modify `.toBuilder()`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/testdata/serializedContainer.json
```diff
@@ -32,7 +32,8 @@
         "artifactId": "jarartifact",
         "version": "LATEST",
         "type": "jar"
-      },
+    },
```
chanseokoh: Indent +2.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
```diff
@@ -104,4 +112,21 @@ public int hashCode() {
...
+  void updateVersion() {
+    if (!fixedVersion) {
+      // todo need method to get latest nonrelease version instead for alphas and betas
+      ArtifactVersion remoteVersion = ArtifactRetriever.DEFAULT.getLatestArtifactVersion(
+          mavenCoordinates.getGroupId(), mavenCoordinates.getArtifactId());
+      if (remoteVersion != null) {
+        String updatedVersion = remoteVersion.toString(); 
+        mavenCoordinates = mavenCoordinates.toBuilder().setVersion(updatedVersion).build();
+      }
+      fixedVersion = true;
```
chanseokoh: Mark it fixed even if `remoteVersion` is `null`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/testdata/serializedContainer.json
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFile.java
elharo: Yes. The goal is to check once and only once on first load. After that we don't want the version to change. This is important for serialization and reproducibility.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
```diff
@@ -187,4 +187,19 @@ public void testEquals() {
     assertNotEquals(libraryFile1, null);
     assertNotEquals(libraryFile1, libraryFile1.getMavenCoordinates().toStringCoordinates());
   }
+  
+  @Test
+  public void testUpdateVersion() {
```
briandealwis: It looks like this will cause a network access?  Can that be avoided?

(Or can we label these tests that require network access.  Perhaps even something as simple as `@Category({ NetworkInterface.class })` might work for now?)

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinatesTest.java
```diff
@@ -159,6 +160,7 @@ public void testSetEmptyClassifier() {
         .setArtifactId("a")
         .setClassifier("")
         .build();
-    assertThat(mavenCoordinates.getClassifier(), is(""));
+    assertTrue(mavenCoordinates.getClassifier().isEmpty());
```
briandealwis: Was there a reason for this change?  The original comparison to "" will cause JUnit to report the classifier string value, whereas this change will only report that it was not true.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -80,9 +80,10 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
     
     SubMonitor subMonitor = SubMonitor.convert(monitor,
         Messages.getString("adding.app.engine.libraries"), //$NON-NLS-1$
-        2);
+        3);
     
     Library masterLibrary = collectLibraryFiles(javaProject, libraries);
```
briandealwis: Would it make sense to pass a monitor into `collectLibraryFiles()` and `computeEntries()`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -64,6 +78,53 @@ public static Library getLibrary(String id) {
...
+            library.setLibraryFiles(libraryFiles);
+            library.setResolved(false);
+            break;
+          }
+        }
+        clientApis.add(library);
+      }
+      
+      return clientApis;
+    } catch (IOException ex) {
```
briandealwis: Should we catch `ClassCastException` in case the .json contains invalid data?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
chanseokoh: Maybe a good time to think about a new policy about tests with network connections. I know there are quite a lot of such tests. Put all of them into `.integration.appengine` going forward? Mark with `@Category`? Any other reasonable policy?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: What would throw a `ClassCastException` here? The `readArray` method might throw a `JSONException` though that's already a runtime exception; and would indicate a program bug since the JSON is loaded from the jar archive.

briandealwis: [getString()](http://docs.oracle.com/javaee/7/api/javax/json/JsonObject.html#getString-java.lang.String-) is documented as throwing CCE if the object is not a string.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
briandealwis: We could use `Assume`.

JUnit 5, which has [conditions](http://junit.org/junit5/docs/current/user-guide/#extensions-conditions), or named tags like `@Tag("network")`.  But it requires Java 8… at least for the tests.

PR #2450, opened 2017-10-03
Various minor fixes

chanseokoh:


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -162,12 +162,6 @@ private void fixTestSourceDirectorySettings(IProject newProject, IProgressMonito
     // 2. Remove "src/test/java" from the Web Deployment Assembly sources.
     deployAssemblyEntryRemoveJob =
         new DeployAssemblyEntryRemoveJob(newProject, new Path("src/test/java"));
-    // Not to be affected by "NonSystemJobSuspender". Not necessary in production, but necessary
-    // for tests to work properly. (As a side note, setting this to system would have not worked
-    // if the deploy assembly update jobs triggered by the above classpath change were non-system
-    // jobs; the update jobs should be visible by "DeployAssemblyEntryRemoveJob" because it joins
-    // the update jobs before removing entries.)
-    deployAssemblyEntryRemoveJob.setSystem(true);
```
chanseokoh: We changed `NonSystemJobSuspender` to `ConvertJobSuspender` (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2112), which now suspends only `ConvertJob`s. No need to be a system job anymore.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -34,22 +34,16 @@
...
    * Library files for Google Client APIs for Java; specifically
    * google-api-client, oAuth, and google-http-client.
    */
   public static final String CLIENT_APIS_GROUP = "clientapis"; //$NON-NLS-1$
-  
-  /**
-   * Library files for all Java servlet applications; specifically
-   * servlet.jar and jsp-api.jar.
-   */
-  public static final String SERVLET_GROUP = "servlet"; //$NON-NLS-1$
```
chanseokoh: Now I think this string should match with the value in `plugin.xml`. (See below.) The field is not used in Java code though.
```xml
   <library
          id="servlet-api-2.5"
          name="Servlet API 2.5"
          group="servlet"
```

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
briandealwis: But that means that the job will show up in the Progress view.  I think we should keep it as a system job.

chanseokoh: Ah, I forgot that.

PR #2452, opened 2017-10-04
Ensure App Engine Standard Runtime exists

### LINKED ISSUE #2427: Importing existing project doesn't work: runtime unbound
To repro:

1. Create AE standard project with wizard
2. New workspace
3. Import project

image

briandealwis:
fix #2427 

Leverage an internal WTP extension point, `org.eclipse.wst.server.core.internalStartup`, to ensure an instance of our _App Engine Standard Runtime_ runtime type is always present.

We use `App Engine Standard Runtime` as the runtime ID to match the runtime that was previously created by `AppEngineStandardFacet#createAppEngineServerRuntime()`: this method creates the runtime using a `null` ID, which causes it to take the _name_ of its runtime-type.  This doesn't seem to have any negative effect, so I've continued with it.

I added some integration tests to explicitly import AES projects for Java 7 and Java 8 using CT4E 1.3.1 and to verify that there are no build errors.  As the _Create New AES Projects_ wizards call `AppEngineStandardFacet#createAppEngineServerRuntime()` and thus create a runtime, the ITs specifically run the `ImportNativeAppEngineStandardProjectTest` separately from a fresh workspace to ensure no runtime exists.


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/RegisterAppEngineStandardRuntimeStartup.java
```diff
@@ -0,0 +1,49 @@
+
```
elharo: Copyright

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/RegisterAppEngineStandardRuntimeStartup.java
```diff
@@ -0,0 +1,49 @@
...
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.wst.server.core.IRuntime;
+import org.eclipse.wst.server.core.IRuntimeType;
+import org.eclipse.wst.server.core.IRuntimeWorkingCopy;
+import org.eclipse.wst.server.core.ServerCore;
+import org.eclipse.wst.server.core.internal.IStartup;
+
+/**
+ * Registers the default instance of our App Engine Standard runtime if not found. Uses the internal
+ * WTP {@link IStartup} to minimize activation to use of WTP-related functionality.
```
elharo: delete "to use"?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/RegisterAppEngineStandardRuntimeStartup.java
```diff
@@ -0,0 +1,49 @@
...
+   * AppEngineStandardFacet#createAppEngineServerRuntime() creates the default runtime using a
+   * {@code null} ID, which causes it to take the name of its runtime-type.
+   */
+  private static final String RUNTIME_ID = "App Engine Standard Runtime";
+
+  @Override
+  public void startup() {
+    IRuntime runtime = ServerCore.findRuntime(RUNTIME_ID);
+    if (runtime != null) {
+      logger.info("Runtime found: " + RUNTIME_ID);
```
elharo: Logging not really needed in this case.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/RegisterAppEngineStandardRuntimeStartup.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * AppEngineStandardFacet#createAppEngineServerRuntime() creates the default runtime using a
+   * {@code null} ID, which causes it to take the name of its runtime-type.
+   */
+  private static final String RUNTIME_ID = "App Engine Standard Runtime";
+
+  @Override
+  public void startup() {
+    IRuntime runtime = ServerCore.findRuntime(RUNTIME_ID);
```
chanseokoh: Why don't we just call the `static` method `createAppEngineServerRuntime()`? The code looks very similar. Even if there is some difference (I don't see a major one) here, I think it's better to have a single code place to create the runtime.

briandealwis: I do remember considering that… but I don't know why I didn't follow through.

briandealwis:
Filed [bug 525582](https://bugs.eclipse.org/bugs/show_bug.cgi?id=525582) re: exposing `internalStartup` as API.

PR #2453, opened 2017-10-04
Move SDK into runtime

### LINKED ISSUE #2350: GCS client library requires the App Engine API library
and it's not added automatically. I.e. it's not listed as a dependency in plugin.xml. Typical symptoms are when running locally

```
HTTP ERROR 500

Problem accessing /hello. Reason:

    Server Error
Caused by:

java.lang.RuntimeException: Google App Engine runtime detected (the environment variable "com.google.appengine.runtime.version" is set), but unable to resolve appengine-sdk classes. For more details see https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/APPENGINE.md
	at com.google.cloud.ServiceOptions.getAppEngineProjectId(ServiceOptions.java:398)
	at com.google.cloud.ServiceOptions.getDefaultProjectId(ServiceOptions.java:306)
	at com.google.cloud.ServiceOptions.getDefaultProject(ServiceOptions.java:284)
	at com.google.cloud.ServiceOptions.<init>(ServiceOptions.java:238)
	at com.google.cloud.storage.StorageOptions.<init>(StorageOptions.java:82)
	at com.google.cloud.storage.StorageOptions.<init>(StorageOptions.java:30)
	at com.google.cloud.storage.StorageOptions$Builder.build(StorageOptions.java:77)
	at com.google.cloud.storage.StorageOptions.getDefaultInstance(StorageOptions.java:128)
	at HelloAppEngine.doGet(HelloAppEngine.java:22)
```

elharo:
@briandealwis @chanseokoh This will be necessary since the Maven graph doesn't include the SDK as a dependency (nor should it). This is probably how we should have fixed #2350 in the first place.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -9,7 +9,7 @@
              
     <library 
           id="appengine-api"
-          group="appengine"
+          group="appengine-api"
           name="App Engine API"
           tooltip="%appengine.api.tooltip"
           siteUri="https://cloud.google.com/appengine/docs/java/" >
@@ -32,7 +32,6 @@
           tooltip="%endpoints.tooltip"
           dependencies="include"
           siteUri="https://cloud.google.com/endpoints/docs/frameworks/java/about-cloud-endpoints-frameworks" >
-      <libraryDependency id="appengine-api" />
       <libraryFile export="true" 
             javadocUri="https://cloud.google.com/endpoints/docs/frameworks/java/javadoc/">
         <mavenCoordinates
@@ -48,7 +47,6 @@
           name="Objectify"
           tooltip="%objectify.tooltip"
           siteUri="https://github.com/objectify/objectify/wiki" >
-      <libraryDependency id="appengine-api" />
       <libraryFile
           export="true"
           javadocUri="https://www.javadoc.io/doc/com.googlecode.objectify/objectify/5.1.21">
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -14,7 +14,7 @@
           tooltip="%appengine.api.tooltip"
           siteUri="https://cloud.google.com/appengine/docs/java/" >
       <libraryFile javadocUri="https://cloud.google.com/appengine/docs/java/javadoc/"
-          export="true">
+          export="false">
         <mavenCoordinates
               artifactId="appengine-api-1.0-sdk"
               groupId="com.google.appengine" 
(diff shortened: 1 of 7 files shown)


chanseokoh:
I am not so confident, but I get the feeling that it makes sense to include the App Engine API as part of the App Engine Runtime container. I believe the API is needed only locally and for App Engine, right? So having that by default in App Engine projects makes sense to me.

That said, to be consistent with Maven projects, we should add the API in `pom.xml` too (probably with the scope of `provided`)?

Lastly, I suggest to update the following Javadoc to include `appengine-api.jar`. BTW, does the name `ServletClasspathProvider` sound OK?
```java
/**
 * Supply Java servlet container classes, specifically servlet-api.jar and jsp-api.jar,
 * to non-Maven projects.
 * <p>
 * The jars are resolved using {@link ILibraryRepositoryService}.
 */
public class ServletClasspathProvider extends RuntimeClasspathProviderDelegate {
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
```diff
@@ -241,8 +241,16 @@ private IClasspathEntry resolveLibraryFileAttachSourceSync(LibraryFile libraryFi
 
     Artifact artifact = repositoryService.resolveArtifact(libraryFile, new NullProgressMonitor());
     IPath libraryPath = new Path(artifact.getFile().getAbsolutePath());
-    IPath sourceAttachmentPath = repositoryService.resolveSourceArtifact(libraryFile,
-        artifact.getVersion(), new NullProgressMonitor());
+    
+    // Not all artifacts have sources; need to work if no source arrtifact is available
```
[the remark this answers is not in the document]
elharo: done

elharo:
It may be needed remotely too, but like the servlet APIs it is automatically provided there.

@briandealwis What do you mean by the App Engine APIs library provider? the ApiClientLibrariesPage and CloudLibraries.CLIENT_APIS_GROUP?

elharo:
In the case of Maven, we do need to do something. Not sure if we should add it to the runtime or the pom.xml.

briandealwis:
The [appengine-api](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/625138aef81eb56494bfe7f02b8cedbb42aa6f49/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml#L11) library definition should be removed, and its associated dependencies.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -9,7 +9,7 @@
              
     <library 
           id="appengine-api"
-          group="appengine"
+          group="appengine-api"
           name="App Engine API"
           tooltip="%appengine.api.tooltip"
           siteUri="https://cloud.google.com/appengine/docs/java/" >
```
chanseokoh: Below here is the `export` attribute. It makes sense to change it to `false`.
```xml
       <libraryFile javadocUri="https://cloud.google.com/appengine/docs/java/javadoc/"
           export="true">
```

elharo: done

PR #2458, opened 2017-10-04
Fix SnakeYaml bundle version

chanseokoh:
Eclipse correctly shows a warning that it can't find the version:

selection_002

The Maven coordinate version is "1.17", but the bundle version is "1.17.0". The bundle's `MANIFEST.MF` says it:
```
Bundle-Description: YAML 1.1 parser and emitter for Java
...
Bundle-Name: SnakeYAML
...
Bundle-SymbolicName: org.yaml.snakeyaml
Bundle-Version: 1.17.0
Created-By: Apache Maven Bundle Plugin
```

### DIFF
--- a/eclipse/ide-target-platform/category.xml
+++ b/eclipse/ide-target-platform/category.xml
@@ -37,5 +37,5 @@
    <bundle id="com.fasterxml.jackson.core.jackson-core" version="0.0.0"/>
    <bundle id="ch.qos.logback.slf4j" version="0.0.0"/>
    <bundle id="org.slf4j.log4j" version="0.0.0"/>
-   <bundle id="org.yaml.snakeyaml" version="1.17"/>
+   <bundle id="org.yaml.snakeyaml" version="1.17.0"/>
 </site>
--- a/features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml
+++ b/features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml
@@ -78,7 +78,7 @@
          id="org.yaml.snakeyaml"
          download-size="0"
          install-size="0"
-         version="1.17"
+         version="1.17.0"
          unpack="false"/>
 
 </feature>


chanseokoh:
Talked to @elharo. Things have been working fine, so we will avoid changing things. Closing.

chanseokoh:
Also, it may have been the case that there was a reason for this.

eclipse/ide-target-platform/category.xml
```diff
@@ -37,5 +37,5 @@
    <bundle id="com.fasterxml.jackson.core.jackson-core" version="0.0.0"/>
    <bundle id="ch.qos.logback.slf4j" version="0.0.0"/>
    <bundle id="org.slf4j.log4j" version="0.0.0"/>
-   <bundle id="org.yaml.snakeyaml" version="1.17"/>
+   <bundle id="org.yaml.snakeyaml" version="1.17.0"/>
```
briandealwis: Should just use 0.0.0 so as to match the version defined in the target platform.  Version 0.0.0 will be replaced by Tycho to that matched at build time.  We only need to specify an explicit version if there may be multiple bundles resolved.

features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml
```diff
@@ -78,7 +78,7 @@
          id="org.yaml.snakeyaml"
          download-size="0"
          install-size="0"
-         version="1.17"
+         version="1.17.0"
```
briandealwis: 0.0.0

briandealwis:
The culprit is PDE's [`FeatureErrorReporter`](http://git.eclipse.org/c/pde/eclipse.pde.ui.git/tree/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/builders/FeatureErrorReporter.java?h=master#n463) that performs exact string comparisons on versions.  This decision can make sense — a feature specifies a particular version of a plugin (if not using `0.0.0`), so presumably you should specify the exact version here.

briandealwis:
In the interest of eliminating warnings, I'm +1 to this change.

Issue #2456, opened 2017-10-04
Java 8 pom.xml doesn't depend on JSP API


elharo:
Non-maven projects have this in the runtime. 

```
  <dependencies>
    <!-- Compile/runtime dependencies -->
    <dependency>
      <groupId>javax.servlet</groupId>
      <artifactId>javax.servlet-api</artifactId>
      <version>3.1.0</version>
      <scope>provided</scope>
    </dependency>
    <dependency>
      <groupId>jstl</groupId>
      <artifactId>jstl</artifactId>
      <version>1.2</version>
    </dependency>
```

chanseokoh:
Flex WAR projects don't have it either.

chanseokoh:
Java 7 standard Maven doesn't have it either.

patflynn:
What's the issue exactly?

chanseokoh:
Some projects created by our wizards have the JSP JAR (jsp-api-2.1.jar for Java 7 and javax.servlet.jsp-api-2.3.0.jar for Java 8), while some projects don't.

elharo:
And it's needed for a project that uses JSPs, which many do. The server runtime supplies this jar on the classpath automatically, but you can't compile JSPs or run them locally without it.

chanseokoh:
@briandealwis would you fix #2455 at the same time?

PR #2465, opened 2017-10-06
Add/Update JSP, JSTL, EL as required

### LINKED ISSUE #2455: Add JSTL when creating App Engine flex native project
For the _native_ flex project, we add only the servlet JAR. However, we add JSTL to Maven flex projects and all standard projects.

Fixing it seems simple; just add JSTL to the `CreateAppEngineFlexWtpProject.PROJECT_DEPENDENCIES` list.
```java
  private static final List<MavenCoordinates> PROJECT_DEPENDENCIES;

  static {
    MavenCoordinates.Builder servletApi = new MavenCoordinates.Builder()
        .setGroupId("javax.servlet") //$NON-NLS-1$
        .setArtifactId("javax.servlet-api") //$NON-NLS-1$
        .setVersion("3.1.0"); //$NON-NLS-1$
    PROJECT_DEPENDENCIES = Collections.singletonList(servletApi.build());
  }
```

briandealwis:
Fixes #2455 and #2456 

Adds / update dependencies for JSP, JSTL, and EL.  The situation of whether to include JSP, JSTL, EL implementations is a bit confusing.  It has an impact as the _servlet classpath container_ is used to host the jars that are provided by the servlet container on the server.  I further simplified the `ServletClasspathContainer` to pull in a single `<library>` definition with required API definitions for each servlet runtime rather than pull in the individual libraries.

## Note on Including API Implementations

The StackOverflow [`jstl` tag info](https://stackoverflow.com/tags/jstl/info) notes that a `jstl` implementation should only be installed if the servlet container does not already provide it. JSTL 1.2 has been around since Servlet 2.5 (JavaEE 5) and is still the current release for Servlet 3.1 (JavaEE 7).  I further found a strong recommendation on SO to [_Never carry around loose server-specific JAR files_ like servlet-api, jsp-api, jstl-api, and el-api](https://stackoverflow.com/questions/4076601/how-do-i-import-the-javax-servlet-api-in-my-eclipse-project).

The StackOverflow [`el` tag info](https://stackoverflow.com/tags/el/info) notes:
> Before JSP 2.0 (released Nov 2003), EL was part of JSTL. Since JSP 2.0, standard EL was moved from JSTL to JSP and maintained as part of the JSP specification.

So as long as we have JSP and JSTL implementations (which we do: below), we should have EL implementation too.

## App Engine Standard for Java 7 (Servlet 2.5)

The [docs on configuring JSPs on App Engine standard](https://cloud.google.com/appengine/docs/standard/java/config/webxml#JSPs) say:
> Note that App Engine provides the JavaServer Pages Standard Tag Library (JSTL), and you do not need to install it.

Servlet 2.5 (JavaEE 5) is compatible with JSP 2.1, EL 2.1 (bundled with JSP 2.1), JSTL 1.2. (The best source I could find was the [Wikipedia page for Java EE](https://en.wikipedia.org/wiki/Java_EE_version_history#Java_EE_5_.28May_11.2C_2006.29).)

## App Engine Standard for Java 8 (Servlet 3.1)

There is no separate documentation for AESv8 on JSPs, and experiments show that the `java8` runtime does include a JSTL implementation, though the devappserver warns:
```
Oct 05, 2017 8:37:37 PM com.google.appengine.tools.development.jetty9.JettyContainerService initContext
WARNING: Adding jar org.apache.taglibs.taglibs-standard-spec-1.2.5.jar to WEB-INF/lib.You might want to add a dependency in your project build system to avoid this warning.
```

Servlet 3.1 (JavaEE 7) is compatible with JSP 2.3, EL 3.0, and JSTL 1.2.  JavaEE 7 provides an artifact on Maven Central ([javax : javaee-web-api : 7.0](http://search.maven.org/#artifactdetails%7Cjavax%7Cjavaee-web-api%7C7.0%7Cjar)) which lists explicit dependencies as `jsp-api` 2.3.1, `jstl-api` 1.2.1, and `el-api` 3.0.0.

## App Engine Flexible (Servlet 3.1)

The [docs on configuring JSPs on App Engine flexible](https://cloud.google.com/appengine/docs/flexible/java/configuring-the-web-xml-deployment-descriptor#JSPs) says:
> Note that App Engine provides the JavaServer Pages Standard Tag Library (JSTL), and you do not need to install it.

And the docs specifically about the [Jetty 9 environment](https://cloud.google.com/appengine/docs/flexible/java/dev-jetty9) confirm:
> Jetty 9 which App Engine flexible environment uses as its Servlet container uses Apache Jasper as the default JSP implementation and includes JSTL taglibs.

As noted above, Servlet 3.1 (JavaEE 7) is compatible with JSP 2.3, EL 3.0, and JSTL 1.2, and we use the same API artifact versions.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -66,8 +66,8 @@
     </library>
 
     <library
-          id="servlet-api-2.5"
-          name="Servlet API 2.5"
+          id="appengine-servlet-2.5"
+          name="App Engine: Servlet API 2.5"
```
elharo: Why change the name? It's not app engine specific.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -78,14 +78,6 @@
               artifactId="servlet-api"
               version="2.5" />
       </libraryFile>
-    </library>
-
-    <library
```
elharo: My recollection is that JSPs didn't work locally until we added this.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
```diff
@@ -86,38 +86,36 @@
...
-          id="servlet-api-3.1"
-          name="Servlet API 3.1"
+          id="appengine-servlet-3.1"
+          name="App Engine: Servlet API 3.1"
           group="servlet"
           siteUri="http://www.oracle.com/technetwork/java/javaee/servlet/index.html"
           export="false"
           recommendation="required">
-    
+      <libraryDepency id="appengine-api"/>
```
elharo: libraryDependency (and suggests a test is missing)

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
```diff
@@ -86,38 +86,36 @@
...
-          name="Java Server Pages API 2.3"
-          siteUri="http://www.oracle.com/technetwork/java/javaee/jsp/index.html"
-          export="false"
-          recommendation="optional" >
       <libraryFile>
         <mavenCoordinates
               groupId="javax.servlet.jsp"
               artifactId="javax.servlet.jsp-api"
-              version="2.3.0" />
+              version="2.3.1" />
```
elharo: What does App Engine provide? That's what we need to match.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
briandealwis: The intent is for this library to be a container of dependencies.  Maybe `aes-jre7` / _App Engine standard environment for Java 7_?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
briandealwis: I haven't removed it: I've combined it into the `appengine-servlet-2.5 ` container.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
```diff
@@ -66,11 +66,31 @@
   private static final List<MavenCoordinates> PROJECT_DEPENDENCIES;
 
   static {
-    MavenCoordinates.Builder servletApi = new MavenCoordinates.Builder()
+    // matches the dependencies in javax:javaee-web-api:7.0
+    // FIXME: include javax.el:javax.el-api:3.0.0?
```
briandealwis: Oops, need to drop this.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
briandealwis: I was thinking that after I packed up last night: it would be good to have integration tests that verified JSP, JSTL, EL, and the App Engine SDK.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
briandealwis: I originally had combined the Servlet, JSP, and JSTL+EL deps into `<library>` elements for a Servlet 2.5 container, and a Servlet 3.1 container, and kept the `appengine-api` separate with them all included by the `ServletClasspathContainer`.  But this didn't make sense since it's up to the actual container implementation as to whether it includes implementations of JSP, JSTL, EL.

(Aside, perhaps it shouldn't the library definitions that decide whether they should be 'exported' or not.)

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml
chanseokoh: typo still not fixed

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -66,26 +66,19 @@
...
 
     <library
-          id="servlet-api-2.5"
-          name="Servlet API 2.5"
+          id="appengine-servlet-2.5"
+          name="App Engine: Servlet API 2.5"
           group="servlet"
           siteUri="http://www.oracle.com/technetwork/java/javaee/servlet/index.html"
           export="false">
+      <libraryDependency id="appengine-api"/>
```
elharo: I had planned to get rid of inter-library dependencies now that we're using Maven. Are we still going to need this?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
elharo: Interesting. Where should we indicate whether they're exported?

chanseokoh: If we name it "App Engine: Servlet API 2.5" (as it stands in the PR) implying this `<library>` is specifically for App Engine Java 7, I think we are good; no need to think about if JSTL+EL is supported in a different container or if they are exported.

briandealwis: These libraries are provided by the servlet container, so they don't need to be exported, and they won't be included in the resulting .war.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
briandealwis: The App Engine SDK is provided (hence here).  Prior to this change, `ServletClasspathContainer` determined the corresponding runtime and [explicitly listed the individual library jars to be included](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2465/files#diff-046dccbf1bc4d935129f833a7579b5a5L101).  Now `ServletClasspathContainer` just includes a single `<library>` reference that corresponds to the Servlet API for the runtime.

chanseokoh:
> These libraries are provided by the servlet container, so they don't need to be exported, and they won't be included in the resulting .war.

That might or might not be true. I found some interesting information. It looks like it is our staging (appcfg.sh) that supplies JSTL and EL JARs. (Tested with a AES Java 7 Maven project.)
```
└── WEB-INF
    ...
    ├── lib
    │   ├── _ah_compiled_jsps-0000.jar
    │   ├── repackaged-appengine-ant-1.9.4.jar
    │   ├── repackaged-appengine-ant-launcher-1.9.4.jar
    │   ├── repackaged-appengine-eclipse-jdt-ecj.jar
    │   ├── repackaged-appengine-jakarta-jstl-1.1.2.jar
    │   ├── repackaged-appengine-jakarta-standard-1.1.2.jar
    │   ├── repackaged-appengine-jasper-6.0.29.jar
    │   ├── repackaged-appengine-jasper-el-6.0.29.jar
    │   └── repackaged-appengine-tomcat-juli-6.0.29.jar
    └── web.xml
```
On old SO answer says that [Tomcat has never included JSTL](https://stackoverflow.com/a/6094605/1701388). Jetty may include it accordingly to its doc, but I can't be certain. In that case, the staging may not need to supply JARs, but perhaps it's doing so to force using the supplied JSTL and EL implementations to have maximum portability?

chanseokoh:
@meltsufin what do you think of https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2465#issuecomment-335519526?

chanseokoh:
For AES Java 8 projects, I find that staging puts different versions and implementations for JSTL than Java 7 projects.
```
    ├── lib
    │   ├── _ah_compiled_jsps-0000.jar
    │   ├── org.apache.taglibs.taglibs-standard-impl-1.2.5.jar
    │   ├── org.apache.taglibs.taglibs-standard-spec-1.2.5.jar
    │   ├── org.eclipse.jdt.core.compiler.ecj-4.4.2.jar
    │   ├── org.eclipse.jetty.apache-jsp-9.3.18.v20170406-nolog.jar
    │   ├── org.mortbay.jasper.apache-el-8.0.33.jar
    │   └── org.mortbay.jasper.apache-jsp-8.0.33.jar
    ├── min-quickstart-web.xml
    ├── quickstart-web.xml
    └── web.xml
```

patflynn:
I believe it's right that inclusion of JSP related libraries is done at
staging time based on the presence of jsp files.   The runtime doesn't
automatically have them, although it appears to due to the staging logic.

briandealwis:
@ludoch's recommendation is that JSTL should be user-defined.  For App Engine standard, they added JSTL when compiling JSPs locally so that apps would compile before deploying.  But there is no local compilation on flex.

So JSTL + EL should not be part of our provided libraries in the servlet container.  So two options for our wizards:
  1. install jstl:jstl:1.2 (say) into WEB-INF/lib (for native projects) or add a `<dependency>` to the pom (for Maven), to allow user replacement;
  2. don't add JSTL at all, and let the user find and add a version.

patflynn:
does staging not do anything if it detects an existing jstl library
preconfigured?

chanseokoh:
@patflynn interestingly, if I add JSTL to AES Java 7 with the following dependency
```xml
    <dependency>
      <groupId>javax.servlet</groupId>
      <artifactId>jstl</artifactId>
      <version>1.2</version>
    </dependency>
```
staging ends up having both `jstl-1.2.jar` and the repackaged `jakarta-jstl-1.1.2.jar`.
```
    ├── lib
    │   ├── _ah_compiled_jsps-0000.jar
    │   ├── jstl-1.2.jar
    │   ├── repackaged-appengine-ant-1.9.4.jar
    │   ├── repackaged-appengine-ant-launcher-1.9.4.jar
    │   ├── repackaged-appengine-eclipse-jdt-ecj.jar
    │   ├── repackaged-appengine-jakarta-jstl-1.1.2.jar
    │   ├── repackaged-appengine-jakarta-standard-1.1.2.jar
    │   ├── repackaged-appengine-jasper-6.0.29.jar
    │   ├── repackaged-appengine-jasper-el-6.0.29.jar
    │   └── repackaged-appengine-tomcat-juli-6.0.29.jar
    └── web.xml
```

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.standard.ftl
```diff
@@ -25,28 +25,57 @@
...
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet.jsp</groupId>
+      <artifactId>jsp-api</artifactId>
+      <version>2.1</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet.jsp.jstl</groupId>
+      <artifactId>jstl-api</artifactId>
```
chanseokoh: It looks like this is purely API interfaces, like `servlet-api`. I see errors when trying to use JSTL.
selection_005
Maybe we need to provide an implementation unlike `servlet-api`. But it seems like we anyway want to provide an implementation: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2465#issuecomment-335545290

patflynn:
@chanseokoh 

For standard I don't think we should be providing an implementation since I thought we established that it's already provided through staging.

For Flex I guess we need to get an answer from @meltsufin whether we should or not

elharo:
Code in an Eclipse project often runs inside the IDE before and without staging, most commonly in unit tests. I'm trying to wrap my head around how this could possibly work without including impl classes in the Eclipse project.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
```diff
@@ -66,11 +66,30 @@
   private static final List<MavenCoordinates> PROJECT_DEPENDENCIES;
 
   static {
-    MavenCoordinates.Builder servletApi = new MavenCoordinates.Builder()
+    // matches the dependencies in javax:javaee-web-api:7.0
+    //@formatter:off
+    MavenCoordinates servletApi = new MavenCoordinates.Builder()
```
elharo: could/should these be pulled from plugins/com.google.cloud.tools.eclipse.appengine.standard.java8/plugin.xml? If so, that can probably wait for a future PR though.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.standard.ftl
chanseokoh: The exact error message is

Can not find the tag library descriptor for "http://java.sun.com/jsp/jstl/core" | index.jsp | JSP Problem

chanseokoh: To clarify, if we use `<groupId>jstl</groupId>` and `<artifactId>jstl</artifactId>` (as we currently do), the error goes away. (BTW, the group ID should be `javax.servlet` rather than `jstl`.)

briandealwis: Aside: the [StackOverflow JSTL wiki](https://stackoverflow.com/tags/jstl/info) notes that 
> Note that there's also a `jstl:jstl` dependency, but it's exactly the same file, only with a wrong group ID. Furthermore, there's also a `javax.servlet.jsp.jstl:jstl` dependency, but it is empty.

briandealwis:
> @patflynn: For standard I don't think we should be providing an implementation since I thought we established that it's already provided through staging.

From Ludo's comments, we should be providing an implementation.  That App Engine standard had an was an implemention detail that is now regretted.

At this point, I'll rework this code to (1) remove JSTL & EL from the servlet container definitions, and (2) add `javax.servlet:jstl:1.2` (Apache JSTL implementation as a standalone jar) to `WEB-INF/lib` (native) or `pom.xml` (standard).

patflynn:
@ludoch can you comment on this?   It may be regretted nevertheless staging still seems to copy it in.  We don't want to include two versions do we?

briandealwis:
[`javax.jstl:jstl:1.2`](http://search.maven.org/#artifactdetails%7Cjavax.servlet%7Cjstl%7C1.2%7Cjar) unfortunately bundles Apache Standard Taglib 1.2.0, whereas [the most recent version is 1.2.5](https://tomcat.apache.org/taglibs/standard/).
```sh
$  java -cp jstl-1.2.jar org.apache.taglibs.standard.Version
standard-taglib 1.2.0
```
Dunno whether to `/shrug` or instead pull in the `javax.jstl:jstl-api:1.2.0` for the API spec, and [`org.apache.taglibs:taglibs-standard-impl:1.2.5`](http://search.maven.org/#artifactdetails%7Corg.apache.taglibs%7Ctaglibs-standard-impl%7C1.2.5%7Cbundle).

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
briandealwis: Not in `.standard.java8` at least.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
briandealwis: `<libraryDependency>` is useful to reference other `<library>` definitions, especially `appengine-api` since we have some nice package exclusion/inclusion rules to avoid surfacing repackaged/shaded types from `com.google.appengine.**.repackaged` and internal types from `com.google.apphosting`.

Issue #2327, opened 2017-08-25
Automatic focus on "Console" view when a new log is shown


NicolaSpreafico:
(please ensure you are running the latest version of CT4E with _Help > Check for Updates_.)

- Eclipse version
> Version: Oxygen Release (4.7.0)
> Build id: 20170620-1800

- Cloud Tools for Eclipse version:
> Google Cloud Tools for Eclipse
> Version: 1.3.0.201708031339

- Google Cloud SDK version: _(run `gcloud version`)_

> Google Cloud SDK 165.0.0
> app-engine-java 1.9.54
> app-engine-python 1.9.57
> beta 2017.03.24
> bq 2.0.25
> core 2017.07.28
> gcloud
> gsutil 4.27

- OS:
> Microsoft Windows version 1703 (build SO 15063.540)

- Java version:
> java version "1.8.0_111"
> Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
> Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)

With the previous plugin for Eclipse (Google  Eclipse Plugin) the Console view automatically focused with a new log line.

Now, with CT4E, it simply goes bold but it doesn't focus
![image](https://user-images.githubusercontent.com/12278553/29703857-b037ae02-8977-11e7-900a-a9b21855c673.png)

In the Eclipse toolbar of Console view I do not see a configuration like "focus when content change", so I'm wondering if the configuration I'm looking for was related to Google Eclipse Plugin itself and not Eclipse IDE.
![image](https://user-images.githubusercontent.com/12278553/29703887-d15d32f0-8977-11e7-8f23-bd5e017481ef.png)

Can this behaviour be accomplished with the new CT4E?

elharo:
Can you explain what you mean by focused in this context? Simply come to the front ahead of other panes?

NicolaSpreafico:
Yes, correct, the tab come to front, active, selected, ...

briandealwis:
That's an Eclipse preference: _Preferences > Run/Debug > Console_:

<img width="506" alt="show console" src="https://user-images.githubusercontent.com/202851/29719316-941f3756-8983-11e7-95ed-f4e15e7972db.PNG">

briandealwis:
I don't see any code in GPE that activates the console.

NicolaSpreafico:
[...]

Issue #2140, opened 2017-07-12
Stacktrace hyperlinks in console output


NicolaSpreafico:
I noticed a different handling of Eclipse dev-server logs (Console view).
Of course the color is not the main theme here, but I'm referring to the stacktrace hyperlinks.

Here is my configuration (the 2 plugins has beend installed separately in order to do the first and the second test)

> **Google Plugin for Eclipse**:
> _Version: 3.9.6.v20170329-1604_
> 
> **Cloud Tools for Eclipse**:
> _Version: 1.2.0.201706082053_
> 
> **Eclipse**:
> _Eclipse Java EE IDE for Web Developers.
> Version: Neon.3 Release (4.6.3)
> Build id: 20170314-1500_
> 
> **OS**:
> _Windows 10 ver 1703 (build SO 15063.413)_
> 
> **Java version**:
> _Java version "1.8.0_111"
> Java(TM) SE Runtime Environment (build 1.8.0_111-b14)
> Java HotSpot(TM) 64-Bit Server VM (build 25.111-b14, mixed mode)_

As test project I created a brand new using `appengine-standard-archetype`.
I added a simple code which thrown an exception and print the stacktrace in order to see the difference
```
Logger.getLogger(HelloAppEngine.class.getName()).log(Level.SEVERE, null, new Throwable());
```

This is the dev-server launched with CT4E, all the logs are in black-color, only the urls are hyperlinked and all the stacktrace classes are not enhanced
![image](https://user-images.githubusercontent.com/12278553/28107197-02d6c82c-66e7-11e7-92c2-7834bf276399.png)

This is the dev-server launched with the (deprecated?) Google Plugin for Eclipse, all the logs are in red-color, the urls are NOT hyperlinked (not a big problem) but all the classes are hyperlinked to the source
![image](https://user-images.githubusercontent.com/12278553/28107320-93ec4a1c-66e7-11e7-9753-1d560a7eeb5e.png)

During local debugging having a link directly to the stacktrace can be very handy, is there a way to introduce this kind of behaviour even with the new CT4E?

Thanks

briandealwis:
Coloring the stderr output is #1248.

briandealwis:
`org.eclipse.jdt.debug.ui` provides several console pattern matchers for pulling out hyperlinks to java stack traces and file paths.

```xml
   <extension point="org.eclipse.ui.console.consolePatternMatchListeners">
     […omitted…]
[...]

PR #2471, opened 2017-10-10
Bring deploy console to front (again) when actual deploy starts

chanseokoh:
Takes care of #2327 but for the deploy console only (i.e., not about the local run console).

We are already bringing the console to the front when `DeployJob` starts, so this weren't necessary if it were not for the new Maven JAR deploy flow. This new flow opens up a Maven launch console (for staging with `mvn package`) first, so we need to bring back the deploy console to the front again when the staging is done.

~~I decided to do this in the deployer class.~~


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
```diff
@@ -160,9 +161,10 @@ private void launchDeployJob(IProject project, Credential credential)
     errorStream.setColor(colorProvider.getColor(IDebugUIConstants.ID_STANDARD_ERROR_STREAM));
```
briandealwis: Wouldn't this location be the better place to raise the console?

chanseokoh: It is doing so already. See a few lines above.

briandealwis: Then why not just use the same approach as #2346 and use `setActivateOnWrite()`?

chanseokoh: Oh, that's the way we should do it. Thanks for the tip.

chanseokoh:
Awesome, `setActivateOnWrite(true)` works very well.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
briandealwis: I'd like to at least consult the user's preference as to whether they want the raise-on-write for stdout and stderr.  I personally loathe this behaviour and so have it turned off.

chanseokoh: There is one conceptual problem. As you can see in [this screenshot](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2327#issuecomment-324944085), the preference you are referring to is under the hierarchy of _Run/Debug_ > _Console_. The settings there are meant for consoles from run/debug launch configuration. So, conceptually, applying the settings globally to every console is actually out of the intended scope (although it's not terribly unreasonable). Our console is custom-created programmatically, not from run/debug, so it feels a bit weird and surprising to find that the run/debug preferences apply to a console of a custom job. Likewise, that's the not only settings in that hierarchy. We don't honor, e.g., _Limit console output_. Should our custom console be committed to support all the existing/future preferences of run/debug (rhetorical question)? It's not difficult to check the preferences and apply it, but it irks me a bit because of this reason.

That said, to be conceptually correct, I think we should have our own preferences panel for this.

In any case, can we consider such support as enhancement to be implemented later? If you do not agree, I think it's better to we go back to the original PR to show the console just once more when the deploy starts, as a stop-gap.

briandealwis: Run/Debug is the UI labels for the launch framework, and include consoles, etc.

I'm fine with this for now; if we can cast the process as an `IProcess` then [we'll get this for free](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2140#issuecomment-326387162).

chanseokoh: > Run/Debug is the UI labels for the launch framework, and include consoles, etc.

Ah, I see.

PR #2474, opened 2017-10-11
don't show GRPC libraries for App Engine Standard Java 7 projects

elharo:
@chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
@@ -38,13 +38,26 @@ public class CloudLibrariesTest {
   public void testGetClientApis() {
     List<Library> libraries = CloudLibraries.getLibraries("clientapis");
     Assert.assertTrue(libraries.size() > 10);
+    
+    Library datalossPrevention = null;
+    Library cloudStorage = null;
+    
     for (Library library : libraries) {
       String tooltip = library.getToolTip();
+      if ("googlecloudstorage".equals(library.getId())) {
+        cloudStorage = library;
+      } else if ("datalossprevention".equals(library.getId())) {
+        datalossPrevention = library;
+      }
+      
       Assert.assertNotNull(library.getName() + " has no tooltip", tooltip);
       Assert.assertFalse(tooltip.isEmpty());
       Assert.assertFalse(library.getName() + " has no files", library.getLibraryFiles().isEmpty());
       Assert.assertEquals("clientapis", library.getGroup());
     }
+    
+    Assert.assertEquals("grpc", datalossPrevention.getTransport());
+    Assert.assertEquals("http", cloudStorage.getTransport());
   }
   
   @Test
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
@@ -39,25 +39,26 @@ public class CloudLibrariesTest {
     List<Library> libraries = CloudLibraries.getLibraries("clientapis");
     Assert.assertTrue(libraries.size() > 10);
     
-    Library datalossPrevention = null;
-    Library cloudStorage = null;
-    
     for (Library library : libraries) {
       String tooltip = library.getToolTip();
-      if ("googlecloudstorage".equals(library.getId())) {
-        cloudStorage = library;
-      } else if ("datalossprevention".equals(library.getId())) {
-        datalossPrevention = library;
-      }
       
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -89,9 +89,11 @@ public static Library getLibrary(String id) {
       for (JsonObject api : apis) {
         String name = api.getString("name");
         String id = api.getString("id");
+        JsonArray transports = api.getJsonArray("transports");
         Library library = new Library(id);
         library.setGroup(CLIENT_APIS_GROUP);
         library.setName(name);
+        library.setTransport(transports.getString(0));
```
briandealwis: Why throw away data?  It's not obvious that the `transports` array has a prioritized order.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java
```diff
@@ -70,7 +70,7 @@ public LibrarySelectorGroup(Composite parentContainer, String groupName) {
     Collection<Library> availableLibraries = CloudLibraries.getLibraries(groupName);
     this.availableLibraries = new LinkedHashMap<>();
     for (Library library : availableLibraries) {
-      if (!restrictedEnvironment || !library.getJavaVersion().equals("1.8")) {
+      if (!restrictedEnvironment || "http".equals(library.getTransport())) {
```
briandealwis: I have to admit to finding this encoded knowledge a bit hard to remember. Perhaps the javadocs for `restrictedEnvironment` should be updated.  Or maybe we should pass in an instance of `AppEngineRuntime` — or create a `Runtime` enum for the different platforms.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
```diff
@@ -38,13 +38,26 @@ public void testGetLibraries() {
...
+        datalossPrevention = library;
+      }
+      
       Assert.assertNotNull(library.getName() + " has no tooltip", tooltip);
       Assert.assertFalse(tooltip.isEmpty());
       Assert.assertFalse(library.getName() + " has no files", library.getLibraryFiles().isEmpty());
       Assert.assertEquals("clientapis", library.getGroup());
     }
+    
+    Assert.assertEquals("grpc", datalossPrevention.getTransport());
```
briandealwis: These seem like they should be broken out as separate tests.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
```diff
@@ -38,13 +38,26 @@ public void testGetLibraries() {
   public void testGetClientApis() {
```
briandealwis: Might this be better named `validateClientApis()`?

elharo: It could be. I'm still in the habit of naming all test methods testFoo. Not sure I want to break that convention.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibrariesTest.java
elharo: of course, done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java
elharo: need to refactor a bit, but meanwhile updated javadoc

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
elharo: ATM there is exactly one transport per API. The JSON is a little more general because the API team said this may not always be true.

briandealwis: That would be worth noting here in a comment.

elharo: done

PR #2475, opened 2017-10-11
Refactor MavenCoordinates

chanseokoh:
1. ~~The group ID and artifact ID are mandatory, so it's natural to require them in the builder constructor.~~
1. `MavenCoordinates` constructor: `public` --> `private`
   - However, in another sense, given that `MavenCoordinates` doesn't have setters (immutable), and an instance with a group ID and an artifact ID constructs a complete and sane instance, so it may not be so wrong to have such a working public constructor. Nonetheless, I made it `private`.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
@@ -144,11 +144,11 @@ public class LibraryFileTest {
   
   @Test
   public void testCompareTo() {
-    MavenCoordinates m1 = new MavenCoordinates.Builder("A1", "G1").setVersion("1").build();
-    MavenCoordinates m2 = new MavenCoordinates.Builder("B1", "G1").setVersion("2").build();
-    MavenCoordinates m3 = new MavenCoordinates.Builder("A1", "E1").setVersion("1").build();
-    MavenCoordinates m4 = new MavenCoordinates.Builder("A1", "G1").setVersion("2").build();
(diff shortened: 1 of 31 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
```diff
@@ -34,26 +34,31 @@
...
+import org.junit.Before;
 import org.junit.Test;
 
 public class LibraryFileTest {
 
+  private MavenCoordinates mavenCoordinates;
+
+  @Before
+  public void setUp() {
+    mavenCoordinates = new MavenCoordinates.Builder("groupId", "artifactId").build();
```
elharo: don't need the method, just initialize the field

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
```diff
@@ -126,35 +120,28 @@ public Builder toBuilder() {
...
-      coordinates.classifier = classifier;
-      return coordinates;
-    }
     
     /**
      * @param groupId the Maven group ID
+     * @param artifactId the Maven artifact ID
      */
-    public Builder setGroupId(String groupId) {
+    public Builder(String groupId, String artifactId) {
```
elharo: I wonder if the constructor should be empty with named setters for groupId and artifactId? One of the advantages of the builder pattern is named parameters, which this loses.

chanseokoh: I see. So the pattern does want to have a 0-arg constructor. I'll revert the changes.

PR #2476, opened 2017-10-11
Ensure JSP and JSTL are added to new standard and flex projects

briandealwis:
An alternative implementation for #2455 and #2456:
  - AES native (Java 7 + 8): downloads and installs `jstl:jstl:1.2` into the `src/main/webapp/WEB-INF/lib`; removes `jstl` from the servlet container (where it didn't belong)
  - AES maven (Java 7 + 8): sets up the `pom.xml` to `jstl:jstl:1.2` plus with appropriate `jsp-api` references
  - AEF native: downloads and installs `jstl:jstl:1.2` into `lib/` as was happening previously with `servlet-api` and `jsp-api`
     - this doesn't feel quite right, but we're not creating a proper WTP server runtime for flex
  - AEF maven: adds `jstl:jstl:1.2` to the `pom.xml`

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
@@ -26,11 +26,11 @@ import com.google.cloud.tools.eclipse.appengine.newproject.CodeTemplates;
 import com.google.cloud.tools.eclipse.appengine.newproject.CreateAppEngineWtpProject;
 import com.google.cloud.tools.eclipse.appengine.newproject.Messages;
 import com.google.cloud.tools.eclipse.util.ClasspathUtil;
+import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.ArrayList;
-import java.util.Collections;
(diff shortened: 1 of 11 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
```diff
@@ -66,19 +66,26 @@
...
   static {
-    MavenCoordinates.Builder servletApi = new MavenCoordinates.Builder()
+    // FIXME: servlet-api and jsp-api should be provided by the servlet container and not included
+    MavenCoordinates servletApi = new MavenCoordinates.Builder()
         .setGroupId("javax.servlet") //$NON-NLS-1$
         .setArtifactId("javax.servlet-api") //$NON-NLS-1$
-        .setVersion("3.1.0"); //$NON-NLS-1$
-    PROJECT_DEPENDENCIES = Collections.singletonList(servletApi.build());
+        .setVersion("3.1.0") //$NON-NLS-1$
+        .build();
```
chanseokoh: I was going to say that, since `javax.servlet-api` is being not exported, what you add here would not be exported either. We want to export JSTL, right?

briandealwis: We do.  I'll bite the bullet and put the JSTL library into `WEB-INF/lib`.

PR #2480, opened 2017-10-12
surface error on project creation to user

elharo:
@tgroh @briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectCreator.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectCreator.java
@@ -252,7 +252,6 @@ public class DataflowProjectCreator implements IRunnableWithProgress {
    * platform. If not supported, throw a {@link ProjectCreationException}.
    */
   private String getTargetPlatform() throws ProjectCreationException {
-    // Safe cast by API Contract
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectCreator.java
```diff
@@ -194,31 +195,34 @@ public void run(IProgressMonitor monitor)
             mavenGroupId, mavenArtifactId, "0.0.1-SNAPSHOT", packageString, archetypeProperties,
             projectImportConfiguration, progress.newChild(4));
         break;
-      } catch (CoreException e) {
-        failures.add(e);
+      } catch (CoreException ex) {
+        // todo should we use a MultiStatus here?
```
briandealwis: It would seem so!  Given we're potentially trying multiple archetype versions, it maybe should be a nested set of MultiStatus objects.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowProjectCreator.java
```diff
@@ -194,31 +195,34 @@ public void run(IProgressMonitor monitor)
...
+        failures.add(ex);
       }
     }
-    if (projects.isEmpty()) {
+    if (!failures.isEmpty()) {
```
briandealwis: This change is not correct.  The code above is tries a set of archetype versions.  Providing the last attempt worked (i.e., the project was created) then all is well.

PR #2488, opened 2017-10-12
0.x.0 versions aren't released

### LINKED ISSUE #2481: Sometimes adding certain library adds all transitive deps
For Cloud Libraries, adding a single library to a Maven project usually adds a single dependency in `pom.xml`. For example,
```xml
<dependency>
<artifactId>google-cloud-spanner</artifactId>
<version>0.25.0-beta</version>
</dependency>
```

However, adding Cloud Speech adds about 60 deps. I think it adds most, if not all, transitive deps. Adding Cloud Endpoints similarly adds a lot of deps.

BTW, adding Objectify additionally adds guava, and I'm not sure if this is needed.

elharo:
This fixes the 0.8.0 problem but I don't think that it addresses #2481. I suspect that bug comes about in `Pom.addDependencies`. The mistake there is looping through all the library files in the library.

These used to be only the files declared in plugin.xml, but can now include files added transitively later. That one needs a little more thought.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
+++ b/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
@@ -158,7 +158,7 @@ public class ArtifactRetriever {
     try {
       NavigableSet<ArtifactVersion> versions = availableVersions.get(coordinates);
       for (ArtifactVersion version : versions.descendingSet()) {
-        if (Strings.isNullOrEmpty(version.getQualifier())) {
+        if (version.getMajorVersion() > 0 && Strings.isNullOrEmpty(version.getQualifier())) {
           if (range == null || range.containsVersion(version)) {
             return version;
           }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFileTest.java
@@ -34,6 +34,9 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.SortedSet;
 import java.util.TreeSet;
+
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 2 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -158,7 +158,7 @@ private ArtifactVersion getLatestReleaseVersion(String coordinates, VersionRange
     try {
       NavigableSet<ArtifactVersion> versions = availableVersions.get(coordinates);
       for (ArtifactVersion version : versions.descendingSet()) {
-        if (Strings.isNullOrEmpty(version.getQualifier())) {
+        if (version.getMajorVersion() > 0 && Strings.isNullOrEmpty(version.getQualifier())) {
```
chanseokoh: I feel a little bit uncomfortable, as this changes the current behavior. This is a general class, and Dataflow and library loading from plugin.xml depend on this. For example, if there were a third part library that has no major version greater than 0, this will return null from now on.

elharo: It's a general class but it's not really API. We know which libraries depend on this and they all satisfy this constraint. I.e. it doesn't really change the current behavior because the libraries we currently load don't have zero as a major version for a release. Going forward, I think it's pretty standard that 0.x means a pre-release version.

I do want to add a method that gets the latest unreleased version, or perhaps the latest unreleased version if and only if there's no released version so we can upgrade from beta versions in libraries.json; but at the moment that's not critical.

briandealwis: Late to the party here, but why aren't we using the `<release>` version from the `maven-metadata.xml`?  It clearly shows for `0.25.0-beta` for pubs https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2481#issuecomment-336279816

chanseokoh: Maybe it's always the latest, regardless of qualifiers. That may not be what we want: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2481#issuecomment-336284035

briandealwis: But we're now second-guessing decisions made by the project maintainer.  Many of the [recent google-cloud-* libraries](http://search.maven.org/#search%7Cga%7C1%7Cg%3Acom.google.cloud) are 0.* and are `-beta` or `-alpha`.

This code is used by anything fetched from a `<library>` element that doesn't have a version (viz `com.google.cloud.tools.eclipse.appengine.libraries.model.LibraryFactory.getMavenCoordinates(IConfigurationElement[])`).

chanseokoh: Hmm... I agree on the 0.* part. I think our Dataflow code also uses this method to fetch artifacts. Should we reconsider removing the 0.* check? (The 0.8.0 problem doesn't happen without it.)

About the `<release>` element, if it can be always the most recent release ignoring `-beta` or `-alpha`, I think it's not a good idea to use it.

briandealwis: Do we have evidence of the `<release>` tag being wrong?

I'd rather hard-code artifacts that have exceptions than try to figure this out algorithmically.  The use of qualifier in Maven Central is wide and varied: [just look at the sorting algorithm](https://maven.apache.org/ref/3.3.9/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html)! And note SpringBoot uses a `.RELEASE` convention.

PR #2489, opened 2017-10-13
only load direct dependencies into pom.xml

### LINKED ISSUE #2481: Sometimes adding certain library adds all transitive deps

elharo:
@chanseokoh fix #2481

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
@@ -218,7 +218,6 @@ public class PomTest {
     
     Library library = new Library("objectify");
     library.setLibraryFiles(files);
-    library.resolveDependencies();
     
     List<Library> libraries = new ArrayList<>();
     libraries.add(library);
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
@@ -242,7 +242,6 @@ public class PomTest {
   private static Document parse(InputStream in)
       throws ParserConfigurationException, IOException, SAXException {
     DocumentBuilder builder = factory.newDocumentBuilder();
-    factory.setNamespaceAware(true);
     Document actual = builder.parse(in);
     return actual;
   }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
@@ -218,7 +218,6 @@ public class PomTest {
     
     Library library = new Library("objectify");
     library.setLibraryFiles(files);
-    library.setResolved(false);
     library.resolveDependencies();
     
     List<Library> libraries = new ArrayList<>();
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
@@ -82,25 +82,25 @@ public class PomTest {
     List<LibraryFile> list0 = new ArrayList<>();
     list0.add(file0);
     
-    MavenCoordinates cooridnates1 = new MavenCoordinates.Builder()
+    MavenCoordinates coordinates1 = new MavenCoordinates.Builder()
         .setGroupId("com.example.group1")
         .setArtifactId("artifact1")
         .build();
     List<LibraryFile> list1 = new ArrayList<>();
-    LibraryFile file1 = new LibraryFile(cooridnates1);
+    LibraryFile file1 = new LibraryFile(coordinates1);
     list1.add(file1);
     
-    MavenCoordinates cooridnates2 = new MavenCoordinates.Builder()
+    MavenCoordinates coordinates2 = new MavenCoordinates.Builder()
         .setGroupId("com.example.group2")
         .setArtifactId("artifact2")
         .build();
-    MavenCoordinates cooridnates3 = new MavenCoordinates.Builder()
+    MavenCoordinates coordinates3 = new MavenCoordinates.Builder()
         .setGroupId("com.example.group3")
         .setArtifactId("artifact3")
         .build();
     List<LibraryFile> list2 = new ArrayList<>();
-    LibraryFile file2 = new LibraryFile(cooridnates2);
-    LibraryFile file3 = new LibraryFile(cooridnates3);
+    LibraryFile file2 = new LibraryFile(coordinates2);
+    LibraryFile file3 = new LibraryFile(coordinates3);
     list2.add(file2);
     list2.add(file3);
     
@@ -148,21 +148,21 @@ public class PomTest {
   public void testAddDependencies_withDuplicates() 
       throws CoreException, ParserConfigurationException, IOException, SAXException {
 
-    MavenCoordinates cooridnates1 = new MavenCoordinates.Builder()
+    MavenCoordinates coordinates1 = new MavenCoordinates.Builder()
         .setGroupId("com.example.group1")
         .setArtifactId("artifact1")
         .build();
-    MavenCoordinates cooridnates2 = new MavenCoordinates.Builder()
+    MavenCoordinates coordinates2 = new MavenCoordinates.Builder()
         .setGroupId("com.example.group2")
         .setArtifactId("artifact2")
         .build();
 
     List<LibraryFile> list1 = new ArrayList<>();
-    LibraryFile file1 = new LibraryFile(cooridnates1);
+    LibraryFile file1 = new LibraryFile(coordinates1);
     list1.add(file1);
     
     List<LibraryFile> list2 = new ArrayList<>();
-    LibraryFile file2 = new LibraryFile(cooridnates2);
+    LibraryFile file2 = new LibraryFile(coordinates2);
     list2.add(file1);
     list2.add(file2);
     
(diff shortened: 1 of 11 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -115,15 +117,16 @@ void setSiteUri(URI siteUri) {
   }
 
   public synchronized List<LibraryFile> getLibraryFiles() {
```
briandealwis: Returning different results whether a library is resolved or not seems a bit dangerous, and counter to the model (the extension point has `<library><libraryFile/><libraryFile/>…</library>`.  If you changed `getDirectDependencies()` → `getTransitiveDependencies()` you could get rid of the resolved flag and use whether `transitiveDependencies == null` as the flag.

elharo: I do think this class is do for some more rework since it was desigbned under the assumption (no longer valid) that most libraries listed all their deps in plugin.xml. resolved should probably default to false. 

That said, I'm not sure I want to change more than the minimum we need until 1.4.0 is  out. It feels like it risks introducing other bugs.

briandealwis: That's my point: changing `getLibraryFiles()` to return the transitive dependencies is counter to the original design, and it has introduced bugs.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -115,15 +117,16 @@ void setSiteUri(URI siteUri) {
...
   }
 
   /**
    * @param libraryFiles artifacts associated with this library, cannot be <code>null</code>
    */
   public synchronized void setLibraryFiles(List<LibraryFile> libraryFiles) {
     Preconditions.checkNotNull(libraryFiles);
-    this.libraryFiles = new ArrayList<>(libraryFiles);
+    this.directDependencies = new ArrayList<>(libraryFiles);
+    this.transitiveDependencies = this.directDependencies;
```
chanseokoh: This needs some thoughts, as it blindly overwrites the transitiveDependencies. It's getting hard to track the effect, and it's not clear what the contract of this method should be. However, I think we should set `resolved = false` here. I wish everything were immutable.

elharo: Immutability isn't really feasible for this class because we need to load the transitive dependencies when used. Otherwise we waste a lot of time and network bandwidth loading every dependency of every library even though the user doesn't touch it.

resolved should not be set to false here because this is only called as part of initialization of a new library object. 

I might be able to not set transitiveDependencies here and instead initialize it lazily when getLibraryFiles is called. I could even call resolveDependencies from inside getLibraryFiles though that will have the side effect of making a few more of our tests ht the network. Still that hasn't been a big problem so far, so maybe that makes sense.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: getLibraryFiles() originally returned the complete graph of dependencies. I don't think that's changed. 

Let me explore using `transitiveDependencies == null` as the flag. For sanity, I'll probably rename the two methods to 

```
getDirectDependencies
getAllDependencies
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
chanseokoh: At least the code is same (I see it is just renaming) and there is no behavioral change, so I think this won't break things, but looks like we should have set `resolved = false` to be correct.

chanseokoh: Never mind. I think I misunderstood how things work here.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -173,44 +171,53 @@ void setTransport(String transport) {
     this.transport = transport;
   }
   
-  public synchronized boolean isResolved() {
-    return this.resolved;
+  @VisibleForTesting
+  synchronized boolean isResolved() {
```
chanseokoh: No one calls this method except tests.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -81,9 +81,9 @@ static Library create(IConfigurationElement configurationElement) throws Library
           library.setExport(Boolean.parseBoolean(exportString));
         }
         String dependencies = configurationElement.getAttribute("dependencies"); //$NON-NLS-1$
-        if ("include".equals(dependencies)) { //$NON-NLS-1$
+        if (!"include".equals(dependencies)) { //$NON-NLS-1$
           // Load the dependencies from Maven Central later
```
chanseokoh: The comment doesn't seem right, if we look at the code of `setResovled()`. The method makes `library` "final" so that it won't attempt to resolve transitive dependencies.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -173,44 +171,53 @@ void setTransport(String transport) {
...
+    return transitiveDependencies != null;
   }
   
   /**
    * @param resolved true iff this library contains its complete dependency graph
    */
   synchronized void setResolved(boolean resolved) {
```
chanseokoh: No one calls the method with the param value of `false`, so I think we can remove the parameter too. Maybe the method name can be changed after that.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: vestigial; removed

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -194,10 +200,49 @@ public void testAddDependencies_withDuplicates()
...
+    Element artifactId = getOnlyChild(dependency, "artifactId");
+    Assert.assertEquals("objectify", artifactId.getTextContent());
+    Element version = getOnlyChild(dependency, "version");
+    Assert.assertEquals("5.1.21", version.getTextContent());
+  }
 
   private static Document parse(InputStream in)
       throws ParserConfigurationException, IOException, SAXException {
     DocumentBuilder builder = factory.newDocumentBuilder();
+    factory.setNamespaceAware(true);
```
briandealwis: Why was this necessary?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -173,44 +170,48 @@ void setTransport(String transport) {
...
+
   /**
-   * @param resolved true iff this library contains its complete dependency graph
+   * Returns the non-transitive dependencies of this library. Useful when a separate system such as
+   * Maven will resolve the transitive dependencies later. 
    */
-  synchronized void setResolved(boolean resolved) {
-    this.resolved = resolved;
+  public List<LibraryFile> getDirectDependencies() {
+    return new ArrayList<>(this.directDependencies);
```
briandealwis: `ImmutableList`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -114,16 +105,22 @@ void setSiteUri(URI siteUri) {
     this.siteUri = siteUri;
   }
 
-  public synchronized List<LibraryFile> getLibraryFiles() {
-    return new ArrayList<>(libraryFiles);
+  /**
+   * Returns the complete list of all transitive dependencies for this library.
```
briandealwis: Maybe note that this can result in network traffic and may be relatively expensive.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
elharo: It isn't. My mistake. reverted.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: Not a bad idea. Right now I'd prefer not to make further changes.

PR #2494, opened 2017-10-13
Refactor BuildPath methods

chanseokoh:
Currently, `listNativeLibrary()` and `computeEntres()` return either
  - a singleton list/array of the master container; or
  - an empty list/array.

So, to reduce confusion, I'm making them return just the master container (no list/array) or null.

Also we don't make use of the return value of `addNativeLibrary()` (except tests), so I made it `void`.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -144,17 +144,15 @@ public static Library collectLibraryFiles(IJavaProject javaProject, List<Library
     return masterLibrary;
   }
 
-  private static List<IClasspathEntry> computeEntries(IJavaProject javaProject, Library library,
+  /**
+   * @return a {@link IClasspathEntry} created from {@code library} if {@code javaProject} does not
```
elharo: a --> an

chanseokoh:
Sure, no need to merge it soon.

Issue #2500, opened 2017-10-16
Error logged when adding Cloud Pub/Sub library


chanseokoh:
~~Looking at the message, I think it has something to do with the `null` artifact qualifier.~~

```
Oct 16, 2017 6:26:19 PM com.google.cloud.tools.eclipse.appengine.libraries.model.Library resolveDependencies
SEVERE: Could not load library MavenCoordinates [repository=central, io.netty:netty-handler:jar:null:4.1.14.Final]
org.eclipse.core.runtime.CoreException: Could not resolve dependencies
	at com.google.cloud.tools.eclipse.util.DependencyResolver$1.call(DependencyResolver.java:87)
	at com.google.cloud.tools.eclipse.util.DependencyResolver$1.call(DependencyResolver.java:1)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:176)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:151)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:99)
	at com.google.cloud.tools.eclipse.util.DependencyResolver.getTransitiveDependencies(DependencyResolver.java:95)
	at com.google.cloud.tools.eclipse.appengine.libraries.model.LibraryFactory.loadTransitiveDependencies(LibraryFactory.java:119)
	at com.google.cloud.tools.eclipse.appengine.libraries.model.Library.resolveDependencies(Library.java:207)
	at com.google.cloud.tools.eclipse.appengine.libraries.model.Library.getAllDependencies(Library.java:114)
	at com.google.cloud.tools.eclipse.appengine.libraries.repository.LibraryClasspathContainerResolverService.resolveLibraryFiles(LibraryClasspathContainerResolverService.java:187)
	at com.google.cloud.tools.eclipse.appengine.libraries.repository.LibraryClasspathContainerResolverService.resolveContainer(LibraryClasspathContainerResolverService.java:140)
	at com.google.cloud.tools.eclipse.appengine.libraries.LibraryClasspathContainerInitializer.initialize(LibraryClasspathContainerInitializer.java:84)
   ...
```

[...]

Issue #2493, opened 2017-10-13
Investigate Objectify dependencies


elharo:
They are currently not loaded transitively. This means we're likely adding too much to a pom.xml in maven projects. 

Also everything except Guava is marked optional or test or provided.

PR #2502, opened 2017-10-17
Update versions of Maven dependencies too. 

### LINKED ISSUE #2499: Library versions are updated only when adding libraries to native projects (i.e., no chance for Maven users)
(From https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2481#issuecomment-336262415. Forgot to create an issue then.)

We lazily update library versions to the most recent releases. However, the only point we update them is when adding libraries to native projects. That is, if users always work on Maven projects, they will be bound to use the pre-defined versions in `libraries.json`, as they won't have a chance to update versions.

elharo:
@chanseokoh fix #2499 #2493 

Fixing the Maven issue unexpectedly led into an unrecognized bug where transitive dependency versions were incorrectly updated. We should have been only updating versions of direct dependencies. 

This also has the side effect of working around, if not exactly fixing, #2502

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
@@ -143,6 +143,9 @@ public class BuildPath {
         // if someone edited the serialized data behind Eclipse's back
         if (library != null && !dependentIds.contains(id)) { 
           dependentIds.add(library.getId());
+          // todo we're not distinguishing direct here;
+          // we should be setting all transitive and separately direct. 
+          //
           masterFiles.addAll(library.getAllDependencies());
           subMonitor.worked(1);
         }
@@ -158,7 +161,11 @@ public class BuildPath {
     
     List<LibraryFile> resolved = Library.resolveDuplicates(new ArrayList<LibraryFile>(masterFiles));
     subMonitor.worked(8);
+    
     masterLibrary.setLibraryFiles(resolved);
+    
+    masterLibrary.setResolved();
+    
     subMonitor.worked(1);
 
     return masterLibrary;
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
@@ -143,9 +143,6 @@ public class BuildPath {
         // if someone edited the serialized data behind Eclipse's back
         if (library != null && !dependentIds.contains(id)) { 
           dependentIds.add(library.getId());
-          // todo we're not distinguishing direct here;
-          // we should be setting all transitive and separately direct. 
-          //
           masterFiles.addAll(library.getAllDependencies());
           subMonitor.worked(1);
         }
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainerInitializer.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainerInitializer.java
@@ -101,7 +101,7 @@ public class LibraryClasspathContainerInitializer extends ClasspathContainerInit
     return true;
   }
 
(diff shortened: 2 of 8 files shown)


chanseokoh:
I verified this doesn't fix #2499. See https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2500#issuecomment-337092390.

chanseokoh:
Oops. Made a comment on a wrong PR.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainerInitializer.java
```diff
@@ -102,6 +101,7 @@ private static boolean jarPathsAreValid(LibraryClasspathContainer container) {
     return true;
   }
 
+  // todo may be unused
```
chanseokoh: This should have had `@Override`. I vaguely remember @briandealwis added this method.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -194,7 +194,14 @@ synchronized void setResolved() {
    * transitive dependency graph.
    */
   private synchronized List<LibraryFile> resolveDependencies() {
+    
+    if (transitiveDependencies != null) {
+      return transitiveDependencies; 
+    }
```
chanseokoh: Sensible optimization, but just letting you know this is redundant at the current state, because this method is called only when `transitiveDependencies != null` by `getAllDependencies()`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
```diff
@@ -194,7 +194,14 @@ synchronized void setResolved() {
...
+    
     List<LibraryFile> transitiveDependencies = new ArrayList<>();
+    
+    // todo check resolved
```
chanseokoh: Probably not needed?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -158,7 +158,11 @@ public static Library collectLibraryFiles(IJavaProject javaProject, List<Library
     
     List<LibraryFile> resolved = Library.resolveDuplicates(new ArrayList<LibraryFile>(masterFiles));
     subMonitor.worked(8);
+    
     masterLibrary.setLibraryFiles(resolved);
+    
+    masterLibrary.setResolved();
```
chanseokoh: So, transitive dependency versions get updated without this? (I fail to follow the overall pipeline.)

elharo: Yes, exactly, which we don't want to happen. We want the latest version of objectify, but if the latest version of objectify asks for Guava 20.0, we don't want to give it Guava 23 instead which is what was happening prior to this PR.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainerInitializer.java
elharo: Guess I turned off that warning. Fixed.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: removed

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/Library.java
elharo: removed

chanseokoh:
Surprisingly, this seems to fix #2500 too.

PR #2508, opened 2017-10-17
rename getLatestArtifactVersion --> getLatestReleaseVersion

elharo:
This is a hopefully safe and mostly automated refactoring in advance of adding a method that gets the latest version, released or otherwise.


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/ArtifactRetriever.java
```diff
@@ -136,15 +136,16 @@ private ArtifactRetriever(String repositoryUrl) {
...
+    VersionRange range = null;
+    return getLatestReleaseVersion(idToKey(groupId, artifactId), range);
   }
 
   /**
    * Returns the latest published release artifact version in the version range,
    * or null if there is no such version.
    */
-  public ArtifactVersion getLatestArtifactVersion(
+  public ArtifactVersion getLatestReleaseVersion(
```
chanseokoh: Now that the name becomes identical to the private method below, this feels like an extra indirection that is unnecessary. We can get rid of this method if we push `idToKey()` down to the private method and make it public.

elharo: done

chanseokoh: Commit not pushed.

chanseokoh:
Hmm... there is a compilation error.

PR #2511, opened 2017-10-18
Include plugin.xml in .cte13 bundle

chanseokoh:
BTW, we should remove `.appengine` from the bundle name.


briandealwis:
oh jeepers

chanseokoh:
Probably adding `PluginXmlTest` can catch this.

elharo:
I was just thinking that. I wonder if there's a way to write such a test that tests all our bundles?

briandealwis:
I initially put the updater into `.compat`, but it required some significant refactoring as we have `.appengine.facets` depend ending on `.compat` to pull in `GpeMigrator`, etc., so I backed it out and went for this simpler solution.

I'll push up the refactoring, but it can wait for 1.4.

briandealwis:
Refactoring in #2517

PR #2516, opened 2017-10-18
True getLatestVersion method

elharo:
This method gets the latest version irrespective of qualifiers.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ArtifactRetrieverIntegrationTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ArtifactRetrieverIntegrationTest.java
@@ -64,8 +64,7 @@ public class ArtifactRetrieverIntegrationTest {
     Assert.assertNotNull(version);
     
     // version 0.26 or later
-    double x = Double.parseDouble(version.getMajorVersion() + "." + version.getMinorVersion());
-    Assert.assertTrue(x > 0.25);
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ArtifactRetrieverIntegrationTest.java
```diff
@@ -56,9 +56,20 @@ public void testGetGuava19() throws InvalidVersionSpecificationException {
...
-  
+
   @Test
   public void testGetLatestArtifactVersion() {
+    ArtifactVersion version = ArtifactRetriever.DEFAULT.getLatestVersion(
+        "com.google.cloud", "google-cloud-dns");
+    Assert.assertNotNull(version);
+    
+    // version 0.26 or later
+    double x = Double.parseDouble(version.getMajorVersion() + "." + version.getMinorVersion());
```
chanseokoh: Just pointing out that this won't work if they release '0.1xx' later (extremely unlikely though).

briandealwis: Ah, I had looked at this earlier thinking _I know there must be an issue with this_.  Would be safer to compare the components separately as integers.

chanseokoh: Generally, this double comparison doesn't work. Another case is like when comparing 0.26 and 0.9. (0.9 is an earlier version.)

PR #2521, opened 2017-10-19
Fix plugin info in MANIFEST.MF

chanseokoh:


chanseokoh:
`BasePluginXmlTest` has evolved to include tests for `build.properties`, so I think it is not terrible wrong to add tests for `MANIFEST.MF`. Moreover, in our setting, we put localized messages of `MANIFEST.MF` into `plugin.properties`, which `BasePluginXmlTest` loads and tests, so it is easy to do so there.

However, we also have an annoyance due to `BasePluginXmlTest` becoming a uber test class: #1281

chanseokoh:
For now, I can add a test to `BasePluginXmlTest`. We can think about #1281 later.

chanseokoh:
Add a MANIFEST.MF localization test. Let's wait and see if any bundle fails.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -156,4 +163,32 @@ private void assertPropertyDefined(String name) {
...
+    boolean localizedMessageExists = false;
+    boolean bundleLocalizationDefined = false;
+
+    Attributes attributes = getManifestAttributes();
+    for (Entry<?, ?> entry : attributes.entrySet()) {
+      String key = entry.getKey().toString();
+      String value = entry.getValue().toString();
+      assertPropertyDefined(value);
+
+      localizedMessageExists |= value.startsWith("%");
```
elharo: This is hard to follow. Is there a more obvious, less clever way to write this test?

chanseokoh: I'll unroll this into a regular `if` statement.

chanseokoh:
PTAL

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -156,4 +163,36 @@ private void assertPropertyDefined(String name) {
...
+        bundleLocalization = value;
+      }
+    }
+
+    if (localizedMessageExists) {
+      assertEquals("plugin", bundleLocalization);
+    }
+  }
+
+  private Attributes getManifestAttributes() throws IOException {
```
elharo: static?

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
```diff
@@ -4,7 +4,7 @@ Bundle-Name: Google Cloud Tools for App Engine Standard Java 8 Tests
 Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.standard.java8.test
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.standard.java8
 Bundle-Version: 0.1.0.qualifier
-Bundle-Vendor: Google, Inc.
+Bundle-Vendor: Google Inc.
```
elharo: If we do care about this, we should probably add a test for it to the BasePluginXmlTest.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
chanseokoh: Done.

plugins/com.google.cloud.tools.eclipse.appengine.standard.java8.test/META-INF/MANIFEST.MF
chanseokoh: Added a test. Let's see if all the tests will succeed.

PR #2522, opened 2017-10-19
Add Eclipse version in bug report template

### LINKED ISSUE #2477: Bug report template should ask for Eclipse version too
> (please ensure you are running the latest version of CT4E with _Help > Check for Updates_.)
> 
> - Cloud Tools for Eclipse version:
> - Google Cloud SDK version: _(run `gcloud version`)_
> - OS:
> - Java version:

chanseokoh:
Fixes #2477.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,6 +2,7 @@
 
 - Cloud Tools for Eclipse version:
 - Google Cloud SDK version: _(run `gcloud version`)_
+- Eclipse version: _(e.g, 4.5.2.v20160212-1500 or Neon.3)_
 - OS:
 - Java version:
 
--- a/plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java
@@ -28,6 +28,7 @@ import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.swt.program.Program;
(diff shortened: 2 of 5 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -2,6 +2,7 @@
 
 - Cloud Tools for Eclipse version:
 - Google Cloud SDK version: _(run `gcloud version`)_
+- Eclipse version: _(e.g, 4.5.2.v20160212-1500 or Neon.3)_
```
elharo: if we're autofilling this, no need to add the e.g....

chanseokoh: This is the template used when users open an issue on GitHub directly, not through our menu.

chanseokoh:
Build fails on Travis. I guess I need to configure something more for Tycho.

plugins/com.google.cloud.tools.eclipse.bugreport/src/com/google/cloud/tools/eclipse/bugreport/ui/BugReportCommandHandler.java
```diff
@@ -75,4 +78,8 @@ private static String getCloudSdkVersion() {
       return ex.toString();
     }
   }
+
+  private static String getEclipseVersion() {
+    return Platform.getBundle("org.eclipse.platform").getVersion().toString();
```
briandealwis: You should handle that the `org.eclipse.platform` bundle is not present (i.e., the bundle returned is `null`).

Although this string isn't the Eclipse version exactly (see _Help > About_, which reports the product name, product version, and build-id), this is simple to compute and is good enough.  Whereas obtaining the _Help > About_ info isn't very easy and requires accessing internal classes.

chanseokoh: FYI, this returns, e.g., `4.5.2.v20160212-1500`, so I think we get enough info. I'll check null.

chanseokoh:
PTAL

PR #2527, opened 2017-10-23
Refactor Maven coordinates UI

chanseokoh:
Separating out the "Create as Maven Project" checkbox and the surrounding UI group from `MavenCoordinatesUi` into a new class `MavenCoordinatesWizardUi`.

This is in preparation for #2470, where we will only support creating a Maven project.

There is no visual or functional change. I verified that the wizards work fine, as they should.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenCoordinatesUi.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenCoordinatesUi.java
@@ -21,16 +21,10 @@ import com.google.cloud.tools.eclipse.util.MavenCoordinatesValidator;
 import com.google.cloud.tools.eclipse.util.status.StatusUtil;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
-import org.eclipse.jface.dialogs.DialogPage;
-import org.eclipse.jface.dialogs.IMessageProvider;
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenCoordinatesUi.java
```diff
@@ -39,50 +33,30 @@
```
elharo: Looking at this now I'm a little surprised this class doesn't extend Composite or Group.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/maven/MavenCoordinatesWizardUi.java
```diff
@@ -0,0 +1,129 @@
...
+    asMavenProjectButton = new Button(container, SWT.CHECK);
+    asMavenProjectButton.setText(Messages.getString("CREATE_AS_MAVEN_PROJECT")); //$NON-NLS-1$
+    asMavenProjectButton.addSelectionListener(new SelectionAdapter() {
+      @Override
+      public void widgetSelected(SelectionEvent event) {
+        updateEnablement();
+      }
+    });
+
+    coordinatesGroup = new Group(container, SWT.NONE);
```
elharo: Would it make sense for MavenCoordinatesUi to extend Group instead?

PR #2533, opened 2017-10-25
Inject appengine-api-1.0-sdk JAR to local apps

### LINKED ISSUE #2531: Should we export appengine-api-1.0-sdk JAR? Or place it under `WEB-INF/lib`?
Stemmed from #2530.

AFAIK, appengine-api-1.0-sdk doesn't have to exported, much like the servlet JAR; it is provided in prod. However, the dev appserver doesn't provide it — I verified NoClassDefFoundErrors when trying to use it. So, the JAR should be exported locally, or somehow the dev appserver should be able to load it at the same level where apps are loaded.

Before 1.4.0, I believe the JAR was being exported (whether or not we intended) for native projects. We had separate containers for libraries, and I believe some base containers (like the Cloud Core Libraries) must have had the appengine-api-1.0-sdk JAR in them. All of the JARs in library containers must have been exported.

After 1.4.0, we pushed the JAR into the App Engine Standard Runtime container and made it not exported. Users will start to hit #2530. This happens for both native and Maven projects.

The easiest (and maybe a common) solution is to mark the JAR to be exported. The deploy staging scrubs off the JAR by the deploy staging, so this won't cause any trouble. It's just that exporting the JAR is conceptually not correct.

chanseokoh:
Fixes #2531.

Puts (copies) appengine-api-1.0-sdk JAR into locally published users' apps, just before starting dev_appserver1.

- Doesn't put the JAR if apps already have one in `WEB-INF/lib`.
- We can keep the current `pom.xml` template and the non-export setting of the JAR.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
@@ -27,28 +27,40 @@ import com.google.cloud.tools.appengine.cloudsdk.process.ProcessExitListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
 import com.google.cloud.tools.appengine.cloudsdk.process.ProcessStartListener;
 import com.google.cloud.tools.appengine.cloudsdk.serialization.CloudSdkVersion;
+import com.google.cloud.tools.eclipse.appengine.libraries.model.CloudLibraries;
+import com.google.cloud.tools.eclipse.appengine.libraries.model.Library;
+import com.google.cloud.tools.eclipse.appengine.libraries.model.LibraryFile;
+import com.google.cloud.tools.eclipse.appengine.libraries.repository.ILibraryRepositoryService;
 import com.google.cloud.tools.eclipse.appengine.localserver.Activator;
 import com.google.cloud.tools.eclipse.appengine.localserver.Messages;
 import com.google.cloud.tools.eclipse.sdk.MessageConsoleWriterOutputLineListener;
 import com.google.cloud.tools.eclipse.util.status.StatusUtil;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import java.io.File;
+import java.io.IOException;
 import java.net.InetAddress;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -356,6 +374,41 @@ public boolean isInUse(InetAddress addr, int port) {
...
+    } catch (IOException e) {
+      String message = Messages.getString("cannot.copy.appengine.api.sdk.jar");
+      throw new CoreException(StatusUtil.error(LocalAppEngineServerBehaviour.class, message, e));
+    }
+  }
+
+  @VisibleForTesting
+  static boolean appEngineApiSdkJarExists(Path directory) {
+    for (String file : directory.toFile().list()) {
+      String lowercase = file.toLowerCase(Locale.US);
```
elharo: should work in practice. Could use a regexp instead, but I'm not sure it matters

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -356,6 +374,41 @@ public boolean isInUse(InetAddress addr, int port) {
...
+      }
+    } catch (IOException e) {
+      String message = Messages.getString("cannot.copy.appengine.api.sdk.jar");
+      throw new CoreException(StatusUtil.error(LocalAppEngineServerBehaviour.class, message, e));
+    }
+  }
+
+  @VisibleForTesting
+  static boolean appEngineApiSdkJarExists(Path directory) {
+    for (String file : directory.toFile().list()) {
```
elharo: Please rename: file --> fileName (I initially thought this was a file object)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -356,6 +374,41 @@ public boolean isInUse(InetAddress addr, int port) {
...
+      String lowercase = file.toLowerCase(Locale.US);
+      if (lowercase.startsWith("appengine-api-1.0-sdk") && lowercase.endsWith(".jar")) {
+        return true;
+      }
+    }
+    return false;
+  }
+
   private static int ifNull(Integer value, int defaultValue) {
```
elharo: very close to MoreObjects.firstNonNull

chanseokoh: I'll leave it this, as this is an existing code not relevant to this PR.

PR #2536, opened 2017-10-25
Do not put WEB-INF/lib/jstl-1.2.jar for Maven projects

### LINKED ISSUE #2535: Don't put WEB-INF/lib/jstl-1.2.jar if creating Maven projects
image

I'm on it.

chanseokoh:
Fixes #2535.

Both flex and standard (regardless of Java 7 or Java 8) add JSTL 1.2, so I decided to push down it to the base class.

I suggest to review from bottom to top of the screen — easy to see which code got moved around.

Let's see if tests will pass.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/META-INF/MANIFEST.MF
```diff
@@ -11,6 +11,7 @@ Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies
 Import-Package: com.google.cloud.tools.eclipse.test.util,
  com.google.cloud.tools.eclipse.test.util.project,
  com.google.cloud.tools.eclipse.test.util.ui,
+ com.google.common.cache;version="15.0.0",
```
elharo: version is off

chanseokoh: Good catch.

PR #2538, opened 2017-10-26
Refactoring and code cleanups

chanseokoh:
Refactoring, plus several tests as a bonus.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexWarFacet.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexWarFacet.java
@@ -20,6 +20,7 @@ import com.google.common.base.Preconditions;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jst.common.project.facet.core.JavaFacet;
+import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
 import org.eclipse.wst.common.project.facet.core.IFacetedProject;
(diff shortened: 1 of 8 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineFlexWarFacet.java
```diff
@@ -61,6 +62,7 @@ public static void installAppEngineFacet(IFacetedProject facetedProject,
     FacetUtil facetUtil = new FacetUtil(facetedProject);
     if (installDependentFacets) {
       facetUtil.addJavaFacetToBatch(JavaFacet.VERSION_1_8);
+      facetUtil.addWebFacetToBatch(WebFacetUtils.WEB_31);
```
chanseokoh: Adding the WTP facet is the correct behavior.

No one called this method before, this doesn't break anything.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/CreateAppEngineFlexWtpProject.java
```diff
@@ -90,40 +88,24 @@ public String getDescription() {
...
-      addDependenciesToProject(newProject, subMonitor.newChild(50));
-    }
+        monitor);
     return mostImportantFile;
   }
 
-  /**
-   * Add Java 8 and Dynamic Web Module facet
-   */
-  private void configureFacets(IProject project, IProgressMonitor monitor) throws CoreException {
```
chanseokoh: This is exactly what `AppEngineFlexWarFacet.installAppEngineFacet()` above does. No need to duplicate.

PR #2545, opened 2017-10-30
Add flex JAR template

chanseokoh:
No UI, so this will be invisible to users.

The app created from this works when deployed.


elharo:
I agree that HttpServer based is not the ultimate goal we want to reach with this template. This is a step on the way and not yet user visible.

patflynn:
Ah so the plan is to merge this and change it to be spring boot app before launch?

That's better, but I don't fully understand how this is an incremental step in that direction. Why not make it spring boot based from the start?

chanseokoh:
Fear not. I'll add a couple Spring Boot template files in this PR too.

elharo:
I agree that HttpServer is an uncommon approach, and we should aim for something that's closer to what users are likely to want.

I'm still not convinced that SpringBoot is the way to go. Flex is far more general than that. I don't object to a SpringBoot wizard, but it shouldn't be the only option we offer. I don't want to give users the incorrect idea that Flex is about SpringBoot.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.flex.spring.boot.ftl
```diff
@@ -0,0 +1,73 @@
...
+
+  <groupId>${projectGroupId}</groupId>
+  <artifactId>${projectArtifactId}</artifactId>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.showDeprecation>true</maven.compiler.showDeprecation>
```
elharo: Do we really want to show deprecation?
In general I prefer to keep samples minimal.

chanseokoh: The original and current standard env `pom.xml` had this, and this is currently in every `pom.xml` we generate. I'll remove all of them here. Just let me know if you change your mind.

patflynn:
great. Thanks for adding the spring boot files. 

Remove the HttpServer based project. It's a dead-end for users and adds to the maintenance burden.

@saturnism @lesv FYI if you guys want to validate the boot app.

elharo:
I'm not sure about the HttpServer. It's not serving the same market as SpringBoot, but there are many apps for which it's exactly what's needed to run effectively on flex. 

There are also many apps for which SpringBoot is clearly overkill. We don't want to lock people into it or suggest that SpringBoot is the blessed and only way to build a flex app. 

I would like to see a simple Jetty based template too, though I'm not sure if we have resources to commit to it right now; and I certainly don't want to block this work waiting for something that can be added later.

Issue #2523, opened 2017-10-19
Problem when removing Google APIs container and adding new libraries later


chanseokoh:
1. Create a Java project.
1. Add Endpoints.
1. Right-click on the _Google APIs_ container > _Build Path_ > _Remove from Build Path_. The container will be gone.
1. Add Objectify.

Open the container, and note that the Endpoints revived.

The problem is that the `_libraries.container` and `master-container.container` JSONs are not purged when removing the container. If there is a way to hook into the event of a container being removed by any means, we could potentially remove the files together. Not sure if there is such an event we can use.

But I can think of another easy way: when adding the container to a project that does not already have it, just purge those JSONs. I am trying this way, and I think I can make it work soon.

chanseokoh:
Hmm... seems like we can detect classpath changes: https://stackoverflow.com/questions/10173053/how-to-detect-changes-to-eclipse-classpath-container-contents

elharo:
This should be resolved by @briandealwis's work on a more reasonable way to remove libraries.

elharo:
#2523 helped, but didn't completely fix this. If you follow the steps above, the endpoints library comes up selected in the Add Library dialog and must be manually deselected.

Still a low priority fix though.

chanseokoh:
Yeah, I think this is a low priority, given that at least the library is pre-checked.

briandealwis:
I had to [back out that commit](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2553#issuecomment-344054071).

PR #2553, opened 2017-11-02
Add support for editing master container

### LINKED ISSUE #1464: BuildPath should know which libraries are already added
E.g. if Objectify is already in the project it should come up preselected:

image

Possibly it should be disabled if unchecking it here will not remove it.

briandealwis:
Adds support for creating and editing the master container.
  - Change `CloudLibrariesPage#setSelection()` to retrieve library lists and set selections
  - Removes `IClasspathContainerPageExtension2` from `CloudLibrariesPage` since we're not creating multiple containers and `getSelection()` is called anyways when creating a new container definitions
  - Merge `AppEngineLibrariesPage` (formerly labelled _App Engine Libraries_) into `ApiClientLibrariesPage` (called _Google Cloud Libraries for Java_), since they're now all part of the same master container.
    - but we test that the project is an App Engine project
  - Fix up labels
  - Push the master-container library list serialization/deserialization onto `BuildPath` to avoid referencing `LibraryClasspathContainerSerializer` directly (it felt weird)
    - Adds the container path to the `BuildPath` library-list methods (even though it's still ignored)
  - Restore `BuildPath.listNativeLibraries()` to be side-effect free by explicitly serializing the master container library list elsewhere
  
There's a tension here that's not yet resolved: these containers are identified as ".appengine", and we now include the App Engine libraries (currently Objectify, Endpoints) in the list.  But this may not make sense since the classpath container uses the _Google Cloud Libraries for Java_ title, and can be added to any arbitrary project.

*AESv8 project*
screen shot 2017-11-02 at 12 52 34 pm

*Plain Java project*
screen shot 2017-11-02 at 12 53 27 pm

Fixes #1464, #2400


plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java
```diff
@@ -76,12 +79,12 @@ public LibrarySelectorGroup(Composite parentContainer, String groupName) {
...
     }
-    createContents(parentContainer);
+    createContents(parentContainer, groupLabel);
   }
 
-  private void createContents(Composite parentContainer) {
+  private void createContents(Composite parentContainer, String groupLabel) {
     Group apiGroup = new Group(parentContainer, SWT.NONE);
-    apiGroup.setText(Messages.getString("appengine.libraries.group")); //$NON-NLS-1$
+    apiGroup.setText(groupLabel); // $NON-NLS-1$
```
chanseokoh: Remove the comment.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java
```diff
@@ -93,7 +96,15 @@ private void createContents(Composite parentContainer) {
...
     }
-    GridLayoutFactory.swtDefaults().generateLayout(apiGroup);
+    GridLayoutFactory.fillDefaults().generateLayout(apiGroup);
+  }
+
+  /**
+   * Return the available libraries.
+   */
+  @VisibleForTesting
+  Collection<? extends Library> getLibraries() {
```
chanseokoh: `getAvailableLibraries`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -239,4 +210,45 @@ public void done(IJobChangeEvent event) {
...
+   */
+  public static void saveLibraryList(IJavaProject project, IPath containerPath,
+      List<Library> libraries, IProgressMonitor monitor) throws CoreException {
+    LibraryClasspathContainerSerializer serializer = new LibraryClasspathContainerSerializer();
+    List<String> libraryIds = new ArrayList<>();
+    for (Library library : libraries) {
+      libraryIds.add(library.getId());
+    }
+    try {
+    serializer.saveLibraryIds(project, containerPath, libraryIds);
```
chanseokoh: Indentation.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -118,7 +118,8 @@ public LibraryClasspathContainer loadContainer(IJavaProject javaProject, IPath c
     }
   }
 
-  public void saveLibraryIds(IJavaProject javaProject, List<String> libraryIds)
+  // todo containerPath argument is not used
```
chanseokoh: Then why add the argument?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
```diff
@@ -73,7 +74,7 @@
   @Override
   public IStatus resolveAll(IJavaProject javaProject, IProgressMonitor monitor) {
     try {
-      IStatus status = Status.OK_STATUS;
+      MultiStatus status = StatusUtil.multi(this, Messages.getString("TaskResolveLibrariesError"));
```
chanseokoh: Why start with an error?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -19,105 +19,167 @@
...
-    setTitle(Messages.getString(group + "-title"));  //$NON-NLS-1$
-    setDescription(Messages.getString(group + "-description"));  //$NON-NLS-1$
-    setImageDescriptor(AppEngineImages.appEngine(64));
-    this.group = group;
+  private boolean isMavenProject;
+  private IClasspathEntry oldEntry;
+  private IClasspathEntry newEntry;
+
+  protected CloudLibrariesPage(String pageId) {
+    super(pageId); // $NON-NLS-1$
```
chanseokoh: Remove the comment.

elharo:
We do need to be careful about which projects have access to Endpoints. It's standard only: "Cloud Endpoints Frameworks supports only the App Engine standard environment. The App Engine flexible environment is not supported in Cloud Endpoints Frameworks."

I don't think we get that right right now.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -19,105 +19,167 @@
...
 
   private static final Logger logger = Logger.getLogger(CloudLibrariesPage.class.getName());
-  private LibrarySelectorGroup librariesSelector;
+  private final LibraryClasspathContainerSerializer serializer =
+      new LibraryClasspathContainerSerializer();
+
+  /** The library groups to be displayed; pairs of (id, title). */
+  private Map<String, String> groups;
+  /** Selected libraries that are not shown but should be preserved. */
+  private List<Library> hiddenLibraries = Collections.emptyList();
```
chanseokoh: My guess is that this is the list of libraries that have already been added before opening this page? If so, I think `existingLibraries` or such fits better. `hiddenLibraries` confused me. Also, existing libraries are ticked by default, right? In so, `hiddenLibraries` doesn't sound quite right.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -19,105 +19,167 @@
...
-    IClasspathContainerPageExtension, IClasspathContainerPageExtension2 {
+    IClasspathContainerPageExtension {
 
   private static final Logger logger = Logger.getLogger(CloudLibrariesPage.class.getName());
-  private LibrarySelectorGroup librariesSelector;
+  private final LibraryClasspathContainerSerializer serializer =
+      new LibraryClasspathContainerSerializer();
+
+  /** The library groups to be displayed; pairs of (id, title). */
+  private Map<String, String> groups;
```
chanseokoh: `libraryGroups`? An example in a comment would help, e.g., "App Engine library group and Cloud library group".

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: This is complicated.  I'll log any libraries that are discarded in setSelectedLibraries().

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
briandealwis: To match `loadLibraryIds()`, and in symmetry with the loadContainer()/saveContainer() methods.

I'm not sure whether we should be using the container path any more, or if there's any value in trying to support multiple containers, or different containerIds (e.g., we currently specify 'appengine').

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
briandealwis: It's a `MultiStatus` so it actually starts with OK.  It becomes not-OK only when a non-OK child is added.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
chanseokoh: I see, so this means "unrecognized libraries"? Then I think that makes sense.

briandealwis:
@elharo are you sure?  Endpoints provides a page on [_Getting Started with Endpoints on App Engine Flexible Environment_](https://cloud.google.com/endpoints/docs/openapi/get-started-app-engine).

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: I thought so too, but now I'm not sure.  I can't think of a situation where this will happen and we'd want this behaviour.

elharo:
This is specifically for the framework 2.0 and the library jar we're adding here: https://cloud.google.com/endpoints/docs/frameworks/about-cloud-endpoints-frameworks

That page, I think, is about using the REST API.

If I'm wrong about that, then those two pages contradict each other and I need to track someone down and get one of them fixed.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/MessagesTest.java
```diff
@@ -35,15 +35,8 @@ public void testClientApisTitle() {
...
-    Assert.assertEquals("Additional jars used by Google Client APIs for Java",
-        Messages.getString("clientapis-description"));
-  }
-
-  @Test
-  public void testDescription() {
-    Assert.assertEquals(
-        "Additional jars commonly used in App Engine standard environment applications",
-        Messages.getString("appengine-description"));
+    Assert.assertEquals("Additional jars for applications using Google Client APIs for Java",
```
chanseokoh: I don't think there is something called "Google Client APIs for Java". Instead, there is "Google API Client Library for Java", which is different from "Google Cloud Client Library for Java", so I'm inclining to change it to the latter.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/META-INF/MANIFEST.MF
```diff
@@ -12,16 +12,20 @@ Export-Package: com.google.cloud.tools.eclipse.appengine.libraries.ui
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
  com.google.cloud.tools.eclipse.appengine.libraries,
  com.google.cloud.tools.eclipse.appengine.libraries.model,
+ com.google.cloud.tools.eclipse.appengine.libraries.persistence,
  com.google.cloud.tools.eclipse.appengine.ui,
  com.google.cloud.tools.eclipse.util,
+ com.google.cloud.tools.eclipse.util.status,
  com.google.common.annotations;version="[20.0.0,21.0.0)",
  com.google.common.base,
+ com.google.common.collect;version="15.0.0",
```
chanseokoh: `;version="[20.0.0,21.0.0)",`

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/plugin.xml
```diff
@@ -3,11 +3,6 @@
 <plugin>
 
    <extension point="org.eclipse.jdt.ui.classpathContainerPage">
-     <classpathContainerPage
-       id="com.google.cloud.tools.eclipse.appengine.libraries"
-       name="%appEngineLibrariesTitle"
```
chanseokoh: Should be able to delete the message.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/ApiClientLibrariesPage.java
```diff
@@ -16,12 +16,34 @@
...
+import com.google.common.collect.Maps;
+import java.util.Map;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.IJavaProject;
 
 public class ApiClientLibrariesPage extends CloudLibrariesPage {
 
   public ApiClientLibrariesPage() {
-    super(CloudLibraries.CLIENT_APIS_GROUP);
+    super("clientapis");
```
chanseokoh: You don't want to use `CLIENT_APIS_GROUP` which is exactly "clientapis"?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/ApiClientLibrariesPage.java
```diff
@@ -16,12 +16,34 @@
...
 
+import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
 import com.google.cloud.tools.eclipse.appengine.libraries.model.CloudLibraries;
+import com.google.cloud.tools.eclipse.appengine.ui.AppEngineImages;
+import com.google.common.collect.Maps;
+import java.util.Map;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.IJavaProject;
 
 public class ApiClientLibrariesPage extends CloudLibrariesPage {
```
chanseokoh: So now `ApiClientLibrariesPage` is the only subclass of `CloudLibrariesPage`. The latter was to reuse common base, so I think we can just merge the former into the latter.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/ApiClientLibrariesPage.java
```diff
@@ -16,12 +16,34 @@
...
 
   public ApiClientLibrariesPage() {
-    super(CloudLibraries.CLIENT_APIS_GROUP);
+    super("clientapis");
+    setTitle(Messages.getString("clientapis-title")); //$NON-NLS-1$
+    setDescription(Messages.getString("apiclientlibrariespage-description")); //$NON-NLS-1$
+    setImageDescriptor(AppEngineImages.appEngine(64));
```
chanseokoh: It's an App Engine icon. Let's use a GCP icon if we have. (I think we have.)

chanseokoh:
I got an NPE when I tried to add Cloud DNS on a freshly created App Engine standard Java8 native project.

```
	at com.google.cloud.tools.eclipse.appengine.libraries.ui.CloudLibrariesPage.finish(CloudLibrariesPage.java:119)
	at org.eclipse.jdt.internal.ui.wizards.buildpaths.ClasspathContainerWizard.performFinish(ClasspathContainerWizard.java:95)
	at org.eclipse.jdt.internal.ui.wizards.buildpaths.newsourcepage.AddLibraryToBuildpathAction$1.performFinish(AddLibraryToBuildpathAction.java:98)
	at org.eclipse.jface.wizard.WizardDialog.finishPressed(WizardDialog.java:799)
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
chanseokoh: It may happen if the library definition file changes and some libraries got removed?

chanseokoh:
Looks like ticking or removing existing libraries doesn't work for Maven projects. That may require more PRs, I guess.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/messages.properties
```diff
@@ -1,6 +1,5 @@
+apiclientlibrariespage-description=Additional jars for applications using Google Client APIs for Java
 appengine-title=App Engine Standard Environment Libraries
-appengine-description=Additional jars commonly used in App Engine standard environment applications
 clientapis-title=Google Client APIs for Java
```
chanseokoh: I prefer "Google Cloud Client Libraries for Java", to be consistent.

selection_033

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -77,38 +77,21 @@ public static void addMavenLibraries(IProject project, List<Library> libraries,
...
-      LibraryClasspathContainerSerializer serializer = new LibraryClasspathContainerSerializer();
-      // We could create the LibraryClasspathContainer and serialize out the classpath information,
-      // but our LibraryClasspathContainerInitializer requires source paths to be resolved too,
-      // or it tosses the serialized classpath information
-      serializer.saveLibraryIds(javaProject, libraryIds);
-    } catch (IOException ex) {
-      throw new CoreException(
-          StatusUtil.error(BuildPath.class, "Error saving master library list", ex));
-    }
+    IPath containerPath = null; // not used, and masterEntry may be null
```
chanseokoh: I actually didn't get what `masterEntry` has something to do with `containerPath`, if it's null. `masterEntry` isn't get passed into `saveLibraryList()` either. Care to explain?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -239,4 +210,45 @@ public void done(IJobChangeEvent event) {
     job.schedule();
   }
 
+  /**
+   * Load the list of library dependencies saved for this project.
+   */
+  public static List<Library> loadLibraryList(IJavaProject project, IPath containerPath,
+      IProgressMonitor monitor) throws CoreException {
```
chanseokoh: `monitor` is not used.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -239,4 +210,45 @@ public void done(IJobChangeEvent event) {
...
+      }
+    }
+    return selectedLibraries;
+  }
+
+  /**
+   * Save the list of library dependencies for this project.
+   */
+  public static void saveLibraryList(IJavaProject project, IPath containerPath,
+      List<Library> libraries, IProgressMonitor monitor) throws CoreException {
```
chanseokoh: `monitor` not used.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -148,11 +131,11 @@ public static Library collectLibraryFiles(IJavaProject javaProject, List<Library
         }
       }
     } catch (IOException | CoreException ex) {
-      logger.log(Level.WARNING, "Error loading previous libraries", ex);
+      logger.log(Level.WARNING, "Error loading previous libraries", ex); //$NON-NLS-1$
     }
     
     Library masterLibrary = new Library(CloudLibraries.MASTER_CONTAINER_ID);
-    masterLibrary.setName("Google APIs"); //$NON-NLS-1$
+    masterLibrary.setName(Messages.getString("google.api.libraries")); //$NON-NLS-1$
```
chanseokoh: This is the container name we see in the package explorer. There seems little value of translating this name "Google APIs"?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -178,17 +160,6 @@ private static IClasspathEntry computeEntry(IJavaProject javaProject, Library li
         Messages.getString("computing.entries"), //$NON-NLS-1$
         11);
```
chanseokoh: nit: 11 --> 1

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
chanseokoh: Oh, there won't be such libraries, as `loadLibraryList` will filter them out since `CloudLibraries.getLibrary(libraryId) == null`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/messages.properties
elharo: "Google Cloud Client Libraries for Java" is a very specific group that does not include the Objectify or Endpoints libraries.  Not sure whether that matters here or not.

briandealwis: Thoughts on this?  Here are a few options:
  - Google Client APIs for Java
  - GCP Libraries for Java Projects
  - Java APIs for GCP Projects

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
briandealwis: > Care to explain?

Not really :-) I kept/introduced them thinking that we may want to introduce other instances of this classpath container. Removing these paths.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
briandealwis: The ordering may change in a RTL language.

briandealwis:
> Looks like initially ticking existing libraries or removing them doesn't work for Maven projects. That may require more PRs, I guess.

Filed #2567.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/MessagesTest.java
elharo: We have to put them in one container. Do we have to put them in one library though? I.e. can we have multiple libraries attached to the same container?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/messages.properties
elharo: As I said, APIs is overloaded and we shouldn't use it if we can help it.
Maybe just "Google Libraries for GCP" or "Google Cloud Platform Libraries."
I'm sort of mixing together libraries for running on GCP (Objectify, endpoints) with libraries that access GCP.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryContainerStateLocationProvider.java
```diff
@@ -41,4 +41,14 @@
    */
   IPath getContainerStateFile(IJavaProject javaProject, String id, boolean create)
       throws CoreException;
+
+  /**
+   * Remove the state file if present.
+   * 
+   * @param javaProject
```
elharo: add description  or remove @param

plugins/com.google.cloud.tools.eclipse.ui/src/com/google/cloud/tools/eclipse/ui/util/images/SharedImages.java
```diff
@@ -21,11 +21,17 @@
...
+  public static final ImageDescriptor GCP_IMAGE_DESCRIPTOR = AbstractUIPlugin
+      .imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui", "icons/obj16/gcp.png"); //$NON-NLS-1$ //$NON-NLS-2$
+
   public static final ImageDescriptor CLOUDSDK_IMAGE_DESCRIPTOR =
-      AbstractUIPlugin.imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui",
-          "icons/obj16/cloudsdk.png");
+      AbstractUIPlugin.imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui", //$NON-NLS-1$
+          "icons/obj16/cloudsdk.png"); //$NON-NLS-1$
+  public static final ImageDescriptor GCP_WIZARD_IMAGE_DESCRIPTOR = AbstractUIPlugin
+      .imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui", "icons/wizban/gcp.png"); //$NON-NLS-1$ //$NON-NLS-2$
```
elharo: wizban? wizard banner?

briandealwis: Yeah, it's the term used in the eclipse plugins

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
briandealwis: This should probably be renamed to "Google Cloud Platform Libraries" then

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/messages.properties
briandealwis: I like "Google Cloud Platform Libraries".

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/MessagesTest.java
briandealwis: Do you mean: "do we have to edit them all at once in one screen"?  Although the container page is a wizard page, the `ClasspathContainerWizard` only supports a single page.

briandealwis:
Hmm I can't reproduce with an App Engine Standard Java 8 project, @elharo.  What kind of project did you create?

(But I did find a different bug where the initially selected libraries is listed again, so I've added another test)

elharo:
App Engine Standard Java 8 project, no maven. defaults everywhere except for objectify

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/MessagesTest.java
elharo: Prior to this PR we have two libraries here that both load into the same Google APIs classpath container:

image

I'm wondering if it's possible to keep that arrangement? Are we required to have a 1:1 relationship between libraries and classpath containers?

chanseokoh:
> I got an NPE when I tried to add Cloud DNS on a freshly created App Engine standard Java8 native project.

I can reproduce this at the old commit https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2553/commits/cf05dde879572e4a8e724c7975a178ced8f1158d but not at the latest update. Looks like this is fixed.

briandealwis:
Ah, so @elharo is adding a new _Google Cloud Platform Libraries_ container where there is already a _Google Cloud Platform Libraries_ container.  JDT doesn't provide any way to prevent this: you can, for example, add a second _JRE System Library_ to your project.  Adding a new _Web App Libraries_ container fails, though it seems to be a side0effect rather than anything the `WebAppLibrariesContainerPage` does deliberately.

Although we're instructed to create a new classpath entry (called with `setSelection(null)`), since we really only support a single container, there's no reason that we can't just load up the existing list.

plugins/com.google.cloud.tools.eclipse.ui/src/com/google/cloud/tools/eclipse/ui/util/images/SharedImages.java
```diff
@@ -21,11 +21,17 @@
...
 public class SharedImages {
 
+  public static final ImageDescriptor GCP_IMAGE_DESCRIPTOR = AbstractUIPlugin
+      .imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui", "icons/obj16/gcp.png"); //$NON-NLS-1$ //$NON-NLS-2$
+
   public static final ImageDescriptor CLOUDSDK_IMAGE_DESCRIPTOR =
-      AbstractUIPlugin.imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui",
-          "icons/obj16/cloudsdk.png");
+      AbstractUIPlugin.imageDescriptorFromPlugin("com.google.cloud.tools.eclipse.ui", //$NON-NLS-1$
+          "icons/obj16/cloudsdk.png"); //$NON-NLS-1$
```
chanseokoh: nit: +1 blank

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryContainerStateLocationProvider.java
```diff
@@ -41,4 +41,14 @@
    */
   IPath getContainerStateFile(IJavaProject javaProject, String id, boolean create)
       throws CoreException;
+
+  /**
+   * Remove the state file if present.
+   * 
+   * @param javaProject the project the {@link LibraryClasspathContainer} belongs to
+   * @param id the file name/ID for the {@link LibraryClasspathContainer} state file
```
chanseokoh: nit: the param name different from the actual name

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -175,17 +178,31 @@ private File getContainerStateFile(IJavaProject javaProject, String fileId, bool
...
-      if (!containerFile.exists() && create) {
-        containerFile.create(new ByteArrayInputStream(new byte[0]), true, null);
+      IFile containerFile = folder.getFile(id + ".container"); //$NON-NLS-1$
+      return containerFile;
+    }
+
+    @Override
+    public void removeContainerStateFile(IJavaProject javaProject, String id) throws CoreException {
+      IFile stateFile = getFile(javaProject, id, false);
+      if (stateFile != null && stateFile.exists()) {
```
chanseokoh: nit: I don't think `getFile()` can ever return null, but not a big deal.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -120,22 +122,23 @@ public LibraryClasspathContainer loadContainer(IJavaProject javaProject, IPath c
...
+    if (librariesFile == null) {
+      logger.warning("Master libraries file cannot be created, save failed"); //$NON-NLS-1$
       return;
     }
-    try (Writer out = Files.newBufferedWriter(stateFile.toPath(), StandardCharsets.UTF_8)) {
+    try (Writer out = Files.newBufferedWriter(librariesFile.toPath(), StandardCharsets.UTF_8)) {
       out.write(gson.toJson(libraryIds.toArray()));
     }
+    // delete the container state cache file since the library list has changed
+    stateLocationProvider.removeContainerStateFile(javaProject, CloudLibraries.MASTER_CONTAINER_ID);
```
chanseokoh: This may be the only place of this MASTER constant? I wonder if we can git rid of the constant to de-couple it from this class so that this class can be general to deal with any container. Or, is the class already the master container-specific?

briandealwis:
ok, pushed up a fix for @elharo's add-new-container problems.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java
```diff
@@ -58,30 +58,33 @@
   private final ListenerList/* <ISelectedChangeListener> */ listeners = new ListenerList/* <> */();
 
   public LibrarySelectorGroup(Composite parentContainer, String groupName) {
-    this(parentContainer, groupName, true);
+    this(parentContainer, groupName, Messages.getString("appengine.libraries.group"), true); //$NON-NLS-1$
```
chanseokoh: Would you rename the key to remove `appengine`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/plugin.properties
```diff
@@ -1,5 +1,4 @@
 pluginName=Cloud Tools For Eclipse App Engine Libraries
 providerName=Google Inc.
 
-appEngineLibrariesTitle = App Engine Libraries
-googleCloudClientLibrariesTitle = Google Cloud Client Libraries for Java
\ No newline at end of file
+googleCloudPlatformLibrariesTitle = Google Cloud Platform Libraries
```
elharo: nit: probably don't want space around =

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java
briandealwis: I'll push this up to the callers.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
briandealwis: I'm not happy with this access here either, but it seemed the easiest solution without some bigger refactoring.  The issue here is that the file locations are controlled by an internal class, the `LibraryContainerStateLocationProvider`, so the outsiders can't know where these files.

The roles and responsibilities of the objects involved have become blurry. `BuildPath` collects library dependencies, but we access definitions via `CloudLibraries`.  `BuildPath` configures classpath entries, but resolution is done by `LibraryClasspathContainerResolverService`.  `LibraryClasspathContainerSerializer` shouldn't need to knows where to persist its data, it should just know how to persist data to a stream or at least to a provided file path.  But it needs to know this information as it's also responsible for creating the actual JDT container instance.  I think this could benefit from a rethink.

In the meantime, the only place that persists the library list is in `BuildPath#saveLibraryList()` (and a test), so I think it would make sense to move this `removeContainerState()` to there.

chanseokoh: OK, I always fail to follow what's happening with library classes. Sounds fair.

briandealwis: did some refactoring, but more reorganization will need to wait for another day.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/LibraryClasspathContainer.java
```diff
@@ -30,6 +30,8 @@
   private final String description;
   private final List<IClasspathEntry> classpathEntries;
   private final List<LibraryFile> libraryFiles;
+  public static final String CONTAINER_PATH_PREFIX =
+  "com.google.cloud.tools.eclipse.appengine.libraries";
```
chanseokoh: nit: indent +2

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/persistence/LibraryClasspathContainerSerializer.java
```diff
@@ -118,24 +120,33 @@ public LibraryClasspathContainer loadContainer(IJavaProject javaProject, IPath c
     }
   }
 
+  public void resetContainer(IJavaProject javaProject, IPath containerPath)
+      throws IOException, CoreException {
```
chanseokoh: nit: unnecessary`IOException`

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -18,106 +18,197 @@
...
-  private final String group;
-
-  protected CloudLibrariesPage(String group) {
-    super(group + "-page"); //$NON-NLS-1$
-    setTitle(Messages.getString(group + "-title"));  //$NON-NLS-1$
-    setDescription(Messages.getString(group + "-description"));  //$NON-NLS-1$
-    setImageDescriptor(AppEngineImages.appEngine(64));
-    this.group = group;
+  private boolean isMavenProject;
+  private IClasspathEntry oldEntry;
```
chanseokoh: `oldEntry` is assigned but never referenced. Remove it?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -18,106 +18,197 @@
...
-  public IClasspathEntry[] getNewContainers() {
-    List<Library> libraries = new ArrayList<>(librariesSelector.getSelectedLibraries());
-    if (libraries == null || libraries.isEmpty()) {
-      return null;
+  public void setSelection(IClasspathEntry containerEntry) {
+    // Although null means "new entry", we only support a single entry
+    // and since we use a fixed container path, we treat null as edit-current-entry
+    // FIXME: might be useful to read in current dependencies for maven projects
+    oldEntry = containerEntry;
+    if (project == null || isMavenProject) {
```
chanseokoh: `project` can never be null at this point. `project` --> `oldEntry`? Even so, there is no need for a class field.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -18,106 +18,197 @@
...
-        Library masterLibrary = BuildPath.collectLibraryFiles(project, libraries, monitor.newChild(7));
-        IClasspathEntry masterEntry = BuildPath.listNativeLibrary(project, masterLibrary, monitor.newChild(3));
-        if (masterEntry != null) {
-          return new IClasspathEntry[] {masterEntry};
-        } else {
-          return new IClasspathEntry[0];
-        }
-      }
+      List<Library> savedLibraries =
+          BuildPath.loadLibraryList(project, new NullProgressMonitor());
```
chanseokoh: nit: fits in one line.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -18,106 +18,197 @@
...
   }
 
-  @Override
-  public IClasspathEntry getSelection() {
-    // Since this class implements IClasspathContainerPageExtension2,
-    // Eclipse calls getNewContainers instead.
-    logger.log(Level.WARNING, "Unexpected call to getSelection()");
-    return null;
+  @VisibleForTesting
+  List<Library> getVisibleLibraries() {
```
chanseokoh: This is used only in testing. You can remove this and do

```java
    Predicate<Control> isObjectify = new Predicate<Control>() {
      @Override
      public boolean apply(Control control) {
        return control instanceof Button && ((Button) control).getText().equals("Objectify");
      }
    };
    assertNotNull(CompositeUtil.findControl(shellTestResource.getShell(), isObjectify));
```
or
```java
    assertNull(CompositeUtil.findControl(shellTestResource.getShell(), isObjectify));
```

chanseokoh: You should be able to remove `CloudLibrariesGroup.getAvailableLibraries()` too.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -18,106 +18,197 @@
...
-    
+    setSelectedLibraries(initialSelection);
+    composite.setLayout(new RowLayout(SWT.HORIZONTAL));
     setControl(composite);
   }
 
   @Override
   public boolean finish() {
-    return true;
+    final List<Library> libraries = getSelectedLibraries();
```
chanseokoh: nit: remove `final`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -18,106 +18,197 @@
...
   }
 
-  @Override
-  public void initialize(IJavaProject project, IClasspathEntry[] currentEntries) {
-    this.project = project;
+  @VisibleForTesting
+  void setSelectedLibraries(List<Library> selectedLibraries) {
+    initialSelection = new ArrayList<>(selectedLibraries);
+    List<Library> remaining = new ArrayList<>(selectedLibraries);
+    if (!librariesSelectors.isEmpty()) {
```
chanseokoh: I think this check is unnecessary and actually logging a warning in that case is desirable?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -239,4 +199,56 @@ public void done(IJobChangeEvent event) {
...
+    }
+    return selectedLibraries;
+  }
+
+  /**
+   * Save the list of library dependencies for this project.
+   */
+  public static void saveLibraryList(IJavaProject project, List<Library> libraries,
+      IProgressMonitor monitor) throws CoreException {
+    SubMonitor progress = SubMonitor.convert(monitor, libraries.size() + 10);
```
chanseokoh: nit: +5 instead of +10?

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: Gone and good riddance.  I can't remember why I had it.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: That SWT grovelling code is a bit icky, don't you think?

But I can rework the tests to just look for the library group rather than objectify specifically.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: Hmm, there was a reason, but I don't recollect and the tests aren't failing.  The comment is also out of date, so I'll toss that too.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: It extends to column 102.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
briandealwis: Ok.  This should never happen, and there should be warnings as part of the classpath container initializer.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
briandealwis: Forgot to add a `worked()` below.

briandealwis:
It turns out fixing #2523 is [pretty easy](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2553/commits/bc19868a6a77ed7fb517e772dd7b2b2837af5eb0).

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
```diff
@@ -178,12 +178,16 @@ public void setSelection(IClasspathEntry containerEntry) {
...
       return;
     }
-    try {
-      List<Library> savedLibraries = BuildPath.loadLibraryList(project, new NullProgressMonitor());
-      setSelectedLibraries(savedLibraries);
-    } catch (CoreException ex) {
-      logger.log(Level.WARNING,
-          "Error loading selected library IDs for " + project.getElementName(), ex); //$NON-NLS-1$
+    if (containerEntry != null) {
+      // ignore persisted library lists if this is a new container
```
chanseokoh: Cool!

I think the location of the comment is confusing though.

briandealwis: I have to back this out: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2553#issuecomment-344054071

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/CloudLibrariesPage.java
chanseokoh: Sorry, I actually meant just removing the this `if` statement rather than removing `remaining` and the related logging, but NVM.

Now, it is really true we don't need this `if`, because the `for` loop will just exit right away.

chanseokoh:
I believe you can now remove `CloudLibrariesGroup.getAvailableLibraries()` too.

PR #2559, opened 2017-11-02
install YEdit with CT4E

elharo:
@briandealwis This installs YEdit along with CT4E. You suggested making this an optional separate install instead. So far I haven't figured out the magic code to make that happen. Suggestions? Do I need to make a new feature parallel to features/com.google.cloud.tools.eclipse.suite.e45.feature/?


briandealwis:
To make it appear as a separate feature you'll need to reference in the `org.dadacoalition.yedit.feature` to the `metadata.p2.inf` file in a similar fashion as the `com.google.cloud.tools.eclipse.suite.e45.feature` feature. (Normally we would just add it to the `category.xml` file but we wanted to customise the category metadata to include a version.)

elharo:
Not getting the categories right yet:

image

image

gcp-repo/metadata.p2.inf
```diff
@@ -41,3 +41,27 @@ units.1.requires.1.greedy=true
 units.1.provides.1.namespace=org.eclipse.equinox.p2.iu
 units.1.provides.1.name=com.google.cloud.tools.eclipse.category.definition
 units.1.provides.1.version=1.0.0
+
+# Define the YEdit' category IU
```
chanseokoh: unnecessary single quote

gcp-repo/metadata.p2.inf
```diff
@@ -41,3 +41,27 @@ units.1.requires.1.greedy=true
...
+units.2.id=com.google.cloud.tools.eclipse.category
+units.2.version=$version$
+units.2.properties.1.name=org.eclipse.equinox.p2.type.category
+units.2.properties.1.value=true
+units.2.properties.2.name=org.eclipse.equinox.p2.name
+units.2.properties.2.value=YEdit
+units.2.properties.3.name=org.eclipse.equinox.p2.provider
+units.2.properties.3.value=Google Inc.
```
chanseokoh: Shouldn't be Google.

gcp-repo/metadata.p2.inf
```diff
@@ -41,3 +41,27 @@ units.1.requires.1.greedy=true
...
+units.2.properties.4.name=org.eclipse.equinox.p2.contact
+units.2.properties.4.value=https://github.com/oyse/yedit
+units.2.properties.5.name=org.eclipse.equinox.p2.doc.url
+units.2.properties.5.value=https://github.com/oyse/yedit/wiki
+units.2.licenses.0 =\
+YEdit is made available under the Eclipse Public License
+units.2.licenses.0.location=https://www.eclipse.org/legal/epl-v10.html
+units.2.requires.1.namespace=org.eclipse.equinox.p2.iu
+units.2.requires.1.name=org.dadacoalition.yedit.feature
```
chanseokoh: From above this should be `.yedit.feature.feature.group` instead of `.yedit.feature`?

gcp-repo/metadata.p2.inf
elharo: I'm not sure. It's coming out of our repo now. It depends on exactly what "provider"means here. Per docs, "A property key (value "org.eclipse.equinox.p2.provider") for a String property containing information about the vendor or provider of the installable unit."

features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
```diff
@@ -29,7 +29,6 @@
       <import feature="org.eclipse.wst.web_ui.feature" version="3.7.1.v201602111638" match="greaterOrEqual"/>
    </requires>
 
-   <!-- appengine-plugins-core -->
```
briandealwis: This comment is a pain in the neck as any change using the feature editor results in its being thrown away.

elharo:
seems mostly working now

image

features/com.google.cloud.tools.eclipse.suite.e45.feature/feature.xml
elharo: I wondered why that happened. Let's just let it go.

elharo:
However I can't seem to install just one. Selecting either one alone installs both.

elharo:
Also, the yedit bundle is not signed. Should we sign it with our own key or...?

gcp-repo/metadata.p2.inf
briandealwis: I think Orbit would count as a provider, since they re-create the .jar as a bundle.  Since we're just redistributing, my 2¢ is that it should be _YEdit Project_ (as per the feature).

But we can sign it: a jar can have multiple signers.

briandealwis:
> However I can't seem to install just one. Selecting either one alone installs both.

How are you seeing this?  I can select just one or the other and proceed without having the other selected.

screen shot 2017-11-03 at 3 39 19 pm

gcp-repo/metadata.p2.inf
```diff
@@ -56,7 +56,7 @@ units.2.properties.1.value=true
 units.2.properties.2.name=org.eclipse.equinox.p2.name
 units.2.properties.2.value=YEdit
 units.2.properties.3.name=org.eclipse.equinox.p2.provider
-units.2.properties.3.value=Google Inc.
+units.2.properties.3.value=Orbit
```
briandealwis: Hmm, this doesn't seem right: we're pulling this from YEdit directly.  It's not from Orbit.

elharo:
Maybe I just needed to restart. It's working for me now.

gcp-repo/metadata.p2.inf
briandealwis: ping

elharo: What do you think this should be?

briandealwis: I think _YEdit Project_ as per the [YEdit feature](https://github.com/oyse/yedit/blob/master/org.dadacoalition.yedit.feature/feature.xml#L6).

elharo: Done

PR #2561, opened 2017-11-03
Upgrade Google APIs and libraries to 1.23.0

chanseokoh:
We don't have to merge this until we get comfortable upgrading them. Putting up as a PR for reference and record.


chanseokoh:
I've tested all the API usages I can think of: GCS bucket creation/check, App Engine application check, account/project selector, login/logout, Dataflow-enabled project check. I am going to merge this soon.

PR #2565, opened 2017-11-07
Update com.google.api versions in MANIFEST.MF

chanseokoh:
Something I noticed after #2561.

Probably there was a reason we pinned a specific version range `[1.22.0,1.23.0)` (and `[1.23.0,1.24.0)` after #2561). I just noticed many got slipped, probably because the quick-fix to add missing import packages does not cap the upper limit. @briandealwis any comments?


chanseokoh:
Oops, builds are failing.

chanseokoh:
Shouldn't have touched the version for exporting.

Issue #1374, opened 2017-02-09
Cache project query result


akerekes:
Databinding can trigger several queries very quickly, which in a slower connection could cause bad UX so we could add a caching mechanism to the project retrieval logic. This cache would be refreshed whenever the user clicks on the [Refresh button](#1352) or when we explicitly want to trigger it.

chanseokoh:
Just to be sure, the project selector itself is independent of databinding (like the current `AccountSelector`), right? It's the deploy preferences that will want to wire up databinding, right?

akerekes:
Currently the project selector depends on another class to provide the list of projects like a repository, and the caching could be added to this repository.

chanseokoh:
Another question: if we hadn't used databinding, could we avoid introducing this caching? Or, will this caching still be desired even in the absent of databiniding?

akerekes:
I haven't investigated it as currently databinding saves us from writing a lot of code.

akerekes:
The two calls to `projectRepository.getProjects()`:
 - when databinding is added to the account selector, the selection change listener triggers the project list query corresponding to the account selected
 - to display validation errors when we open the dialog, we call `DataBindingContext.updateTargets()` which while triggers the validation and in case of error the error message to be displayed, but also triggers the querying of the project list.

We may be able to avoid the second query if the account selector only fires the selection changed event if there's a change to the selection.

PR #2568, opened 2017-11-08
Service key and project ID emulation settings tab for App Engine local run

chanseokoh:
Only the UI, which is even not visible (not wired). No validation yet —what works is just selecting/retrieving/entering/saving values through UI. The service key path input is particularly incomplete; the _browse..._ button does nothing.

A lot of part is straightforward. However, there is one trick to make things work reasonably sound in the case where an account and a project ID are saved in a run config but the account is not logged in; it was especially tricky to make it right in the run config dialog.

selection_001

It's loading projects synchronously, but it's not so bad even with a lot of projects. If we do it async in the future, I think we should be very careful. Unlike the deploy panel, I can see many potential pitfalls when doing async in the run config dialog. It's easy to go into the current mess of the Dataflow run config.


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/messages.properties
```diff
@@ -38,3 +38,10 @@ devappserver.not.available=There is no App Engine development server available
 cannot.copy.appengine.api.sdk.jar=Cannot copy appengine-api-1.0-sdk.jar
 server.already.in.operation=Server is already in operation
 server.port=server port: {0,number,\#}
+
+gcp.emulation.tab.name=GCP
```
elharo: Cloud Platform

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
```diff
@@ -0,0 +1,237 @@
...
+  @Override
+  public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
+    // No particular default values to set in a newly created configuration.
+  }
+
+  @Override
+  public void initializeFrom(ILaunchConfiguration configuration) {
+    accountEmailModel = getAttribute(configuration, ATTRIBUTE_ACCOUNT_EMAIL, ""); //$NON-NLS-1$
+    gcpProjectIdModel = getAttribute(configuration, ATTRIBUTE_GCP_PROJECT, ""); //$NON-NLS-1$
+    String serviceKey = getAttribute(configuration, ATTRIBUTE_SERVICE_KEY, ""); //$NON-NLS-1$
```
briandealwis: So what happens if the user also adds a `GOOGLE_APPLICATION_CREDENTIALS` or `GOOGLE_CLOUD_PROJECT ` value to the environment tab?  Which wins?

One approach would be for this tab to actually store and retrieve its values from the launch config's environment data.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
```diff
@@ -0,0 +1,237 @@
...
+
+  @VisibleForTesting
+  GcpEmulationTab(IGoogleLoginService loginService, ProjectRepository projectRepository) {
+    this.loginService = loginService;
+    this.projectRepository = projectRepository;
+  }
+
+  @Override
+  public void createControl(Composite parent) {
+    Composite composite = new Composite(parent, SWT.NONE);
```
briandealwis: I wonder if this, and dataflow, should instead have a read-only representation with an _Edit…_ button that pops this up as a dialog.  That would avoid the dynamic background updates etc.

chanseokoh: That's an attractive idea. I actually thought about having a read-only text field for the project (the account is just for the project selector), which can display the currently saved project even if the user is not logged in. Then we have a "Select Project..." button next to it, which pops up the project selector dialog. I really like this idea considering the benefits from the implementation perspective. The UI is fast because there won't be a long-running project fetch, no async issue in the run config, and a very clean model in the case of users not logged in. @elharo what do you think of this approach?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
```diff
@@ -0,0 +1,237 @@
...
+import org.eclipse.swt.events.ModifyEvent;
+import org.eclipse.swt.events.ModifyListener;
+import org.eclipse.swt.graphics.Image;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.Text;
+import org.eclipse.ui.PlatformUI;
+
+public class GcpEmulationTab extends AbstractLaunchConfigurationTab {
```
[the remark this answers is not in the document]
chanseokoh: I agree the name sounds a bit odd. I did thought out `GcpProjectTab`, but that sounded too broad, where this tab only deals with providing the service key and the project ID for the convenience of users' local development. Any more suggestions?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
chanseokoh: I have a comprehensive plan, which I am already doing for Dataflow — it already has an account selector, whose login info is being passed as `GOOGLE_APPLICATION_CREDENTIALS`. Basically, I will do the following:

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/DataflowPipelineLaunchDelegate.java#L183

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
chanseokoh: @elharo likes the idea too, so we can seriously consider going that route. We agreed that we first move forward with this PR (hoping to be able to release this soon) and implement the idea later. I don't think it will be more difficult to do it later than doing it now; it will be changes only in the UI.

patflynn:
I'm noticing that this PR seems to focus on the App Engine standard use-case. I've brought up concerns about that strategy in the design doc that I think should be resolved before we go far down this path. 

The focus of this feature was for library users in vanilla (non app-engine std environments)

as a user looking at this UI I would ask myself a lot of questions that would require complex answers. That seems bad. 

e.g. 
- Why is it asking me for a project for local dev, it never used to?
- What's a service account? Why do I need one I'm using emulators?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
briandealwis: What are you going to do for standard Java launches?

If you store values in the environment, then I think you're done.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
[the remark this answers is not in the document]
briandealwis: GcpRuntimeCredentialsTab?

chanseokoh:
> I'm noticing that this PR seems to focus on the App Engine standard use-case.

In fact, we are only focusing on the App Engine local run now. This doesn't mean we ignore non-App Engine cases. We just do this for App Engine initially. However, this tab component is designed to be universal — it can be reused in other run configs (such as plain Java run configs). I'll keep this universality in mind.

> as a user looking at this UI I would ask myself a lot of questions that would require complex answers.

I've asked the same question in the design doc. My comment is on the phrase "for selecting the project ID". We can continue our discussion there. I also push for adding something for clarity.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
chanseokoh: > What are you going to do for standard Java launches?

I think we can just do the same thing. We anyway have to extend the standard Java launches to add this tab. I just hope we can also extend the launch so that we can extract values from our tab.

> If you store values in the environment, then I think you're done.

That's true, but I'm afraid programmatically adding environment values there may confuse users and actually costs us more complexity in implementation. I think doing this is much simpler, and I don't see any issue.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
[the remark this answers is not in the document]
chanseokoh: How about `GcpLocalRunTab`? It's not only for credentials; we have the project field.

chanseokoh:
> However, this tab component is designed to be universal — it can be reused in other run configs (such as plain Java run configs). I'll keep this universality in mind.

For further clarification, there is nothing App Engine-specific in this PR. The tab is for setting a service key and a project ID, which can be applied to any run config.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
chanseokoh: Oh, you actually meant adding env vars to the launch config attribute, right? I thought you meant adding env vars to the _Environments_ tab UI. I agre that is what we should do, now knowing that it seems not possible to extend the vanilla-Java launch delegate.

chanseokoh: > One approach would be for this tab to actually store and retrieve its values from the launch config's environment data.

After some fiddling, I realized this doesn't work nicely, unless we tightly interwire the _Environments_ tab and our tab. For example, suppose the current `GOOGLE_APPLICATION_CREDENTIALS` is `foo.json`. This value will be loaded in the service key field in our tab. The user will also be able to see it in the _Environments_ tab. Now, in the _Environments_ tab, suppose the user modifies this value to `bar.json`. When the user presses _Apply_, what happens is that the value is immediately reverted back to `foo.json`, because the framework will call `performApply()` of our tab as well (even though our tab is invisible). This will overwrite `bar.json` with `foo.json` that is in the service key text field. This problem hit me, as I closed the run config right way not knowing that the value was reverted.

Basically, it seems not a good idea for a tab to make changes to a model of a different tab, unless you are willing to update UIs of the other tab simultaneously. I don't think we want to extend the _Environments_ tab to hack this. It won't work in general cases either.

elharo:
Let's not overthink this. If both the GCP and environment tabs set the credentials/project ID, pick one when launching the server. It's not worth extra effort to keep them in sync.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpEmulationTab.java
briandealwis: You'll need to go a bit deeper and implement the `activated()` and `deactivated()` APIs.  That way you'll know when to ignore your text field.

chanseokoh: OK, that sounded like a bit of abusing the tab framework, but looks like we can utilize it to make things work reasonably acceptable. This results in some minute annoyances (confirming save even when values have restored to its original state, and another due to #2571), but I think they are pretty insignificant issues.

chanseokoh:
@briandealwis the last commit changes this to save/load values to/from environment variables. PTAL.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+    composite.addDisposeListener(new ImageDisposer(gcpIcon));
+  }
+
+  private void updateProjectSelector() {
+    Credential credential = accountSelector.getSelectedCredential();
+    if (credential == null) {
+      projectSelector.setProjects(new ArrayList<GcpProject>());
+    } else {
+      try {
+        List<GcpProject> gcpProjects = projectRepository.getProjects(credential);
```
briandealwis: This is potentially really slow.  Perhaps it should be wrapped in a `BusyIndicator`.  See `PipelineArgumentsTab#updatePipelineOptionsForm()`. (Or the callers should be wrapped.)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+          }
+        }
+      }
+    });
+
+    // Project row
+    Label projectLabel = new Label(composite, SWT.LEAD);
+    projectLabel.setText(Messages.getString("label.project")); //$NON-NLS-1$
+
+    Composite projectSelectorComposite = new Composite(composite, SWT.NONE);
```
briandealwis: I wonder if it makes sense to extract this into a `AccountProjectSelectionPanel` or something.  We use this in a couple of places, like the `AppEngineDeployDialog`.  And #1374 seems like it relevant here too — maybe we should turn `ProjectRepository` into an injectable service that caches the credential-projects for some period of time.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+    GridDataFactory.fillDefaults().span(2, 1).applyTo(projectSelectorComposite);
+    GridDataFactory.fillDefaults().applyTo(filterField);
+    GridDataFactory.fillDefaults().grab(true, false).hint(SWT.DEFAULT, 200)
+        .applyTo(projectSelector);
+    GridLayoutFactory.fillDefaults().spacing(0, 0).generateLayout(projectSelectorComposite);
+
+    setControl(composite);
+
+    gcpIcon = SharedImages.GCP_IMAGE_DESCRIPTOR.createImage();
+    composite.addDisposeListener(new ImageDisposer(gcpIcon));
```
briandealwis: You can do this by overriding `AbstractLaunchConfigurationTab.dispose()`

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+      return configuration.getAttribute(attribute, defaultValue);
+    } catch (CoreException e) {
+      logger.log(Level.WARNING, "Can't get value from launch configuration.", e); //$NON-NLS-1$
+      return defaultValue;
+    }
+  }
+
+  @VisibleForTesting
+  static Map<String, String> getEnvironmentMap(ILaunchConfiguration configuration) {
+    Map<String, String> emptyMap = new HashMap<>();
```
briandealwis: `Collections.emptyMap()`

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+  @Override
+  public void initializeFrom(ILaunchConfiguration configuration) {
+    accountEmailModel = getAttribute(configuration, ATTRIBUTE_ACCOUNT_EMAIL, ""); //$NON-NLS-1$
+
+    Map<String, String> environmentMap = getEnvironmentMap(configuration);
+    gcpProjectIdModel = Strings.nullToEmpty(environmentMap.get(PROJECT_ID_ENVIRONMENT_VARIABLE));
+    String serviceKey = Strings.nullToEmpty(environmentMap.get(SERVICE_KEY_ENVIRONMENT_VARIABLE));
+
+    initializingUiValues = true;
+    accountSelector.selectAccount(accountEmailModel);
```
briandealwis: I think it's worth mentioning here that the account-selection-event listener updates the project list and waits until complete.  Which isn't great for UI liveness, but …

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Indeed, we may decide to do so in the future. For now, the deploy dialog uses databinding, and things work quite differently there (e.g., auto-selecting a single account, it has a refresh button, it has the App Engine application checking and error messages, it has the project creation link, etc), so the project selector there is very specialized. So, this would require a substantial work to factor out, so we'd probably want to consider it when the next time we need the bare minimum combination in this PR.

The cache idea is nice too.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+      @Override
+      public void run() {
+        updateProjectSelector();
+
+        if (!initializingUiValues) {
+          boolean emptySelection = accountSelector.getSelectedEmail().isEmpty();
+          boolean savedEmailAvailable = accountSelector.isEmailAvailable(accountEmailModel);
+          if (!emptySelection || savedEmailAvailable) {
+            accountEmailModel = accountSelector.getSelectedEmail();
+            gcpProjectIdModel = ""; //$NON-NLS-1$
```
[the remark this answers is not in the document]
chanseokoh: I've fiddled with this issue for quite some time. That behavior would be desirable, but if is not possible to know ahead if the user will ever change back to the previous account to restore the preserved project, or the user really wanted to change the account and then just let the project not selected (in the latter case, we shouldn't provide a wrong project ID when actually running the server.) So, I think this is what we should do, unfortunately.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+      }
+      if (!serviceKeyInput.getText().isEmpty()) {
+        environmentMap.put(SERVICE_KEY_ENVIRONMENT_VARIABLE, serviceKeyInput.getText());
+      } else {
+        environmentMap.remove(SERVICE_KEY_ENVIRONMENT_VARIABLE);
+      }
+      configuration.setAttribute(ILaunchManager.ATTR_ENVIRONMENT_VARIABLES, environmentMap);
+
+      if (environmentTab != null) {
+        // Needed to make the "Environment" tab to re-initialize its UI from the changes made here.
```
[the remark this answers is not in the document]
chanseokoh: It's not clear if it is a bug. Its comment says it won't do anything when activated/deactivated, which is OK from their perspective because everything just works when isolated. It's not working because we are modifying their model in a different tab.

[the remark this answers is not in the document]
briandealwis: But it's not _their_ model: it's an `ILaunchConfiguration` that is shared across the tabs.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Turns out, the tab lifecycle doesn't seem to match SWT lifecycle here.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+        environmentMap.remove(PROJECT_ID_ENVIRONMENT_VARIABLE);
+      }
+      if (!serviceKeyInput.getText().isEmpty()) {
+        environmentMap.put(SERVICE_KEY_ENVIRONMENT_VARIABLE, serviceKeyInput.getText());
+      } else {
+        environmentMap.remove(SERVICE_KEY_ENVIRONMENT_VARIABLE);
+      }
+      configuration.setAttribute(ILaunchManager.ATTR_ENVIRONMENT_VARIABLES, environmentMap);
+
+      if (environmentTab != null) {
```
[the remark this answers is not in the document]
chanseokoh: That makes sense. But `deactivated()` always calls `performApply()`, so there is no change anyway. Also, I think doing it here is on the safe side and fine.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Oh, right, we need a mutable map.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,290 @@
...
+    // Account row
+    new Label(composite, SWT.LEAD).setText(Messages.getString("label.account")); //$NON-NLS-1$
+    accountSelector = new AccountSelector(composite, loginService);
+    accountSelector.addSelectionListener(new Runnable() {
+      @Override
+      public void run() {
+        updateProjectSelector();
+
+        if (!initializingUiValues) {
+          boolean emptySelection = accountSelector.getSelectedEmail().isEmpty();
```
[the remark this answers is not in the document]
chanseokoh: Just realized `somethingSelected` works better here.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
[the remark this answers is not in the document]
chanseokoh: Filed: https://bugs.eclipse.org/bugs/show_bug.cgi?id=527142

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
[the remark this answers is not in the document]
chanseokoh: Actually, I think it's enough to move it into `deactivated()`. `performApply()` is called too frequently, and I can't think of any case where we have to update the environment tab before deactivation.

chanseokoh:
I believe all the comments are addressed. PTAL.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
briandealwis: What does this mean — what is the effect? Because with this `ImageDisposer`, `gcpIcon` is still left pointing at an invalid object.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+  // We set up intermediary models between a run configuration and UI components for certain values,
+  // because, e.g., the account selector cannot load an email if it is not logged in. In such a
+  // case, although nothing is selected in the account selector, we should not clear the email saved
+  // in the run configuration.
+  private String accountEmailModel;
+  private String gcpProjectIdModel;
+  // To prevent updating above models when programmatically setting up UI components.
+  private boolean initializingUiValues;
+
+  private boolean activated;  // to avoid https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2568#discussion_r150128582
```
briandealwis:
```
/**
 * true if this tab is the currently visible tab
 * @see #performApply(ILaunchConfigurationWorkingCopy) for details
 */
`

File: plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java [hunk 15]
```diff
@@ -0,0 +1,304 @@
...
+    // Account row
+    new Label(composite, SWT.LEAD).setText(Messages.getString("label.account")); //$NON-NLS-1$
+    accountSelector = new AccountSelector(composite, loginService);
+    accountSelector.addSelectionListener(new Runnable() {
+      @Override
+      public void run() {
+        updateProjectSelector();
+
+        if (!initializingUiValues) {
+          boolean somethingSelected = !accountSelector.getSelectedEmail().isEmpty();
```
briandealwis: `accountSelected`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+          boolean savedEmailAvailable = accountSelector.isEmailAvailable(accountEmailModel);
+          if (somethingSelected || savedEmailAvailable) {
+            accountEmailModel = accountSelector.getSelectedEmail();
```
briandealwis: This doesn't seem right: if no account is currently selected, but our saved-email (`accountEmailModel`) is available, then we throw away our saved-email.

Ohhh! I see: `savedEmailAvailable==true` means the selector has been populated, and so we now update the `accountEmailModel` to keep it in sync with changes.  A comment would help here, and below for the project selector.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+    Composite projectSelectorComposite = new Composite(composite, SWT.NONE);
+    final Text filterField = new Text(projectSelectorComposite,
+        SWT.BORDER | SWT.SEARCH | SWT.ICON_SEARCH | SWT.ICON_CANCEL);
+
+    projectSelector = new ProjectSelector(projectSelectorComposite);
+    projectSelector.addSelectionChangedListener(new ISelectionChangedListener() {
+      @Override
+      public void selectionChanged(SelectionChangedEvent event) {
+        if (!initializingUiValues) {
+          boolean somethingSelected = !projectSelector.getSelectProjectId().isEmpty();
```
briandealwis: `projectSelected`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  private void updateProjectSelector() {
+    final Credential credential = accountSelector.getSelectedCredential();
+    if (credential == null) {
+      projectSelector.setProjects(new ArrayList<GcpProject>());
+      return;
+    }
+
+    BusyIndicator.showWhile(Display.getCurrent(), new Runnable() {
```
briandealwis: Should be able to use `serviceKeyInput.getDisplay()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+  @Override
+  public void initializeFrom(ILaunchConfiguration configuration) {
+    accountEmailModel = getAttribute(configuration, ATTRIBUTE_ACCOUNT_EMAIL, ""); //$NON-NLS-1$
+
+    Map<String, String> environmentMap = getEnvironmentMap(configuration);
+    gcpProjectIdModel = Strings.nullToEmpty(environmentMap.get(PROJECT_ID_ENVIRONMENT_VARIABLE));
+    String serviceKey = Strings.nullToEmpty(environmentMap.get(SERVICE_KEY_ENVIRONMENT_VARIABLE));
+
+    initializingUiValues = true;
+    // Selecting an account loads projects into the project selector synchronously (via a listener).
```
briandealwis: Should this be wrapped in a `BusyIndicator`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+    accountSelector.selectAccount(accountEmailModel);
+    projectSelector.selectProjectId(gcpProjectIdModel);
+    serviceKeyInput.setText(serviceKey);
+    initializingUiValues = false;
+  }
+
+  @Override
+  public void performApply(ILaunchConfigurationWorkingCopy configuration) {
+    if (!activated) {  // to avoid https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2568#discussion_r150128582
```
briandealwis: It's probably best to include a potted summary here, in case these projects are moved elsewhere, or Github disappears.

_must avoid updating the environment map unless we're active as performApply() is when the user makes changes in other tabs, like the EnvironmentTab, and thus we could clobber changes made in the EnvironmentTab_

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -0,0 +1,304 @@
...
+      return configuration.getAttribute(attribute, defaultValue);
+    } catch (CoreException e) {
+      logger.log(Level.WARNING, "Can't get value from launch configuration.", e); //$NON-NLS-1$
+      return defaultValue;
+    }
+  }
+
+  @VisibleForTesting
+  static Map<String, String> getEnvironmentMap(ILaunchConfiguration configuration) {
+    Map<String, String> emptyMap = new HashMap<>();  // should be mutable
```
briandealwis: This comment is unexpected?  An explanation would be useful

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Right. This covers the case where users explicitly de-selected any selection. In this case, we toss the saved value. (Account deselection is not implemented yet, but we do have an issue for this enhancement). I'll add a comment.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Selecting an account itself should be instant. It is the project selector that makes network calls. So, I put `BusyIndicator` around the project selector listener.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Wanted to signify we shouldn't use `Collections.emptyMap()` which is immutable. I'll update the comment.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Did some more testing. Looks like the tab's `dispose()` is matched with class instantiation, which is different that SWT's `dispose()` is matched with `createControl()`. I finally decided to make `gcpIcon` instance-scoped; `gcpIcon` is now a `final` field, created only once when the class is created.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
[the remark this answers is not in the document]
briandealwis: Ah, I meant:

  1. I select account A1 and project P
  2. I then select A2 which has access to P

P should stay selected.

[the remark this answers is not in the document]
chanseokoh: The case is not so different, and it is difficult to make it right, because it is assumed that the UI and the model are always in sync. Suppose the user changed to A1.5 between A1 and A2. If we could magically anticipate that the user will eventually change to A2 later, we could not clear P. However, if we do not clear P when selecting A1.5 and the user applies the change and exits, we will be wrongly providing P at runtime.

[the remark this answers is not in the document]
briandealwis: I'm not suggesting retaining the last-selected project when switching A1 &rarr; A1.5 &rarr; A2.  It makes sense that it's cleared.

That said, I'm not sure how often this will occur in practice.  It doesn't happen with any of my current projects.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -68,6 +67,8 @@
   private static final String SERVICE_KEY_ENVIRONMENT_VARIABLE =
       "GOOGLE_APPLICATION_CREDENTIALS"; //$NON-NLS-1$
 
+  private final Image gcpIcon = SharedImages.GCP_IMAGE_DESCRIPTOR.createImage();
```
briandealwis: This approach isn't a good practice as the instance may be created but never used (and hence the image is created unnecessarily), and there's no guarantee that the instance will be created on the SWT thread.  Better would be for `getImage()` to create the image lazily and `dispose()` to do a check-if-null.

chanseokoh: OK, that makes sense. BTW, the image doesn't need to be created on the UI thread.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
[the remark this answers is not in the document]
chanseokoh: OK, let's leave this as is. To make your scenario work, the account selector listener has to be coupled with the project selector (because it can only determine the situation after all the projects of A2 are loaded), which doesn't sound worth the effort.

In the end, all the complications and shortcomings will be eliminated by #2571.

PR #2573, opened 2017-11-10
Fix Guava import version

chanseokoh:


chanseokoh:
I remember @briandealwis added some kind of test a while ago (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/pom.xml#L447), but I guess this isn't really working.

briandealwis:
It is frustrating.  The rule is succeeding as it tests the version that is resolved rather than the potential versions that can be resolved.  We're resolving to Guava 20.0, which satisfies both the package constraint (`version="15.0.0"`) and passes the dependency rule.

elharo:
Then let's add a more straightforward test that simply reads the files directly rather than trying to resolve the dependency.

chanseokoh:
The test is not flexible, but it should serve its purpose. One caveat is that not every bundle has `PluginXmlTest`. (Some bundles don't have `plugin.xml`.)

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -189,6 +192,35 @@ public final void testBundleVendor() throws IOException {
...
+    Pattern pattern = Pattern.compile(regexPrefix + "[^;,]+");
+
+    Matcher matcher = pattern.matcher(value);
+    while (matcher.find()) {
+      int nextCharOffset = matcher.end();
+      if (nextCharOffset == value.length()) {
+        fail("Import-Package version not defined");
+      }
+      String stringAfterMatch = value.substring(nextCharOffset);
+      if (!stringAfterMatch.startsWith(";version=" + versionString)) {
```
briandealwis: Hmm, this doesn't seem to match the callers above?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -189,6 +192,35 @@ public final void testBundleVendor() throws IOException {
...
+  private void checkVersionStringsInManifestMf(
+      String value, String prefixToCheck, String versionString) {
+    String regexPrefix = prefixToCheck.replaceAll(".", "\\\\.");
+    Pattern pattern = Pattern.compile(regexPrefix + "[^;,]+");
+
+    Matcher matcher = pattern.matcher(value);
+    while (matcher.find()) {
+      int nextCharOffset = matcher.end();
+      if (nextCharOffset == value.length()) {
+        fail("Import-Package version not defined");
```
briandealwis: This is called above for Require-Bundles too.

chanseokoh:
Oops, the errors due to pushing too prematurely to head home.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -189,6 +192,35 @@ public final void testBundleVendor() throws IOException {
     assertEquals("Google Inc.", vendor);
   }
 
+  @Test
+  public final void testGuavaImportVersions() throws IOException {
+    checkVersionStringsInManifestMf(
```
briandealwis: This looks really useful.  Could I suggest a refactoring?
  - call this `testDependencyVersions()`
  - add methods `checkImportPackageVersion(package-pattern, version-range)` and `checkRequireBundleVersion(bundle-name, version-range)` (e.g., `checkImportPackageVersion("com.google.common.*", "[20,21)")`) — embedding `"Require-Bundle"` or `"Import-Package"` and `"version"` vs `"bundle-version"` is error prone
  - use `org.osgi.framework.VersionRange#valueOf(String)` to parse and compare the version ranges
  - the patterns should allow `\s` for white-space

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -189,6 +192,35 @@ public final void testBundleVendor() throws IOException {
...
+  }
+
+  private void checkVersionStringsInManifestMf(
+      String attributeName, String prefixToCheck, String versionString) throws IOException {
+    String value = getManifestAttributes().getValue(attributeName);
+    if (value != null) {
+      String regexPrefix = prefixToCheck.replaceAll("\\.", "\\\\.");
+      Pattern pattern = Pattern.compile(regexPrefix + "[^;,]*");
+
+      Matcher matcher = pattern.matcher(value);
```
briandealwis: Might be easier to use `Splitter.on(',')` and then apply the pattern to each import directive?

chanseokoh: I would have gone that way if it worked. Commas can be in, e.g., `version="[20.0.0,21.0.0)"`.

briandealwis: Ah good point :-)

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
chanseokoh: Good idea. But I am not thrilled to spend more time on this to make it general and flexible ATM. If there is unnecessary white-space, let's just remove it and be consistent.

PR #2580, opened 2017-11-14
bump request size to 700 and paginate

### LINKED ISSUE #2578: Project deploy window does not list all the projects
In our Google Cloud organization we organized all our projects under folders and subfolders
https://cloud.google.com/resource-manager/docs/creating-managing-folders

Here is my configuration
```
Eclipse Java EE IDE for Web Developers.

Version: Oxygen.1a Release (4.7.1a)
Build id: 20171005-1200
```

```
Google Cloud Tools for Eclipse
Version: 1.4.1.201710261936

Copyright 2016-2017 Google Inc.
```

```
> gcloud version
Google Cloud SDK 179.0.0
app-engine-java 1.9.59
app-engine-python 1.9.62
bq 2.0.27
core 2017.11.06
gcloud
gsutil 4.28
```

When I open the "`Deploy to App Engine standard`" modal, the listed projects are related **ONLY to the root folder of the organization**. The projects in subfolder are not listed and I do not see any option that can allow to select a folder or subfolder.

For the moment I'm deploying the project using directly Maven goal, but this is missing functionality in the plugin.

----------------------------------------------------------------------------

**EDIT**: At first I thought that the problem is that all the projects inside folder are missing, but after a better look some of them are listed, so it does not seem an in-folder problem.

I am missing projects from the plugin list, projects that are inside a sub-folder or the root folder as well.

I have access to hundreds of projects, can the solution of this issue be simpler than the folder organization and simply the plugin list the first X projects instead all of them?

elharo:
fix #2578 
@chanseokoh Bump maximum number of projects to 700

Do we want to loop with the refresh token to make multiple requests instead?

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
+++ b/plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
@@ -38,7 +38,7 @@ import java.util.List;
  */
 public class ProjectRepository {
 
-  private static final int PROJECT_LIST_PAGESIZE = 300;
+  private static final int PROJECT_LIST_PAGESIZE = 700;
   private static final String PROJECT_DELETE_REQUESTED = "DELETE_REQUESTED";
 
   private final IGoogleApiFactory apiFactory;
--- a/plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
+++ b/plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
@@ -38,7 +38,10 @@ import java.util.List;
  */
 public class ProjectRepository {
 
+  // API may return fewer than 700 results in one request, but 600 and change is the
+  // most we've seen in any one account to date.
   private static final int PROJECT_LIST_PAGESIZE = 700;
+  
   private static final String PROJECT_DELETE_REQUESTED = "DELETE_REQUESTED";
 
   private final IGoogleApiFactory apiFactory;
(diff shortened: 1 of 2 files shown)


chanseokoh:
Pagination would be a desirable enhancement, but it will require quite some work. It should also take into account the filter input. Not even sure if that's even possible. We may end up having to query the entire projects to make the filter working?

briandealwis:
I remember writing code that loops — it's in `GcpProjectServicesJob`. I wonder at what point will the remote API force pagination?

elharo:
May not be related but when I used my corp account I saw this:

image

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
```diff
@@ -38,7 +38,10 @@
  */
 public class ProjectRepository {
 
-  private static final int PROJECT_LIST_PAGESIZE = 300;
+  // API may return fewer than 700 results in one request, but 600 and change is the
```
chanseokoh: I can't understand the second part of the comment clearly.

briandealwis: `600 and change` is slang for "600-odd"

briandealwis:
I get the same _error occurred while retrieving…_ when selecting the `clout-tools-for-eclipse-kokoro` project.

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
chanseokoh: Ah, thanks.

chanseokoh:
If we are talking about the "An error occurred while ... : Operation not allowed", that's because you don't have a permission (probably App Engine-related) on that project.

elharo:
Yes, that's the error. Unrelated to this PR I think.

I do wonder if perhaps we should filter out those projects from the list.

briandealwis:
Although it would be great to filter them out, we can only check one project at a time.  That would not be great with 600-odd projects.

I opened #2582

chanseokoh:
Tests are failing. They look relevant to the recent changes.

Issue #2430, opened 2017-09-27
Mini project selector and GCS bucket field enabled when using defaults in Dataflow launch config


chanseokoh:
This is after fixing #2410.

If the _Use default Dataflow options_ is checked, all the controls are supposed to be disabled. However, the mini project combo selector is enabled, and this is fairly confusing. The GCS bucket field is (mostly?) disabled initially, but it can be made enabled when ticking/unticking the checkbox.

![selection_014](https://user-images.githubusercontent.com/10523105/30932515-b37bdcfe-a395-11e7-8acb-4070d9181471.png)

briandealwis:
/me bangs head on wall

This situation arises as loading the default from the preferences into the `RunOptionsDefaultsComponent` which sets the project, which triggers a verification job. When the verification job completes, it triggers a validation of the `RunOptionsDefaultsComponent`, which seeing a valid account and project, re-enables the staging location.

That will account for the periodic test failures in #2378.

briandealwis:
I realize that with #2375 the toggle-to-use-defaults doesn't actually make sense as the defaults may be incomplete.  The check button should really be a push button to _Load Defaults_.

chanseokoh:
The enablement of `RunOptionsDefaultsComponent` is currently broken, and I see it will be more so after #3044. It's really time to fix this.

chanseokoh:
The [i2430-correct-enablement branch](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/i2430-correct-enablement) is a fix, but I need to devise a test.

PR #2585, opened 2017-11-15
Reload pipeline descriptions as required

### LINKED ISSUE #2232: Warning logged in Dataflow QA
```
Pipeline Options Type com.google.cloud.dataflow.sdk.options.DirectPipelineOptions not found in PipelineOptions type hierarchy in project dataflow; Hierarchy Focus: PipelineOptions [in PipelineOptions.class [in org.apache.beam.sdk.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
Super types:
  HasDisplayData [in HasDisplayData.class [in org.apache.beam.sdk.transforms.display [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
Sub types:
  ApplicationNameOptions [in ApplicationNameOptions.class [in org.apache.beam.sdk.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
    BigQueryOptions [in BigQueryOptions.class [in org.apache.beam.sdk.io.gcp.bigquery [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    CloudResourceManagerOptions [in CloudResourceManagerOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    DirectOptions [in DirectOptions.class [in org.apache.beam.runners.direct [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-direct-java/2.0.0/beam-runners-direct-java-2.0.0.jar]]]
    GcsOptions [in GcsOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    PubsubOptions [in PubsubOptions.class [in org.apache.beam.sdk.io.gcp.pubsub [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    StreamingOptions [in StreamingOptions.class [in org.apache.beam.sdk.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
      BigQueryOptions [in BigQueryOptions.class [in org.apache.beam.sdk.io.gcp.bigquery [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
        DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      CloudResourceManagerOptions [in CloudResourceManagerOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      PubsubOptions [in PubsubOptions.class [in org.apache.beam.sdk.io.gcp.pubsub [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
        DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  BigQueryOptions [in BigQueryOptions.class [in org.apache.beam.sdk.io.gcp.bigquery [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  CloudResourceManagerOptions [in CloudResourceManagerOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
  DataflowPipelineDebugOptions [in DataflowPipelineDebugOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  DataflowPipelineWorkerPoolOptions [in DataflowPipelineWorkerPoolOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  DataflowWorkerLoggingOptions [in DataflowWorkerLoggingOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  DirectOptions [in DirectOptions.class [in org.apache.beam.runners.direct [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-direct-java/2.0.0/beam-runners-direct-java-2.0.0.jar]]]
  GcpOptions [in GcpOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
    BigQueryOptions [in BigQueryOptions.class [in org.apache.beam.sdk.io.gcp.bigquery [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    CloudResourceManagerOptions [in CloudResourceManagerOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    GcsOptions [in GcsOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    PubsubOptions [in PubsubOptions.class [in org.apache.beam.sdk.io.gcp.pubsub [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  GcsOptions [in GcsOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  GoogleApiDebugOptions [in GoogleApiDebugOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
    GcpOptions [in GcpOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
      BigQueryOptions [in BigQueryOptions.class [in org.apache.beam.sdk.io.gcp.bigquery [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
        DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      CloudResourceManagerOptions [in CloudResourceManagerOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      GcsOptions [in GcsOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
        DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      PubsubOptions [in PubsubOptions.class [in org.apache.beam.sdk.io.gcp.pubsub [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
        DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
          TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  PubsubOptions [in PubsubOptions.class [in org.apache.beam.sdk.io.gcp.pubsub [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  StreamingOptions [in StreamingOptions.class [in org.apache.beam.sdk.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
    BigQueryOptions [in BigQueryOptions.class [in org.apache.beam.sdk.io.gcp.bigquery [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    CloudResourceManagerOptions [in CloudResourceManagerOptions.class [in org.apache.beam.sdk.extensions.gcp.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-extensions-google-cloud-platform-core/2.0.0/beam-sdks-java-extensions-google-cloud-platform-core-2.0.0.jar]]]
    DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
      TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
    PubsubOptions [in PubsubOptions.class [in org.apache.beam.sdk.io.gcp.pubsub [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-io-google-cloud-platform/2.0.0/beam-sdks-java-io-google-cloud-platform-2.0.0.jar]]]
      DataflowPipelineOptions [in DataflowPipelineOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        DataflowWorkerHarnessOptions [in DataflowWorkerHarnessOptions.class [in org.apache.beam.runners.dataflow.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
        TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  TestPipelineOptions [in TestPipelineOptions.class [in org.apache.beam.sdk.testing [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
    TestDataflowPipelineOptions [in TestDataflowPipelineOptions.class [in org.apache.beam.runners.dataflow [in /Users/elharo/.m2/repository/org/apache/beam/beam-runners-google-cloud-dataflow-java/2.0.0/beam-runners-google-cloud-dataflow-java-2.0.0.jar]]]
  UnknownPipelineOptions [in ProxyInvocationHandler$UnknownPipelineOptions.class [in org.apache.beam.sdk.options [in /Users/elharo/.m2/repository/org/apache/beam/beam-sdks-java-core/2.0.0/beam-sdks-java-core-2.0.0.jar]]]
```

briandealwis:
A configuration tab's `isValid()` may be called even when the tab is not active (i.e., after a `deactivated()`/`performApply()` and — more importantly — before an `activated()`/`initializeFrom()`), and it may be called frequently.  This PR improves loading and detecting changed launch configuration details.

Fixes #2232

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/MajorVersionPipelineRunnerConfigurationTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/MajorVersionPipelineRunnerConfigurationTest.java
@@ -16,18 +16,24 @@
 
 package com.google.cloud.tools.eclipse.dataflow.core.launcher;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
 
 import com.google.cloud.tools.eclipse.dataflow.core.project.MajorVersion;
 import java.util.Arrays;
 import java.util.Set;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.debug.core.ILaunchConfiguration;
 import org.hamcrest.CoreMatchers;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 
 /**
  * Test that each {@link MajorVersion} has associated configuration settings in the
@@ -66,4 +72,21 @@ public class MajorVersionPipelineRunnerConfigurationTest {
     Set<PipelineRunner> runners = PipelineRunner.inMajorVersion(majorVersion);
     assertThat(runners, CoreMatchers.hasItem(defaultRunner));
   }
+
+  /** Ensure the default runner chosen is appropriate for the majorVersion. */
+  @Test
+  public void testFromLaunchConfiguration_defaultRunner() throws CoreException {
+    ILaunchConfiguration empty = mock(ILaunchConfiguration.class, new Answer() {
+      @Override
+      public Object answer(InvocationOnMock invocation) throws Throwable {
+        // for getAttribute(attr, default)
+        Object[] arguments = invocation.getArguments();
+        return arguments.length == 2 ? arguments[1] : null;
+      }
+    });
+    PipelineLaunchConfiguration lc =
+        PipelineLaunchConfiguration.fromLaunchConfiguration(empty, majorVersion);
+    assertEquals(PipelineLaunchConfiguration.defaultRunner(majorVersion), lc.getRunner());
+  }
+
 }
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.eclipse.dataflow.core.launcher;
 import static com.google.cloud.tools.eclipse.dataflow.core.launcher.PipelineRunner.DIRECT_PIPELINE_RUNNER;
 import static com.google.cloud.tools.eclipse.dataflow.core.launcher.PipelineRunner.DIRECT_RUNNER;
 
-import com.google.cloud.tools.eclipse.dataflow.core.DataflowCorePlugin;
 import com.google.cloud.tools.eclipse.dataflow.core.launcher.options.PipelineOptionsHierarchy;
 import com.google.cloud.tools.eclipse.dataflow.core.launcher.options.PipelineOptionsProperty;
 import com.google.cloud.tools.eclipse.dataflow.core.launcher.options.PipelineOptionsType;
@@ -38,7 +37,6 @@ import java.util.Set;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
-import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 
 /**
  * A POJO that contains options specific to launching a dataflow pipeline.
@@ -49,8 +47,6 @@ public class PipelineLaunchConfiguration {
    */
   static final Set<String> PROVIDED_PROPERTY_NAMES = ImmutableSet.of("runner");
 
-  private Optional<String> eclipseProjectName;
-
   private boolean useDefaultLaunchOptions;
 
   private PipelineRunner runner;
@@ -58,17 +54,23 @@ public class PipelineLaunchConfiguration {
   private Optional<String> userOptionsName;
 
   /**
-   * Construct a DataflowPipelineLaunchConfiguration from the provided {@link ILaunchConfiguration}.
+   * Construct a DataflowPipelineLaunchConfiguration from the provided {@link ILaunchConfiguration}
+   * using the provided project version to determine suitable defaults.
    */
   public static PipelineLaunchConfiguration fromLaunchConfiguration(
-      ILaunchConfiguration launchConfiguration) throws CoreException {
-    PipelineLaunchConfiguration configuration = createDefault();
+      ILaunchConfiguration launchConfiguration, MajorVersion projectVersion) throws CoreException {
+    PipelineLaunchConfiguration configuration =
+        new PipelineLaunchConfiguration(defaultRunner(projectVersion));
     configuration.setValuesFromLaunchConfiguration(launchConfiguration);
     return configuration;
   }
 
   public static PipelineLaunchConfiguration createDefault() {
-    return new PipelineLaunchConfiguration(defaultRunner(MajorVersion.ONE));
+    return createDefault(MajorVersion.ONE);
+  }
+
+  public static PipelineLaunchConfiguration createDefault(MajorVersion version) {
+    return new PipelineLaunchConfiguration(defaultRunner(version));
   }
 
   public static PipelineRunner defaultRunner(MajorVersion majorVersion) {
@@ -87,7 +89,6 @@ public class PipelineLaunchConfiguration {
     this.useDefaultLaunchOptions = true;
 
     this.userOptionsName = Optional.absent();
-    this.eclipseProjectName = Optional.absent();
   }
 
   public PipelineRunner getRunner() {
@@ -122,16 +123,6 @@ public class PipelineLaunchConfiguration {
     this.useDefaultLaunchOptions = useDefaultLaunchOptions;
   }
 
-  public String getEclipseProjectName() {
-    return eclipseProjectName.orNull();
-  }
-
-  private void setEclipseProjectName(ILaunchConfiguration configuration) throws CoreException {
-    eclipseProjectName = Optional.fromNullable(
-        configuration.getAttribute(
-            IJavaLaunchConfigurationConstants.ATTR_PROJECT_NAME, (String) null));
-  }
-
   private void setValuesFromLaunchConfiguration(ILaunchConfiguration configuration)
       throws CoreException {
     setRunner(PipelineRunner.fromRunnerName(configuration.getAttribute(
@@ -146,8 +137,6 @@ public class PipelineLaunchConfiguration {
 
     setUserOptionsName(configuration.getAttribute(
         PipelineConfigurationAttr.USER_OPTIONS_NAME.toString(), getUserOptionsName()));
-
-    setEclipseProjectName(configuration);
   }
 
   /**
@@ -163,13 +152,6 @@ public class PipelineLaunchConfiguration {
         PipelineConfigurationAttr.ALL_ARGUMENT_VALUES.toString(), argumentValues);
     configuration.setAttribute(
         PipelineConfigurationAttr.USER_OPTIONS_NAME.toString(), userOptionsName.orNull());
-
-    try {
-      setEclipseProjectName(configuration);
-    } catch (CoreException e) {
-      DataflowCorePlugin.logWarning("CoreException while trying to retrieve"
-          + " project name from Configuration Working Copy");
-    }
   }
 
   /**
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
@@ -33,6 +33,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.debug.core.ILaunchConfiguration;
@@ -272,6 +273,24 @@ public class PipelineLaunchConfiguration {
     } else {
       return !Strings.isNullOrEmpty(argumentValues.get(propertyName));
     }
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    } else if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    PipelineLaunchConfiguration other = (PipelineLaunchConfiguration) obj;
+    return Objects.equals(argumentValues, other.argumentValues)
+        && Objects.equals(runner, other.runner)
+        && Objects.equals(userOptionsName, other.userOptionsName);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(argumentValues, runner, userOptionsName);
   }
 
   @Override
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.eclipse.dataflow.ui.launcher;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Matchers.anyString;
@@ -31,6 +32,9 @@ import com.google.cloud.tools.eclipse.test.util.ui.CompositeUtil;
 import com.google.cloud.tools.eclipse.test.util.ui.ShellTestResource;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Lists;
+import java.lang.reflect.InvocationTargetException;
+import java.util.HashMap;
+import java.util.Map;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.runtime.CoreException;
@@ -87,6 +91,32 @@ public class PipelineArgumentsTabTest {
       tab.initializeFrom(configuration);  // Should not throw NPE.
 
       ProjectUtils.waitForProjects();  // Suppress some non-terminated-job error logs
+    }
+
+    @Test
+    public void testReinitialize()
+        throws CoreException, InvocationTargetException, InterruptedException {
+      IWorkspaceRoot workspaceRoot = mock(IWorkspaceRoot.class);
+      when(workspaceRoot.getProject(anyString())).thenReturn(mock(IProject.class));
+      ILaunchConfigurationDialog dialog = mock(ILaunchConfigurationDialog.class);
+
+      ILaunchConfiguration configuration1 = mock(ILaunchConfiguration.class);
+      Map<String, Object> attributes1 = new HashMap<String, Object>();
+      when(configuration1.getAttributes()).thenReturn(attributes1);
+      ILaunchConfiguration configuration2 = mock(ILaunchConfiguration.class);
+      Map<String, Object> attributes2 = new HashMap<String, Object>();
+      when(configuration2.getAttributes()).thenReturn(attributes2);
+
+      PipelineArgumentsTab tab = new PipelineArgumentsTab(workspaceRoot);
+      tab.setLaunchConfigurationDialog(dialog);
+      tab.createControl(shellResource.getShell());
+
+      assertTrue(tab.reinitialize(configuration1));
+      assertFalse("cache should be ok", tab.reinitialize(configuration1));
+      attributes1.put("foo", "bar");
+      assertTrue("config changed, cache should be discarded", tab.reinitialize(configuration1));
+      assertTrue("different config; cache should be discarded", tab.reinitialize(configuration2));
+      assertTrue("config changed, cache should be discarded", tab.reinitialize(configuration1));
     }
   }
 
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
@@ -20,8 +20,13 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.anyBoolean;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyMap;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -39,6 +44,7 @@ import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.debug.ui.ILaunchConfigurationDialog;
 import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 import org.eclipse.swt.widgets.Button;
@@ -53,6 +59,8 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 import org.junit.runners.Suite;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 
 @RunWith(Suite.class)
 @Suite.SuiteClasses({
@@ -94,30 +102,54 @@ public class PipelineArgumentsTabTest {
     }
 
     @Test
-    public void testReinitialize()
+    public void testReload()
         throws CoreException, InvocationTargetException, InterruptedException {
       IWorkspaceRoot workspaceRoot = mock(IWorkspaceRoot.class);
       when(workspaceRoot.getProject(anyString())).thenReturn(mock(IProject.class));
       ILaunchConfigurationDialog dialog = mock(ILaunchConfigurationDialog.class);
 
-      ILaunchConfiguration configuration1 = mock(ILaunchConfiguration.class);
-      Map<String, Object> attributes1 = new HashMap<String, Object>();
-      when(configuration1.getAttributes()).thenReturn(attributes1);
-      ILaunchConfiguration configuration2 = mock(ILaunchConfiguration.class);
-      Map<String, Object> attributes2 = new HashMap<String, Object>();
-      when(configuration2.getAttributes()).thenReturn(attributes2);
+      ILaunchConfigurationWorkingCopy configuration1 = mockLaunchConfiguration();
+      ILaunchConfigurationWorkingCopy configuration2 = mockLaunchConfiguration();
 
       PipelineArgumentsTab tab = new PipelineArgumentsTab(workspaceRoot);
       tab.setLaunchConfigurationDialog(dialog);
       tab.createControl(shellResource.getShell());
 
-      assertTrue(tab.reinitialize(configuration1));
-      assertFalse("cache should be ok", tab.reinitialize(configuration1));
-      attributes1.put("foo", "bar");
-      assertTrue("config changed, cache should be discarded", tab.reinitialize(configuration1));
-      assertTrue("different config; cache should be discarded", tab.reinitialize(configuration2));
-      assertTrue("config changed, cache should be discarded", tab.reinitialize(configuration1));
+      assertTrue(tab.reload(configuration1));
+      assertFalse("cache should be ok", tab.reload(configuration1)); //$NON-NLS-1$
+      configuration1.setAttribute(PipelineConfigurationAttr.USER_OPTIONS_NAME.toString(), "bar"); //$NON-NLS-1$
+      assertTrue("config changed, cache should be discarded", tab.reload(configuration1)); //$NON-NLS-1$
+      assertTrue("different config; cache should be discarded", tab.reload(configuration2)); //$NON-NLS-1$
+      assertTrue("config changed, cache should be discarded", tab.reload(configuration1)); //$NON-NLS-1$
     }
+  }
+
+  private static ILaunchConfigurationWorkingCopy mockLaunchConfiguration() throws CoreException {
+    final Map<String, Object> attributes = new HashMap<>();
+    Answer<Object> getAttributes = new Answer<Object>() {
+      @Override
+      public Object answer(InvocationOnMock invocation) throws Throwable {
+        String key = invocation.getArgumentAt(0, String.class);
+        return attributes.containsKey(key) ? attributes.get(key) : invocation.getArgumentAt(1, Object.class);
+      }};
+    Answer<Void> setAttributes = new Answer<Void>() {
+      @Override
+      public Void answer(InvocationOnMock invocation) throws Throwable {
+        attributes.put(invocation.getArgumentAt(0, String.class),
+            invocation.getArgumentAt(1, Object.class));
+        return null;
+      }};
+    ILaunchConfigurationWorkingCopy launch = mock(ILaunchConfigurationWorkingCopy.class);
+    doReturn(attributes).when(launch).getAttributes();
+    doAnswer(getAttributes).when(launch).getAttribute(anyString(), anyString());
+    doAnswer(getAttributes).when(launch).getAttribute(anyString(), anyInt());
+    doAnswer(getAttributes).when(launch).getAttribute(anyString(), anyBoolean());
+    doAnswer(getAttributes).when(launch).getAttribute(anyString(), anyMap());
+    doAnswer(setAttributes).when(launch).setAttribute(anyString(), anyString());
+    doAnswer(setAttributes).when(launch).setAttribute(anyString(), anyInt());
+    doAnswer(setAttributes).when(launch).setAttribute(anyString(), anyBoolean());
+    doAnswer(setAttributes).when(launch).setAttribute(anyString(), anyMap());
+    return launch;
   }
 
   @RunWith(Parameterized.class)
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
@@ -47,6 +47,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
@@ -83,11 +84,9 @@ import org.eclipse.ui.forms.events.IExpansionListener;
 /**
  * A tab specifying arguments required to run a Dataflow Pipeline.
  * 
- * As computing the pipeline options hierarchy can be expensive, we try to cache values.
- * {@link #reinitialize(ILaunchConfiguration)} is responsible for loading information derived from
- * an {@link ILaunchConfiguration}. We null out all computed information in
- * {@link #deactivated(ILaunchConfigurationWorkingCopy)}, called when the user switches to another
- * tab.
+ * Computing the pipeline options hierarchy can be expensive, so we try to avoid doing so.
+ * {@link #reload(ILaunchConfiguration)} is responsible for loading information derived from an
+ * {@link ILaunchConfiguration}.
  */
 public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   private static final Joiner MISSING_GROUP_MEMBER_JOINER = Joiner.on(", "); //$NON-NLS-1$
@@ -112,9 +111,8 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   private final PipelineOptionsHierarchyFactory pipelineOptionsHierarchyFactory =
       new ClasspathPipelineOptionsHierarchyFactory();
 
-  private ILaunchConfiguration currentConfiguration;
-  private int currentConfigurationHash;
   private IProject project;
+  private MajorVersion majorVersion;
   private PipelineLaunchConfiguration launchConfiguration;
 
   /*
@@ -124,7 +122,6 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
    * able.
    */
   private PipelineOptionsHierarchy hierarchy;
-
 
   public PipelineArgumentsTab() {
     this(ResourcesPlugin.getWorkspace().getRoot());
@@ -308,24 +305,11 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   }
 
   @Override
-  public void deactivated(ILaunchConfigurationWorkingCopy workingCopy) {
-    // super.deactivated() calls performApply(), saving the PipelineLaunchConfiguration
-    // values into the ILaunchConfigurationWorkingCopy.
-    super.deactivated(workingCopy);
-    // null out all computed values that were derived from the ILaunchConfiguration;
-    // will force reinitialize() to recompute all values
-    currentConfiguration = null;
-    launchConfiguration = null;
-    project = null;
-    hierarchy = null;
-  }
-
-  @Override
   public void initializeFrom(ILaunchConfiguration configuration) {
     try {
-      reinitialize(configuration);
+      reload(configuration);
 
-      updateRunnerButtons(getMajorVersion(project));
+      updateRunnerButtons(majorVersion);
 
       defaultOptionsComponent.setUseDefaultValues(launchConfiguration.isUseDefaultLaunchOptions());
       defaultOptionsComponent.setPreferences(getPreferences());
@@ -343,29 +327,29 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   }
 
   /**
-   * Reload computed information about the dataflow pipeline from the values in the given
-   * {@link ILaunchConfiguration}. As this is called from {{@link #isValid(ILaunchConfiguration)}},
-   * which is frequently called, try to avoid recomputing this information if unnecessary.
+   * Reload any computed information only if the launch configuration has changed in some meaningful
+   * way. This must be a fast check as this method is called from
+   * {{@link #isValid(ILaunchConfiguration)}}, which is called frequently.
    * 
    * @return true if values were reloaded, or false if the configuration was up-to-date
    */
   @VisibleForTesting
-  boolean reinitialize(ILaunchConfiguration configuration)
+  boolean reload(ILaunchConfiguration configuration)
       throws CoreException, InvocationTargetException, InterruptedException {
-    // assume that Map#hashCode() hashes its elements, and possibly its elements' elements
-    int configurationHash = configuration.getAttributes().hashCode();
-    if (currentConfiguration == configuration && currentConfigurationHash == configurationHash) {
+    // recompute the features of interest from the provided launch configuration
+    IProject project = findProject(configuration);
+    MajorVersion majorVersion = getMajorVersion(project);
+    PipelineLaunchConfiguration launchConfiguration =
+        PipelineLaunchConfiguration.fromLaunchConfiguration(configuration, majorVersion);
+    if (Objects.equals(project, this.project) && Objects.equals(majorVersion, this.majorVersion)
+        && Objects.equals(launchConfiguration, this.launchConfiguration)) {
+      // our features of interest are the same
       return false;
     }
-    // null out in case of exception
-    currentConfiguration = null;
-    project = findProject(configuration);
-    MajorVersion majorVersion = getMajorVersion(project);
-    launchConfiguration =
-        PipelineLaunchConfiguration.fromLaunchConfiguration(configuration, majorVersion);
-    updateHierarchy(majorVersion);
-    currentConfiguration = configuration;
-    currentConfigurationHash = configurationHash;
+    this.project = project;
+    this.majorVersion = majorVersion;
+    this.launchConfiguration = launchConfiguration;
+    updateHierarchy();
     return true;
   }
 
@@ -419,13 +403,13 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
    * @throws InterruptedException if the update is interrupted
    * @throws InvocationTargetException if an exception occurred during the update
    */
-  private void updateHierarchy(final MajorVersion majorVersion)
+  private void updateHierarchy()
       throws InvocationTargetException, InterruptedException {
     getLaunchConfigurationDialog().run(true, true, new IRunnableWithProgress() {
       @Override
       public void run(IProgressMonitor monitor)
           throws InvocationTargetException, InterruptedException {
-        hierarchy = getPipelineOptionsHierarchy(majorVersion, monitor);
+        hierarchy = getPipelineOptionsHierarchy(monitor);
       }
     });
   }
@@ -438,8 +422,7 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
     }
   }
 
-  private PipelineOptionsHierarchy getPipelineOptionsHierarchy(
-      MajorVersion majorVersion, IProgressMonitor monitor) {
+  private PipelineOptionsHierarchy getPipelineOptionsHierarchy(IProgressMonitor monitor) {
     if (project != null && project.isAccessible()) {
       try {
         return pipelineOptionsHierarchyFactory.forProject(project, majorVersion, monitor);
@@ -541,7 +524,7 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   @Override
   public boolean isValid(ILaunchConfiguration configuration) {
     try {
-      reinitialize(configuration);
+      reload(configuration);
       return validatePage();
     } catch (CoreException | InvocationTargetException | InterruptedException ex) {
       return false;
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
@@ -59,6 +59,7 @@ import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.debug.ui.AbstractLaunchConfigurationTab;
+import org.eclipse.jdt.launching.IJavaLaunchConfigurationConstants;
 import org.eclipse.jface.operation.IRunnableWithProgress;
 import org.eclipse.jface.window.Window;
 import org.eclipse.swt.SWT;
@@ -81,12 +82,19 @@ import org.eclipse.ui.forms.events.IExpansionListener;
 
 /**
  * A tab specifying arguments required to run a Dataflow Pipeline.
+ * 
+ * As computing the pipeline options hierarchy can be expensive, we try to cache values.
+ * {@link #reinitialize(ILaunchConfiguration)} is responsible for loading information derived from
+ * an {@link ILaunchConfiguration}. We null out all computed information in
+ * {@link #deactivated(ILaunchConfigurationWorkingCopy)}, called when the user switches to another
+ * tab.
  */
 public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   private static final Joiner MISSING_GROUP_MEMBER_JOINER = Joiner.on(", "); //$NON-NLS-1$
 
   private static final String ARGUMENTS_SEPARATOR = "="; //$NON-NLS-1$
 
+  private final IWorkspaceRoot workspaceRoot;
   private Executor displayExecutor;
 
   private ScrolledComposite composite;
@@ -100,11 +108,14 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   private TextAndButtonComponent userOptionsSelector;
   private PipelineOptionsFormComponent pipelineOptionsForm;
 
-  private PipelineLaunchConfiguration launchConfiguration;
-
   private final DataflowDependencyManager dependencyManager = DataflowDependencyManager.create();
   private final PipelineOptionsHierarchyFactory pipelineOptionsHierarchyFactory =
       new ClasspathPipelineOptionsHierarchyFactory();
+
+  private ILaunchConfiguration currentConfiguration;
+  private int currentConfigurationHash;
+  private IProject project;
+  private PipelineLaunchConfiguration launchConfiguration;
 
   /*
    * TODO: By default, this may include all PipelineOptions types, including custom user types that
@@ -114,7 +125,6 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
    */
   private PipelineOptionsHierarchy hierarchy;
 
-  private final IWorkspaceRoot workspaceRoot;
 
   public PipelineArgumentsTab() {
     this(ResourcesPlugin.getWorkspace().getRoot());
@@ -262,7 +272,9 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
 
   @Override
   public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
-    launchConfiguration = PipelineLaunchConfiguration.createDefault();
+    project = findProject(configuration);
+    MajorVersion version = getMajorVersion(project);
+    launchConfiguration = PipelineLaunchConfiguration.createDefault(version);
     launchConfiguration.toLaunchConfiguration(configuration);
   }
 
@@ -296,21 +308,24 @@ public class PipelineArgumentsTab extends AbstractLaunchConfigurationTab {
   }
 
   @Override
+  public void deactivated(ILaunchConfigurationWorkingCopy workingCopy) {
+    // super.deactivated() calls performApply(), saving the PipelineLaunchConfiguration
+    // values into the ILaunchConfigurationWorkingCopy.
+    super.deactivated(workingCopy);
+    // null out all computed values that were derived from the ILaunchConfiguration;
+    // will force reinitialize() to recompute all values
+    currentConfiguration = null;
+    launchConfiguration = null;
+    project = null;
+    hierarchy = null;
+  }
+
+  @Override
   public void initializeFrom(ILaunchConfiguration configuration) {
     try {
-      launchConfiguration = PipelineLaunchConfiguration.fromLaunchConfiguration(configuration);
+      reinitialize(configuration);
 
-      IProject project = getProject();
-      MajorVersion majorVersion = MajorVersion.ONE;
-      if (project != null && project.isAccessible()) {
-         majorVersion = dependencyManager.getProjectMajorVersion(project);
-         if (majorVersion == null) {
-            majorVersion = MajorVersion.ONE;
-         }
-      }
-
-      updateRunnerButtons(majorVersion);
-      updateHierarchy(majorVersion);
+      updateRunnerButtons(getMajorVersion(project));
 
       defaultOptionsComponent.setUseDefaultValues(launchConfiguration.isUseDefaultLaunchOptions());
       defaultOptionsComponent.setPreferences(getPreferences());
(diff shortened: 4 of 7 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/MajorVersionPipelineRunnerConfigurationTest.java
```diff
@@ -66,4 +72,21 @@ public void testDefaultRunnerInConfiguredRunners() {
...
+
+  /** Ensure the default runner chosen is appropriate for the majorVersion. */
+  @Test
+  public void testFromLaunchConfiguration_defaultRunner() throws CoreException {
+    ILaunchConfiguration empty = mock(ILaunchConfiguration.class, new Answer() {
+      @Override
+      public Object answer(InvocationOnMock invocation) throws Throwable {
+        // for getAttribute(attr, default)
+        Object[] arguments = invocation.getArguments();
+        return arguments.length == 2 ? arguments[1] : null;
```
chanseokoh: See if this works: `when(empty.getAttribute(anyObject(), anyObject())).thenAnswer(AdditionalAnswers.returnsSecondArg());`

chanseokoh:
Let's not close #2232. I know at least another case where #2232 happens, and I verified that this PR doesn't cover that case.

chanseokoh:
Oops, I was on a wrong branch. This seems to fix #2232.

chanseokoh:
But again, I was able to reproduce #2232 with https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2232#issuecomment-320368325.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
```diff
@@ -49,26 +48,30 @@
...
-   * Construct a DataflowPipelineLaunchConfiguration from the provided {@link ILaunchConfiguration}.
+   * Construct a DataflowPipelineLaunchConfiguration from the provided {@link ILaunchConfiguration}
+   * using the provided project version to determine suitable defaults.
    */
   public static PipelineLaunchConfiguration fromLaunchConfiguration(
-      ILaunchConfiguration launchConfiguration) throws CoreException {
-    PipelineLaunchConfiguration configuration = createDefault();
+      ILaunchConfiguration launchConfiguration, MajorVersion projectVersion) throws CoreException {
+    PipelineLaunchConfiguration configuration =
+        new PipelineLaunchConfiguration(defaultRunner(projectVersion));
```
chanseokoh: ` ... = createDefault(projectVersion)` looks simpler.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -262,7 +269,9 @@ public void clear() {
 
   @Override
   public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
-    launchConfiguration = PipelineLaunchConfiguration.createDefault();
+    project = findProject(configuration);
+    MajorVersion version = getMajorVersion(project);
```
chanseokoh: Set it to the field `majorVersion`?

plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/MajorVersionPipelineRunnerConfigurationTest.java
briandealwis: Ooo, that's much nicer.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
briandealwis: Not appropriate here — `setDefaults()` is called when a new launch configuration is created, e.g., using the create-new button from the Run/Debug Configurations dialog.  It's intended to pre-populate the launch config with suitable defaults.

briandealwis:
> I can still reproduce #2232 with #2232 (comment).

Hmm, me too.

briandealwis:
`updateHierarchy()` computes the hierarchy in the background, and isValid() either needs to wait for it *or* we can just return `false` since we're not in a valid state.
```
private void updateHierarchy()
      throws InvocationTargetException, InterruptedException {
      @Override
      public void run(IProgressMonitor monitor)
          throws InvocationTargetException, InterruptedException {
        hierarchy = getPipelineOptionsHierarchy(monitor);
      }
    });
  }
```

chanseokoh:
I thought we made `updateHierarchy()` synchronous. My cursory testing aligns with it. (For example, if I put `Thread.sleep(10000)` inside `run()`, `updateHierarchy()` returns only after 10 sec. Not clearly documented, but I think `run()` here runs synchronously, even though `fork` is true. Maybe we should just set if `false`. I believe making everything synchronous was the decision about #2289.

Also, I'd like to point out https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2232#issuecomment-320368325 is not because of concurrency. I left a comment back then: https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2232#issuecomment-323175689

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
chanseokoh: If so, is it OK that we save `project` just above (and `launchConfiguration` below)? It looks inconsistent that you save `project` and `launchConfiguration` while you don't for `majorVersion`.

chanseokoh:
Perhaps https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2232#issuecomment-320368325 is a bit different problem, and you might want to tackle it in a separate PR?

elharo:
Is this PR still in play or is it now moot?

chanseokoh:
I know this is still in play.

briandealwis:
PTAL: I've pushed up a substantial revision that addresses the above points.  The crux of these problem is that we were guessing a project's dataflow version and often getting it wrong.

Highlights of the changes:
  - `PipelineArgumentsTab` and `PipelineLaunchConfiguration` no longer record the pipeline runner until one has been selected
  - `PipelineArgumentsTab#setDefaults()` no longer sets default information
  - Added a new method `DataflowDependencyManager#getActualProjectMajorVersion()` as its `getProjectMajorVersion()` treats an absence as all possible versions, whereas we want to know an exact version.
  - Added support for visibly enabling/disabling the subcomponents of the `PipelineArgumentsTab` when provided an invalid project

chanseokoh:
I'll look into it, but it's going to take some time.

plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/MajorVersionPipelineRunnerConfigurationTest.java
```diff
@@ -66,4 +72,15 @@ public void testDefaultRunnerInConfiguredRunners() {
     Set<PipelineRunner> runners = PipelineRunner.inMajorVersion(majorVersion);
     assertThat(runners, CoreMatchers.hasItem(defaultRunner));
   }
+
+  @Test
+  public void testFromLaunchConfiguration_defaultRunner() throws CoreException {
+    ILaunchConfiguration empty =
+        mock(ILaunchConfiguration.class, AdditionalAnswers.returnsSecondArg());
+    IProject project = mock(IProject.class);
```
chanseokoh: This is unused. Also, could you rename `lc` to `pipelineLaunchConfig`?

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+    public void testSetDefaults() {
+      ILaunchConfigurationWorkingCopy configuration = mock(ILaunchConfigurationWorkingCopy.class);
+      pipelineArgumentsTab.setDefaults(configuration);
+      // we don't set any defaults
+      Mockito.verifyZeroInteractions(configuration);
+    }
+
+    @Test
+    public void testReload()
+        throws CoreException, InvocationTargetException, InterruptedException {
```
chanseokoh: unnecessary `throws`

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+      assertTrue("config changed, cache should be discarded", //$NON-NLS-1$
+          pipelineArgumentsTab.reload(configuration1));
+      assertTrue("different config; cache should be discarded", //$NON-NLS-1$
+          pipelineArgumentsTab.reload(configuration2));
+      assertTrue("config changed, cache should be discarded", //$NON-NLS-1$
+          pipelineArgumentsTab.reload(configuration1));
+    }
+
+    @Test
+    public void testInitializeFrom_accessibleProject() throws CoreException {
```
chanseokoh: unnecessary `throws`

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+
+      ProjectUtils.waitForProjects(); // Suppress some non-terminated-job error logs
     }
 
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2165
     @Test
-    public void testInitializeForm_noExceptionForNonAccessibleProject() throws CoreException {
-      IWorkspaceRoot workspaceRoot = mock(IWorkspaceRoot.class);
-      when(workspaceRoot.getProject(anyString())).thenReturn(mock(IProject.class));
+    public void testInitializeFrom_noExceptionForNonAccessibleProject() throws CoreException {
```
chanseokoh: Unnecessary `throws`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+      assertEquals(0, pipelineArgumentsTab.runnerButtons.size());
+      assertFalse(pipelineArgumentsTab.defaultOptionsComponent.isEnabled());
+      assertFalse(pipelineArgumentsTab.userOptionsSelector.isEnabled());
+
+      ProjectUtils.waitForProjects(); // Suppress some non-terminated-job error logs
+    }
+
+    @Test
+    public void testIsValid_errorOnNonDataflowProject()
+        throws CoreException, InvocationTargetException, InterruptedException {
```
chanseokoh: Unnecessary `throws`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+      doReturn(null).when(dependencyManager).getProjectMajorVersion(project1);
+
+      pipelineArgumentsTab.isValid(configuration1);
+      assertEquals("Project is not configured for Dataflow",
+          pipelineArgumentsTab.getErrorMessage());
+    }
+
+    @Test
+    public void testIsValid_errorOnNonProject()
+        throws CoreException, InvocationTargetException, InterruptedException {
```
chanseokoh: Unnecessary `throws`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -432,6 +432,11 @@ public void selectionChanged(SelectionChangedEvent event) {
     stagingLocationInput.addModifyListener(listener);
   }
 
+  public boolean isEnabled() {
```
chanseokoh: This is used only in tests. Probably don't have to be `public` but `@VisibleForTesting`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/page/component/TextAndButtonComponent.java
```diff
@@ -54,6 +53,11 @@ public void setText(String value) {
     text.setText(value);
   }
 
+  public boolean isEnabled() {
```
chanseokoh: This is used only in tests. Probably don't have to be `public` but `@VisibleForTesting`.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
```diff
@@ -122,20 +123,44 @@ public void setUseDefaultLaunchOptions(boolean useDefaultLaunchOptions) {
...
+    if (this == obj) {
+      return true;
+    } else if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    PipelineLaunchConfiguration other = (PipelineLaunchConfiguration) obj;
+    return Objects.equals(majorVersion, other.majorVersion)
+        && Objects.equals(argumentValues, other.argumentValues)
+        && Objects.equals(runner, other.runner)
+        && Objects.equals(userOptionsName, other.userOptionsName);
```
chanseokoh: Should we also compare the `useDefaultLaunchOptions` field?

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
```diff
@@ -122,20 +123,44 @@ public void setUseDefaultLaunchOptions(boolean useDefaultLaunchOptions) {
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(majorVersion, argumentValues, runner, userOptionsName);
```
chanseokoh: Should we factor in `useDefaultLaunchOptions`?

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
```diff
@@ -122,20 +123,44 @@ public void setUseDefaultLaunchOptions(boolean useDefaultLaunchOptions) {
...
-        configuration.getAttribute(
-            IJavaLaunchConfigurationConstants.ATTR_PROJECT_NAME, (String) null));
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(getClass())
+        .add("runnerName", runner != null ? runner.getRunnerName() : "(default runner)")
+        .add("argumentValues", argumentValues).toString();
   }
 
   private void setValuesFromLaunchConfiguration(ILaunchConfiguration configuration)
```
chanseokoh: Noticed this should just have been a constructor:
```java
  private PipelineLaunchConfiguration(MajorVersion majorVersion, ILaunchConfiguration configuration)
      throws CoreException {
    this.majorVersion = Preconditions.checkNotNull(majorVersion);

    // never set a default runner unless actually specified
```
I was hunting for a while who can call this when, but if we make this a constructor, it becomes obvious. `fromLaunchConfiguration` then could simply be
```java
  public static PipelineLaunchConfiguration fromLaunchConfiguration(MajorVersion majorVersion,
      ILaunchConfiguration launchConfiguration) throws CoreException {
    return new PipelineLaunchConfiguration(majorVersion, launchConfiguration);
  }
```
Maybe we should just ditch the static `fromLaunchConfiguration()`. This another-static-constructor method doesn't help me understand code, IMO.

But then, this refactoring doesn't have to be done in this PR.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
```diff
@@ -146,30 +171,23 @@ private void setValuesFromLaunchConfiguration(ILaunchConfiguration configuration
...
   /**
    * Stores the Dataflow Pipeline-specific options of this LaunchConfiguration inside the provided
    * {@link ILaunchConfigurationWorkingCopy}.
    */
   public void toLaunchConfiguration(ILaunchConfigurationWorkingCopy configuration) {
-    configuration.setAttribute(
-        PipelineConfigurationAttr.RUNNER_ARGUMENT.toString(), runner.getRunnerName());
+    if (runner != null) {
+      configuration.setAttribute(PipelineConfigurationAttr.RUNNER_ARGUMENT.toString(),
+          runner.getRunnerName());
```
chanseokoh: Note to myself: save a runner in LC after selecting a project, save, close, unset a project, save, close, and see if PLC is constructed without a runner.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManager.java
```diff
@@ -125,7 +125,23 @@ public boolean hasTrackedDataflowDependency(IProject project) {
         || Artifact.RELEASE_VERSION.equals(dependency.getVersion()));
   }
 
+  /**
+   * Returns a possible version range for the given project.
+   * 
+   * @throws IllegalStateException if the version range is invalid
```
chanseokoh: I think the Javadoc should be something callers can understand. Unclear to me as a caller when a range is invalid and how to avoid it.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManager.java
```diff
@@ -125,7 +125,23 @@ public boolean hasTrackedDataflowDependency(IProject project) {
...
   public VersionRange getDataflowVersionRange(IProject project) {
+    VersionRange actual = getActualDataflowVersionRange(project);
+    return actual != null ? actual : allVersions();
+  }
+
+  /**
+   * Returns the encoded version range for the given project or {@code null} if not found.
+   * 
+   * @return the version range or {@code null} if not found
+   * @throws IllegalStateException if the version range is invalid
```
chanseokoh: I think the Javadoc should be something callers can understand. Unclear to me as a caller when a range is invalid and how to avoid it.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+    private IProject project2;
+    private ILaunchConfigurationWorkingCopy configuration2;
+
+    @Before
+    public void setUp() throws CoreException {
+      workspaceRoot = mock(IWorkspaceRoot.class);
+      dependencyManager = mock(DataflowDependencyManager.class);
+
+      project1 = mockProject("project1");
+      when(workspaceRoot.getProject(eq("project1"))).thenReturn(project1);
```
chanseokoh: Nit: you don't need `eq` when all arguments are constants.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -61,35 +77,150 @@
...
+      configuration1 = mockLaunchConfiguration();
+      when(configuration1.getAttribute(eq(IJavaLaunchConfigurationConstants.ATTR_PROJECT_NAME),
+          anyString())).thenReturn("project1");
+      doReturn(MajorVersion.ONE).when(dependencyManager).getProjectMajorVersion(project1);
+
+      project2 = mockProject("project2");
+      when(workspaceRoot.getProject(eq("project2"))).thenReturn(project2);
```
chanseokoh: Nit: you don't need `eq` when all arguments are constants.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
chanseokoh: Hmm... maybe the reason for the static method was that the original author didn't like making a constructor throw an exception?

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManager.java
chanseokoh: Looking at the code a bit, I think `IllegalStateException` should never be thrown, unless our code is buggy. It's when, e.g., doing `.createFromVersionSpec("nonsensical string")`, but we are sane to always pass a well-formed version string.

briandealwis:
> @chanseokoh: I was following #2232 (comment), and when I hit Apply after entering a Dataflow project in the Main tab, I got this logged:

Oh geesh.  Ok, here the issue is that the user hasn't actually opened the Pipeline Arguments tab, so no changes have been made and thus there are no changes for this tab to apply.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
```diff
@@ -117,6 +118,15 @@ public void setPreferences(DataflowPreferences preferences) {
     }
   }
 
+  public boolean isEnabled() {
```
chanseokoh: Used only in tests. Maybe making it non-public?

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -298,50 +312,97 @@ PipelineRunner getSelectedRunner() {
...
+      updateHierarchy();
+      return true;
     } catch (CoreException | InvocationTargetException | InterruptedException ex) {
       // TODO: Handle
-      DataflowUiPlugin.logError(ex, 
-          "Error while initializing from existing configuration"); //$NON-NLS-1$
+      DataflowUiPlugin.logError(ex, "Error while initializing from existing configuration"); //$NON-NLS-1$
+      project = null;
+      launchConfiguration = null;
+      return true;
```
chanseokoh: Maybe `return false`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -352,49 +413,40 @@ void updateRunnerButtons(MajorVersion majorVersion) {
...
     }
   }
 
-  private PipelineOptionsHierarchy getPipelineOptionsHierarchy(
-      MajorVersion majorVersion, IProgressMonitor monitor) {
-    IProject project = getProject();
-    if (project != null && project.isAccessible()) {
+  private PipelineOptionsHierarchy getPipelineOptionsHierarchy(IProgressMonitor monitor) {
+    if (launchConfiguration != null) {
+      Verify.verify(project != null && project.isAccessible());
```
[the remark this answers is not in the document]
briandealwis: But if the project is null then the launchConfiguration will be null.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
briandealwis: Returning false means the values were not changed.  Added a comment.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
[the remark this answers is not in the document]
chanseokoh: Ah, I missed that.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
briandealwis: `eq()` is necessary if you use pattern matchers, as we do in `testSuppressingUpdates()`

      when(workspaceRoot.getProject(anyString())).thenReturn(mock(IProject.class));

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowDependencyManager.java
briandealwis: This code pulls out the version specified in the user's `pom.xml` for the Dataflow Java SDK (`com.google.cloud.dataflow:google-cloud-dataflow-java-sdk-all`), and it's possible the user has invalid values.  I'll update the javadoc to make this more explicit.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/PipelineLaunchConfiguration.java
briandealwis: It should.  I *suspect* I didn't include this as held as have another PR (that needs dusting off) that removes this support (from [#2430](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2430#issuecomment-336514168)).

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
chanseokoh: Still pending.

chanseokoh:
I can still reproduce #2232.

1. Create a Dataflow v1 project and a v2 project.
1. Create a new launch config. Whatever the project value in the _Main_ tab is, set it to the v1 project.
1. Select the _Pipeline Arguments_ tab.
1. Come back to _Main_.
1. Select the v2 project.
1. Select the _Pipeline Arguments_ tab.

briandealwis:
This PR as is currently does handle [@chanseokoh's sequence](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2585#issuecomment-358112464) correctly, but in a bit of a round-about fashion.  In verifying it, I realized there's a better way, but it's worth explaining the interactions here.

When I run [@chanseokoh's sequence](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2585#issuecomment-358112464), the launch config dialog (LCD) shows two tabs in error: the _Main_ and _Pipeline Arguments_ tab.  As the _Main_ tab is active, its error is shown: _Main type not specified_).
screen shot 2018-01-17 at 3 35 16 pm

If I specify a type, then the other error in the _Pipeline Arguments_ tab's error is shown, that an invalid runner is selected:
screen shot 2018-01-17 at 3 37 45 pm

Which seems legitimate: a runner had been selected — and encoded into this `ILaunchConfiguration` — but with the project change that runner is not valid for the new project.  When the user selects that tab, the `PipelineArgumentsTab` detects the invalid runner (`PipelineArgumentsTab#updateRunnerButtons()`) and selects a valid runner by _selecting the corresponding runner radio button_ (and logs that it was changed).  This button-selection is important.

Eventually our `initializeFrom()` calls `updateLaunchConfigurationDialog()` (inherited from `AbstractLaunchConfigurationTab`), which initiates a somewhat complicated sequence of actions including calling `performApply(ILaunchConfigurationWorkingCopy)` on the active tab.  `PipelineArgumentTab#performApply()` writes out the selected runner button to the provided `ILaunchConfigurationWorkingCopy`, thus repairing the launch configuration.

I've skipped some details in the previous paragraph: it's actually more complicated.

   - `PipelineArgumentsTab` has modification and selection listeners on all of the UI components in the tab that calls `updateLaunchConfigurationDialog()`
   - `updateLaunchConfigurationDialog()` calls `performApply(ILaunchConfigurationWorkingCopy)` on the active tab and then calls `isValid(ILaunchConfiguration)` on all the tabs to reflect errors
   - Updating the UI components in `initializeFrom(ILaunchConfiguration)` to match the `ILaunchConfig` *currently* triggers these modification listeners and thus results in multiple calls to `updateLaunchConfigurationDialog()`, and thus triggering calls to `performApply()`.
   - Because our actions in `initializeFrom() were calling `updateLaunchConfigurationDialog()` and triggering `performApply()`, **we were writing out stale UI settings to the `ILaunchConfigurationWorkingCopy`**.
   - So I introduced a field `activated` to record whether the tab's UI has been updated to reflect the changes in the _current_ ILaunchConfiguration`; `performApply()` uses this flag to determine whether the UI settings should be reflected into the provided `ILaunchConfigurationWorkingCopy`
      - (I'm going to rename this flag to `uiUpToDate`)
      - As `activated` is only set to `true` at the end of `initializeFrom()`, most of the calls to `performApply()` are (correctly) ignored.  Eventually `activated` is set to `true` and the next `updateLaunchConfigurationDialog()` will trigger the `performApply()` which writes out the changed runner.
   - But `activated` isn't set to `true` until after the UI is up-to-date, and the code to update the pipeline options uses the pipeline settings as pulled in from the `ILaunchConfiguration`, which uses the old runner

### 30,000 ft view

If we step back, activating our `PipelineArgumentsTab` follows three steps:

   1. we load state from an ILaunchConfiguration (`isValid()` and `initializeFrom()`)
   2. we update the UI to reflect that state, including corrections (`initializeFrom()`)
   3. we update the ILaunchConfigurationWorkingCopy from the UI (`performApply()`)

The crux of the problem is that updating the UI in (2) may use invalid state (invalid pipeline runner pulled from the `ILaunchConfiguration`).  And even if we were able to correct the state (i.e., change the pipeline runner in our state — we can't change the ILaunchConfiguration until `performApply()`), updating our UI components triggered calls to `isValid()` which would re-load the state from the `ILaunchConfiguration` and so over-write our state corrections (though not our corrections in the UI).

### The easy fix

It turns out `PipelineArgumentsTab` has support for suppressing these calls to `updateLaunchConfigurationDialog()` from our modification/selection listeners.  So the easy fix is to

   1. suppress these calls during the UI updating in `initializeFrom()`
   2. correct the state when detected in `updateRunnerButtons()`
   3. ensure a call to `updateLaunchConfigurationDialog()` happens after updating the UI so as to trigger a `performApply()` to push the corrections into the `ILaunchConfigurationWorkingCopy`.

I've implemented this now and @chanseokoh's sequence now runs with no warnings. \o/

### The better fix

These problems are really because our `PipelineArgumentsTab` serves two purposes: to validate the settings recorded in an `ILaunchConfiguration`, and to drive a UI to reflect and modify the settings recorded in an `ILaunchConfiguration`.  We're using fields to hold onto the state derived from the `ILaunchConfiguration`, and these fields are being modified between calls to `initializeFrom()` and `isValid()`, and so we need to coordinate whether these fields are stale or not.  

We could avoid having this coordination problem by moving the local state from being fields and instead pass it around as method arguments. But we're actually maintaining a lot of information, and as some of it's expensive to compute we'll need to cache.

elharo:
If this is ready to go, I say let's take the easy fix. We can file a tech debt bug to improve it, but it may not be worth burning time on.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
chanseokoh: Weird. I can't compile unless this is `public`, but Travis doesn't seem to have any problem. Should we revert in to `public`? `@VisibleForTesting` could remain.

briandealwis:
Weird, I see it too though I'm sure it didn't show up previously.

briandealwis:
@chanseokoh found an error:
> go to the Pipleline tab, makes some changes to be able to Apply there. Then selecting a different runner does not enable Apply again, i.e., the tab doesn't notice there is a change.

Cause is that I removed the call to `updateLaunchConfigurationDialog()` from `updatePipelineOptionsForm()` and explicitly called it from `initializeFrom()`, but missed that there were two other locations that called `updatePipelineOptionsForm()`.

PR #2586, opened 2017-11-15
distinguish app engine permissions issues

### LINKED ISSUE #2582: Improve project-selector AppEngine-detection message on permission failure
When selecting a project for deployment, we check that the project has been configured for App Engine, and show a message if not.  

screen shot 2017-11-14 at 1 49 04 pm

If the user has insufficient permissions, we show an error message which can be a bit confusing:

screen shot 2017-11-14 at 1 47 24 pm

elharo:
@briandealwis fix #2582

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppEngineApplicationQueryJobTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppEngineApplicationQueryJobTest.java
@@ -27,6 +27,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import com.google.api.client.auth.oauth2.Credential;
+import com.google.cloud.tools.eclipse.projectselector.ApplicationPermissionsException;
 import com.google.cloud.tools.eclipse.projectselector.ProjectRepository;
 import com.google.cloud.tools.eclipse.projectselector.ProjectRepositoryException;
 import com.google.cloud.tools.eclipse.projectselector.ProjectSelector;
@@ -101,6 +102,24 @@ public class AppEngineApplicationQueryJobTest {
     verify(projectSelector).setStatusLink(EXPECTED_MESSAGE_WHEN_NO_APPLICATION, EXPECTED_LINK);
 
     assertEquals(AppEngine.NO_APPENGINE_APPLICATION, project.getAppEngine());
+  }
+  
+
+  @Test
+  public void testRun_accountDoesNotHavePermission()
+      throws ProjectRepositoryException, InterruptedException {
+    when(projectRepository.getAppEngineApplication(credential, "projectId"))
+        .thenThrow(new ApplicationPermissionsException("testException", null));
+
+    queryJob.schedule();
+    queryJob.join();
+
+    verify(isLatestQueryJob).apply(queryJob);
+    verify(projectSelector).isDisposed();
+    verify(projectSelector).setStatusLink(
+        "This account does not have permission for the App Engine application in project projectId.", null);
+
+    assertNull(project.getAppEngine());
   }
 
   @Test
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
@@ -74,6 +74,7 @@ projectselector.project.not.selected=Select a project
 projectselector.filter=Filter projects by name or ID
 projectselector.retrieveproject.error.message=An error occurred while retrieving projects: {0}
 projectselector.retrieveapplication.error.message=An error occurred while retrieving App Engine application in project {0}:\n{1}
+projectselector.retrieveapplication.permissionerror.message=This account does not have permission for the App Engine application in project {0}.
 projectselector.missing.appengine.application.link=This project does not have an App Engine application which is \
  required for deployment. <a href="{0}">Create an App Engine application in the Cloud Console</a>.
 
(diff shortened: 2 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppEngineApplicationQueryJobTest.java
```diff
@@ -102,6 +103,24 @@ public void testRun_projectHasNoApplication()
...
+    when(projectRepository.getAppEngineApplication(credential, "projectId"))
+        .thenThrow(new ApplicationPermissionsException("testException", null));
+
+    queryJob.schedule();
+    queryJob.join();
+
+    verify(isLatestQueryJob).apply(queryJob);
+    verify(projectSelector).isDisposed();
+    verify(projectSelector).setStatusLink(
+        "This account does not have permission for the App Engine application in project projectId.", null);
```
briandealwis: `for` → `to access`?  `for` seems nebulous, but perhaps that's a good thing?

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
```diff
@@ -134,6 +134,14 @@ public AppEngine getAppEngineApplication(Credential credential, String projectId
     } catch (GoogleJsonResponseException ex) {
       if (ex.getStatusCode() == HttpStatusCodes.STATUS_CODE_NOT_FOUND) {
         return AppEngine.NO_APPENGINE_APPLICATION;
+      } else if (ex.getStatusCode() == HttpStatusCodes.STATUS_CODE_FORBIDDEN) {
+        String message = ex.getLocalizedMessage();
```
briandealwis: I wondered if we should do some additional checking to ensure this detail has `status == PERMISSION_DENIED`, but the status field isn't actually listed in the [status description](https://cloud.google.com/appengine/docs/admin-api/reference/rest/Shared.Types/Status).  Hmm, and actually this error object doesn't match the status object description.

The JSON returned is:
```
{
  "error": {
    "code": 403,
    "message": "Operation not allowed",
    "errors": [
      {
        "message": "Operation not allowed",
        "domain": "global",
        "reason": "forbidden"
      }
    ],
    "status": "PERMISSION_DENIED"
  }
}
```

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -74,6 +74,7 @@ projectselector.project.not.selected=Select a project
 projectselector.filter=Filter projects by name or ID
 projectselector.retrieveproject.error.message=An error occurred while retrieving projects: {0}
 projectselector.retrieveapplication.error.message=An error occurred while retrieving App Engine application in project {0}:\n{1}
+projectselector.retrieveapplication.permissionerror.message=This account does not have permission for the App Engine application in project {0}.
```
briandealwis: I was testing this out, and I wonder if it might be better to put the account name here too?

briandealwis: _'briandealwis@gmail.com' does not have permission to access the App Engine application in project 'cloud-tools-for-eclipse-kokoro'_ seems better to me.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppEngineApplicationQueryJobTest.java
elharo: Maybe "to deploy to"?
The account may be able to access the app, just not replace it.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
elharo: That would be nice. I don't see the account available in this method though, and I didn't want to make big changes just to push it in here. Is there some easy way to grab it from a credential object?

plugins/com.google.cloud.tools.eclipse.projectselector/src/com/google/cloud/tools/eclipse/projectselector/ProjectRepository.java
elharo: I'm pushing internally to get documentation to clarify the contract.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
briandealwis: We don't seem to be able to get that easily :-(

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppEngineApplicationQueryJobTest.java
briandealwis: I prefer "for" then since we're not actually trying to deploy at this particular case.  It's probably the case that if I can't retrieve the app details, I can't deploy an app either?

chanseokoh: Just for clarity, an "application" here means the [Application resource](https://cloud.google.com/appengine/docs/admin-api/creating-an-application) that may exist at most one per GCP project. So the query here is to check if the project has ever created the application, to start using App Engine.

elharo:
Doesn't seem like we can do a lot for the comments right now. I'm going to commit this, and we can improve it later if we figure out how.

PR #2587, opened 2017-11-15
Add browse-file dialog for service key in GcpLocalRunTab / add service key file validation

chanseokoh:
- The _Browse..._ button for the service key file field is fully functioning.
- Added file validation for the field.


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -200,14 +203,16 @@ public void modifyText(ModifyEvent event) {
     });
     Button browse = new Button(composite, SWT.NONE);
     browse.setText(Messages.getString("button.browse")); //$NON-NLS-1$
+    browse.addSelectionListener(new FileFieldSetter(serviceKeyInput,
+        new String[] {"*.json", "*"})); //$NON-NLS-1$ //$NON-NLS-2$
```
elharo: why * AND *.json? Why not just *.json?

Also this might be a little easier to read with an extr local variable for the file field setter.

chanseokoh: If we can be certain the file should always be `.json`, we can git rid of `*`. I put `*` as a fallback because otherwise, there is no way users can select a non-`.json` file. (They could still enter anything in the input field directly though.)

elharo: Pantheon will always create a .json file. I think that's good enough.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -200,14 +203,16 @@ public void modifyText(ModifyEvent event) {
     });
     Button browse = new Button(composite, SWT.NONE);
     browse.setText(Messages.getString("button.browse")); //$NON-NLS-1$
+    String[] filterExtensions = new String[] {"*.json"}; //$NON-NLS-1$
```
briandealwis: Does `GOOGLE_APPLICATION_CREDENTIALS` support P12 files?  I know we can download them from Pantheon.

chanseokoh: My gut feeling is that it supports P12 files or such. However, the assumption here based on my discussion with @elharo is that, we'll need only service accounts (`.json`) for the Cloud libraries.

elharo: P12 is old and busted. JSON is the new hotness:

image

PR #2590, opened 2017-11-16
hide mutable private field

elharo:


plugins/com.google.cloud.tools.eclipse.ui/src/com/google/cloud/tools/eclipse/ui/util/event/FileFieldSetter.java
```diff
@@ -56,7 +59,7 @@ public void widgetSelected(SelectionEvent event) {
       filterPath = filterPath.removeLastSegments(1);
     }
     dialog.setFilterPath(filterPath.toString());
-    dialog.setFilterExtensions(filterExtensions);
+    dialog.setFilterExtensions(Arrays.copyOf(filterExtensions, filterExtensions.length));
```
chanseokoh: Maybe this one doesn't need a copy? We don't mutate `filterExtensions` and I don't expect `FileDialog` to mutate it either.

elharo: An excess of caution perhaps. Unless FileDialog promises never to mutate this field in the future, there's a risk.

PR #2591, opened 2017-11-16
Default to java8 runtime

### LINKED ISSUE #2572: Convert to App Engine Standard menu should configure Java 8 instead of Java 7
Currently, it's always Java 7.

It may be OK for Java 7 projects to remain on the Java 7 runtime, although they will have no problem being deployed to the Java 8 runtime.

elharo:
@briandealwis fix #2572

A possible fix: this PR now simply removes all automatic downgrades to the java7 runtime. 
The proposal is that if there's a `<runtime>java8</runtime>` in appengine-web.xml, then we leave it alone. The user can manually edit that file to remove it if they really want the legacy runtime.

However we allow people to deploy Java 7 and Servlet 2.5 apps to the Java 8 runtime.


elharo:
Yes, my default compliance level is set to 1.7 when I run in hosted mode. Not sure why. It's 1.8 in the Host environment (Eclipse Neon).

PR #2592, opened 2017-11-16
Turn OK MultiStatus -> Status.OK_STATUS

### LINKED ISSUE #2588: LibraryClasspathContainerResolverJob leaves confusing status message when successfully finished
The job finishes normally, but it leaves a confusing error message that does not go away from the Eclipse status bar.

selection_043

It's because `LibraryClasspathContainerResolverService.resolveAll()` starts with the message. The message is intended to be an error message, so we should set it only when there is an error.

```java
  @Override
  public IStatus resolveAll(IJavaProject javaProject, IProgressMonitor monitor) {
    try {
      MultiStatus status = StatusUtil.multi(this, Messages.getString("TaskResolveLibrariesError"));
```

briandealwis:
Fixes #2588

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/cte13/CloudToolsEclipseProjectNotifier.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/cte13/CloudToolsEclipseProjectNotifier.java
@@ -145,6 +145,7 @@ public class CloudToolsEclipseProjectNotifier implements IStartup {
       IStatus result = CloudToolsEclipseProjectUpdater.updateProject(project, progress.newChild(1));
       status.merge(result);
     }
-    return status;
+    // rewrite if OK as otherwise Progress View shows the updating projects message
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/cte13/CloudToolsEclipseProjectNotifier.java
```diff
@@ -145,6 +145,7 @@ private IStatus upgradeProjects(Collection<IProject> projects, SubMonitor progre
       IStatus result = CloudToolsEclipseProjectUpdater.updateProject(project, progress.newChild(1));
       status.merge(result);
     }
-    return status;
+    // rewrite if OK as otherwise Progress View shows the updating projects message
```
elharo: project's

briandealwis: In this case it's literally "Updating projects for …"

PR #2596, opened 2017-11-21
Add button to create service key in GcpLocalRunTab

chanseokoh:
- The button is enabled only when a project is selected.
- Maybe we can move the button right to "Browse...".
- The file is saved to the user's home. (We can think about the location.) The name is "app-engine-default-service-account-key-<project name>.json". If a file with the same name exists, it won't attempt to create a new button and show an error message.
- If created successful, the input field is filled with the created file. A decor message explain what has happened.
selection_007
- I've mulled over how I can show an error message effective. I decided to show it as a decor message too:
selection_003
selection_005


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/messages.properties
```diff
@@ -48,3 +49,6 @@ project.filter.hint=Filter projects by name or ID
 error.file.does.not.exist={0} does not exist.
 error.is.a.directory={0} is a directory.
 error.is.not.readable={0} is not readable.
+service.key.created=A service key for the App Engine default service account created and set:\n{0}
```
elharo: "Created a service account key for the App Engine app default service account"
(same language as Pantheon)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/messages.properties
```diff
@@ -48,3 +49,6 @@ project.filter.hint=Filter projects by name or ID
...
+service.key.already.exists=Failed to create a service account key. A file already exists:\n{0}
```
elharo: Could not create service account key because the file already exists:

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -358,4 +395,64 @@ static String getAttribute(ILaunchConfiguration configuration,
       return emptyMap;
     }
   }
+
+  @VisibleForTesting
+  Path getServiceAccountKeyPath() {
+    Preconditions.checkState(!projectSelector.getSelection().isEmpty());
+
+    String projectId = projectSelector.getSelectProjectId();
+    String filename = "app-engine-default-service-account-key-" //$NON-NLS-1$
```
elharo: I wonder if we should make our own dot directory in home. What do you think?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -358,4 +395,64 @@ static String getAttribute(ILaunchConfiguration configuration,
...
+        + projectId + ".json"; //$NON-NLS-1$
+    return Paths.get(System.getProperty("user.home")).resolve(filename); //$NON-NLS-1$
+  }
+
+  @VisibleForTesting
+  void createServiceAccountKey(Path keyFile) {
```
elharo: This logic should probably be separated from the UI classes.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Can't really think of a good name, when considering this is not specific to our plugin and potentially may contain non-App Engine service accounts in the future.

I also feel unsure trying to hide it (dot directory) is better, as this is an important file.

elharo: I think cluttering the user's home directory is a bad idea. dot directories are pretty standard for this sort of thing on Linux and not uncommon on Mac OS X. Not sure about Windows. Maybe we can ask around for suggestions.

elharo: Crazy idea: what if we generated new keys for each local run and deleted them when we're done?

chanseokoh: > Crazy idea: what if we generated new keys for each local run and deleted them when we're done?

We will need give some thoughts to make this robust in case a run does not stop cleanly, not to litter the key space. There seems to be a hard limit on the number of keys (about 8~10) per service account.

briandealwis: We could store this in the workspace plugin state location (e.g., `<workspace>/.metadata/.plugins/com.google.cloud.tools.eclipse.xxx`).  Or in the Eclipse configuration directory, which is shared across workspaces (`org.eclipse.core.runtime.Platform#getConfigurationLocation()` is the easiest accessor).

If we're thinking of a non-Eclipse location, typical OS-specific locations are:
  - on MacOS: `${user.home}/Library/Application Support/<appName>`
  - on Linux: `${XDG_DATA_HOME}/<appName>` since it's a data file (?) which defaults to `${user.home}/.local/share` according to [XDG](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html); alternatively we could use `${XDG_CONFIG_HOME}` if we think of this more as a config file
  - on Windows: `${LOCALAPPDATA}/<appName>` [source](http://environmentvariables.org/LocalAppData)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -98,15 +115,18 @@
   private boolean activated;
 
   public GcpLocalRunTab(EnvironmentTab environmentTab) {
-    this(environmentTab, PlatformUI.getWorkbench().getService(IGoogleLoginService.class),
+    this(environmentTab,
+        PlatformUI.getWorkbench().getService(IGoogleLoginService.class),
+        PlatformUI.getWorkbench().getService(IGoogleApiFactory.class),
         new ProjectRepository(PlatformUI.getWorkbench().getService(IGoogleApiFactory.class)));
```
briandealwis: I wonder if we should push this service-fetching into the tab group.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -206,14 +229,28 @@ public void modifyText(ModifyEvent event) {
     String[] filterExtensions = new String[] {"*.json"}; //$NON-NLS-1$
     browse.addSelectionListener(new FileFieldSetter(serviceKeyInput, filterExtensions));
 
+    serviceKeyDecoration = new ControlDecoration(serviceKeyInput, SWT.LEAD | SWT.TOP);
+    serviceKeyDecoration.hide();
+
+    createServiceKey = new Button(composite, SWT.NONE);
```
briandealwis: Would it be worth having a hover on this button to explain that it creates a new App Engine service account key, same as what the app will be executed on App Engine?  Just in case they have multiple service accounts…

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -206,14 +229,28 @@ public void modifyText(ModifyEvent event) {
...
+    createServiceKey.setText(Messages.getString("create.new.service.key")); //$NON-NLS-1$
+    createServiceKey.addSelectionListener(new SelectionAdapter() {
+      @Override
+      public void widgetSelected(SelectionEvent event) {
+        createServiceAccountKey(getServiceAccountKeyPath());
```
briandealwis: Should wrap in a `getLaunchConfigurationDialog().run()`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -358,4 +395,64 @@ static String getAttribute(ILaunchConfiguration configuration,
...
+
+    } catch (IOException e) {
+      logger.log(Level.SEVERE, e.getMessage(), e);
+      String message = Messages.getString("cannot.create.service.key",  //$NON-NLS-1$
+          e.getLocalizedMessage());
+      showServiceKeyDecorationMessage(message, true /* errorImage */);
+    }
+  }
+
+  private void showServiceKeyDecorationMessage(String message, boolean errorImage) {
```
briandealwis: s/errorImage/isError/?

plugins/com.google.cloud.tools.eclipse.googleapis/.classpath
```diff
@@ -10,6 +10,7 @@
 	<classpathentry exported="true" kind="lib" path="lib/google-api-services-compute-v1-rev160-1.23.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/google-api-services-dataflow-v1b3-rev219-1.23.0.jar"/>
 	<classpathentry exported="true" kind="lib" path="lib/google-api-services-datastore-v1-rev34-1.23.0.jar"/>
+	<classpathentry exported="true" kind="lib" path="lib/google-api-services-iam-v1-rev222-1.23.0.jar"/>
```
briandealwis: Did you move this?  Just wondering if there was a reason.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Good idea. Personally, I think all of the controls in this UI needs more explanation.

plugins/com.google.cloud.tools.eclipse.googleapis/.classpath
chanseokoh: Yeah, I noticed there are in an alphabetical order.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Good idea. I think a `BusyIndicator` will work fine. This is a one API call, and we have very short timeouts set.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Thought about that, and my thinking has been that, if a class surely knows how to get something, why pass the burden of acquiring it to callers, unless there is a merit (there doesn't seem to me)? I may be wrong, but this is a UI class, so I thought `PlatformUI.getWorkbench()` must always succeed. (Otherwise, how one could possibly show a UI tab?)

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Discussed this for a while with @elharo. The Eclipse configuration directory seems a workable option now.

chanseokoh:
Still doing some testing.

chanseokoh:
Moved the button next to the "Browse..." button. Things work well.

selection_008

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
briandealwis: The downside is that it makes and bakes in an assumption about where these instances are from.  So it's kinda-injected, but not quite.  Anyways, this is done elsewhere in the code and it's something that can be tackled later.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -403,7 +406,7 @@ Path getServiceAccountKeyPath() {
     String projectId = projectSelector.getSelectProjectId();
     String filename = "app-engine-default-service-account-key-" //$NON-NLS-1$
         + projectId + ".json"; //$NON-NLS-1$
-    return Paths.get(System.getProperty("user.home")).resolve(filename); //$NON-NLS-1$
+    return Paths.get(Platform.getConfigurationLocation().getURL().getPath()).resolve(filename);
```
briandealwis: The convention seems to be to create a directory to hold files, usually reverse-DNS-based or bundle symbolic name. Since we could generate non-AppEngine keys, perhaps just `com.google.cloud.tools.eclipse`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -234,24 +231,30 @@ public void modifyText(ModifyEvent event) {
...
+    createServiceKey.setToolTipText(
+        Messages.getString("create.new.service.key.tooltip")); //$NON-NLS-1$
     createServiceKey.addSelectionListener(new SelectionAdapter() {
       @Override
       public void widgetSelected(SelectionEvent event) {
-        createServiceAccountKey(getServiceAccountKeyPath());
+        BusyIndicator.showWhile(createServiceKey.getDisplay(), new Runnable() {
+          @Override
+          public void run() {
+            createServiceAccountKey(getServiceAccountKeyPath());
```
briandealwis: What if they've provided a path in the text?  Should we use it?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: `Platform.getConfigurationLocation()` already gives `.metadata/.plugins/org.eclipse.pde.core/pde-junit/`.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: We'll need a slightly different UI if we do that. We may think about it later.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
chanseokoh: Oh, wait. That was in a testing.

chanseokoh: I think it'll still be under `org.eclipse.pde.core`.

briandealwis: It will be the same as the `osgi.configuration.area` property.  In my IDE instance, for example (_Help > About > Configuration_), it's `file:/usr/local/installs/e4-dev/Eclipse.app/Contents/Eclipse/configuration/`.

briandealwis: And for completeness, the contents:
```
$ ls -1 /usr/local/installs/e4-dev/Eclipse.app/Contents/Eclipse/configuration/
config.ini
org.eclipse.core.runtime/
org.eclipse.e4.ui.css.swt.theme/
org.eclipse.equinox.app/
org.eclipse.equinox.simpleconfigurator/
org.eclipse.equinox.source/
org.eclipse.help.base/
org.eclipse.oomph.setup/
org.eclipse.osgi/
org.eclipse.ui.intro.universal/
org.eclipse.update/
org.eclipse.wst.server.discovery/
```

briandealwis: It *might* be worth putting project keys in project-specific directories, rather than encoding the project ID in the file name.  We may create other project-specific files and this would be a place to put them?

chanseokoh: Updated to create a key under `com.google.cloud.tools.eclipse`.

I'm sticking to putting a project ID in a filename rather than creating a directory. I think it's possible directories are YAGNI.

PR #2597, opened 2017-11-22
Check com.google.api versions in MANIFEST.MF

chanseokoh:


chanseokoh:
Hold on. Seems like this is not working.

chanseokoh:
Ah, it does work. I was wrongly modifying a `MANIFEST.MF` in test bundles.

PR #2600, opened 2017-11-27
Clear error message before attempting to create new bucket

### LINKED ISSUE #2543: In Pipeline Arguments tab in Dataflow run config, update error message after successfully creating bucket
To create a bucket, you'll type a unique, non-existing bucket name into the staging location field. This first step will correctly show an error message "Could not fetch bucket <...>". In case of the project wizard dialog, when you then click the _Create Bucket_ button, the error message will be replaced by an info message "Created staging location at gs://<...>". However, when you do the same in the _Pipeline Arguments_ tab in a Dataflow run configuration, the error message is not cleared and stays here. There is no message that the bucket is successfully created, so this is confusing.

briandealwis:
The create-bucket management is handled by the `RunOptionsDefaultsComponent` which uses `MessageTarget` facade to handle reporting messages to the various UI components.

The wizard page is an instance of `NewDataflowProjectWizardDefaultRunOptionsPage` and it uses the  `DialogPageMessageTarget` to set info and error messages to the page container.  These errors are channelled through a single method, `WizardPage#setMessage(String message, int messageType)`, which only tracks the last message.  Thus the _Could not fetch bucket_ error message is subsequently overwritten by the _Created staging location_ info message.

On the other hand, the `PipelineArgumentsTab` is an `AbstractLaunchConfigurationTab` which manages info and warning messages separately from error messages.

Since _Create Bucket_ is a user action, we should be ok to call `MessageTarget#clear()` before creating the bucket to clear any messages shown.

Fixes #2543


briandealwis:
PTAL @elharo — I made one minor change in response to test failures.

PR #2614, opened 2017-11-30
Add support for removing previously-selected libraries from Maven poms

### LINKED ISSUE #2567: Support editing and removing libraries for Maven projects
Following on #2553, the Google APIs container page allows adding new libraries to a Maven project, but doesn't support editing or removing libraries.  

We could slurp up the libraries from the `pom.xml` to set the current selection.  Removal should be relatively easy.  But what should happen if they use a different version of the libraries?  Do we preserve the versions of the existing libraries?  Or write out everything with the new versions?

We could punt on supporting Maven projects entirely and instead shown an error message directing the users to the POM editor, and `setPageComplete(false)`.

briandealwis:
- Adds a new method to `BuildPath#loadMavenLibraries()` to resolve the set of libraries whose requirements are matched by the project's `<dependency> definitions.
- Adds a new parameter to `BuildPath#addMavenLibraries()` to pass the set of candidate libraries that were considered, so as to be able to determine which libraries were *not* selected.  We only match a `Library` when all of its dependencies are matched, and conversely we only remove `<dependency>` definitions when all the dependencies of a `Library` are matched.
- Make `PomTest` more compact by simplifying construction of `Library` definitions

I have some tests to exercise `BuildPath#addMavenLibraries()` and `loadMavenLibraries()` but I'm not entirely sure they add much, and they trigger m2e as they create a project and write out to the pom.xml.

Fixes #2567

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
@@ -16,10 +16,9 @@
 
 package com.google.cloud.tools.eclipse.appengine.libraries;
 
-import static com.google.cloud.tools.eclipse.appengine.libraries.model.MavenCoordinates.coordinates;
-
 import com.google.cloud.tools.eclipse.appengine.libraries.model.Library;
 import com.google.cloud.tools.eclipse.appengine.libraries.model.LibraryFile;
+import com.google.cloud.tools.eclipse.appengine.libraries.model.MavenCoordinates;
 import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
 import java.io.IOException;
 import java.io.InputStream;
@@ -49,58 +48,56 @@ import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
 public class PomTest {
-
-  @Rule
-  public final TestProjectCreator projectCreator = new TestProjectCreator();
+  
+  @Rule public final TestProjectCreator projectCreator = new TestProjectCreator();
 
   // todo we're doing enough of this we should import or write some utilities
   private static final DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
-
-  @BeforeClass
+  
+  @BeforeClass 
   public static void configureParser() {
     factory.setNamespaceAware(true);
   }
-
+  
   private Pom pom;
   private IFile pomFile;
-
+  
   @Before
   public void setUp() throws SAXException, IOException, CoreException {
     IProject project = projectCreator.getProject();
     pomFile = project.getFile("pom.xml");
     InputStream in = Files.newInputStream(Paths.get("testdata/testpom.xml").toAbsolutePath());
     pomFile.create(in, IFile.FORCE, null);
-
+   
     pom = Pom.parse(pomFile);
   }
-
+  
   @Test
-  public void testAddDependencies()
+  public void testAddDependencies() 
       throws CoreException, ParserConfigurationException, IOException, SAXException {
-
+    
     Library library0 =
-        new Library("id0",
-            new LibraryFile(coordinates("com.example.group0", "artifact0", "1.2.3")));
+        newLibrary("id0", new LibraryFile(coordinates("com.example.group0", "artifact0", "1.2.3")));
     Library library1 =
-        new Library("id1", new LibraryFile(coordinates("com.example.group1", "artifact1")));
+        newLibrary("id1", new LibraryFile(coordinates("com.example.group1", "artifact1")));
     Library library2 =
-        new Library("id2", new LibraryFile(coordinates("com.example.group2", "artifact2")),
+        newLibrary("id2", new LibraryFile(coordinates("com.example.group2", "artifact2")),
             new LibraryFile(coordinates("com.example.group3", "artifact3")));
     List<Library> libraries = Arrays.asList(library0, library1, library2);
-
+    
     pom.addDependencies(libraries);
-
+    
     Assert.assertEquals(1, pomFile.getHistory(null).length);
-
+    
     InputStream contents = pomFile.getContents();
     Document actual = parse(contents);
-
+    
     NodeList dependencies = actual.getElementsByTagName("dependencies");
     Assert.assertEquals(1, dependencies.getLength());
     NodeList children = ((Element) dependencies.item(0)).getElementsByTagName("dependency");
-
+    
     Assert.assertEquals(4, children.getLength());
-
+    
     Element child0 = (Element) children.item(0);
     Element groupId = getOnlyChild(child0, "groupId");
     Assert.assertEquals("com.example.group0", groupId.getTextContent());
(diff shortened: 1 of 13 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -173,4 +173,12 @@ private static void resolveTransitiveDependencies(ImmutableMap<String, Library>
       library.setLibraryDependencies(transitiveDependencies);
     }
   }
+
+  /**
+   * @return
+   */
+  public static Object getAllLibraries() {
+    // TODO Auto-generated method stub
```
elharo: do you want to fill this in? or at least remove the auto boilerplate?

briandealwis: Oops.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -58,21 +59,41 @@
...
   private static final Logger logger = Logger.getLogger(BuildPath.class.getName());
   
-  public static void addMavenLibraries(IProject project, List<Library> libraries,
-      IProgressMonitor monitor) throws CoreException {
-    if (libraries.isEmpty()) {
+  /**
+   * Set the Maven dependencies to the selected libraries, and optionally remove any traces of
+   * unselected libraries.
+   * 
+   * @param availableLibraries if provided, then remove dependencies that correspond to unselected
```
elharo: The argument name feels off. Are we removing everything in this list or everything in this list that isn't in selected?

Also, please document nullability.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -81,24 +89,77 @@ static Pom parse(IFile pomFile) throws SAXException, IOException, CoreException
...
+  void addDependencies(List<Library> selected) throws CoreException {
+    addDependencies(selected, null);
+  }
+
+  /**
+   * Adjust dependencies required for the list of selected libraries, adding any necessary
+   * dependencies and removing any unnecessary dependencies.
+   * 
+   * @param selected the set of libraries to be included
+   * @param availableLibraries if not {@code null}, then remove any dependencies that match
```
elharo: again, availableLibraries doesn't seem like the right name. previouslySelectedLibraries? alreadySelectedLIbraries? currentLibraries?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -137,6 +198,67 @@ void addDependencies(List<Library> libraries) throws CoreException {
     }   
   }
 
+  /**
+   * Remove any dependencies that were required by the previously-selected libraries and that are
+   * not required by the currently selected libraries.
+   * 
+   * @param selectedLibraries the currently selected libraries
+   * @param previousLibraries the previously-selected libraries
```
elharo: This is a much better name for this.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -137,6 +198,67 @@ void addDependencies(List<Library> libraries) throws CoreException {
...
+      Node node = children.item(i);
+      if (node.getNodeType() == Node.ELEMENT_NODE) {
+        Element dependency = (Element) node;
+        String groupId = getValue(dependency, "groupId");
+        String artifactId = getValue(dependency, "artifactId");
+        String encoded = groupId + "|" + artifactId;
+        currentDependencies.put(encoded, node);
+      }
+    }
+    Verify.verify(currentDependencies.isEmpty() == (dependencies.getChildNodes().getLength() == 0));
```
elharo: do you really think this could happen? If so, we should throw a more explicit exception.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -137,6 +198,67 @@ void addDependencies(List<Library> libraries) throws CoreException {
...
+
+    Set<String> selectedDependencies = new HashSet<>();
+    for (Library library : selectedLibraries) {
+      for (LibraryFile libraryFile : library.getDirectDependencies()) {
+        MavenCoordinates coordinates = libraryFile.getMavenCoordinates();
+        String encoded = coordinates.getGroupId() + "|" + coordinates.getArtifactId();
+        selectedDependencies.add(encoded);
+      }
+    }
+    Verify.verify(selectedDependencies.isEmpty() == selectedLibraries.isEmpty());
```
elharo: do you really think this could happen? If so, we should throw a more explicit exception.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -137,6 +198,67 @@ void addDependencies(List<Library> libraries) throws CoreException {
...
+    Verify.verify(currentDependencies.isEmpty() == (dependencies.getChildNodes().getLength() == 0));
+
+    Set<Node> nodesToRemove = new HashSet<>();
+    for (Library library : previousLibraries) {
+      // true if all coordinates for this library are found
+      boolean allFound = true;
+      Set<Node> libraryNodes = new HashSet<>();
+      for (LibraryFile file : library.getDirectDependencies()) {
+        MavenCoordinates coord = file.getMavenCoordinates();
+        String encoded = coord.getGroupId() + "|" + coord.getArtifactId();
```
elharo: why | instead of :?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -148,7 +148,9 @@ protected void addAdditionalDependencies(IProject newProject, AppEngineProjectCo
     SubMonitor progress = SubMonitor.convert(monitor, 12);
     if (config.getUseMaven()) {
       enableMavenNature(newProject, progress.newChild(5));
-      BuildPath.addMavenLibraries(newProject, config.getAppEngineLibraries(), progress.newChild(7));
+      // pass availableLibraries=null since there are no libraries to unselect
+      BuildPath.addMavenLibraries(newProject, config.getAppEngineLibraries(),
+          null /* availableLibraries */, progress.newChild(7));
```
elharo: Perhaps could send an empty list instead.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
briandealwis: This is the  the set of all libraries that were available to be selected from.  We use this list to determine which dependency-sets can be removed.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
briandealwis: No reason. Will switch.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
briandealwis: It's also incorrect :-(

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
briandealwis: No, I don't think this should happen.  `Verify` == assertion.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -255,4 +295,20 @@ private static Element getOnlyChild(Element element, String name) {
     return (Element) children.item(0);
   }
 
+  private Library newLibrary(String libraryId, LibraryFile... libraryFiles) {
```
elharo: static

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -255,4 +295,20 @@ private static Element getOnlyChild(Element element, String name) {
...
+    Library library = new Library(libraryId);
+    library.setLibraryFiles(Arrays.asList(libraryFiles));
+    return library;
+  }
+
+  private MavenCoordinates coordinates(String groupId, String artifactId) {
```
elharo: static

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -255,4 +295,20 @@ private static Element getOnlyChild(Element element, String name) {
...
+    return new MavenCoordinates.Builder().setGroupId(groupId).setArtifactId(artifactId).build();
+  }
+
+  private MavenCoordinates coordinates(String groupId, String artifactId, String version) {
```
elharo: static

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -81,24 +89,76 @@ static Pom parse(IFile pomFile) throws SAXException, IOException, CoreException
...
+  void addDependencies(List<Library> selected) throws CoreException {
+    updateDependencies(selected, null);
+  }
+
+  /**
+   * Adjust the pom to add any required dependencies for the selected libraries and remove any
+   * unnecessary dependencies for the removed libraries.
+   * 
+   * @param selectedLibraries the set of libraries to be included
+   * @param removedLibraries these libraries dependencies are removed providing they are not
```
elharo: library dependencies

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/MavenCoordinates.java
```diff
@@ -28,6 +28,16 @@
   public static final String JAR_TYPE = "jar";
   public static final String MAVEN_CENTRAL_REPO = "central";
 
+  public static MavenCoordinates coordinates(String groupId, String artifactId) {
```
elharo: These methods may be a little too error prone to include in the main, non-test API. In particular, nothing breaks if you swap the groupID and artifactId arguments.

briandealwis: ok

PR #2618, opened 2017-12-04
Pull org.apache.commons.compress bundle in

chanseokoh:
I've not specified the bundle version (i.e., 0.0.0), as I thought we don't need to pin a certain version. Not sure if it'd be better to pin down.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -146,7 +146,10 @@
       <version>0.4.1</version>
     </dependency>
     <dependency>
-      <!-- bundle: org.apache.commons.compress -->
+      <!--
+        dependency used by appengine-plugins-core
+        bundle: org.apache.commons.compress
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -145,6 +145,12 @@
       <artifactId>appengine-plugins-core</artifactId>
       <version>0.4.1</version>
     </dependency>
+    <dependency>
+      <!-- bundle: org.apache.commons.compress -->
```
briandealwis: Do you think it's worth mentioning that this version comes from appengine-plugins-core?  We haven't done this for the other dependencies and I somewhat regret it (e.g., what restrictions do we have for updating jackson?).

chanseokoh: Good idea.

PR #2619, opened 2017-12-04
Do not fail if "gsutil rm" fails when the destination folder does not exist

chanseokoh:
We clean up the GCS bucket first before uploading binaries to make sure old binaries from previous builds get cleaned up first. However, when doing a build for the first time, the folder doesn't exist, and `gsutil rm` fails with a non-zero exit code.

### DIFF
--- a/release/release_after_kokoro.sh
+++ b/release/release_after_kokoro.sh
@@ -230,7 +230,7 @@ echo "THIS WILL DELETE ALL FILES IN '$DESTINATION' IF THEY EXIST!"
 ask_proceed
 
 set -x
-gsutil -m rm $DESTINATION/** || true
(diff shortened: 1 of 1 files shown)


release/release_after_kokoro.sh
```diff
@@ -230,7 +230,7 @@ echo "THIS WILL DELETE ALL FILES IN '$DESTINATION' IF THEY EXIST!"
 ask_proceed
 
 set -x
-gsutil -m rm $DESTINATION/** && \
+gsutil -m rm $DESTINATION/** || true
```
briandealwis: Do we need this `|| true`?  We're not using `set -o errexit` that I can see.

chanseokoh: Good catch!

Issue #2602, opened 2017-11-27
Flakey _Widget is disposed_ failures in integration tests on Travis


briandealwis:
Saw this twice now.  This failure was on Neon.

```
———————————————————————————
 T E S T S
———————————————————————————
Running com.google.cloud.tools.eclipse.integration.appengine.ImportNativeAppEngineStandardProjectTest
[Watchdog] > importAppEngineStandardJava7_from1_3_1(com.google.cloud.tools.eclipse.integration.appengine.ImportNativeAppEngineStandardProjectTest)
Window manager warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x4002bb (Import )
Window manager warning: meta_window_activate called by a pager with a 0 timestamp; the pager needs to be fixed.
16:21:59.129 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Next >' and with style 'SWT.PUSH')
16:21:59.135 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Next >' and with style 'SWT.PUSH')
16:21:59.136 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Next >' and with style 'SWT.PUSH')
16:21:59.154 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Next >' and with style 'SWT.PUSH')
16:21:59.156 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Next >' and with style 'SWT.PUSH')
16:21:59.741 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Finish' and with style 'SWT.PUSH')
16:21:59.741 [main] WARN  o.e.s.s.finder.widgets.SWTBotButton - Widget is not enabled: (of type 'Button' and with mnemonic 'Finish' and with style 'SWT.PUSH')
[...]

PR #2623, opened 2017-12-05
Test Mars using GTK2

briandealwis:
Set `SWT_GTK3=0` to cause tests to run against GTK2.

Mitigates #2602.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,9 @@ env:
   matrix:
     # ECLIPSE_TARGET is the targeted platform: mars is the default target
     # JAVA_RUNTIME specifies the execution runtime for tests; must match an <id> in .travis-toolchains.xml
+    # SWT_GTK3=0 to force GTK2 on Mars as its GTK3 support is buggy
     - JAVA_RUNTIME=JavaSE-1.7 SWT_GTK3=0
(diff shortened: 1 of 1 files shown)


briandealwis:
Feel free to trigger rebuilds and lets see if we can get any further widget-disposed test failures.

briandealwis:
A number of the ITs failed on Neon due to `SWTBot#activeShell()` throwing a `WidgetNotFoundException` in situations where there definitely should be a main window.  What's odd is that these failures are sporadic: the Travis PR and Merge builds are failing at different places.  It's almost as if another window is showing.

briandealwis:
At least it's consistently failing on Neon.  I'll revise the patch to only use GTK2 for Mars.

briandealwis:
Fail on Mars, but it seems a bit spurious:
```
testPartialValidity_account_project(com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest)  Time elapsed: 0.052 sec  <<< FAILURE!
java.lang.AssertionError: should be complete with account and project
	at com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest.testPartialValidity_account_project(RunOptionsDefaultsComponentTest.java:452)
```

PR #2626, opened 2017-12-05
Fix opt-in dialog closing bug

### LINKED ISSUE #2622: Analytics submission can be blocked forever if opt-in dialog is forcibly closed due to parent closing
When the `OptInDialog` is closed, we release the lock, resuming any pending submission jobs. The places we release the lock are `okPressed()`, `cancelPressed()`, and `handelShellCloseEvent()`.

```java
  @Override
  protected void okPressed() {
    ...
    answeredSignal.release();
  }

  @Override
  protected void cancelPressed() {
    ...
    answeredSignal.release();
  }

  /**
   * When the dialog closes in other ways than pressing the buttons.
   */
  @Override
  protected void handleShellCloseEvent() {
    ...
    answeredSignal.release();
  }
```

I thought `handleShellCloseEvent()` covered all the other cases than the cases of the buttons pressed, but I just realized it doesn't. For example, if the dialog is closed because its parent dialog is closed, the method is not called.

So, I need to find another method to override that can cover any case of dialog closing. Unfortunately, there does not seem such a method, which is surprising.

chanseokoh:
Fixes #2622.

Misc: removed Javadocs that do not make sense as public description for the methods.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/CompositeUtil.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/CompositeUtil.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.eclipse.test.util.ui;
 
 import com.google.common.base.Predicate;
+import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Label;
@@ -29,6 +30,15 @@ public class CompositeUtil {
       @Override
       public boolean apply(Control control) {
         return type.isInstance(control);
+      }
+    });
+  }
+
+  public static Button findButton(Composite composite, final String text) {
+    return (Button) findControl(composite, new Predicate<Control>() {
+      @Override
+      public boolean apply(Control control) {
+        return control instanceof Button && ((Button) control).getText().equals(text);
       }
     });
   }
@@ -53,7 +63,7 @@ public class CompositeUtil {
    */
   @SuppressWarnings("unchecked")
   public static <T> T findControlAfterLabel(Composite composite, Class<T> type, String label) {
-    return (T) CompositeUtil.findControl(composite, new AfterLabel<T>(type, label));
+    return (T) CompositeUtil.findControl(composite, new AfterLabel<>(type, label));
   }
 
   private static class AfterLabel<T> implements Predicate<Control> {
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialogTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialogTest.java
@@ -17,17 +17,15 @@
 package com.google.cloud.tools.eclipse.usagetracker;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 
 import com.google.cloud.tools.eclipse.test.util.ui.CompositeUtil;
 import com.google.cloud.tools.eclipse.test.util.ui.ShellTestResource;
+import org.eclipse.jface.window.Window;
 import org.eclipse.swt.events.FocusAdapter;
 import org.eclipse.swt.events.FocusEvent;
 import org.eclipse.swt.widgets.Button;
-import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.swtbot.swt.finder.widgets.SWTBotButton;
 import org.junit.Before;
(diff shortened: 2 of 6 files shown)


chanseokoh:
Wait, let me see if I can improve the interface.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -276,15 +277,16 @@ private void showOptInDialogIfNeeded(final Shell parentShell) throws Interrupted
     if (shouldShowOptInDialog()) {
       Display display = PlatformUI.getWorkbench().getDisplay();
 
-      final OptInDialog[] dialogHolder = new OptInDialog[1];
+      final Boolean[] optInAnswer = new Boolean[1];
```
elharo: any reason not to use a little b boolean here instead?

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialogTest.java
```diff
@@ -17,60 +17,84 @@
...
-    assertFalse(getIsOptInYes());
+  public void testReturnCode_dialogClosed() {
+    scheduleClosingDialogAfterOpen(CloseAction.CLOSE_SHELL);
+    assertNotEquals(Window.OK, dialog.open());
   }
 
-  private boolean getIsOptInYes() throws InterruptedException {
-    final Boolean[] answerHolder = new Boolean[1];
-    Thread thread = new Thread(new Runnable() {
+  private enum CloseAction { PRESS_OK, PRESS_CANCEL, CLOSE_SHELL };
```
briandealwis: Should you test for `dispose()` (`DISPOSE_SHELL`)?  I guess it doesn't matter now with your blocking-dialog pattern.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/CompositeUtil.java
```diff
@@ -33,6 +34,15 @@ public boolean apply(Control control) {
     });
   }
 
+  public static Button findButton(Composite composite, final String text) {
```
briandealwis: TBH we should minimize these types of methods and use SWTBot's helpers (e.g., `SWTBot#buttonWithLabel()`).

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialogTest.java
```diff
@@ -17,60 +17,84 @@
...
+    shell.addFocusListener(new FocusAdapter() {
       @Override
-      public void run() {
-        try {
-          answerHolder[0] = dialog.isOptInYes();
-        } catch (InterruptedException ex) {}
+      public void focusGained(FocusEvent e) {
+        switch (closeAction) {
+          case PRESS_OK:
+            Button okButton = CompositeUtil.findButton(dialog.getShell(), "Share");
```
briandealwis: Why not use SWTBot directly?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialog.java
```diff
@@ -36,18 +34,12 @@
  */
 public class OptInDialog extends Dialog {
```
briandealwis: Would be worth adding javadoc to indicate that SHARE &rarr; `Window.OK` and DO_NOT_SHARE &rarr; `Window.CANCEL`.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialogTest.java
chanseokoh: As we discussed, I couldn't make SWTBot find buttons with `buttonWithLabel()`.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/CompositeUtil.java
chanseokoh: Having trouble making `buttonWithLabel()`. Not sure why it doesn't work.

plugins/com.google.cloud.tools.eclipse.usagetracker.test/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialogTest.java
chanseokoh: Tried to simulate disposing the shell, but I faced some difficulty with it. Let's settle with `CLOSE_SHELL`.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/OptInDialog.java
chanseokoh: Will add the Javadoc.

PR #2637, opened 2017-12-06
Remove leading '/' when converting URL with drive letter on Windows

### LINKED ISSUE #2636: Creating new service account key may not be working on Windows
Had a chance to look into Kokoro and found the following failure. The problem is that `Paths.get("/T:/src/github/...")` throws an error. We make this call in the main bundle, so we should probably fix this.

```
java.nio.file.InvalidPathException: Illegal char <:> at index 2: /T:/src/github/google-cloud-eclipse/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/target/work/configuration/
	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at java.nio.file.Paths.get(Paths.java:84)
	at com.google.cloud.tools.eclipse.appengine.localserver.ui.GcpLocalRunTabTest.testGetServiceAccountKeyPath(GcpLocalRunTabTest.java:453)
```

chanseokoh:
Fixes #2636.

I've added a Windows-specific test. Let's see if it will succeed on Kokoro.


chanseokoh:
I just realized my assumption was wrong. Sadly, waste of time.

PR #2638, opened 2017-12-06
Convert URL to Path in a robust way

### LINKED ISSUE #2636: Creating new service account key may not be working on Windows

chanseokoh:
Fixes #2636.

I thought `Paths.get(url.toURI())` did not work as desired, but there was a different reason it was not working.

I believe `Platform.getConfigurationLocation().getURL().toURI()` will never throw an exception unless Eclipse is buggy.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
@@ -32,6 +32,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -241,7 +242,10 @@ public class GcpLocalRunTab extends AbstractLaunchConfigurationTab {
         BusyIndicator.showWhile(createServiceKey.getDisplay(), new Runnable() {
           @Override
           public void run() {
-            createServiceAccountKey(getServiceAccountKeyPath());
+            Path keyPath = getServiceAccountKeyPath();
+            if (keyPath != null) {
+              createServiceAccountKey(keyPath);
+            }
           }});
       }
     });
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ui/GcpLocalRunTab.java
```diff
@@ -413,9 +417,14 @@ Path getServiceAccountKeyPath() {
...
-    
-    String configurationLocation = Platform.getConfigurationLocation().getURL().getPath();
-    return Paths.get(configurationLocation + "/com.google.cloud.tools.eclipse/" + filename);
+
+    try {
+      Path configurationLocation = Paths.get(Platform.getConfigurationLocation().getURL().toURI());
+      return configurationLocation.resolve("com.google.cloud.tools.eclipse").resolve(filename);
+    } catch (URISyntaxException e) {
+      logger.log(Level.SEVERE, "Cannot convert configuration location into URI", e); //$NON-NLS-1$
+      return null;
```
briandealwis: Since this should never happen, is it worth either marking this as a thrown exception, or wrapping it into a `CoreException`?

chanseokoh: I don't feel like making an exception publically visible to callers, specifically because it is never meant to happen.

chanseokoh:
The previously offending tests are now passing on Kokoro.

Issue #2651, opened 2017-12-07
Expensive PipelineArgumentsTab.updatePipelineOptionsForm() is called twice when changing radio selection


chanseokoh:
We add the selection-changed listener below to the radio buttons. As the Javadoc says, the intention is to update the tab when the runner selection is changed.

```java
  /**
   * When the Runner selection is changed, update the underlying launch configuration, update the
   * PipelineOptionsForm to show all available inputs, and re-render the tab.
   */
  private class UpdateLaunchConfigAndRequiredArgsSelectionListener extends SelectionAdapter {
    ...
    public void widgetSelected(SelectionEvent e) {
      launchConfiguration.setRunner(runner);
      updatePipelineOptionsForm();
    }
  }
```

However, `widgetSelected()` is called not only when a button is selected but also when a button is de-selected. Since `updatePipelineOptionsForm()` is very expensive, we should basically do

```java
    @Override
    public void widgetSelected(SelectionEvent e) {
      Button radio = (Button) e.getSource();
      if (radio.getSelection()) {
        launchConfiguration.setRunner(runner);
        updatePipelineOptionsForm();
      }
    }
```

briandealwis:
ZOMG good find: this is why we see these _Pipeline Options Type XXX not found_ warnings logged (#2232).

chanseokoh:
Fixed by #2652.

PR #2652, opened 2017-12-08
Show UI progress during expense jobs / do not perform wrong update / simplify code

### LINKED ISSUE #2650: Refactor PipelineArgumentsTab.updatePipelineOptionsForm()
I think the future is completely unnecessary here and solely remains there to increase complexity with no good reason.

`getLaunchConfigurationDialog().run()` returns only after `optionsHierarchyFuture.set(launchConfiguration.getOptionsHierarchy(hierarchy)` is done, which is the most expense call that we wrap. Once the future set, the UI will run the future listener code almost immediately. We should just run the listener code after `run()`.

```java
  private void updatePipelineOptionsForm() {
    final SettableFuture<...> optionsHierarchyFuture = SettableFuture.create();
    optionsHierarchyFuture.addListener(new Runnable() {
      @Override
      public void run() {
        ...
      }
    }, displayExecutor);

    try {
      // blocking call (regardless of "fork"), returning only after the inner runnable completes
        @Override
        public void run(IProgressMonitor monitor)
            throws InvocationTargetException, InterruptedException {
          optionsHierarchyFuture.set(launchConfiguration.getOptionsHierarchy(hierarchy));
        }
      });
    } catch (...) {
      ...
    }
  }
```

chanseokoh:
Fixes #2650 #2651.


chanseokoh:
@briandealwis I think this won't conflict with your PR.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -399,36 +393,23 @@ public String getName() {
...
-              updateLaunchConfigurationDialog();
-            } catch (InterruptedException | ExecutionException ex) {
-              DataflowUiPlugin.logError(ex, "Exception while updating available Pipeline Options"); //$NON-NLS-1$
-            }
-          }
-        });
-      }
-    }, displayExecutor);
     try {
+      // TODO: use "Holder" instead of "AtomicReference" when Java 8 is available.
```
briandealwis: I don't see a Holder class in Java 8?

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -399,36 +393,23 @@ public String getName() {
...
         public void run(IProgressMonitor monitor)
             throws InvocationTargetException, InterruptedException {
-          optionsHierarchyFuture.set(launchConfiguration.getOptionsHierarchy(hierarchy));
+          monitor.beginTask(Messages.getString("updating.pipeline.options"), //$NON-NLS-1$
+              IProgressMonitor.UNKNOWN);
+          optionsHierarchy.set(launchConfiguration.getOptionsHierarchy(hierarchy));
         }
       });
+
+      pipelineOptionsForm.updateForm(launchConfiguration, optionsHierarchy.get());
```
briandealwis: This `updateForm` was expensive and hence the `BusyIndicator`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
chanseokoh: You're right. Can't find a good class. I think I should just update comment to say "this is just a holder; no atomicity is required."

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
chanseokoh: I removed it after noticing that it doesn't work on my machine, but maybe it's only my machine? I think I rather put this into `run()`, which does work on my machine.

chanseokoh: Hmm... putting it `run()` is worse, as it consumes 100% of UI operations and the progress bar is frozen.

chanseokoh: Well, actually putting it into `run()` seems better. The 100% UI consumption seems inevitable anyway. (Basically, there is nothing much you can do if drawing UI itself is taking time.)

briandealwis: You can't put it into run() as you're forking the job (i.e., it's in a non-UI thread).  Could you add some variable comments to the run (i.e., `true /*fork*/`)?

chanseokoh: I used `syncExec()`. I'll add comments to variables.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -352,10 +345,13 @@ void updateRunnerButtons(MajorVersion majorVersion) {
   private void updateHierarchy(final MajorVersion majorVersion)
       throws InvocationTargetException, InterruptedException {
     // blocking call (regardless of "fork"), returning only after the inner runnable completes
-    getLaunchConfigurationDialog().run(true, true, new IRunnableWithProgress() {
+    getLaunchConfigurationDialog().run(true /*fork*/, true /*cancelable*/,
+        new IRunnableWithProgress() {
       @Override
       public void run(IProgressMonitor monitor)
           throws InvocationTargetException, InterruptedException {
+        monitor.beginTask(Messages.getString("loading.pipeline.options.hierarchy"), //$NON-NLS-1$
```
briandealwis: This should be moved into `getPipelineOptionsHierarchy()` — `beginTask()` should be called by the actual worker.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -68,11 +76,22 @@ public void testGetName() {
...
       IWorkspaceRoot workspaceRoot = mock(IWorkspaceRoot.class);
       when(workspaceRoot.getProject(anyString())).thenReturn(mock(IProject.class));
 
       ILaunchConfigurationDialog dialog = mock(ILaunchConfigurationDialog.class);
+      doAnswer(new Answer<Void>() {
+        @Override
+        public Void answer(InvocationOnMock invocation)
+            throws InvocationTargetException, InterruptedException {
+          IRunnableWithProgress runnable = invocation.getArgumentAt(2, IRunnableWithProgress.class);
+          runnable.run(new NullProgressMonitor());
```
briandealwis: Interesting… my only concern here is that any SWT work performed by this runnable will succeed in tests.

briandealwis: Would this be worth extracting into a helper class?

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
chanseokoh: I just realized that `beginTask()` should be called only once on a given monitor instance, to mark the begging of a main task:
```
	 * Notifies that the main task is beginning.  This must only be called once
	 * on a given progress monitor instance.
```
So, it cannot be called it inside `getPipelineOptionsHierarchy()`. You could then guess `getPipelineOptionsHierarchy()` could do `subTask()` or `Submonitor.convert()`, but they don't work like `beginTask()`. (For example, `Submonitor.convert()` can't affect the amount work of the top-level dialog progress bar.)

I think it makes sense to `beginTask()` here to mark the beginning of a main task and specifically to declare the amount work of `IProgressMonitor.UNKNOWN` — it makes the top-level progress bar to swing left and right.

chanseokoh: If `getPipelineOptionsHierarchy()` want to do anything with the given `monitor`, it can always decided to do there with `SubMonitor`. Here, I think it's conceptually sane to begin a task on its own.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -170,4 +163,14 @@ public boolean apply(Control control) {
       });
     }
   }
+
+  private static class RunnableExecutor implements Answer<Void> {
```
briandealwis: Could you please add javadoc here that this it intended to mock `IRunnableContext#run(fork, cancelable,runnable)`, as implemented by major wizard dialogs (e.g., ILaunchConfigurationDialog).

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -396,7 +396,7 @@ public String getName() {
 
   private void updatePipelineOptionsForm() {
     try {
-      // This is mere a reference holder; atomicity not required.
+      // This is merely a reference holder; atomicity not required.
```
briandealwis: or a mere reference holder :-)

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
chanseokoh: Sure, no problem.

briandealwis: The other thing to add is that unlike `IRunnableWithContext` implementations, the runnable will be executed on the active thread.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
briandealwis: It goes against the contract of `IProgressMonitor`.  Either pass through the monitor unmolested, OR you create a child monitor.

Why can't `getPipelineOptionsHierarchy()` be changed to:

```
  private PipelineOptionsHierarchy getPipelineOptionsHierarchy(IProgressMonitor monitor) {
    monitor.beginTask(xxx, IProgressMonitor.UNKNOWN);
    …
```

chanseokoh: > Why can't getPipelineOptionsHierarchy() be changed to:

What if the code is like this:

```java
  getPipelineOptionsHierarchy(majorVersion, monitor);  // some dumb code
  anotherMethodThatCallsBeginTask(monitor);
  hierarchy = getPipelineOptionsHierarchy(majorVersion, monitor);
}
```
You end up calling `beginTask()` three times, whereas you should call it once. I think each method should not mess up with what their caller did and confine their scope to their own boundary, i.e., using `subTask` or `SubMonitor`.

chanseokoh: I think `beginTask()` is special, in that it marks the beginning of a task. You can call it if you are really beginning a task, even though you are not doing anything.

briandealwis: In your example, you *must* pass in a child monitor of the original monitor.  Yes, each one calls `beginTask()`, but it's up to the *actual container that is reporting progress* to decide how sub-(sub-)*-monitors are reported.

(You can use `SubMonitor#convert()` to avoid needing a `beginTask()`.)

chanseokoh: Ah, I see. I got this wrong.

> (You can use SubMonitor#convert() to avoid needing a beginTask().)

As I said,

> (For example, SubMonitor.convert() can't affect the amount work of the top-level dialog progress bar.)

doing `convert()` doesn't give the result we want here. The progress bar won't swing. I'll make `getPipelineOptionsHierarchy()` call `beginTask()`.

plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -68,11 +77,15 @@ public void testGetName() {
...
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2165
     @Test
-    public void testInitializeForm_noExceptionForNonAccessibleProject() throws CoreException {
+    public void testInitializeForm_noExceptionForNonAccessibleProject()
+        throws CoreException, InvocationTargetException, InterruptedException {
       IWorkspaceRoot workspaceRoot = mock(IWorkspaceRoot.class);
       when(workspaceRoot.getProject(anyString())).thenReturn(mock(IProject.class));
 
       ILaunchConfigurationDialog dialog = mock(ILaunchConfigurationDialog.class);
+      doAnswer(new SynchronousIRunnableContextExecutor())
```
briandealwis: Now that's a mouthful :-)

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTab.java
```diff
@@ -352,11 +346,14 @@ void updateRunnerButtons(MajorVersion majorVersion) {
...
       throws InvocationTargetException, InterruptedException {
     // blocking call (regardless of "fork"), returning only after the inner runnable completes
-    getLaunchConfigurationDialog().run(true, true, new IRunnableWithProgress() {
+    getLaunchConfigurationDialog().run(true /*fork*/, true /*cancelable*/,
+        new IRunnableWithProgress() {
       @Override
       public void run(IProgressMonitor monitor)
           throws InvocationTargetException, InterruptedException {
-        hierarchy = getPipelineOptionsHierarchy(majorVersion, monitor);
+        SubMonitor subMonitor = SubMonitor.convert(
```
briandealwis: I believe you need to call done() on the monitor at the end here.  That's the protocol on IProgressMonitor (`beginTask()` must have a corresponding `done()`), and the same is said on `SubMonitor#convert()`

> It is not necessary to call done() on the result, but the caller is responsible for calling done() on the argument. Calls beginTask on the argument.

Though the examples of use on `SubMonitor` are confusing.

chanseokoh: We used `SubMonitor` everywhere, but we almost never called `done()`.  This is quite new to me. Maybe we should file an issue.

chanseokoh: I was reading comments of `SubMonitor`:

```
 * <p><b>BEST PRACTISES:</b></p>
 * 
 * <p>We recommend that newly-written methods follow the given contract:</p>
 * <ul>
 * <li>It WILL call beginTask on the IProgressMonitor.</li>
 * <li>It WILL accept a null argument, indicating that no progress should be reported and the operation cannot be cancelled.</li>
 * <li>It WILL NOT call done on the IProgressMonitor, leaving this responsibility up to the caller.</li>
 * </ul>
 * <p>If you wish to follow these conventions, you may copy and paste the following text into your method's JavaDoc:</p>
 * 
 * <pre>@param monitor the progress monitor to use for reporting progress to the user. It is the caller's responsibility
 *        to call done() on the given monitor. Accepts <code>null</code>, indicating that no progress should be
 *        reported and that the operation cannot be cancelled.</pre>
```

Note

```
you may copy and paste the following text into your method's JavaDoc:</p>
```

Note that following Javadoc of `convert()` is documentation for callers of `convert()`. That means, the caller in the Javadoc refers to the caller of the method that uses `convert()`. (In our case as an example, is the caller of `run()`.)

> It is not necessary to call done() on the result, but the caller is responsible for calling done() on the argument. Calls beginTask on the argument.

Reading this again, it all makes sense. We shouldn't call `done()` as the Javadoc says. It is the responsibility of the caller of our method `run()` to call `done()`.

PR #2662, opened 2017-12-11
fix warning by democking

elharo:
@briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
@@ -43,6 +43,7 @@ public class BannedElementTest {
         null);
   }
   
+  @SuppressWarnings("unlikely-arg-type")
   @Test
   public void testEquals() {
     BannedElement element1 = new BannedElement("message");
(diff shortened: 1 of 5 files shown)


chanseokoh:
Some of the warnings may be due to `anyMap()`, which I recently learnt we can use `anyMap(type.class, type.class)`.

elharo:
`anyMap(String.class, String.class)` does not seem to compile.

elharo:
Looks like what we need is `Matchers.anyMapOf`

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/BannedElementTest.java
```diff
@@ -43,6 +43,7 @@ public void testBannedElementConstructor_nullLocation() {
         null);
   }
   
+  @SuppressWarnings("unlikely-arg-type")
```
briandealwis: Hmm, maybe we should settle on some common settings because I don't see any warnings here, and I've checked that I have all _unlikely_ error/warnings on.

elharo: image

This is actually a very good warning. In almost all cases it would catch a real bug. Here it doesn't only because we are specifically testing equals returning false when comparing a different object type.

PR #2664, opened 2017-12-11
more tests on MANIFEST.MF

elharo:


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -191,6 +191,32 @@ public final void testBundleVendor() throws IOException {
...
+    Attributes manifest = getManifestAttributes();
+    assertEquals("1.0", manifest.getValue("Manifest-Version"));
+    assertEquals("2", manifest.getValue("Bundle-ManifestVersion"));
+  }
+
+  @Test
+  public final void testBundleClasspath() throws IOException {
+    Attributes manifest = getManifestAttributes();
+    assertEquals(".", manifest.getValue("Bundle-ClassPath"));
+    assertEquals("2", manifest.getValue("Bundle-ManifestVersion"));
```
chanseokoh: duplicate test

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/META-INF/MANIFEST.MF
```diff
@@ -1,4 +1,5 @@
 Manifest-Version: 1.0
+Bundle-ClassPath: .
```
briandealwis: This addition is not necessary as the default is "."

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -191,6 +191,32 @@ public final void testBundleVendor() throws IOException {
...
+  public final void testManifestVersion() throws IOException {
+    Attributes manifest = getManifestAttributes();
+    assertEquals("1.0", manifest.getValue("Manifest-Version"));
+    assertEquals("2", manifest.getValue("Bundle-ManifestVersion"));
+  }
+
+  @Test
+  public final void testBundleClasspath() throws IOException {
+    Attributes manifest = getManifestAttributes();
+    assertEquals(".", manifest.getValue("Bundle-ClassPath"));
```
briandealwis: This test will break on bundles that include jars, like googleapis.  The common error is specifying jar files and forgetting to include ".".

It would also be useful to verify that all elements here are specified in the `build.properties` > `bin.includes`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/META-INF/MANIFEST.MF
elharo: The question is whether we want to rely on the defaults or not. Most of our bundles do have this.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/BasePluginXmlTest.java
```diff
@@ -191,6 +192,32 @@ public final void testBundleVendor() throws IOException {
...
+    Attributes manifest = getManifestAttributes();
+    assertEquals("1.0", manifest.getValue("Manifest-Version"));
+    assertEquals("2", manifest.getValue("Bundle-ManifestVersion"));
+  }
+
+  @Test
+  public final void testBundleClasspath() throws IOException {
+    Attributes manifest = getManifestAttributes();
+    assertTrue(manifest.getValue("Bundle-ClassPath").contains("."));
+    assertEquals("2", manifest.getValue("Bundle-ManifestVersion"));
```
chanseokoh: You already have this assertion in `testManifestVersion()` above.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/META-INF/MANIFEST.MF
briandealwis: I don't see how this redundant information helps.  This default will never change as it would result in widespread breakage.

elharo: In that case, we should take it out of all the MANIFEST.MF where we already have this.

briandealwis: And add a test to fail if it's found!

PR #2665, opened 2017-12-11
Tweak target descriptors

briandealwis:
- Use Orbit's latest release repo to get fixes/patches (recommended practice, even with older releases)
- Use WTP release train repositories (e.g., `/webtools/repository/oxygen`) rather than specific versions; we're using the main Eclipse release train repositories (e.g., `/releases/oxygen`).
- Drop org.eclipse.datatools since it's not used: I included it as it was referenced in GPE
- Use consistent ordering of repositories

### DIFF
--- a/eclipse/mars/gcp-eclipse-mars.tpd
+++ b/eclipse/mars/gcp-eclipse-mars.tpd
@@ -27,7 +27,6 @@ location "http://download.eclipse.org/releases/mars/" {
 	org.eclipse.m2e.wtp.sdk.feature.feature.group
 	org.eclipse.mylyn.commons.feature.group
 	org.eclipse.jpt.jpa.feature.feature.group
-	//org.eclipse.datatools.sdk.feature.feature.group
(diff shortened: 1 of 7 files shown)


eclipse/mars/gcp-eclipse-mars.tpd
```diff
@@ -27,7 +27,7 @@ location "http://download.eclipse.org/releases/mars/" {
 	org.eclipse.m2e.wtp.sdk.feature.feature.group
 	org.eclipse.mylyn.commons.feature.group
 	org.eclipse.jpt.jpa.feature.feature.group
-	org.eclipse.datatools.sdk.feature.feature.group
+	//org.eclipse.datatools.sdk.feature.feature.group
```
chanseokoh: Why not removing entirely (in all three `.tpd` files)? Otherwise, a comment would help.

briandealwis: sure.  since they're not in the generated `.target` files, I'll just tweak them here directly.

Issue #2378, opened 2017-09-12
Flaky RunOptionsDefaultsComponentTest


chanseokoh:
Seen this twice on Kokoro today. The offending assertion was added into an existing test in #2375.

```
testEnablement_existingStagingLocation(com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest)  Time elapsed: 0.094 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest.testEnablement_existingStagingLocation(RunOptionsDefaultsComponentTest.java:264)
```

chanseokoh:
Also happens on Travis.

briandealwis:
Annoying.  I'm actively improving these tests with #2362, but I should be able to avoid this by manually invoking `#validate()`.

briandealwis:
`master` failed on `eclipse-mars` with:
```
testEnablement_selectedProject(com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest)  Time elapsed: 0.075 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest.testEnablement_selectedProject(RunOptionsDefaultsComponentTest.java:293)
```

chanseokoh:
I saw this timeout error several times on Travis and Kokoro. At first I thought it could be a rare transient network error, but now this happens multiple times, we need to do something.

```
Running com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest
Tests run: 18, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 33.69 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.dataflow.ui.preferences.RunOptionsDefaultsComponentTest
[...]

PR #2666, opened 2017-12-11
Install Spring Boot deps before tests; move to 1.5.6-RELEASE

briandealwis:
Cause Spring Boot dependencies to be downloaded to the local Maven repository by the build-time Maven instance rather than M2Eclipse at test time.  The hope is that this will avoid any repository corruption as tracked in #2284.

This change updates the two Spring Boot test projects to be consistent and use 1.5.6-RELEASE.


briandealwis:
Failure due to #2378.  Will merge with HEAD (with fix from #2672).

chanseokoh:
Looking at [this Kokoro log](https://pantheon.corp.google.com/logs/viewer?project=cloud-tools-for-eclipse-kokoro&minLogLevel=0&expandAll=false&timestamp=2017-12-13T23:16:12.983607262Z&advancedFilter=resource.type%3D%22project%22%0Aresource.labels.project_id%3D%22cloud-tools-for-eclipse-kokoro%22%0AlogName%3D%22projects%2Fcloud-tools-for-eclipse-kokoro%2Flogs%2Fkokoro-google-cloud-eclipse-gcp_windows-presubmit-2-20171213-144028%22&dateRangeStart=2017-12-13T22:36:38.741Z&dateRangeEnd=2017-12-13T23:36:38.741Z&interval=PT1H) from #2682 which failed, seems like a lot of dependencies are still missing.

Maybe we should try simplifying the zip test projects. For example,
- Ditch `springboot-appengine-standard.zip` and replace it with a non-Spring project. `ImportMavenAppEngineStandardProjectTest.runImport()` is the only consumer of the zip, which doesn't have to be Spring-specific.
- There might be some way to minimize `web-fragment-example.zip` or `spring-boot-test.zip` too.

chanseokoh:
`org.springframework:spring-web:jar:4.3.10.RELEASE` from `web-fragment-example.zip` is subsumed by `org.springframework.boot:spring-boot-starter-web:1.5.6.RELEASE`, so no need to update that.

elharo:
Simplifying the zip test projects sounds like a very good idea.

PR #2672, opened 2017-12-12
De-flake RunOptionsDefaultsComponentTest

chanseokoh:
Fixes one cause of https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2378#issuecomment-351155572, https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2378#issuecomment-351186875, and https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2378#issuecomment-351182009.


chanseokoh:
That requires doing `while (Display.getCurrent().readAndDispatch())` too in a loop, which I am not comfortable putting into `join()` of the test target class.

FYI, the reason we need event dispatch is because the future listeners should be able to run (which will in turn schedule jobs due to UI changes, e.g., project selection change).

chanseokoh:
Let's try to get this in early, as this is the main reason of Kokoro failures.

PR #2675, opened 2017-12-13
Code cleanup

chanseokoh:
`AppEngineStandardProjectWizard` should work, as its sibling `AppEngineFlexProjectWizard` test is already working fine. They both have the same base class.

Also fixed an issue number. That particular issue has been migrated to a new issue number.


plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/flex/AppEngineFlexProjectWizardTest.java
```diff
@@ -21,10 +21,10 @@
 
 public class AppEngineFlexProjectWizardTest {
 
-  private AppEngineFlexProjectWizard wizard = new AppEngineFlexProjectWizard();
+  private final AppEngineFlexProjectWizard wizard = new AppEngineFlexProjectWizard();
 
   @Test
-  public void testWindowTitle() {
+  public void testTitleSet() {
```
briandealwis: `testWindowTitle` seems more to the point, since it's testing the actual value of the window title, not just that it's set.

PR #2679, opened 2017-12-13
Fix up some unnecessary imports and throws

briandealwis:
Some trivial tweaks to remove unnecessary imports and throws.


briandealwis:
I just saw this happen!  I don't know *why* it's happening, but one workaround is to collapse and re-expand the item.

screen shot 2017-12-13 at 3 54 10 pm

PR #2683, opened 2017-12-14
Install dependency JARs that test projects require before starting Kokoro build 

chanseokoh:
Hopefully this avoids #2284.

Also updating one of the test projects to remove Spring repos (`https://repo.spring.io/libs-release`).

Lastly, reverting the manual dependency install attempt from #2666, which will be no longer necessary with this PR.


chanseokoh:
Things look good. This does not prevent every download (because some test projects are not Maven) in integration tests, but most of the downloads are suppressed.

PR #2390, opened 2017-09-15
Check if project supports Dataflow

briandealwis:
Use the [Cloud Service Management APIs](https://cloud.google.com/service-management/overview) to check if a project supports the Dataflow APIs.  Currently advisory-only: although an error is shown, the page-complete status is not affected when a project does not support dataflow as it's not required for local-only pipelines.

Fixes #2001

PR #2686, opened 2017-12-15
Build with ErrorProne on Travis

### LINKED ISSUE #2608: ErrorProne checks on each PR
Is there a Github hook for this?

briandealwis:
This PR adds a new profile `errorprone` that invokes the ErrorProne compiler, and fixes some errors identified by ErrorProne.  It changes the Travis build to use this `errorprone` profile.

We can't replace Tycho's use of ECJ as the `tycho-compiler-plugin` is specifically brought in as part of the `eclipse-plugin` packaging, and doesn't offer an option to skip compilation.  Although this double-compilation (ecj + errorprone/javac) increases the compilation time, the additional compilation time is dwarfed by testing time.

I had to rework `com.google.cloud.tools.eclipse.googleapis` and `.test.dependencies` slightly to bring in their dependent libraries using Maven's `<dependencies>` to keep the Travis build happy.  I had tried this approach for #2390 but we were getting a substantial number of dependencies copied in the `lib/` and I wasn't aware of the `maven-dependency-plugin`'s `<includeGroupIds>` option to limit dependencies.

(Aside: I'm still a bit puzzled why the `maven-compiler-plugin` needs these `<dependencies>` but `tycho-compiler-plugin` does not, as the `maven-compiler-plugin` has no problem resolving classes that are in other bundles.  I guess the `tycho-compiler-plugin` must recompute the classpath from the `Bundle-Classpath` before passing onto the ECJ compiler?  Expressing these as `<dependencies>` ensures the `maven-compiler-plugin` sees them always.)

Fixes #2608.  

### Time impacts 
Baseline: `mvn package` (just the `tycho-compiler-plugin` using ECJ):
```
[INFO] Total time: 01:15 min
[INFO] Final Memory: 200M/493M
```

With ErrorProne: `mvn -Perrorprone package` (ECJ + ErrorProne)
```
[INFO] Total time: 01:42 min
[INFO] Final Memory: 246M/453M
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,8 @@ before_script:
   - sleep 3 # give xvfb some time to start
   - metacity --sm-disable --replace &
   - sleep 3 # give metacity some time to start
-script: ./mvnw -V -B -Ptravis --fail-at-end verify ${MAVEN_FLAGS}
+script: ./mvnw -V -B --fail-at-end verify ${MAVEN_FLAGS}
+     -Ptravis -Perrorprone
      ${ECLIPSE_TARGET:+-Declipse.target=${ECLIPSE_TARGET}}
      --toolchains=.travis-toolchains.xml
      -Dtoolchain.java.runtime=${JAVA_RUNTIME}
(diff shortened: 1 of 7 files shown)


briandealwis:
> With this, tests will fail if they hit an errorprone warning, right?

Yup.

> Can all errorprone warnings be suppressed if it's a false positive?

ErrorProne uses `@SuppressWarnings` to disable checks.  The error messages from ErrorProne point to a web page with detail.  For example, [GcpProjectTest.java](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2686/files#diff-caa65d3ce642eaddf0885d3f18a15163R75) triggered the [`SelfEquals`](http://errorprone.info/bugpattern/SelfEquals) pattern, which explains how to disable the check.  I don't know that all patterns can be disabled, but the only one we've hit so far can be :-)

chanseokoh:
Cool!

PR #2689, opened 2017-12-20
Respect virtual component rules for resolving files in WEB-INF

### LINKED ISSUE #2564: WebProjectUtil.getWebInfDirectory() doesn't always work as expected
I've got curious about https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1557#issuecomment-342517225; I can reproduce it whenever importing the [bookshelf-standard sample](https://github.com/GoogleCloudPlatform/getting-started-java/tree/master/bookshelf-standard/2-structured-data) into Eclipse.

It turns out `WebProjectUtil.getWebInfDirectory()` sometimes doesn't work, and in this case it was returning `/target/m2e-wtp/web-resources/WEB-INF` instead of the correct `/src/main/webapp/WEB-INF`.

Surprisingly, the result depends on the order of XML elements in `.settings/org.eclipse.wst.common.component`. For example, if we have `<wb-resource deploy-path="/" source-path="/src/main/webapp" tag="defaultRootSource"/>` before `<wb-resource deploy-path="/" source-path="/target/m2e-wtp/web-resources"/>` like below

```xml
        ...
        <wb-resource deploy-path="/" source-path="/src/main/webapp" tag="defaultRootSource"/>
        <wb-resource deploy-path="/" source-path="/target/m2e-wtp/web-resources"/>
        <wb-resource deploy-path="/WEB-INF/classes" source-path="/src/main/java"/>
        ...
    </wb-module>
</project-modules>
```

then it returns "/src/main/webapp", which is what we want. But if the two `wb-resource` elements are switched, then "/target/m2e-wtp/web-resources" is returned.

Our code makes use of `getRootFolder()`, and it may be just returning the first root (`deploy-path="/"`).

```java
    if (component != null && component.exists()) {
      IVirtualFolder root = component.getRootFolder();
      ...
        return (IFolder) root.getFolder(WEB_INF).getUnderlyingFolder();
```

There should be a way to get the path tagged with `defaultRootSource`. Not sure if that's possible.

briandealwis:
Use WTP's virtual component model to resolve files and where to place new files and directories, rather than guess the right `WEB-INF` directory.

The attached `test-dynamic-web-project-dynamicrootsource.zip` is a simple _Dynamic Web Project_ with the following `.settings/org.eclipse.wst.common.component`:
```xml
<project-modules id="moduleCoreId" project-version="1.5.0">
    <wb-module deploy-name="test-dynamic-web-project-dynamicrootsource">
        <wb-resource deploy-path="/" source-path="target/m2e-wtp/web-resources"/>
        <wb-resource deploy-path="/" source-path="/src/main/webapp" tag="defaultRootSource"/>
        <wb-resource deploy-path="/WEB-INF/classes" source-path="/src/main/java"/>
        <property name="context-root" value="test-dynamic-web-project-dynamicrootsource"/>
        <property name="java-output-path" value="/test-dynamic-web-project-dynamicrootsource/build/classes"/>
    </wb-module>
</project-modules>
```
I originally tried creating the test projects programmatically for the `defaultRootSource`-tagged case, but it proved exceedingly annoying.

Fixes #2564


chanseokoh:
Too many tests are failing. I haven't looked into it, but I guess this change has a big impact.

chanseokoh:
I believe https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2564#issuecomment-342574475 should be fixed too.

briandealwis:
> I believe #2564 (comment) should be fixed too.

- `DeployAssemblyEntryRemoveJob.run()` uses the virtual component API `removeLink()` which is fine.
- `WebXmlValidator.resolveJsp()` is checking for the existence of an `IFile` by calling `IVirtualFile#getUnderlyingFile()`, which is ok as it will look in the associated directories for the file.  But the code could just use `IVirtualResource#exists()` directly.

chanseokoh:
Hmm... maybe we are not on the same page. Let's talk offline for a bit.

chanseokoh:
I believe #1557 is a separate issue. I remember the warning of #2564 was harmless for the bookshelf example in #1557.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtilTest.java
```diff
@@ -0,0 +1,373 @@
...
+    // should be found alongside the web.xml
+    IFolder libFolder = WebProjectUtil.createWebInfFolder(project, new Path("lib"), null);
+    assertNotNull(libFolder);
+    assertTrue(libFolder.exists());
+    assertEquals("web/WEB-INF/lib", libFolder.getProjectRelativePath().toString());
+  }
+
+
+  @Test
+  public void testFindWebInfFile_dwpProject() {
```
chanseokoh: `dwp` is quite unfamiliar. `dynamicWebProject` (or just `wtpProject`)?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -83,12 +132,33 @@ public static IFolder getWebInfDirectory(IProject project) {
    * @return the file location or {@code null} if not found
```
chanseokoh: Can you update the Javadoc to explicitly say

"Note that this can return a file outside the project source directory; it can return a file in a build directory (for example, under `target/m2e-wtp/web-resources/WEB-INF`), which can get cleaned frequently."

I can imagine a developer tries to use this function to find and manipulate `web.xml` in the source repo, not knowing that it can return `web.xml` in a build directory. Also, in this regard, please update the Javadoc to explain this subtle difference from what can happen with `createWebInfFile()`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -45,22 +53,63 @@
...
+   * component model. Specifically, we create files and directories in the
+   * {@code &lt;wb-resource&gt;} with the {@code defaultRootSource} tag, if found. We do not
+   * overwrite the file.
+   * 
+   * @param project the hosting project
+   * @param filePath the path of the file within the project's {@code WEB-INF}
+   * @param contents the content for the file
+   * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=448544">Eclipse bug 448544</a>
+   */
+  public static IFile createWebInfFile(IProject project, IPath filePath, InputStream contents,
```
chanseokoh: I gave some thoughts on this, and I think this isn't perfect yet. About the reason why we go through all the hassle of respecting the WTP virtual component model, you said earlier that

> ... there *may* be multiple directories where it (`appengine-web.xml`) could exist. .... You *could* have a `WebContent`, `web`, `src/main/webapp` all in the same project.
>
> In this case we're only creating the `appengine-web.xml` *if* it doesn't already exist.

However, it will still create `appengine-web.xml` in this case if the file isn't at the `defaultRootSource`. For example,

1. `src/main/webapp` is tagged as `defaultRootSource`.
1. `src/main/webapp/WEB-INF/appengine-web.xml` doesn't exist.
1. `WebContent` is another web-resource source-path.
1. `WebContent/WEB-INF/appengine-web.xml` exists.

Then you'll create a brand new `appengine-web.xml`, even though the project already had its own file.

But I'd say at least we are not making matters worse from now, so let's create a low priority bug rather than making this PR more complicated. Having multiple web resource folders sounds like a corner case.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -45,22 +53,63 @@
...
+   * @param project the hosting project
+   * @param folderPath the path of the folder within the project's {@code WEB-INF}
+   * @param contents the content for the file
+   * @see <a href="https://bugs.eclipse.org/bugs/show_bug.cgi?id=448544">Eclipse bug 448544</a>
    */
-  public static IFolder getWebInfDirectory(IProject project) {
-    // Try to obtain the directory as if it was a Dynamic Web Project
+  public static IFolder createWebInfFolder(IProject project, Path folderPath,
+      IProgressMonitor monitor) throws CoreException {
+    SubMonitor progress = SubMonitor.convert(monitor, 2);
```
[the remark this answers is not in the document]
briandealwis: ok

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
briandealwis: > Then you'll create a brand new appengine-web.xml, even though the project already had its own file.

The behaviour you're describing is deliberate: I'm assuming that the caller will use `findInWebInf()` first and only call `createWebInfFile()` if the file is not found.  Since `createWebInfFile()` is provided the file-contents, it didn't seem helpful to search the various `<wb-resource>` directories again.  And with m2e's content-filtering, it might be useful to be able to explicitly create a file that m2e might then re-filter.

I'll change the javadoc for `createWebInfFile()` to make it explicit that resolving is done outside.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CreateAppEngineWtpProject.java
```diff
@@ -231,12 +231,8 @@ private void addJstl12ToClasspath(IProject newProject, IProgressMonitor monitor)
...
 
     // locate WEB-INF/lib
-    IFolder webInfFolder = WebProjectUtil.getWebInfDirectory(newProject);
-    IFolder libFolder = webInfFolder.getFolder("lib"); //$NON-NLS-1$
-    if (!libFolder.exists()) {
-      libFolder.create(true, true, subMonitor.newChild(5));
-    }
-
+    IFolder libFolder =
+        WebProjectUtil.createWebInfFolder(newProject, new Path("lib"), subMonitor.newChild(5)); //$NON-NLS-1$
```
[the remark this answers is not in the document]
briandealwis: I see what you mean.

chanseokoh:
Can you take a look at the Kokoro failure that seems Windows-specific? (Ignore Travis for now. I think it has some technical problems.) It looks like it is failing to delete a project. Perhaps closing `foo.txt` explicitly will fix it.

```
Contains: Problems encountered while deleting files.
Contains: Could not delete: T:\src\github\google-cloud-eclipse\plugins\com.google.cloud.tools.eclipse.appengine.facets.test\target\work\data\test-dynamic-web-project-dynamicrootsource\src\main\webapp\WEB-INF\foo.txt.
java.nio.file.FileSystemException: T:\src\github\google-cloud-eclipse\plugins\com.google.cloud.tools.eclipse.appengine.facets.test\target\work\data\test-dynamic-web-project-dynamicrootsource\src\main\webapp\WEB-INF\foo.txt: The process cannot access the file because it is being used by another process.
```

briandealwis:
Ah, kokoro discovered a file leak.  Helpful!

briandealwis:
Kokoro failure is #2704 and will be fixed by #2706.

briandealwis:
Uh oops, I somehow missed a number of other calls to `IFile.getContents()`.

briandealwis:
PTAL: Kokoro errors fix.

PR #2700, opened 2018-01-04
Rough sketch of core Cloud SDK installation

chanseokoh:
The basic idea: whoever needs a `CloudSDK` calls `CloudSdkManager.getCloudSdk()`. If the Cloud SDK has never been installed by our tools, it will trigger installation. `getCloudSdk()` is basically non-blocking, so if the Cloud SDK is not ready, it returns `null`. So, for example, if that happens, callers would tell the user that the Cloud SDK is not ready yet and the user should try it later.

We can discuss what we can/should do if installation fails.

~~The initial sketch doesn't do anything about installation progress reporting~~, but I am positive we should do it. One way could be to use the usual Eclipse job. That way, we might be able to better report failure, but still, all we can do might just be to surface Java exceptions and expect the user to analyze them and take proper actions.


plugins/com.google.cloud.tools.eclipse.sdk/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,8 @@ Require-Bundle: com.google.cloud.tools.appengine;bundle-version="0.4.2",
  org.eclipse.e4.core.di;bundle-version="1.5.0"
 Import-Package: com.google.common.base;version="[20.0.0,21.0.0)",
  com.google.common.collect;version="[20.0.0,21.0.0)",
+ org.eclipse.core.runtime,
```
elharo: I don't see where this is used.

chanseokoh:
I think we'll soon add Eclipse-specific things, e.g., once we decided how to handle failure and report progress.

plugins/com.google.cloud.tools.eclipse.sdk/META-INF/MANIFEST.MF
chanseokoh: I was tinkering with the Eclipse `Job`. I pushed a commit that sketches the use of it to report progress through the Eclipse `Job` framework.

patflynn:
The null/'try again later' thing sounds a bit weird.  I think you'd have to either block and install, or provide progress information asynchronously.

chanseokoh:
ㄱ

chanseokoh:
Added more fine-grained progress reporting through the Eclipse job framework.

selection_003

chanseokoh:
When a failure occurs during installation, the exception will be reported through the Eclipse job exception handling.

problem occurred _005

The error is logged and can be seen in the _Error Log_ view.

event details _006

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,65 @@
...
+public class CloudSdkManager {
+
+  /**
+   * Returns {@code CloudSdk} only if there exists an up-to-date Cloud SDK with the App Engine Java
+   * component installed at the managed location. Otherwise, returns {@code null} after triggering a
+   * background installation task. Callers who get {@code null} may need to try this method at later
+   * points.
+   *
+   * Note that even if the Cloud SDK is already installed, this method executes {@code gcloud} twice
+   * externally to check components and versions, which may incur a noticeable delay. If possible,
```
elharo: delete "externally"

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class CloudSdkManager {
+
+  /**
+   * Returns {@code CloudSdk} only if there exists an up-to-date Cloud SDK with the App Engine Java
+   * component installed at the managed location. Otherwise, returns {@code null} after triggering a
+   * background installation task. Callers who get {@code null} may need to try this method at later
+   * points.
+   *
+   * Note that even if the Cloud SDK is already installed, this method executes {@code gcloud} twice
```
elharo: Executing twice on every call seems excessive. I doubt we need to check for updates more than once a day.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,65 @@
...
+   *     component installed at the managed location; {@code null} otherwise
+   * @throws ManagedSdkVerificationException if unable to verify SDK installation; possible causes
+   *     include corrupted installation or a problem executing binaries
+   */
+  public static CloudSdk getCloudSdk() throws ManagedSdkVerificationException {
+    try {
+      ManagedCloudSdk managedSdk = ManagedCloudSdk.newManagedSdk();
+      if (!managedSdk.isInstalled()
+          || !managedSdk.hasComponent(SdkComponent.APP_ENGINE_JAVA)
+          || !managedSdk.isUpToDate()) {
```
elharo: The isUpTodate check should probably be removed since that's the one that hits the network.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,65 @@
...
+        new CloudSdkInstallJob(managedSdk).schedule();
+        return null;
```
elharo: The null return worries me a bit. It seems error prone. If we threw a checked exception instead we could guarantee this was handled.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,65 @@
...
+      }
+
+      Path sdkHome = managedSdk.getSdkHome();
+      CloudSdk sdk = new CloudSdk.Builder().sdkPath(sdkHome).build();
+      return sdk;
+
+    } catch (UnsupportedOsException e) {
+      throw new RuntimeException("Cloud Tools for Eclipse supports Windows, Linux, and Mac only.");
```
elharo: Do we think this is unreachable?

chanseokoh: I'm not sure, honestly, and what could be the best way to handle this.

But first, I did not want to make callers to handle `UnsupportedOsException` every time. 

Secondly, if this happens, it means appengine-plugins-core cannot auto-install the Cloud SDK at all, which means CT4E should have a way for the user to manually configure the Cloud SDK location if we were to handle this case nicely in practice. But our plan is to basically remove such manual configuration UX. So, if this ever happens, we rather need to fix appengine-plugins-core to support the OS. In that sense, CT4E can't support such an OS.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: Yeah, that would reduce two calls to one call. Or, should we also take out `hasComponent()` which calls `gcloud`? I can think of a corner case if we take it out, but I also think there is a way to prevent that.

Regarding checking for the updates once a day, we could

- check for updates whenever Eclipse starts (we could start installing the SDK right after the start-up if it has never been installed, but it may also be a good idea to defer installation)

and

- schedule a job that wakes up every 24 hours and checks for updates.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: I think it's a good idea. We could rather check updates periodically in background as you mentioned earlier.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: Practical idea.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: It turns out `hasComponent()` (effectively `gcloud components list`) hits the network too.

```
$ gcloud components list
ERROR: (gcloud.components.list) Failed to fetch component listing from server. Check your network settings and try again.
$ 
```

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: Let's tackle this in a later PR, as it will require a lot of new code and I think how we will tackle this exactly is still to be discussed.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Returns {@code CloudSdk} if there exists an up-to-date Cloud SDK with the App Engine Java
+   * component installed at the managed location. Otherwise, throws {@code
+   * CloudSdkNotReadyException} after triggering a background installation task. Callers who get the
+   * exception may need to try this method at later points.
+   *
+   * Note that even if the Cloud SDK is already installed, this method executes {@code gcloud} twice
+   * to check components and versions, which may incur a noticeable delay. If possible, avoid
+   * unnecessary repetitive calls.
```
briandealwis: Shouldn't this multiple-invocation be an internal detail?  On the first fetch, we may perform some checks, but no checks should be required on subsequent fetches.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
```diff
@@ -0,0 +1,84 @@
...
+ * Wraps {@code CloudSdkInstallTask} in {@code Job} mainly to report install progress to users.
+ */
+public class CloudSdkInstallJob extends Job {
+
+  public static final String JOB_FAMILY = CloudSdkInstallJob.class.getName();
+
+  /** Scheduling rule to prevent running CloudSdkInstallTask concurrently. */
+  private static final MutexRule MUTEX_RULE = new MutexRule();
+
+  private final ManagedCloudSdk managedSdk;
```
briandealwis: Could make sense for `ManagedCloudSdk` to implement `ISchedulingRule`.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
```diff
@@ -0,0 +1,84 @@
...
+
+  public CloudSdkInstallJob(ManagedCloudSdk managedSdk) {
+    super("Installing the Google Cloud SDK... (may take up to several minutes)");
```
briandealwis: I don't think the `(may take up to…)` should be part of the job name.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  protected IStatus run(IProgressMonitor monitor) {
+    CloudSdkInstallTask installTask = new CloudSdkInstallTask(managedSdk, monitor);
+    Future<Void> futureTask = Executors.newSingleThreadExecutor().submit(installTask);
+
+    try {
+      while (true) {
+        if (monitor.isCanceled()) {
+          // By the design of the appengine-plugins-core SDK downloader, cancellation support is
+          // implemented through the Java thread interruption facility.
```
briandealwis: Ahhh!  This implementation had seemed really puzzling until I saw this comment.

Jobs do execute on a thread (`getThread()`).  When a Job is cancelled, the Jobs framework will call `canceling()`: this Job can interrupt itself there.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallTask.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override
+  public Void call() throws ManagedSdkVerificationException, IOException, InterruptedException,
+      SdkInstallerException, CommandExecutionException, CommandExitException {
+    SubMonitor subMonitor = SubMonitor.convert(monitor, 50);
+
+    try {
+      if (!managedSdk.isInstalled()) {
+        subMonitor.setTaskName("Installing the core SDK...");
```
briandealwis: Rather than use `setTaskName()`, which replaces the current task name, it might be better to use `subTask()`.  So a single `setTaskName("Installing Google Cloud SDK")` and then multiple `subTask()` calls for `"Installing Core SDK"`, `"Installing App Engine Java component"`, etc.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
```diff
@@ -0,0 +1,84 @@
...
+
+  public CloudSdkInstallJob(ManagedCloudSdk managedSdk) {
+    super("Installing the Google Cloud SDK... (may take up to several minutes)");
+    this.managedSdk = managedSdk;
+    setUser(true);
+    setRule(MUTEX_RULE);
+  }
+
+  @Override
+  protected IStatus run(IProgressMonitor monitor) {
```
briandealwis: Should we have another isUpdateToDate() check here — in case several jobs were somehow scheduled?

PR #2707, opened 2018-01-09
always pick latest Maven plugin version

### LINKED ISSUE #2014: Use latest appengine-maven-plugin in generated pom.xml
Currently the version is fixed to be "1.3.1" in the template `pom.xml.standard.ftl` and `pom.xml.flex.ftl`. We can check the latest version with something like this:

```java
    String latestVersion = MavenUtils.resolveLatestReleasedArtifactVersion(
        monitor,
        "com.google.cloud.tools", //$NON-NLS-1$
        "appengine-maven-plugin", //$NON-NLS-1$
        "maven-plugin", //$NON-NLS-1$
        DEFAULT_VERSION);
```

elharo:
@briandealwis fix #2014

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
@@ -41,6 +41,7 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.osgi.framework.Version;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
@@ -216,6 +217,13 @@ public class CodeTemplatesTest {
     Assert.assertEquals("my-project-artifact-id", artifactId.getTextContent());
     Element version = (Element) root.getElementsByTagName("version").item(0);
     Assert.assertEquals("98.76.54", version.getTextContent());
+    
+    Element pluginVersion =
+        (Element) root.getElementsByTagName("appengine.maven.plugin.version").item(0);
+    Version v = new Version(pluginVersion.getTextContent());
+    Assert.assertTrue(v.getMajor() >= 1);
+    Assert.assertTrue(v.getMinor() >= 3);
+    Assert.assertTrue(v.getMicro() >= 2);
   }
 
   private Document buildDocument(IFile xml)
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -195,6 +196,15 @@ private static void createPomXml(IProject project, AppEngineProjectConfig config
     properties.put("projectGroupId", config.getMavenGroupId()); //$NON-NLS-1$
     properties.put("projectArtifactId", config.getMavenArtifactId()); //$NON-NLS-1$
     properties.put("projectVersion", config.getMavenVersion()); //$NON-NLS-1$
+    
+    String mavenPluginVersion = ArtifactRetriever.DEFAULT.getLatestReleaseVersion(
+        "com.google.cloud.tools", "appengine-maven-plugin") //$NON-NLS-1$ $NON-NLS-2$
+        .toString(); 
+    if (mavenPluginVersion == null) {
+      mavenPluginVersion = "1.3.2";  //$NON-NLS-1$
```
briandealwis: Should we make `getLatestReleaseVersion()` take a fallback version?

elharo: Possibly, though personally I've always found that pattern sort of confusing. I don't usually expect a method argument to be unused or conditionally used.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
```diff
@@ -216,6 +217,13 @@ private void validatePomXml()
     Assert.assertEquals("my-project-artifact-id", artifactId.getTextContent());
     Element version = (Element) root.getElementsByTagName("version").item(0);
     Assert.assertEquals("98.76.54", version.getTextContent());
+    
+    Element pluginVersion =
+        (Element) root.getElementsByTagName("appengine.maven.plugin.version").item(0);
+    Version v = new Version(pluginVersion.getTextContent());
+    Assert.assertTrue(v.getMajor() >= 1);
+    Assert.assertTrue(v.getMinor() >= 3);
+    Assert.assertTrue(v.getMicro() >= 2);
```
chanseokoh: This will fail if 1.4.0 is released?

PR #2709, opened 2018-01-10
Use latest versions of libraries where possible

elharo:


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -107,6 +107,10 @@ static Library create(IConfigurationElement configurationElement) throws Library
         MavenCoordinates mavenCoordinates = getMavenCoordinates(
             libraryFileElement.getChildren(ELEMENT_NAME_MAVEN_COORDINATES));
         LibraryFile libraryFile = loadSingleFile(libraryFileElement, mavenCoordinates);
+        // Pin servlet versions; update everything else
+        if (!libraryFile.getMavenCoordinates().getGroupId().startsWith("javax.servlet")) {
+          libraryFile.updateVersion();
+        }
```
chanseokoh: The library stuff always confuses me. I think this is not really effective, because when we call `Library.getAllDependencies()` on it, it will call `resolveDependencies()` which in turn calls `LibraryFile.updateVersion()`. Do you remember the recent test failure due to the Endpoints version update issue? Whichever version we specify in `plugin.xml`, I think they can eventually be upgraded.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
```diff
@@ -107,6 +107,11 @@ static Library create(IConfigurationElement configurationElement) throws Library
         MavenCoordinates mavenCoordinates = getMavenCoordinates(
             libraryFileElement.getChildren(ELEMENT_NAME_MAVEN_COORDINATES));
         LibraryFile libraryFile = loadSingleFile(libraryFileElement, mavenCoordinates);
+        // Pin servlets, JSP, and guava; update everything else
+        if (!libraryFile.getMavenCoordinates().getGroupId().startsWith("javax.servlet")
```
briandealwis: I'd prefer modifying the plugin.xml definitions to support expressing pinning: there's no traceability from the library definitions to this code.  Especially since there are multiple javax.servlet* groups (e.g., javax.servlet, javax.servlet.jsp, javax.servlet.jsp.jstl), and who knows which group we'll need to select in the future.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -18,7 +18,7 @@
         <mavenCoordinates
               artifactId="appengine-api-1.0-sdk"
               groupId="com.google.appengine" 
-              version="1.9.59" />
+              version="1.9.60" />
```
briandealwis: Maybe we should make our update-to-latest more explicit and support using `version` as `LATEST` or `RELEASE`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/LibraryFactory.java
elharo: Good idea. Done.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
briandealwis: Good point.  In these cases the user won't be able to fetch the artifacts either.  I wonder if we wouldn't be better treating this as a fatal error and abort the current operation.

chanseokoh: Perhaps it is possible that, although fetching a latest version from MC fails, a fallback version exists locally and the user can continue the operation.

elharo: Not necessarily. It's very common to have many old versions sitting around in your local repo.

PR #2718, opened 2018-01-11
Remove unnecessary Cloud SDK checks

### LINKED ISSUE #2708: Installing App Engine facets probably doesn't require Cloud SDK
When installing the (std and flex) facets, we check if the Cloud SDK is installed through `AppEngineFacetInstallDelegate.execute()`. Because of this check, users cannot create a new project without the Cloud SDK. I have been thinking the SDK is required to create the App Engine Server Runtime.

However, I got really suspicious of this, so took a shot on testing this. I moved my local Cloud SDK directory, took down the Cloud SDK checks in code, deleted the App Engine Server Runtime, started a fresh workspace, and was able to create a new project without an error. Creating a new project successfully creates the App Engine Server Runtime if it doesn't exist. Of course, local run fails because CT4E can't find any Cloud SDK locally.

chanseokoh:
Fixes #2708. Removes a lot of code. Creating projects, converting GPE projects, converting to App Engine standard projects, and installing the facets work well.


plugins/com.google.cloud.tools.eclipse.appengine.compat/src/com/google/cloud/tools/eclipse/appengine/compat/gpe/GpeConvertCommandHandler.java
```diff
@@ -17,40 +17,30 @@
...
-      if (CloudSdkPrompter.getCloudSdkLocation(shell) == null) {
-        // no further action required: user chose to not configure
-        return null;
-      }
-
       IFacetedProject facetedProject = ProjectFacetsManager.create(project,
           true /* convert to faceted project if necessary */, null /* no monitor here */);
       if (AppEngineStandardFacet.hasFacet(facetedProject)) {
-        throw new IllegalStateException("Convert menu enabled for App Engine projects");
+        throw new IllegalStateException("Convert menu enabled for App Engine standard projects");
```
briandealwis: This is an odd message: "Convert menu should not be enabled …"?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
```diff
@@ -129,12 +110,5 @@ protected void retrieveConfigurationValues() {
   @Override
   public void init(IWorkbench workbench, IStructuredSelection selection) {
     this.workbench = workbench;
-    if (config.getCloudSdkLocation() == null) {
-      File location = CloudSdkPrompter.getCloudSdkLocation(getShell());
-      // if the user doesn't provide the Cloud SDK then we'll error in performFinish() too
-      if (location != null) {
-        config.setCloudSdkLocation(location);
```
briandealwis: Looks like you can get rid of AppEngineProjectConfig.cloudSdkLocation too

chanseokoh: Yeah, that has been removed.

PR #2720, opened 2018-01-11
Test to see if TRAVIS_SECURE_ENV_VARS is set

briandealwis:


briandealwis:
`TRAVIS_SECURE_ENV_VARS=false`, even though I'm a team member.  So `TRAVIS_SECURE_ENV_VARS=true` only for PRs on the repository, and never for forked repositories regardless of the state of the owner.

PR #2721, opened 2018-01-12
test if TRAVIS_SECURE_ENV_VARS is set

incompetentclod:


briandealwis:
The associated build log shows that the build environment (from `env`) does not include any `$travis_XXX_key` or `$travis_XXX_iv` values, and `$TRAVIS_SECURE_ENV_VARS` is `false`.

Issue #2711, opened 2018-01-10
SocketTimeoutException in verify-feature-completeness when attempting to install ADT


briandealwis:
Seeing occasional build failures on Travis in `verify-feature-completeness`, where we attempt to install CT4E alongside ADT, due to a timeout fetching from ADT's repository.  I suppose at some point we can get rid of this test as I doubt anybody is using ADT any more.

(This was on the Mars built target.)
```
[INFO] --- tycho-eclipserun-plugin:1.0.0:eclipse-run (default) @ verify-feature-completeness ---
[INFO] Adding repository file:/home/travis/build/GoogleCloudPlatform/google-cloud-eclipse/build/verify-feature-completeness//../../eclipse/ide-target-platform/target/repository
[INFO] Toolchain in tycho-eclipserun-plugin: JDK[/usr/lib/jvm/java-7-openjdk-amd64/jre]
[INFO] Expected eclipse log file: /home/travis/build/GoogleCloudPlatform/google-cloud-eclipse/build/verify-feature-completeness/target/eclipserun-work/data/.metadata/.log
[INFO] Command line:
[...]

PR #2722, opened 2018-01-12
Add .appengine.localserver dependency to prevent spurious NPEs

### LINKED ISSUE #2715: Stop logging could-not-find-runtime
Our build logs for `.deploy.test`, `.deploy.ui.test`,  are cluttered with entries like the following:
```
!ENTRY org.eclipse.core.jobs 4 2 2018-01-10 21:50:19.329
!MESSAGE An internal error occurred during: "Install App Engine runtimes in "test0.6152019114061741"".
!STACK 0
java.lang.NullPointerException: Could not find App Engine Standard runtime type
	at com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet.createAppEngineServerRuntime(AppEngineStandardFacet.java:411)
	at com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet.createAppEngineFacetRuntime(AppEngineStandardFacet.java:423)
	at com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet.installAllAppEngineRuntimes(AppEngineStandardFacet.java:380)
	at com.google.cloud.tools.eclipse.appengine.facets.StandardFacetInstallDelegate$AppEngineRuntimeInstallJob.run(StandardFacetInstallDelegate.java:114)
```

This happens as we add an App Engine Standard facet to a project, which attempts to install an App Engine server runtime on the project.  Our runtime definitions are in `.appengine.localserver`. Because Tycho attempts to minimize the bundles required, Tycho doesn't include `.appengine.localserver` when running these tests.

We could force these test projects to include `.appengine.localserver`, so the definition is available.

Or we could fix `.installAllAppEngineRuntimes()` to instead just do a simple log string that the runtime definition could not be found.

briandealwis:
Adds `com.google.cloud.tools.eclipse.appengine.localserver` as a bundle dependency to several of our test projects to ensure the App Engine Standard Runtime runtime and server definitions are found. This change avoids the spurious _ java.lang.NullPointerException: Could not find App Engine Standard runtime type_ errors in the build logs.

Fixes #2715

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/META-INF/MANIFEST.MF
@@ -6,6 +6,7 @@ Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.deploy.test
 Bundle-Vendor: Google Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.deploy
+Comment: .appengine.localserver required for WTP server runtime definitions
 Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies,
  com.google.cloud.tools.eclipse.appengine.localserver
 Import-Package: com.google.cloud.tools.eclipse.appengine.facets,
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/META-INF/MANIFEST.MF
@@ -6,7 +6,8 @@ Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.deploy.test
 Bundle-Vendor: Google Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.deploy
-Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/META-INF/MANIFEST.MF
```diff
@@ -6,7 +6,8 @@ Bundle-SymbolicName: com.google.cloud.tools.eclipse.appengine.deploy.test
 Bundle-Vendor: Google Inc.
 Bundle-Version: 0.1.0.qualifier
 Fragment-Host: com.google.cloud.tools.eclipse.appengine.deploy
-Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies
+Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies,
+ com.google.cloud.tools.eclipse.appengine.localserver
```
elharo: Import-Package wouldn't work?

briandealwis: No, since we're looking to pull in the extension.  An alternative is to pull this in via a `tycho—platform-configuration` `<extraRequirements>` in  pom.xml.

elharo: Okay. It would be nice to have a comment in file indicating this since I know I'm going to look at this a year from now and have no idea why we did it this way. E.g.
Comment: these next two are required instead of imported because...

briandealwis:
Failures due to #2711.

Issue #2690, opened 2017-12-20
Copious debug output from TimeoutAwareConnectionFactoryTest


briandealwis:
I noticed the Travis always complains that the logs are too long to be displayed.  I was looking for a test failure and noticed that the server for the TimeoutAwareConnectionFactoryTest is dumping debug-level information which takes up pages.

```
Running com.google.cloud.tools.eclipse.googleapis.internal.TimeoutAwareConnectionFactoryTest
16:11:17.179 [main] DEBUG org.eclipse.jetty.util.log - Logging to Logger[org.eclipse.jetty.util.log] via org.eclipse.jetty.util.log.Slf4jLog
16:11:17.187 [main] INFO  org.eclipse.jetty.util.log - Logging initialized @5418ms
16:11:17.194 [main] DEBUG o.e.j.u.component.ContainerLifeCycle - org.eclipse.jetty.server.Server@301e76a9 added {qtp1546991174{STOPPED,8<=0<=200,i=0,q=0},AUTO}
16:11:17.254 [main] DEBUG o.e.j.u.component.ContainerLifeCycle - HttpConnectionFactory@71046bb7{HTTP/1.1} added {HttpConfiguration@593bf558{32768/8192,8192/8192,https://:0,[]},POJO}
16:11:17.256 [main] DEBUG o.e.j.u.component.ContainerLifeCycle - ServerConnector@16d29dd9{null}{0.0.0.0:0} added {org.eclipse.jetty.server.Server@301e76a9,UNMANAGED}
16:11:17.256 [main] DEBUG o.e.j.u.component.ContainerLifeCycle - ServerConnector@16d29dd9{null}{0.0.0.0:0} added {qtp1546991174{STOPPED,8<=0<=200,i=0,q=0},AUTO}
16:11:17.257 [main] DEBUG o.e.j.u.component.ContainerLifeCycle - ServerConnector@16d29dd9{null}{0.0.0.0:0} added {org.eclipse.jetty.util.thread.ScheduledExecutorScheduler@7c39ae4c,AUTO}
16:11:17.257 [main] DEBUG o.e.j.u.component.ContainerLifeCycle - ServerConnector@16d29dd9{null}{0.0.0.0:0} added {org.eclipse.jetty.io.ArrayByteBufferPool@5d791d43,POJO}
[...]

Issue #2557, opened 2017-11-02
Drop support for Mars


elharo:
This issue is, for now, for tracking thoughts. If this is going to cause major problems for anyone, please do comment below.

When Mars usage drops low enough, eliminate support for it.

Disadvantages:

  * Lose anyone still using Java 7
  * Lose anyone still committed to Mars

Advantages:

  * Can move appengine-plugins-core and rest of build to Java 8
  * Moving to Java 8 lets us move to Tycho 1.0
  * cut 1/3 of our Travis CI and flakes
  * Can use Submonitor.split() instead of newChild
  * LDP Dockertools requires Neon: https://marketplace.eclipse.org/content/eclipse-docker-tooling#group-details
  * Bring Guava to 21 or later
  * `private final WritableList/* <GcpProject> */ input; // Generics supported only in Neon+`

[LSP support](https://developers.redhat.com/blog/2016/11/24/deliver-support-for-new-languages-in-eclipse-ide-faster-with-generic-editor-and-language-servers/) is also interesting but would require us to drop Neon too.

Maybe when Photon is released in 2018?

briandealwis:
> cut 1/3 of our Travis CI and flakes

We'd probably add Photon instead.

elharo:
I suppose then we'd cut a quarter of the flakes we'd get if we kept all four. :-)

PR #2667, opened 2017-12-11
Let Kokoro pull Eclipse Oyxgen bundles from GCS

chanseokoh:

PR #2723, opened 2018-01-12
Improve SLF4j/Logback configuration

### LINKED ISSUE #263: slf4j warnings

briandealwis:
Fix #263, #2690

We're seeing copious amounts of unrelated HTTP traffic debugging traces from Jetty in our logs (#2690).  We're also seeing _DEBUG_ output from m2e-related operations like:
```
17:51:23.619 [Worker-7] DEBUG o.e.a.i.i.DefaultLocalRepositoryProvider - Using manager EnhancedLocalRepositoryManager with priority 10.0 for /home/travis/.m2/repository
```

Our custom `logback.xml` configuration file in `com.google.cloud.tools.eclipse.test.logback` configures Logback to only log _INFO_ messages, so this indicates that our `logback.xml` is not being used in all situations.  Ultimately the configuration approach we've used requires adding a dependency for each test project to our `.test.logback` project if there is any code that may somehow use SLF4j.

I realized that we can use a different approach to configuring Logback.  It turns out we can specify a system property to specify the configuration file as a URL, so we can use Eclipse-style `platform:/plugin/BUNDLEID/path/to/logback.xml` URLs to resolve the file within a bundle.  Because we've fixed up our test projects to include our `.test.dependencies` bundle, we can put the `logback.xml` in that bundle and then reference it from our tests.

So this PR:
  - moves the `logback.xml` to to `.test.dependencies`
  - configures tycho-surefire to reference the `logback.xml` from `.test.dependencies`
  - pulls in the SLF4j API facades for Apache Commons Logging (used by org.apache.* and freemarker) and Log4j (used by SWTBot) that redirect those logging APIs to the SLF4j binding; as a result no more log4j configuration errors
  - removes the `.test.logback` bundle

### Aside 1: Why use Logback

I did look at just using SLF4j's Simple Logger.  It has a few advantages in that it can be configured entirely by system properties.  But it doesn't include timestamps by default, which are useful for diagnosing problems.  And the default timestamp is just milliseconds from start, which is different from other timestamps shown.  It was easier to configure logback with a single file.

I did look at trying to use SLF4j's java.util.logging binding, but it wasn't in Orbit.  See below.

### Aside 2: We use SLF4j from Orbit

I did try to pull in the SLF4j bundles directly from Maven Central, as we could then use the java.util.logging binding as that binding is not in Orbit.  But the MC bundles don't use reverse-DNS bundle naming (e.g., the SLF4j API bundle is `slf4j.api`) and unfortunately older versions of m2e used `Require-Bundle: org.slf4j.api` which only matches the SLF4j API bundle in Orbit.  m2e now uses an Import-Package [since Neon](https://bugs.eclipse.org/bugs/show_bug.cgi?id=506473), so we can revisit this when we drop support for Mars (#2557).


briandealwis:
Kokoro is failing as I had to update the target platforms, and Oxygen.2 has since been released.  But our Oxygen mirror (#2667) only holds Oxygen.1.

briandealwis:
Failure due to #2711.

PR #2728, opened 2018-01-16
Use a Job in LaunchHelper to call IServer#start() asynchronously

### LINKED ISSUE #2727: Eclipse Oxygen very slow server startup when debug on GAE projects
I have a Java GAE (Google App Engine) project in Eclipse Oxygen Release 4.7.2 and Google Cloud Tools for Eclipse plugin version 1.5.0.

I have simple a Servlet class in my code:

```
package it.ale.test;

import java.io.IOException;

import javax.servlet.ServletException;
import javax.servlet.annotation.WebServlet;
import javax.servlet.http.HttpServlet;
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletResponse;

@WebServlet(name = "HelloAppEngine", urlPatterns = { "/hello" }, loadOnStartup = 1)
public class HelloAppEngine extends HttpServlet {

    @Override
    public void init() throws ServletException {

        super.init();

        System.out.println("Print something.");

    }

    @Override
    public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {

        response.setContentType("text/plain");
        response.setCharacterEncoding("UTF-8");

        response.getWriter().print("Hello App Engine!\r\n");

    }

}
```

when I put a **debug breakpoint** on any line of the _"public void init()"_ function, example on the line:

`System.out.println("Print something.");`

and I start the server in DEBUG mode **(Debug As -> App Engine)**, the server startup is incredible SLOW and I reach the error message:

> server app engine Standard at localhost was unable to start within 240 seconds. If the server requires more time, try increasing the timeout in the server editor.

briandealwis:
Fixes #2727 by causing `LaunchHelper`, which normally runs in the UI thread, to start the `IServer` instance in a background thread using a `Job`.  The Job is named _Launching {server}_, so that it appears in the progress area; I use "Launching" rather than "Starting" as WTP already creates a "Starting" job internally.

screen shot 2018-01-16 at 10 17 36 am

Problems are reported normally: the use of several jobs has no effect.

screen shot 2018-01-16 at 10 17 01 am

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
@@ -35,7 +35,11 @@ import java.util.logging.Logger;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.debug.core.ILaunch;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
@@ -130,8 +134,7 @@ public class LaunchHelper {
     return serverWorkingCopy.save(false, progress.newChild(2));
   }
 
-  protected void launch(IServer server, String launchMode, SubMonitor progress)
-      throws CoreException {
+  protected void launch(final IServer server, final String launchMode, SubMonitor progress) {
     // Explicitly offer to save dirty editors to avoid the puzzling prompt-to-save in
     // IServer#start() that prompts the user *as the server continues to launch*.
     // ServerUIPlugin.saveEditors() respects the "Save editors before starting the server"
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
```diff
@@ -130,16 +134,29 @@ private IServer createServer(IModule[] modules, SubMonitor progress) throws Core
...
     // Explicitly offer to save dirty editors to avoid the puzzling prompt-to-save in
     // IServer#start() that prompts the user *as the server continues to launch*.
     // ServerUIPlugin.saveEditors() respects the "Save editors before starting the server"
     // preference.
     if (!ServerUIPlugin.saveEditors()) {
       return;
     }
-    server.start(launchMode, progress);
+
+    // Launch in a Job. IServer#start()'s javadoc is incorrect: the server is not launched
```
elharo: Have you filed a bug or sent a patch to fix this Javadoc upstream?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/launching/LaunchHelper.java
```diff
@@ -130,16 +134,29 @@ private IServer createServer(IModule[] modules, SubMonitor progress) throws Core
...
+    // asynchronously but instead respects the serverType's `synchronousStart` setting.
```
chanseokoh: If so, how about setting false to `synchronousStart`?

chanseokoh: Oh, I read in #2727 why it is set synchronous.

briandealwis:
Hmm, I came across [WTP bug 203142](https://bugs.eclipse.org/bugs/show_bug.cgi?id=203142), which emphatically states that `IServer#start()` should be asynchronous.  In re-reading through the Server code, it seems the description of the `serverType`'s `synchronousStart` attribute is incorrect, which says that if true then publish and start are done serially.  The code in `Server#getPublishAndStartAsynchChainedJob()`, called when `synchronousStart=false` (or is unspecified), still ensures the publish and start steps are serialized.

I'll revert the above change and instead remove `synchronousStart`.

Issue #2724, opened 2018-01-12
Auto-install and update Cloud SDK


chanseokoh:
Listing some tasks. There is no consistency about the granularity of the items; some may be too specify and others may be too big. There may be missing items too.

- [ ] Add the feature debug flag
- [ ] Define new preferences / set initial preferences correctly
- [ ] New preference page UI
- [ ] Core install/update infrastructure
- [ ] Install at deploy
- [ ] Install at local run
- [ ] Periodic/occasional update feature when auto-managing
- [ ] Ensure updates are blocked while local run or deploy is in progress and vice versa
- [ ] Update notification when not configured to not auto-update
- [ ] Install in the background at other opportunistic times (e.g., when installing std/flex facets, Eclipse restarts, once a day, etc.)

PR #2734, opened 2018-01-17
Add feature flag for Managed Cloud SDK development

chanseokoh:
#2724.

Travis will probably fail until the Endpoint version issue is fixed.


plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class CloudSdkManager {
+
+  private static final String OPTION_MANAGED_CLOUD_SDK =
+      "com.google.cloud.tools.eclipse.sdk/managed.cloud.sdk";
+
+  public final static boolean MANAGED_SDK;
+
+  static {
+    WorkbenchPlugin activator = WorkbenchPlugin.getDefault();
```
briandealwis: Not a pressing concern since this is short-lived code, but  usually you should use your own ServiceReference or ServiceTracker: if the workbench is stopped or uninstalled, its services are released.  Plus WorkbenchPlugin is an internal class.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.eclipse.osgi.service.debug.DebugOptions;
+import org.eclipse.ui.internal.WorkbenchPlugin;
+
+public class CloudSdkManager {
+
+  private static final String OPTION_MANAGED_CLOUD_SDK =
+      "com.google.cloud.tools.eclipse.sdk/managed.cloud.sdk";
+
+  public final static boolean MANAGED_SDK;
```
briandealwis: Not really relevant here since this flag will be short-lived, but the DebugOptions class is intended to be lightweight.  Rather than a static, why not a `isManaged()` method that queries the value?  The advantage is that there are no changes requires for consumers if we decide to allow dynamically toggling the option via the _Tracing_ preference page.

plugins/com.google.cloud.tools.eclipse.sdk/.options
```diff
@@ -0,0 +1,4 @@
+# Uncomment for development. To be loaded in a hosted debug Eclipse instance,
+# open the "gcloud-eclipse-tools" run configuration, go to the "Tracing" tab,
+" and check "Enable tracing" and "com.google.cloud.tools.eclipse.sdk".
+#com.google.cloud.tools.eclipse.sdk/managed.cloud.sdk=true
```
briandealwis: If you don't have this uncommented then it won't show up in the tracing tab.  It should be set to `false` to reflect the default.

plugins/com.google.cloud.tools.eclipse.sdk/.options
```diff
@@ -0,0 +1,4 @@
+# Set true for development. To be loaded in a hosted debug Eclipse instance,
+# open the "gcloud-eclipse-tools" run configuration, go to the "Tracing" tab,
+# and check "Enable tracing" and "com.google.cloud.tools.eclipse.sdk".
+com.google.cloud.tools.eclipse.sdk/managed.cloud.sdk=false
```
briandealwis: Could make this a bit more descriptive? `enable.managed.cloud.sdk`?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class CloudSdkManager {
+
+  private static final String OPTION_MANAGED_CLOUD_SDK =
+      "com.google.cloud.tools.eclipse.sdk/managed.cloud.sdk";
+
+  public static boolean isManaged() {
+    BundleContext context = FrameworkUtil.getBundle(CloudSdkManager.class).getBundleContext();
+    DebugOptions debugOptions = context.getService(context.getServiceReference(DebugOptions.class));
+    return debugOptions.getBooleanOption(OPTION_MANAGED_CLOUD_SDK, false);
```
briandealwis: It is possible for `getService()` to return `null`.

chanseokoh: Really in practice? Anyway, I'll check null.

PR #2735, opened 2018-01-18
Remove _libraries.container file when Google Cloud Libraries container is removed

### LINKED ISSUE #2523: Problem when removing Google APIs container and adding new libraries later

briandealwis:
Fixes #2523

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Activator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Activator.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
 package com.google.cloud.tools.eclipse.appengine.libraries;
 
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Activator.java
```diff
@@ -0,0 +1,69 @@
+
```
elharo: Copyright

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Activator.java
```diff
@@ -0,0 +1,69 @@
...
+    private void visitChildren(IJavaElementDelta delta) {
+      for (IJavaElementDelta childDelta : delta.getAffectedChildren()) {
+        visit(childDelta);
+      }
+    }
+  };
+
+  @Override
+  public void start(BundleContext context) throws Exception {
```
elharo: probably doesn't need to throw Exception

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Activator.java
```diff
@@ -0,0 +1,69 @@
...
+    JavaCore.addElementChangedListener(listener, ElementChangedEvent.POST_CHANGE);
+  }
+
+  @Override
+  public void stop(BundleContext context) throws Exception {
```
elharo: probably doesn't need to throw Exception

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPath.java
```diff
@@ -286,4 +287,46 @@ public static void saveLibraryList(IJavaProject project, List<Library> libraries
...
+    SubMonitor progress = SubMonitor.convert(monitor, 5);
+    LibraryClasspathContainerSerializer serializer = new LibraryClasspathContainerSerializer();
+    if (!serializer.hasLibraryIds(javaProject)) {
+      // nothing to do since no library list file found
+      return;
+    }
+    progress.worked(1);
+
+    // Check if we can find our master library container
+    IClasspathEntry[] rawClasspath;
```
elharo: might be clearer if we used one try block and pushed this declaration and following into it

Issue #2743, opened 2018-01-22
New preference page UI for managed Cloud SDK feature


chanseokoh:

chanseokoh:
Fixed by #2746.

PR #2746, opened 2018-01-22
Create new managed Cloud SDK preferences and UI

### LINKED ISSUE #2742: Define user preferences for managed Cloud SDK feature
We need new preferences for

- auto-manged vs. custom SDK selection
- auto-update

Default values should also be set.

chanseokoh:
Fixes #2742 #2743.

- Defines two new preference properties.
- New preference page UI

selection_001

The UI is prevented from being enabled by the feature-flag.


briandealwis:
I think we still need some indication to allow manual updates and to trigger an update.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManagerTest.java
```diff
@@ -17,13 +17,26 @@
...
 
+  @After
+  public void tearDown() {
+    CloudSdkManager.forceManagedSdkFeature = false;
+  }
+
   @Test
   public void testManagedSdkOption() {
-    assertFalse(CloudSdkManager.isManaged());
+    assertFalse(CloudSdkManager.managedSdkFeatureEnabled());
```
briandealwis: I think an `is` prefix helps understand the purpose — I had to reread it a couple of times as I imagined that it enabled the management feature and returned the previous state.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/PreferenceConstants.java
```diff
@@ -28,10 +28,26 @@
...
   static final String BUNDLEID = "com.google.cloud.tools.eclipse.sdk";
 
+  public static enum CloudSdkManagement {
+    MANAGED, CUSTOM
+  }
+
+  /**
+   * Preference name for how the Google Cloud SDK is managed.
+   *
+   * @see CloudSdkManagement
```
briandealwis: Should be clear what the acceptable preference values are? `@see CloudSdkManagement#name()`?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/PreferenceConstants.java
```diff
@@ -28,10 +28,26 @@
   // host bundle for the preference
   static final String BUNDLEID = "com.google.cloud.tools.eclipse.sdk";
 
+  public static enum CloudSdkManagement {
```
briandealwis: `CloudSdkManagementOptions`?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/Activator.java
```diff
@@ -27,7 +27,7 @@
     @Override
     public void preferenceChange(PreferenceChangeEvent event) {
       Object newValue = event.getNewValue();
-      if (PreferenceConstants.CLOUDSDK_PATH.equals(event.getKey())
+      if (PreferenceConstants.CLOUD_SDK_PATH.equals(event.getKey())
```
briandealwis: Does this need to check that the `CloudSdkManagement` option is `CUSTOM`?  We should probably retain the preference if the user switches to `MANAGED` should they switch back again.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/Activator.java
```diff
@@ -37,6 +37,7 @@ public void preferenceChange(PreferenceChangeEvent event) {
   @Override
   public void start(BundleContext context) throws Exception {
     PreferenceConstants.getPreferenceNode().addPreferenceChangeListener(listener);
+    PreferenceConstants.initializeDefaults();
```
briandealwis: It seems odd for a constants class to have an `initializeDefaults()` method.  It also seems odd to need an `initializeDefaults()` since `ScopedPreferenceStore` should be consulting the `DefaultScope` too.  We should be able to use the `org.eclipse.equinox.preferences.preferences` extension point to provide an initializer. This has the benefit of not actually changing the preference values unnecessarily.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -75,32 +91,96 @@ public void widgetSelected(SelectionEvent event) {
...
+      updateControlEnablement();
+    }
     sdkLocation.load();
-    fireValueChanged(VALUE, "", "");
+    fireValueChanged(VALUE, "", ""); //$NON-NLS-1$ //$NON-NLS-2$
   }
 
   @Override
   public void loadDefault() {
+    if (CloudSdkManager.managedSdkFeatureEnabled()) {
```
briandealwis: This doesn't seem quite right: I actually think we shouldn't ever reset the sdkLocation (what is a meaningful value?).  But I think we can reset the SDK management option and the auto-update values to the default.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -75,32 +91,96 @@ public void widgetSelected(SelectionEvent event) {
...
+      loadSdkManagement(true /* loadDefault */);
+      autoUpdateCheck.loadDefault();
+      updateControlEnablement();
+    }
```
briandealwis: I wonder if loadDefault() should also have a `fireValueChanged()`.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -75,32 +91,96 @@ public void widgetSelected(SelectionEvent event) {
...
+        @Override
+        public void widgetSelected(SelectionEvent e) {
+          updateControlEnablement();
+        }
+      });
+
+      managedSdkSubArea = new Composite(parent, SWT.NONE);
+      autoUpdateCheck = new BooleanFieldEditor(
+          PreferenceConstants.CLOUD_SDK_AUTO_UPDATE,
+          Messages.getString("AutoUpdate"), BooleanFieldEditor.DEFAULT, managedSdkSubArea);
```
briandealwis: I think this should have a `//$NON-NLS-1$`

elharo:
Trigger an update, maybe. I don't think we need an indication to allow manual updates. If a user requires manual updates for some reason they can select an explicit SDK. That is, there are to options:

1. The automatic managed SDK that requires no user intervention.
2. A manual SDK

Nothing in between.

briandealwis:
Makes sense to me.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
chanseokoh: You mean, the existing code below `sdkLocation.loadDefault()` doesn't feel right? We can remove that. (In fact, we have never given what the default value is for that preference constant.)

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
chanseokoh: Currently `fireValueChanged()` is only about `sdkLocation`, and since I'm going to remove `sdkLocation.loadDefault()`, we won't fire the value changed event.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/Activator.java
chanseokoh: I think we don't need to check the option here. We retain the SDK location value regardless of the managed option, i.e., switching back and forth won't fire the change anyway.

chanseokoh:
New UI:

selection_001

@briandealwis I still need to take care of `org.eclipse.equinox.preferences.preferences` to set the defaults, but I think this will give you how the preferences will work. Basically, we introduce only one pref constant: `CLOUD_SDK_MANAGEMENT`.

briandealwis:
I think the explanatory text needs slight tweaking? It doesn't convey that CT4E will install it automatically.  _local_ can mean different things.
> Google Cloud Tools for Eclipse requires the Google Cloud SDK. If not found, an SDK will automatically be installed and updated periodically.
> [ ] Use the following SDK

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/Activator.java
briandealwis: I think this does need tweaking.  The purpose of this code is so that code can use the E4 DI to do:
```
@Inject CloudSdk sdk;
```
But I introduced this before we had the pluggable path resolver, and most uses will need to use the `CloudSdk.Builder` to add listeners, etc.  We should tear this out.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/PreferenceConstants.java
```diff
@@ -40,4 +52,9 @@ static IPreferenceStore getPreferenceStore() {
   static IEclipsePreferences getPreferenceNode() {
     return InstanceScope.INSTANCE.getNode(BUNDLEID);
   }
+
+  static void initializeDefaults() {
```
briandealwis: Hmm… I think we could actually reduce this to just using the CLOUD_SDK_PATH preference.  We use the Managed SDK only if a Cloud SDK was not found in the usual places.  I'm thinking that if I launch Eclipse with `GOOGLE_CLOUD_SDK_HOME` set to a location, it should override any managed location.

parkjam4:
I think @briandealwis is right that the text could be tweaked a little. What about this idea:
[ ] Allow Eclipse to manage the SDK
[ ] Manually manage the SDK [  path/to/sdk  ]  [ Browse ]

chanseokoh:
@parkjam4 that was basically the first UI that I tried (see the screenshot of https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2746#issue-290607188), but later @elharo suggested removing such radio buttons.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/PreferenceConstants.java
chanseokoh: I think it's good to have an explicit preference for the management option, as there are cases where the behavior becomes unclear. For example, suppose CT4E did not find the SDK in the usual places so it has been set to auto-managed. What if the user later installs an SDK manually? Should CT4E continue to auto-manage its SDK or switch to use the local one? Similarly, what if the user was using a local SDK and then the SDK became temporarily available because the user moved the directory? Should CT4E start downloading and installing its own SDK?

briandealwis:
Sounds like we're converging back on the radio buttons ;)

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/Activator.java
briandealwis: Stripping out CloudSdk injection: #2751

PR #2749, opened 2018-01-25
Minor fixes

chanseokoh:
I had a branch hanging around for too long.


plugins/com.google.cloud.tools.eclipse.dataflow.ui.test/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/PipelineArgumentsTabTest.java
```diff
@@ -102,7 +102,7 @@ public void setUp() throws CoreException, InvocationTargetException, Interrupted
 
       project1 = mockProject("project1");
       javaProject1 = mock(IJavaProject.class);
-      when(workspaceRoot.getProject(eq("project1"))).thenReturn(project1);
```
briandealwis: Ok, I misread Mockito's requirements for argument matchers.  I thought that if any invocation of a method used an arg matcher, then all uses of that method requires  arg martchers.  But it's that if any arg used a matcher within an invocation then all args require an arg matcher.

PR #2751, opened 2018-01-26
Remove CloudSdk injection support

briandealwis:
We had some support to allow using E4 Dependency Injection to write code like `@Inject CloudSdk sdk`.  But since key configuration values requires using the `CloudSdk.Builder`, it actually wasn't that useful.


chanseokoh:
I just merged #2746. I'll take a look at this once you resolve merge conflicts.

PR #2761, opened 2018-01-29
Automatically determine the CLOUD_SDK_MANAGEMENT if not previously set

### LINKED ISSUE #2744: Perform one-time initialization of managed Cloud SDK preferences when the feature first runs
Per our design doc, when the feature is released and runs for the first time, it should check if the user is already using a Cloud SDK in CT4E. In that case, preference should be initialized (once) to use the custom SDK.

The exact criterion for determining the initial values is documented in the design doc.

briandealwis:
Change preference initializer to check if the `CLOUD_SDK_MANAGEMENT` preference has been set and, if not set, determine the appropriate setting.  If the user had previously set the location of the Cloud SDK or the Cloud SDK is found in a well-known location, then we configure the CLOUD_SDK_MANAGEMENT preference to being `MANUAL`.  Otherwise we set it to `AUTOMATIC`.

Fixes #2744


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/TestPreferencesRule.java
```diff
@@ -0,0 +1,60 @@
...
+    return new ScopedPreferenceStore(InstanceScope.INSTANCE, storeId);
+  }
+
+  @Override
+  protected void after() {
+    super.after();
+    try {
+      preferences.removeNode();
+    } catch (BackingStoreException ex) {
+      logger.log(java.util.logging.Level.WARNING, "Failed clearing preferences: " + storeId, ex);
```
chanseokoh: Why the full qualification of `Level`?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
```diff
@@ -57,7 +61,41 @@ static IEclipsePreferences getPreferenceNode() {
...
-    getPreferenceStore().setDefault(CLOUD_SDK_MANAGEMENT,
-        CloudSdkManagementOption.AUTOMATIC.name());
+    initializeDefaultPreferences(getPreferenceStore());
+  }
+
+  @VisibleForTesting
+  public void initializeDefaultPreferences(IPreferenceStore preferences) {
+    preferences.setDefault(CLOUD_SDK_MANAGEMENT, CloudSdkManagementOption.AUTOMATIC.name());
+    if (CloudSdkManager.isManagedSdkFeatureEnabled()
+        && preferences.isDefault(CLOUD_SDK_MANAGEMENT)) {
```
chanseokoh: Shouldn't this be `preferences.getString(CLOUD_SDK_MANAGEMENT).isEmpty()`? I think `isDefault()` is basically `.getString(CLOUD_SDK_MANAGEMENT).equals("AUTOMATIC")` here.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
```diff
@@ -57,7 +61,41 @@ static IEclipsePreferences getPreferenceNode() {
...
+      configureManagementPreferences(preferences, isCloudSdkAvailable());
+    }
+  }
+
+  /** Configure the managed SDK settings given current settings. */
+  @VisibleForTesting
+  static void configureManagementPreferences(
+      IPreferenceStore preferences, boolean cloudSdkAvailable) {
+    // has the user previously set the Cloud SDK path? has it been found in a well-known location?
+    if (!preferences.isDefault(CLOUD_SDK_PATH) || cloudSdkAvailable) {
```
chanseokoh: Ditto. I think this should be `!preferences.getString(...).isEmpty()`.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
```diff
@@ -39,8 +47,21 @@ public CloudSdkPreferenceResolver() {
 
   @Override
   public Path getCloudSdkPath() {
+    // We only consult the Managed Cloud SDK when it has been explicitly configured, which
+    // is done in CloudSdkPreferences.
+    if (!preferences.isDefault(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT)) {
```
chanseokoh: Shouldn't this be `preferences.getString(CLOUD_SDK_MANAGEMENT).isEmpty()`? I think `isDefault()` is basically `.getString(CLOUD_SDK_MANAGEMENT).equals("AUTOMATIC")` here.

Perhaps we don't need this check at all, since we will return the managed Cloud SDK path only when `CloudSdkPreferences.CLOUD_SDK_MANAGEMENT` is set to `"AUTOMATIC"` by the following line, and if it is set to `AUTOMATIC`, it means the preferences are completely set up.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
```diff
@@ -39,8 +47,21 @@ public CloudSdkPreferenceResolver() {
...
+      if(CloudSdkManagementOption.AUTOMATIC.name().equals(preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT))) {
```
chanseokoh: Code style: `if (`

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
```diff
@@ -39,8 +47,21 @@ public CloudSdkPreferenceResolver() {
...
+        System.err.println("Should call: Job.getJobManager().join(FAMILY OBJECT)");
```
chanseokoh: Remove `System.err`.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferencesTest.java
```diff
@@ -17,20 +17,63 @@
...
-    new CloudSdkPreferences().initializeDefaultPreferences();
+    CloudSdkPreferences.configureManagementPreferences(preferences, false /*cloudSdkAvailable*/);
+    assertFalse(preferences.isDefault(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT));
+    assertEquals("AUTOMATIC", preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT));
+  }
+
+  @Test
+  public void testConfigure_cloudSdkAvailable() {
+    IPreferenceStore preferences = preferencesCreator.getPreferenceStore();
+    preferences.putValue(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT, "MANUAL");
```
chanseokoh: I think we should not set a value to reflect the env where the preferences have never configured.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferencesTest.java
```diff
@@ -17,20 +17,63 @@
...
+
+    assertEquals("FOO", preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT));
+    assertEquals(
+        "AUTOMATIC", preferences.getDefaultString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT));
+  }
+
+  @Test
+  public void testConfigure_noCloudSdkAvailable() {
+    IPreferenceStore preferences = preferencesCreator.getPreferenceStore();
     preferences.putValue(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT, "MANUAL");
```
chanseokoh: I think we should not set a value to reflect the env where the preferences have never configured.

briandealwis: Oh jeepers, this shouldn't be here.  Oh the plus side, `CloudSdkPreferences#configureManagementPreferences()` doesn't consult this value, but it means the `isDefault()` test is useless.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferencesTest.java
briandealwis: Oh jeepers, again.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
briandealwis: Hmm… the javadoc for `isDefault()` is a bit ambiguous:
```java
    /**
     * Returns whether the current value of the preference with the given name
     * has the default value.
     *
     * @param name the name of the preference
     * @return <code>true</code> if the preference has a known default value
     * and its current value is the same, and <code>false</code> otherwise
     * (including the case where the preference is unknown to this store)
     */
    public boolean isDefault(String name);
```
The `@return` description's "_the preference has a known default value and its current value is the same_" can be interpreted as either _the set value is the same as the default value_ or _the value returned is taken from the default value_.

 In practice, `ScopedPreferenceStore` and JFace's `PreferenceStore` use the latter interpretation: that the value would be taken from the defaults.

In reality that's the only useful interpretation as it's impossible to do a check like what I'm doing here: if we only have a default then …

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
briandealwis: To avoid the possibly ambiguity in `isDefault()`, I can move the `setDefault()` and instead check if it's set.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/TestPreferencesRule.java
briandealwis: Um, I'm not sure.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
```diff
@@ -57,7 +61,42 @@ static IEclipsePreferences getPreferenceNode() {
 
   @Override
   public void initializeDefaultPreferences() {
-    getPreferenceStore().setDefault(CLOUD_SDK_MANAGEMENT,
-        CloudSdkManagementOption.AUTOMATIC.name());
+    initializeDefaultPreferences(getPreferenceStore());
+  }
+
+  @VisibleForTesting
+  public void initializeDefaultPreferences(IPreferenceStore preferences) {
```
chanseokoh: No need for `public`.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
```diff
@@ -57,7 +61,42 @@ static IEclipsePreferences getPreferenceNode() {
...
+      preferences.setDefault(CLOUD_SDK_MANAGEMENT, CloudSdkManagementOption.AUTOMATIC.name());
+    }
+  }
+
+  /** Configure the managed SDK settings given current settings. */
+  @VisibleForTesting
+  static void configureManagementPreferences(
+      IPreferenceStore preferences, boolean cloudSdkAvailable) {
+    // has the user previously set the Cloud SDK path? has it been found in a well-known location?
+    if (preferences.contains(CLOUD_SDK_PATH) || cloudSdkAvailable) {
```
chanseokoh: For `CLOUD_SDK_PATH`, I think it's much safer and better to check `! ... .isEmpty()` instead. Leaving the field empty doesn't feel like the user has actually set the path. In this case, I think it's nicer to auto-download the SDK if it is locally not available.

briandealwis: ok, fair enough.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
```diff
@@ -39,8 +47,22 @@ public CloudSdkPreferenceResolver() {
 
   @Override
   public Path getCloudSdkPath() {
+    // We only consult the Managed Cloud SDK when it has been explicitly configured, which
+    // is done in CloudSdkPreferences.
+    if (preferences.contains(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT)) {
+      String managementOption = preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT);
+      if (CloudSdkManagementOption.AUTOMATIC.name().equals(managementOption)) {
+        // XXX: Should check and wait on the installation job, if in progress
```
elharo: I'm not sure about this. We might be able to push all of that down into appengine-plugins-core rather than checking explicitly.

PS. Google style is, I think, TODO for todo comments.

plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
```diff
@@ -10,7 +10,8 @@ Bundle-ActivationPolicy: lazy
 Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies,
  org.eclipse.ui.ide,
  org.eclipse.core.jobs,
- org.eclipse.wst.validation
+ org.eclipse.wst.validation,
+ org.eclipse.equinox.preferences;bundle-version="3.5.300"
```
elharo: why bundle here instead of import?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
```diff
@@ -57,7 +62,42 @@ static IEclipsePreferences getPreferenceNode() {
...
+      preferences.setValue(CLOUD_SDK_MANAGEMENT, CloudSdkManagementOption.MANUAL.name());
+    } else {
+      preferences.setValue(CLOUD_SDK_MANAGEMENT, CloudSdkManagementOption.AUTOMATIC.name());
+    }
+  }
+
+  /** Return {@code true} if the Cloud SDK is available. */
+  private static boolean isCloudSdkAvailable() {
+    try {
+      new CloudSdk.Builder().build();
```
elharo: This is weird. We create a cloud sdk to find out if we can? There's no simpler way? And this isn't a method in appengine-plugins-core?

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/TestPreferencesRule.java
```diff
@@ -0,0 +1,61 @@
...
+    return new ScopedPreferenceStore(InstanceScope.INSTANCE, storeId);
+  }
+
+  @Override
+  protected void after() {
+    super.after();
+    try {
+      preferences.removeNode();
+    } catch (BackingStoreException ex) {
+      logger.log(Level.WARNING, "Failed clearing preferences: " + storeId, ex);
```
elharo: do we really need to log this? Our tests are full of too much log junk already. Or maybe log at debug level if we absolutely must.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/TestPreferencesRule.java
```diff
@@ -0,0 +1,61 @@
...
+    return preferences;
+  }
+  
+  public IPreferenceStore getPreferenceStore() {
+    return new ScopedPreferenceStore(InstanceScope.INSTANCE, storeId);
+  }
+
+  @Override
+  protected void after() {
+    super.after();
```
elharo: super.after comes last.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
briandealwis: It's not a trivial change in `CloudSdk`.  Basically the `CloudSdk.Builder` consults a set of resolvers, found using Java's `ServiceLoader`, to find the Cloud SDK location.

plugins/com.google.cloud.tools.eclipse.test.util/META-INF/MANIFEST.MF
briandealwis: I had thought I would be bringing in classes the org.eclipse.core.runtime package, which is split.  But I'm not.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/TestPreferencesRule.java
briandealwis: FINE it is then.

briandealwis:
Weird: the kokoro build reports it succeeded.

plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/TestPreferencesRule.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.logging.Logger;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.ui.preferences.ScopedPreferenceStore;
+import org.junit.rules.ExternalResource;
+import org.osgi.service.prefs.BackingStoreException;
+
+/**
+ * Test utility to create create an empty {@link IEclipsePreferences} or {@link IPreferenceStore}.
```
chanseokoh: `to create create`: duplicate words

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
```diff
@@ -39,8 +47,22 @@ public CloudSdkPreferenceResolver() {
...
   public Path getCloudSdkPath() {
+    // We only consult the Managed Cloud SDK when it has been explicitly configured, which
+    // is done in CloudSdkPreferences.
+    if (preferences.contains(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT)) {
+      String managementOption = preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT);
+      if (CloudSdkManagementOption.AUTOMATIC.name().equals(managementOption)) {
+        // TODO: Should check and wait on the installation job, if in progress
+        try {
+          return ManagedCloudSdk.newManagedSdk().getSdkHome();
+        } catch(UnsupportedOsException ex) {
```
chanseokoh: style: `catch (`

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferences.java
```diff
@@ -57,7 +62,42 @@ static IEclipsePreferences getPreferenceNode() {
...
+      return true;
+    } catch(CloudSdkNotFoundException ex) {
```
chanseokoh: style: `catch (`

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkPreferenceResolver.java
briandealwis: I wonder why these aren't being picked up by the formatter…

PR #2764, opened 2018-01-30
Interface for installing managed Cloud SDK

chanseokoh:
Honestly, I think it is a minor flaw not to do anything about concurrent installing, updating, and using the SDK, but anyway, I am proposing an interface for installing the managed SDK to address such concurrency issues.

So, for deploy as an example, we'll do something like this:

```java
public class DeployJob extends WorkspaceJob {

  @Override
  public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
       ...
      CloudSdkManager.installManagedSdk(stdoutOutputStream);
      ...
      CloudSdkManager.suspendModifyingSdk();
      try {
         ... = stageProject(...);
         ...
         ... = deployProject(...);
         ...
      } finally {
        CloudSdkManager.resumeModifyingSdk();
      }
      ...
}
```

Both `installManagedSdk()` and `suspendModifyingSdk()` are blocking calls, intended to be used by the deploy job and the local run.

`installManagedSdkAsync()` and `updateManagedSdkAsync()` are for scheduling a background install/update job, to be implemented via an Eclipse `Job`.

What do you think?


chanseokoh:
> I think it is a minor flow not to do anything about installing, updating, and using the SDK

Oops, meant to say

I think it is a minor flow not to do anything about installing, updating, and using the SDK *concurrently*

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -43,6 +46,90 @@ public static boolean isManagedSdkFeatureEnabled() {
...
+  private static int suspenderCounts = 0;
+  private static boolean isModifyingSdk = false;
+
+  /**
+   * Suspends potential SDK auto-install or auto-update temporarily. This is to use the Cloud SDK
+   * safely by preventing modifying the SDK while using the SDK. If an install or update has already
+   * started, blocks callers until the install or update is complete. Callers must call {@code
+   * CloudSdkManager#allowModifyingSdk} eventually to lift the suspension.
+   *
+   * Any callers that intend to use {@code CloudSdk} must always call this before staring work, even
```
elharo: Might be less error prone to flip this. E.g. the cloud sdk sets a flag when we start downloading, and it throws an exception or perhaps blocks if you try to do anything while it's downloading.

chanseokoh: It actually works like that (sort of) at the minimum, without this method. However, I also realized we nees to block the other direction as well: block installing or updating if someone is using the SDK, like deploying. Only the deploy job knows when it is done using the SDK.

briandealwis:
I don't think this suspend/resume is a great idea.  If an update took 10 minutes or more, then perhaps, but it's generally done in less than a minute.  I'd prefer that we block the launch or whatever in the same manner how launches are blocked on a build.

screen shot 2018-01-30 at 10 29 26 am

chanseokoh:
Suspend/resume is not for blocking launches or deploys. It's about blocking install/update during launches or deploys.

chanseokoh:
> Suspend/resume is not for blocking launches or deploys. It's about blocking install/update during launches or deploys.

Oh, but you're right that currently the method to suspend is a blocking method.

briandealwis:
I misunderstood the purpose — I thought the suspend and resume methods were to suspend installations/updates that were currently in operation.  My apologies! (`suspendSdkUpdates()?)

My concerns here are around ensuring the Job status(es) shown in the _Progress_ view make sense.  Blocking on a Java monitor won't reflect in the progress view: the job will be marked as running.

`DeployJob` is a `WorkspaceJob`, meaning it has the workspace lock for its execution.  Mixing Java-level locks and Eclipse constructs can result in problems as we discovered with the _DependencyGraph_ issues.  I think we should use an Eclipse `ILock` so that the Jobs mechanism can detect possible deadlock.  

We should ensure our jobs are part of a job family so that we can use `IJobManager.join()`, which will be reflected in the _Progress_ view.

Do we need to differentiate between install and update?  The consuming code just wants to check that we have an installation with the necessary SDK components.  `checkInstallation()`?

I wonder of the implications of holding the workspace lock during installation.  It might be better to have a base class for `DeployJob` that can handle installation and relaxing locks.

chanseokoh:
> My concerns here are around ensuring the Job status(es) shown in the Progress view make sense. Blocking on a Java monitor won't reflect in the progress view: the job will be marked as running.

So, I've looked into `IJobManager` and `ILock`. The only synchronization construct the Job system provides is the `ILock` for mutual exclusion. However, I think what we need is the concurrency monitor (not the Eclipse progress monitor) for maximum efficiency (e.g., can run simultaneous local runs while deferring an SDK update). Without the native support of semaphores or condition variables (which are missing from the Job system), I think it is almost impossible to implement a working monitor. So, I think the Java monitor is a necessary evil, and frankly, I can't think of a better idea in the Eclipse job system to handle these requirements.

chanseokoh:
Not tested. I'll write some tests.

chanseokoh:
Oops, something's not right.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManagerTest.java
```diff
@@ -39,4 +44,63 @@ public void testManagedSdkOption_featureForced() {
     CloudSdkManager.forceManagedSdkFeature = true;
     assertTrue(CloudSdkManager.isManagedSdkFeatureEnabled());
   }
+
+  @Test
+  public void testPreventModifyingSdk_cannotWrite() throws InterruptedException {
```
elharo: Are these going to work if the tests run in parallel? Should this class be marked @Sequential?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/messages.properties
```diff
@@ -0,0 +1,3 @@
+InstallJobName=Installing the Cloud SDK...
+InstallStarting=Installing the Cloud SDK... (may take up to several minutes)
```
elharo: delete "up to"

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/messages.properties
```diff
@@ -0,0 +1,3 @@
...
+sdkModificationLocked=The Cloud SDK is being used or modified; cannot install now.
```
elharo: being used or modified --> in use. or how about
Cannot upgrade because the Cloud SDK is in use

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/jobs/MutexRule.java
```diff
@@ -0,0 +1,36 @@
...
+ * serially.
+ */
+public class MutexRule implements ISchedulingRule {
+
+  @Override
+  public boolean contains(ISchedulingRule rule) {
+    return rule == this;
+  }
+
+ @Override
```
elharo: indent

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManagerTest.java
chanseokoh: Yes! A lot of the tests in this PR won't work in parallel. But I cannot find the `@Sequential` annotation. What's the full package qualification?

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManagerTest.java
```diff
@@ -39,4 +52,149 @@ public void testManagedSdkOption_featureForced() {
...
+
+      CloudSdkManager.allowModifyingSdk();
+      prevented = false;
+      concurrentLauncher.join();
+
+      // Incomplete test, but if it ever fails, something is surely broken.
+      assertTrue(installJob.getResult().isOK());
+      assertTrue(concurrentLauncher.getResult().isOK());
+    } finally {
+      if (prevented) {
```
elharo: The logic here is a little unclear to me. Why might prevented be true? I guess if an exception is thrown before we set prevented = false in which case the test fails?

Does it make sense to simply set allowModifyingSdk to true in an `@After` method instead?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -42,4 +53,88 @@ public static boolean isManagedSdkFeatureEnabled() {
...
+            new CloudSdkInstallJob(null /* no console output */, modifyLock);
+        installJob.schedule();
+      }
+    }
+  }
+
+  public static void updateManagedSdkAsync() {
+    if (isManagedSdkFeatureEnabled()) {
+      if (CloudSdkPreferences.isAutoManaging()) {
+        // TODO(chanseok): to be implemented
```
elharo: Issue #?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/BaseCloudSdkInstallJob.java
```diff
@@ -0,0 +1,92 @@
...
+      if (consoleStream != null) {
+        consoleStream.println(Messages.getString("InstallStarting")); //$NON-NLS-1$
+      }
+      return installSdk();
+    } finally {
+      cloudSdkLock.writeLock().unlock();
+    }
+  }
+
+  protected abstract IStatus installSdk();
```
elharo: Why is this class abstract? Do we need multiple subclasses, and if so for what?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class CloudSdkInstallJob extends BaseCloudSdkInstallJob {
+
+  public CloudSdkInstallJob(MessageConsoleStream consoleStream, ReadWriteLock cloudSdkLock) {
+    super(consoleStream, cloudSdkLock);
+  }
+
+  @Override
+  protected IStatus installSdk() {
+    // TODO(chanseok): to be implemented.
```
elharo: Issue #

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManagerTest.java
chanseokoh: What you guessed is right. However, `allowModifyingSdk()` must match 1-to-1 with the number of `prevent...()`. That is why the Boolean var.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/BaseCloudSdkInstallJob.java
chanseokoh: This is purely for testing, honestly. I wanted to actually run the job in tests, without running the actual install code. If you think this is not worth it, I can revert back the abstract class and remove relevant tests. Or, if there is a better idea...

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManagerTest.java
```diff
@@ -39,4 +52,149 @@ public void testManagedSdkOption_featureForced() {
...
+    boolean prevented = true;
+
+    try {
+      final BaseCloudSdkInstallJob installJob = new FakeInstallJob(Status.OK_STATUS);
+
+      Job concurrentLauncher = new Job("concurrent thread attempting runInstallJob()") {
+        @Override
+        public IStatus run(IProgressMonitor monitor) {
+          try {
+            CloudSdkManager.runInstallJob(null, installJob);
```
briandealwis: Add a comment: `// should block until we allow SDK modifications below`

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -42,4 +53,88 @@ public static boolean isManagedSdkFeatureEnabled() {
...
+      throws CoreException, InterruptedException {
+    if (isManagedSdkFeatureEnabled()) {
+      if (CloudSdkPreferences.isAutoManaging()) {
+        runInstallJob(consoleStream, new CloudSdkInstallJob(consoleStream, modifyLock));
+      }
+    }
+  }
+
+  @VisibleForTesting
+  static void runInstallJob(MessageConsoleStream consoleStream, BaseCloudSdkInstallJob installJob)
```
briandealwis: Why are we passing consoleStream?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/BaseCloudSdkInstallJob.java
chanseokoh: Actually, we will have an install job and an update job, so the subclassing can make sense.

plugins/com.google.cloud.tools.eclipse.sdk.test/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJobTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import org.eclipse.core.runtime.jobs.Job;
+import org.junit.Test;
+
+public class CloudSdkInstallJobTest {
+
+  @Test
+  public void testRun() throws InterruptedException {
+    Job job = new CloudSdkInstallJob(null, new ReentrantReadWriteLock());
```
briandealwis: I wonder if we should just pass a `Lock` into the job, so that it can be mocked and checked that it acquired and released the lock.

chanseokoh:
This PR is blocked. Here for reference. I'm just cleaning up things to make it sane.

chanseokoh:
Things have been integrated in other PRs. Closing.

PR #2769, opened 2018-01-31
GCP status widget

briandealwis:
This PR adds a new GCP status widget to the trim and GCP menu.

screen shot 2018-01-31 at 4 28 14 pm

screen shot 2018-01-31 at 4 28 24 pm

When there's a problem (synthetic data, fortunately):

screen shot 2018-01-31 at 4 39 47 pm

screen shot 2018-01-31 at 4 38 18 pm

If an error occurs during fetching, the icon shows an error:

screen shot 2018-01-31 at 4 35 37 pm
error

It's implemented as a command handler that implements `IElementUpdater` to update text, tooltip, and icon with status changes.  We could conceivably hook this into the mylyn notifications too.

It works by processing a partial download of the [JSON incident log at `status.cloud.google.com`](https://status.cloud.google.com/incidents.json).  As this log is quite big (268k, dating back to Feb 2017) but is sorted in descending order, I just fetch the first 8k of data, which is enough to process about a week back.  The service currently refreshes every 3 minutes, meaning we'll download about 160k of data in an hour, which should be ok.

The user can force a refresh by clicking on the trim or menu item.  This also opens the status page in a browser.

(I used the `java.net.URL` connection rather than the Google HTTP client as I didn't much want to refactor our `.googleapis` bundle to expose the classes.  That can be done as follow-on.)

I did do some other minor tweaking of the toolbar and menu definitions, to pull out the status bar out of .login into .ui, and removing `.appengine` from an ID where it no longer belonged.


chanseokoh:
Oh, wow. This is a surprise!

briandealwis:
I woke early the other morning and thought it would be a fun thing to quickly prototype.

Firebase offers a similar [status page](https://status.firebase.google.com) with the same incidents log.  Unfortunately their service names don't include Firebase, and they use overlapping incident numbers.

plugins/com.google.cloud.tools.eclipse.ui.status/src/com/google/cloud/tools/eclipse/ui/status/PollingStatusServiceImpl.java
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  @Override
+  public GcpStatus getCurrentStatus() {
+    return currentStatus;
+  }
+
+  void refreshStatus() {
+    try {
+      // As of 2018-01-30 the incidents log is 258k! But the incidents appear to be sorted from most
```
elharo: Is this documented anywhere? Can we count on this remaining available at this URL?

plugins/com.google.cloud.tools.eclipse.ui.status/src/com/google/cloud/tools/eclipse/ui/status/ShowGcpStatusHandler.java
```diff
@@ -0,0 +1,84 @@
+package com.google.cloud.tools.eclipse.ui.status;
```
elharo: copyright

plugins/com.google.cloud.tools.eclipse.ui.status/src/com/google/cloud/tools/eclipse/ui/status/WorkbenchStartup.java
```diff
@@ -0,0 +1,29 @@
+package com.google.cloud.tools.eclipse.ui.status;
```
elharo: copyright

briandealwis:
> If I follow this, it's a little weird for a menu item itself to provide the status. Would it make sense for the detailed status to appear in a dialog when the user selects the menu item?

The menu item reflects some state.  The startup class registers a listener with the status-monitoring service (which I should rename) so that status updates trigger updates of tool and menu items that reflect that state.

We currently do the same thing with the Google Login button.

(This is the standard Eclipse mechanism. If you think of a _Bold_ tool item, it reflects the bold state. When some action happens to change into a bold state, it then causes a refreshElements() to update all representations of the bold-action in the UI.)

plugins/com.google.cloud.tools.eclipse.ui.status/src/com/google/cloud/tools/eclipse/ui/status/PollingStatusServiceImpl.java
briandealwis: No, it doesn't seem to be documented anywhere that I can see.  It seems to be regenerated about once a minute, judging from the `Last-Modified` headers.

If I switch to the Google HTTP client code, then we should be able to fetch the entire file with gzip encoding which reduces it to 36k. That would fetch approx 720kb/hour at a 3-minute refresh interval.

Issue #2753, opened 2018-01-29
Core Cloud SDK install and update infrastructure


chanseokoh:
Implementation goes into `CloudSdkManager`. I'm thinking the following.

1. A synchronous install method to be used for deploy and local run, which takes a `MessageConsole` where SDK install output is directed.
1. An async install method as a `Job`, to be used at other potential places of triggering installation. As an Eclipse `Job`, no detailed SDK install output will be reported to users. Rather, the progress will be reported via `IProgressMonitor`.
1. An async update method as a `Job`, which works similar to 2.

1 is mandatory, while 2 is somewhat optional. 3 is mandatory if 1 does not always update.

chanseokoh:
For the initial iteration, I think there should be some locking mechanism so that

- there will be no concurrent install/update
- local run and deploy blocks install/update and vice versa

briandealwis:
Working on adding _update_.

briandealwis:
Closing as Step 3 completed with #2813.

PR #2772, opened 2018-02-01
Simple install interface for managed Cloud SDK

chanseokoh:
Part of #2753.

As we discussed, we will care only about installing the SDK. There won't be an update function.

(Branched out of #2764.)


plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -42,4 +48,42 @@ public static boolean isManagedSdkFeatureEnabled() {
     }
     return false;
   }
+
+  /**
+   * Installs the managed Cloud SDK, if the preferences are configured to auto-managed the SDK.
```
elharo: Installs a Cloud SDK if the preferences are configured to auto-manage the SDK.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  @Override
+  public boolean belongsTo(Object family) {
+    return super.belongsTo(family) || family == CLOUD_SDK_MODIFY_JOB_FAMILY;
+  }
+
+  @Override
+  protected IStatus run(IProgressMonitor monitor) {
+    if (consoleStream != null) {
```
elharo: Not sure this message real helps. It feels to me like run and modifySdk are close to the same thing. Maybe just make `run` abstract here and then you don't need a separate modifySdk method.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -0,0 +1,57 @@
...
+      consoleStream.println(Messages.getString("startModifying")); //$NON-NLS-1$
+    }
+    return modifySdk();
+  }
+
+  protected abstract IStatus modifySdk();
```
elharo: If you do keep `modifySdk` then it probably needs to receive the progress monitor, though if yore unsure we can add this later.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/messages.properties
```diff
@@ -0,0 +1,3 @@
+installJobName=Installing the Cloud SDK...
+startModifying=Installing/upgrading the Cloud SDK... (may take several minutes)
```
elharo: I don't think you need this message, at least not yet

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
chanseokoh: Oh, I see now the two methods are same after simplification. I'll ditch using an abstract class.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/messages.properties
chanseokoh: Removed.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -42,4 +48,41 @@ public static boolean isManagedSdkFeatureEnabled() {
...
+    }
+  }
+
+  @VisibleForTesting
+  static void runInstallJob(MessageConsoleStream consoleStream, CloudSdkInstallJob installJob)
+      throws CoreException, InterruptedException {
+    installJob.schedule();
+    installJob.join();
+
+    IStatus status = installJob.getResult();
```
briandealwis: IIRC if the job isn't a system job then a non-OK result will result in an error dialog…

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/jobs/MutexRule.java
```diff
@@ -0,0 +1,36 @@
...
+
+package com.google.cloud.tools.eclipse.util.jobs;
+
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+
+/**
+ * Scheduling rule to prevent two jobs from running concurrently. Set this to jobs that need to run
+ * serially.
+ */
+public class MutexRule implements ISchedulingRule {
```
briandealwis: One suggestion might be to provide a descriptive string, output in its `toString()`, to see what this mutex rule is about.  It might help us one day.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -42,4 +48,41 @@ public static boolean isManagedSdkFeatureEnabled() {
...
   }
+
+  /**
+   * Installs a Cloud SDK, if the preferences are configured to auto-manage the SDK. Blocks callers
+   * 1) if the managed SDK is being installed concurrently by others; and 2) until the installation
+   * is complete.
+   *
+   * @param consoleStream stream to which the install output is written
+   */
+  public static void installManagedSdk(MessageConsoleStream consoleStream)
```
briandealwis: Do you think just using `OutputStream` might be better?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -42,4 +48,41 @@ public static boolean isManagedSdkFeatureEnabled() {
...
+      throws CoreException, InterruptedException {
+    if (isManagedSdkFeatureEnabled()) {
+      if (CloudSdkPreferences.isAutoManaging()) {
+        runInstallJob(consoleStream, new CloudSdkInstallJob(consoleStream));
```
briandealwis: Does `runInstallJob()` actually buy us anything, or is it just for testing?

briandealwis: I wondered if we could add an optimization to first check if the Managed Cloud SDK is installed before creating and running the job. But I suppose there's a change that an install-in-progress will appear to be present, resulting in using an incomplete install.  That's worth a comment here.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: We have `MessageConsoleWriterListener` which accepts the `MessageConsoleStream`. The other day, I thought the same thing and tried to use `OutputStream` but realized it is a bit cumbersome and needs more code as the native `OutputStream` doesn't have `println`.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: Just for testing, unfortunately.

chanseokoh: Good idea.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/jobs/MutexRule.java
chanseokoh: There seems nothing much to say. There is no field, no internal state. The built-in `toString()` will print the full qualified class name, and that's basically all the info?

chanseokoh: OK, maybe I could say

`return MutexRule.class + ": mutex rule for running jobs mutually exclusively`

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
chanseokoh: Perhaps we run it as a system job for local run and deploy, because they have the console and can cancel/stop in their ways. For the periodic install (not decided when to do), maybe a non-system job.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/jobs/MutexRule.java
briandealwis: I was thinking along the lines of `new MutexRule("ManagedCloudSDK")` and print that string in the toString().

chanseokoh: Good idea.

PR #2777, opened 2018-02-06
Perform Managed Cloud SDK installation as part of pre-launch

### LINKED ISSUE #2755: Install Cloud SDK at App Engine local run
As soon as the user starts the dev appserver, the installation process will start if the SDK is not found (and the preference is set to auto-managed). The detailed Cloud SDK install output will be redirected to the local run console.

I'm thinking of not checking or forcing SDK update here.

For the initial iteration, I think there should be some locking mechanism so that

- there will be no concurrent install/update; and
- local run and deploy blocks install/update and vice versa.

briandealwis:
Fixes #2755


chanseokoh:
BTW, I am taking care of the deploy side.

briandealwis:
PTAL @elharo: tries to install the managed Cloud SDK before launch.

Issue #2732, opened 2018-01-17
build exceptions: Version 8.0 of jst.ear and jst.appclient do not exist


briandealwis:
A recent change?  Seen on the Oxygen builds in a number of tests, although they seem harmless. Maybe we need to include an additional test-time dependency?
```
com.google.cloud.tools.eclipse.appengine.validation.ToServlet25QuickFixTest
testGetLabel(com.google.cloud.tools.eclipse.appengine.validation.ToServlet25QuickFixTest)  Time elapsed: 0 sec
testConvertServlet_sunNamespace(com.google.cloud.tools.eclipse.appengine.validation.ToServlet25QuickFixTest)  Time elapsed: 1.232 sec
testConvertServlet_jcpNamespace(com.google.cloud.tools.eclipse.appengine.validation.ToServlet25QuickFixTest)  Time elapsed: 0.121 sec
Running com.google.cloud.tools.eclipse.appengine.validation.ToServlet25SourceQuickFixTest
!ENTRY org.eclipse.jst.j2ee 4 0 2018-01-17 10:36:40.563
!MESSAGE Plugin org.eclipse.jst.jee, extension org.eclipse.jst.j2ee.J2EEModelProvider
Version 8.0 of project facet jst.ear does not exist.
!ENTRY org.eclipse.jst.j2ee 4 0 2018-01-17 10:36:40.564
!MESSAGE Version 8.0 of project facet jst.ear does not exist.
!STACK 0
java.lang.IllegalArgumentException: Version 8.0 of project facet jst.ear does not exist.
	at org.eclipse.wst.common.project.facet.core.util.internal.Versionable.getVersion(Versionable.java:80)
	at org.eclipse.wst.common.project.facet.core.internal.ProjectFacet.getVersion(ProjectFacet.java:1)
	at org.eclipse.jst.j2ee.model.J2EEModelProviderRegistry.readElement(J2EEModelProviderRegistry.java:85)
	at org.eclipse.wst.common.core.util.RegistryReader.internalReadElement(RegistryReader.java:81)
	at org.eclipse.wst.common.core.util.RegistryReader.readRegistry(RegistryReader.java:122)
	at org.eclipse.jst.j2ee.model.J2EEModelProviderRegistry.getInstance(J2EEModelProviderRegistry.java:43)
	at org.eclipse.jst.j2ee.model.ModelProviderManager.initProviders(ModelProviderManager.java:262)
[...]

PR #2778, opened 2018-02-06
Install Cloud SDK at deploy time

### LINKED ISSUE #2754: Install Cloud SDK at App Engine deploy time
As soon as the user clicks on the `Deploy' button, the installation process will start if the SDK is not found (and the preference is set to auto-managed). The detailed Cloud SDK install output will be redirected to the deploy console.

I'm thinking of not checking or forcing SDK update here.

For the initial iteration, I think there should be some locking mechanism so that

- there will be no concurrent install/update; and
- local run and deploy blocks install/update and vice versa.

chanseokoh:
- Part of #2753, implementing the install process.
- Fixes #2754.

@briandealwis I've changed the interface; the install method now returns `IStatus` and does not throw `InterruptedException`.

(Looks like the Eclipse repo is having an issue today.)


chanseokoh:
It's always the Oxygen build that fails. It can't find a feature. I hope this is not a permanent issue.

```
[INFO] Adding repository http://download.eclipse.org/webtools/repository/oxygen
[ERROR] Failed to resolve target definition /home/travis/build/GoogleCloudPlatform/google-cloud-eclipse/eclipse/oxygen/gcp-eclipse-oxygen.target: Could not find "org.eclipse.jst.web_sdk.feature.feature.group/3.8.0.v201710032159" in the repositories of the current location -> [Help 1]
```

briandealwis:
I discovered in #2732 that the WTP Oxygen repo (`/webtools/repository/oxygen`) wasn't updated to include Oxygen.2.  I emailed the wtp-dev list and I suspect [the attempted workaround](https://dev.eclipse.org/mhonarc/lists/wtp-dev/msg10593.html), which moved Oxygen .2 in place of .1a.  Regenerating the Oxygen .target should be sufficient.

It may be that we should target specific versions in our .tpd/.target rather than these catch-all repositories like 

   http://download.eclipse.org/webtools/downloads/drops/R3.9.2/R-3.9.2-20171201000141/repository

chanseokoh:
OK, let me try the versioned WTP repo for Oxygen.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
```diff
@@ -48,8 +63,51 @@ public boolean belongsTo(Object family) {
...
+      return Status.OK_STATUS;
+
+    } catch (InterruptedException | ClosedByInterruptException e) {
+      return Status.CANCEL_STATUS;
+    } catch (IOException | ManagedSdkVerificationException | SdkInstallerException |
+        CommandExecutionException | CommandExitException e) {
+      return StatusUtil.error(this, "Failed to install the Google Cloud SDK.", e);
+
+    } catch (UnsupportedOsException e) {
+      throw new RuntimeException("Cloud Tools for Eclipse supports Windows, Linux, and Mac only.");
```
elharo: Would it be better to return a StatusUtil.error in these cases rather than throwing a runtime exception?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -55,28 +57,36 @@ public static boolean isManagedSdkFeatureEnabled() {
...
+      IProgressMonitor cancelMonitor) {
     installJob.setSystem(true);
     installJob.schedule();
-    installJob.join();
 
-    IStatus status = installJob.getResult();
-    if (!status.isOK()) {
-      throw new CoreException(status);
+    try {
+      while (!installJob.join(100, cancelMonitor)) {
```
briandealwis: I think we can use `0` to indicate indefinite wait.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkInstallJob.java
briandealwis: Seems reasonable to me.  The message isn't quite right — the managed cloud installation only supports Windows, Linux, and Mac.

briandealwis:
I'm good with it, @googlebot

briandealwis:
With tweaks, an `UnsupportedOsException` produces a dialog like the following as the internal installation job returns an IStatus.ERROR and the DeployJob too.  I'll see if I can tweak that slightly.

screen shot 2018-02-07 at 4 58 42 pm

briandealwis:
I've been fiddling with different approaches to reporting Cloud SDK installation errors.  What I don't like about the situation above is that there was no indication to the user that the Cloud SDK was about to be installed, so the errors seem disconnected.

The root cause is that our `DeployJob` launches and waits for the `CloudSdkInstallJob`, and uses the `CloudSdkInstallJob`'s _result_ to indicate the success/failure of the installation.  But Eclipse UI's `ProgressManager` has put up dialogs on jobs resulting in an `IStatus.ERROR` for a decade or more, regardless of whether they are system or user jobs, and provides no facilities to ignore errors from particular jobs.

There are a few options.

## Output status to the deploy console

I think this is the simplest option: write out some explanatory text where the user will see it.  We already open a console to host deployment-related output from `AppCfg` and `gcloud app deploy`.  We can just write that we're about to install the Google Cloud SDK.  If an error occurs, like the above, then at least there is some visible connection.

This is what the user might see; I've marked the helper text with `[Cloud SDK]`.  These lines are visible for at least 10-20 seconds while the SDK is downloaded.

```
[Cloud SDK] App Engine deployment requires Google Cloud SDK, but it cannot be found.
[Cloud SDK] Attempting installation of Google Cloud SDK...
Downloading https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz
Downloading 9214184 bytes
.........done.
Extracting archive: /Users/bsd/.google-cloud-tools-java/managed-cloud-sdk/downloads/google-cloud-sdk.tar.gz
[…elided…]
Welcome to the Google Cloud SDK!


Your current Cloud SDK version is: 188.0.1
Installing components from version: 188.0.1

+----------------------------------------------------------------------------+
|                    These components will be installed.                     |
+-----------------------------------------------------+------------+---------+
|                         Name                        |  Version   |   Size  |
+-----------------------------------------------------+------------+---------+
| BigQuery Command Line Tool                          |     2.0.28 | < 1 MiB |
| BigQuery Command Line Tool (Platform Specific)      |     2.0.26 | < 1 MiB |
| Cloud SDK Core Libraries (Platform Specific)        | 2017.09.15 | < 1 MiB |
| Cloud Storage Command Line Tool                     |       4.28 | 3.3 MiB |
| Cloud Storage Command Line Tool (Platform Specific) |       4.27 | < 1 MiB |
| Default set of gcloud commands                      |            |         |
| gcloud cli dependencies                             | 2017.10.20 | 1.4 MiB |
+-----------------------------------------------------+------------+---------+

For the latest full release notes, please visit:
  https://cloud.google.com/sdk/release_notes

Performing in place update...

#============================================================#
#= Installing: BigQuery Command Line Tool                   =#
#============================================================#
#= Installing: BigQuery Command Line Tool (Platform Spec... =#
#============================================================#
#= Installing: Cloud SDK Core Libraries (Platform Specific) =#
#============================================================#
#= Installing: Cloud Storage Command Line Tool              =#
#============================================================#
#= Installing: Cloud Storage Command Line Tool (Platform... =#
#============================================================#
#= Installing: Default set of gcloud commands               =#
#============================================================#
#= Installing: gcloud cli dependencies                      =#
#============================================================#

Performing post processing steps...
........................................................done.

Update done!

WARNING:   There are older versions of Google Cloud Platform tools on your system PATH.
  Please remove the following to avoid accidentally invoking these old tools:

  /Users/bsd/Manumitting/Projects/GooglePluginEclipse/SDKs/google-cloud-sdk/bin/dev_appserver.py

  

This will install all the core command line tools necessary for working with
the Google Cloud Platform.

==> Source [/Users/bsd/.google-cloud-tools-java/managed-cloud-sdk/LATEST/google-cloud-sdk/completion.bash.inc] in your profile to enable shell command completion for gcloud.
==> Source [/Users/bsd/.google-cloud-tools-java/managed-cloud-sdk/LATEST/google-cloud-sdk/path.bash.inc] in your profile to add the Google Cloud SDK command line tools to your $PATH.

For more information on how to get started, please visit:
  https://cloud.google.com/sdk/docs/quickstarts


[Cloud SDK] Installing App Engine components for Java


Your current Cloud SDK version is: 188.0.1
Installing components from version: 188.0.1

+----------------------------------------------------+
|        These components will be installed.         |
+------------------------------+---------+-----------+
|             Name             | Version |    Size   |
+------------------------------+---------+-----------+
| gcloud app Java Extensions   |  1.9.61 | 118.9 MiB |
| gcloud app Python Extensions |  1.9.66 |   6.1 MiB |
+------------------------------+---------+-----------+

For the latest full release notes, please visit:
  https://cloud.google.com/sdk/release_notes

#============================================================#
#= Creating update staging area                             =#
#============================================================#
#= Installing: gcloud app Java Extensions                   =#
#============================================================#
#= Installing: gcloud app Python Extensions                 =#
#============================================================#
#= Creating backup and activating new installation          =#
#============================================================#

Performing post processing steps...
.....................................done.

Update done!

WARNING:   There are older versions of Google Cloud Platform tools on your system PATH.
  Please remove the following to avoid accidentally invoking these old tools:

  /Users/bsd/Manumitting/Projects/GooglePluginEclipse/SDKs/google-cloud-sdk/bin/dev_appserver.py

```

## Mark `DeployJob` as a user job

If we mark the `DeployJob` as a user job — which it is — then we will get an operation-in-progress dialog appear that will track the progress of the job.  It provides some indication that installation is progressing.  I'm finding the results are mixed as the dialog doesn't consistently report checking for the Cloud SDK and the progress dialog doesn't advance much during the deployment as we don't have a progress connection to the underlying processes.

screen shot 2018-02-08 at 10 48 14 pm


## `DeployJob` returns `IStatus.CANCEL`

We could have our `DeployJob` treat a `CloudSdkInstallJob` installation error as essentially cancelling the deploy.  Returning `IStatus.CANCEL` would mean only the `CloudSdkInstallJob`'s `IStatus.ERROR` will be reported to the user.  

screen shot 2018-02-08 at 3 15 50 pm

But the installation error will still seem disconnected as the user may not have been aware that the Cloud SDK was to be installed.

## Change `CloudSdkInstallJob`'s result reporting

We could change our `CloudSdkInstallJob` to not return an `IStatus.ERROR` and instead report the installation error as an `IStatus.WARNING`, so that only `DeployJob` reports an error.  Although it does seem odd to treat a `WARNING` as an `ERROR`, it has the right effect: 

screen shot 2018-02-08 at 4 17 17 pm

and the error status is logged too:
```
!ENTRY com.google.cloud.tools.eclipse.appengine.deploy 4 0 2018-02-08 16:15:02.271
!MESSAGE Deploy failed: cannot install Google Cloud SDK
!STACK 1
org.eclipse.core.runtime.CoreException: Google Cloud SDK installation only supported on Windows, Linux, and MacOS.
	at com.google.cloud.tools.eclipse.appengine.deploy.DeployJob.runInWorkspace(DeployJob.java:95)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:39)
!SUBENTRY 1 com.google.cloud.tools.eclipse.sdk 1 0 2018-02-08 16:15:02.274
!MESSAGE Google Cloud SDK installation only supported on Windows, Linux, and MacOS.
```

An alternative to using `IStatus.WARNING` is to recast the `CloudSdkInstallJob` as a `FuturisticJob` which reports a result using a future.

## Only asynchronous installation is a treated as a `Job`

We could have `CloudSdkManager#installManagedSdk()` run the installation directly, rather than execute it as a `Job`.  This code would still wait for any other installation jobs, and would acquire any required locks, etc.

elharo:
I'm not sure, but "Change CloudSdkInstallJob's result reporting" seems like the best option right now.

briandealwis:
I should have noted that in the deploy console output, the only addition I've proposed is the helper text prefixed with "[Cloud SDK]" — everything else will be shown currently.

So I think the helper text is useful.  I'll also use the alternative status reporting.

briandealwis:
Realized that there are situations where the CloudSdkInstallJob is intended to be standalone, such that installation errors should be surfaced to the user.

patflynn:
@ivanporty FYI things to think about.

ivanporty:
@patflynn thanks, I'm periodically tracking this, we're moving in approximately the same direction with IJ.

PR #2780, opened 2018-02-06
Use versioned WTP oxygen repo

chanseokoh:
Currently Oxygen builds are failing.

Let's see if this will work.


chanseokoh:
It works. Let's get this in once the remaining tests succeed to unblock other PRs.

Issue #2716, opened 2018-01-11
Many ArtifactRetriever exception traces in tests


briandealwis:
The test logs contain 28 occurrences of  stack traces like the following.  These tests don't actually need to hit the network.

```
Jan 10, 2018 9:52:43 PM com.google.cloud.tools.eclipse.util.ArtifactRetriever getLatestReleaseVersion
WARNING: Could not retrieve version for artifact groupId:artifactId
java.io.FileNotFoundException: https://repo1.maven.org/maven2/groupId/artifactId/maven-metadata.xml
        at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1692)
        at sun.net.www.protocol.https.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:263)
        at com.sun.org.apache.xerces.internal.impl.XMLEntityManager.setupCurrentEntity(XMLEntityManager.java:646)
        at com.sun.org.apache.xerces.internal.impl.XMLVersionDetector.determineDocVersion(XMLVersionDetector.java:148)
        at com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:812)
        at com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:777)
        at com.sun.org.apache.xerces.internal.parsers.XMLParser.parse(XMLParser.java:141)
        at com.sun.org.apache.xerces.internal.parsers.DOMParser.parse(DOMParser.java:243)
        at com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderImpl.parse(DocumentBuilderImpl.java:338)
        at javax.xml.parsers.DocumentBuilder.parse(DocumentBuilder.java:177)
        at com.google.cloud.tools.eclipse.util.ArtifactRetriever.getMetadataDocument(ArtifactRetriever.java:216)
        at com.google.cloud.tools.eclipse.util.ArtifactRetriever.access$000(ArtifactRetriever.java:59)
        at com.google.cloud.tools.eclipse.util.ArtifactRetriever$1.load(ArtifactRetriever.java:86)
        at com.google.cloud.tools.eclipse.util.ArtifactRetriever$1.load(ArtifactRetriever.java:82)
        at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3628)
[...]

PR #2782, opened 2018-02-08
turn off some expected logging messages in tests

elharo:
silence some warnings for #2716 
I'm debating how generic we should make this.


plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -70,6 +76,15 @@ public void setUp() throws SAXException, IOException, CoreException {
...
     pom = Pom.parse(pomFile);
+    
+    Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+    logger.setLevel(Level.OFF);
+  }
+  
+  @After
+  public void tearDown() throws SAXException, IOException, CoreException {
+    Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+    logger.setLevel(Level.WARNING);
```
briandealwis: Should use `logger.setLevel(null)` so that it inherits from the parent (which ws probably the previous level)?

PR #2794, opened 2018-02-13
Revert

elharo:
@briandealwis Is it time to revisit the need to check in generated code? What if we just .gitignored these files?


briandealwis:
Oops geesh, I forgot this: changes to .prefs need to be put into `eclipse/settings/files/` and then the settings need to be re-applied.

PR #2797, opened 2018-02-13
ignore missing javadoc comment on public classes and methods

briandealwis:
Apply ignore-missing-javadoc setting to all projects.


briandealwis:
(Apply fix from #2794 across all projects.)

briandealwis:
The `eclipse-settings-maven-plugin` does support applying changes within Eclipse.  But it requires publishing the `eclipse/settings` jar to an accessible Maven repository.  This doesn't have to be Maven Central, and could be a repository that we control (e.g., a GCS bucket, like: `https://storage.googleapis.com/cloud-tools-for-eclipse/settings/`).

But I encountered some odd workspace warnings, when I tried this, though I can't be sure this was due to these settings.

PR #2798, opened 2018-02-13
More warnings fixes

elharo:
@briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -51,18 +51,18 @@ protected IPath getModuleDeployDirectory(IModule module) {
...
-    return server.getPublishedResourceDelta(module);
+  protected IModuleResourceDelta[] getPublishedResourceDelta(IModule[] modules) {
+    return server.getPublishedResourceDelta(modules);
   }
 
   @Override
-  protected IModuleResource[] getResources(IModule[] module) {
-    return server.getResources(module);
+  protected IModuleResource[] getResources(IModule[] modules) {
+    return server.getResources(modules);
```
briandealwis: ditto

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -51,18 +51,18 @@ protected IPath getModuleDeployDirectory(IModule module) {
...
-    server.setModulePublishState2(module, publishState);
+  protected void setModulePublishState(IModule[] modules, int publishState) {
+    server.setModulePublishState2(modules, publishState);
   }
 
   @Override
-  protected IModuleResourceDelta[] getPublishedResourceDelta(IModule[] module) {
-    return server.getPublishedResourceDelta(module);
+  protected IModuleResourceDelta[] getPublishedResourceDelta(IModule[] modules) {
+    return server.getPublishedResourceDelta(modules);
```
briandealwis: ditto

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperation.java
```diff
@@ -51,18 +51,18 @@ protected IPath getModuleDeployDirectory(IModule module) {
   }
 
   @Override
-  protected void setModulePublishState(IModule[] module, int publishState) {
-    server.setModulePublishState2(module, publishState);
+  protected void setModulePublishState(IModule[] modules, int publishState) {
+    server.setModulePublishState2(modules, publishState);
```
briandealwis: In WTP, the array represents a module path to the module.  For example, module[0] = a web module, module[1] = utility module; the last module is the actual module.  So these should remain the singular `module`.

elharo:
Not sure I follow re module or modules. If it should be singular we still
have the warning that a field is shadowed, so we'll need to fix that some
other way.

--
Elliotte Rusty Harold

briandealwis:
I think `submodule` would work.

elharo:
If the array represents the modulePath, then how about `modulePath` as the name?

briandealwis:
Sounds good to me!

Issue #1798, opened 2017-04-14
Sporadic test failure in LocalAppEnginePublishOperationTest


briandealwis:
[Test failure](https://travis-ci.org/GoogleCloudPlatform/google-cloud-eclipse/jobs/222070313#L3513) on Java 7 with Mars.  Failing code attempts to obtain the WTP `IModule` for the shared project:
```
    sharedModule = ServerUtil.getModule(sharedProject);
    assertNotNull(sharedModule);
```

Nothing interesting on the transcript:
```
Running com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest
[Watchdog] > testPublishingSubmodules(com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest)
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 2.134 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest
testPublishingSubmodules(com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest)  Time elapsed: 2.132 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:712)
	at org.junit.Assert.assertNotNull(Assert.java:722)
	at com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest.setUp(LocalAppEnginePublishOperationTest.java:72)
```

chanseokoh:
I can confirm this is not new. I have seen this occasionally (quite infrequently fortunately) for a very long time.

chanseokoh:
Failing in the same test class probably because of this, but at a slightly different location:

```
testPublishingSubmodules(com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEnginePublishOperationTest)  Time elapsed: 3.875 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
[...]

PR #2801, opened 2018-02-14
Trigger installation of Google Cloud SDK on startup

### LINKED ISSUE #2750: Install Managed Cloud SDK on startup
On start up, if Managed SDK is enabled  but not installed, then install

briandealwis:
Fixes #2750

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
@@ -1,0 +1,28 @@
+package com.google.cloud.tools.eclipse.sdk.ui;
+
+import com.google.cloud.tools.eclipse.sdk.CloudSdkManager;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.ui.IStartup;
+import org.eclipse.ui.progress.WorkbenchJob;
+
+/** A simple startup task that triggers installation of the Google Cloud SDK if applicable. */
+public class InstallManagedCloudSdkStartup implements IStartup {
+
+  @Override
+  public void earlyStartup() {
+    // Use a WorkbenchJob in case the user decides to exit quickly
+    Job triggerInstallationJob =
+        new WorkbenchJob("Check Google Cloud SDK") {
+          @Override
+          public IStatus runInUIThread(IProgressMonitor monitor) {
+            CloudSdkManager.installManagedSdkAsync();
+            return Status.OK_STATUS;
+          }
+        };
+    triggerInstallationJob.setSystem(true);
+    triggerInstallationJob.schedule(200 /*ms*/); // small wait in case user quits
+  }
+}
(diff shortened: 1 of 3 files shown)


plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
```diff
@@ -0,0 +1,28 @@
+package com.google.cloud.tools.eclipse.sdk.ui;
```
elharo: copyright

elharo: Why is this in the .ui subpackage? because it runs in the UI thread?

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
```diff
@@ -0,0 +1,28 @@
...
+/** A simple startup task that triggers installation of the Google Cloud SDK if applicable. */
+public class InstallManagedCloudSdkStartup implements IStartup {
+
+  @Override
+  public void earlyStartup() {
+    // Use a WorkbenchJob in case the user decides to exit quickly
+    Job triggerInstallationJob =
+        new WorkbenchJob("Check Google Cloud SDK") {
+          @Override
+          public IStatus runInUIThread(IProgressMonitor monitor) {
```
elharo: Remind me, why does this need to run in the UI thread? My Eclipse thread-fu is weak. :-)

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
briandealwis: Hmmm… this should never happen any more.  Ahh! Maybe it's because I created the file from the PDE wizard.

briandealwis: It doesn't make sense for this to happen without the Eclipse Workbench (the UI).

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
briandealwis: earlyStartup() is called on a non-UI worker thread and runs before the UI is actually rendered.  Since we're about to do something relatively expensive, we should wait until the workbench is actually up and running, especially so we don't slow down startup.

(I'll add a comment to this effect.)

(Maybe I should increase the `schedule()` to like 10 seconds.)

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/InstallManagedCloudSdkStartup.java
briandealwis: Problem solved: this happens as I had "Generate comments" unchecked in the _New Java Class_ wizard.
screen shot 2018-02-14 at 1 28 24 pm

briandealwis:
Kokoro failure is #1798

PR #2807, opened 2018-02-14
Expose version in preferences

### LINKED ISSUE #2791: Expose Cloud SDK version
Probably in the Managed SDK preferences we need to show what version of the Cloud SDK is in use, since sometimes the latest version has a bug and users need to roll back a couple of versions.

Perhaps we can put the version right below the SDK location in a non-editable (but copyable) field:

image

elharo:
@briandealwis Fix #2791 

This may be only an interim step until we add a version chooser, but let me know what you think:

image


briandealwis:
Looks useful!  The label should be outdented so that it doesn't look like it's specific to the local SDK.

The version doesn't update to the right value if I launch with `GOOGLE_CLOUD_SDK_HOME` set to a valid installation.  The update code needs to run on initialization and on any change.  The code has become rather hard to understand.
screen shot 2018-02-14 at 3 03 50 pm

elharo:
What do you think about moving the version above the Use Local SDK checkbox?
Or perhaps they should all be left-aligned since the SDK location textfield is used for both managed and non-managed SDK?

briandealwis:
Moving it above is a good idea.  The SDK Location is really only used for the non-managed SDK.

elharo:
PTAL @briandealwis

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -86,15 +90,19 @@ public void widgetSelected(SelectionEvent event) {
         WorkbenchUtil.openInBrowser(PlatformUI.getWorkbench(), event.text);
       }
     });
-
+    
+    sdkVersionLabel = new Label(parent, SWT.LEFT);
+    sdkVersionLabel.setText(Messages.getString("SdkVersion", "Unset"));
```
briandealwis: We need to explicitly set the font size here due to what we're doing in line 86 as the label's font is one pixel too small:
screen shot 2018-02-15 at 9 14 01 pm

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -86,15 +90,19 @@ public void widgetSelected(SelectionEvent event) {
...
     if (CloudSdkManager.isManagedSdkFeatureEnabled()) {
       useLocalSdk = new Button(parent, SWT.CHECK);
       useLocalSdk.setText(Messages.getString("UseLocalSdk")); //$NON-NLS-1$
       useLocalSdk.addSelectionListener(new SelectionAdapter() {
         @Override
-        public void widgetSelected(SelectionEvent e) {
+        public void widgetSelected(SelectionEvent event) {
           if (!useLocalSdk.getSelection()) {
             status = Status.OK_STATUS;
+            updateSelectedVersion();
```
briandealwis: Do you think we could extract the code from `doCheckState()` and have `updateSelectedVersion()` work for both cases?  This special casing, along with the special casing in `doCheckState()` make this a bit hard to reason through.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -128,6 +139,31 @@ public void widgetSelected(SelectionEvent e) {
...
+    String version = Messages.getString("UnknownVersion"); //$NON-NLS-1$
+    if (!CloudSdkManager.isManagedSdkFeatureEnabled()
+        || (useLocalSdk != null && useLocalSdk.getSelection())) {
+      Path path = Paths.get(sdkLocation.getStringValue());
+      try {
+        version = new CloudSdk.Builder().sdkPath(path).build().getVersion().toString();
+      } catch (CloudSdkVersionFileException ex) {
+        version = Messages.getString("NoSdkFound"); //$NON-NLS-1$
+      }
+    } else if (CloudSdkManager.isManagedSdkFeatureEnabled()) {
```
chanseokoh: Did you mean `CloudSdkManager.isAutoManaging()`? `isManagedSdkFeatureEnabled()` is our temporary debug flag.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -128,6 +139,31 @@ public void widgetSelected(SelectionEvent e) {
     Dialog.applyDialogFont(contents);
     return contents;
   }
+  
+  private void updateSelectedVersion() {
+    String version = Messages.getString("UnknownVersion"); //$NON-NLS-1$
+    if (!CloudSdkManager.isManagedSdkFeatureEnabled()
+        || (useLocalSdk != null && useLocalSdk.getSelection())) {
```
chanseokoh: `useLocalSdk` is always `null` if `isManagedSdkFeatureEnabled()` is false (and always non-null otherwise). So, I don't like the null check for testing if our debug flag is on; we should directly check `isManagedSdkFeatureEnabled()` instead.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -244,6 +280,11 @@ public void setFilterPath(File path) {
 
     @Override
     protected boolean doCheckState() {
+      if (useLocalSdk != null && !useLocalSdk.getSelection()) {
```
chanseokoh: ditto. We should directly check the `isManagedSdkFeatureEnabled()` flag.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
briandealwis: Can't check `isAutoManaging()` as this dialog is changing that result.  But the logic here is becoming a bit tortured :-/  It would be worth pulling out the test into a local `isAutoManaging()` method that will use `useLocalSdk` to determine if auto-managing.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
elharo: done

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -128,6 +139,31 @@ public void widgetSelected(SelectionEvent e) {
     Dialog.applyDialogFont(contents);
     return contents;
   }
+  
+  private void updateSelectedVersion() {
+    String version = Messages.getString("UnknownVersion"); //$NON-NLS-1$
+    if (!CloudSdkManager.isManagedSdkFeatureEnabled()
```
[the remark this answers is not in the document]
elharo: done

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
elharo: done

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
elharo: Maybe. My initial attempt to do this didn't quite work, but I may be able to improve it a bit.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
elharo: done

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -86,15 +90,20 @@ public void widgetSelected(SelectionEvent event) {
         WorkbenchUtil.openInBrowser(PlatformUI.getWorkbench(), event.text);
       }
     });
-
+    
+    sdkVersionLabel = new Label(parent, SWT.LEFT);
```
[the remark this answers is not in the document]
elharo: done

PR #2813, opened 2018-02-17
Add support for updating managed Cloud SDK

briandealwis:
Merged @chanseokoh's read-write lock support and added support for updating the Managed Cloud SDK.  Added some logging too.

I haven't looked into how/if we can hook this into Eclipse's built-in Eclipse Update mechanism, so for now I've added a command _Update Managed Google Cloud SDK_ that can be triggered via the _Quick Access_.

Separately must also hook in calls to `preventModifySdk()` and `enableModifySdk()` from the debug launch delegate.


briandealwis:
My intention is for the _Quick Access_ as a temporary measure.  +1 to the pref page button.

PR #2816, opened 2018-02-20
Build against Eclipse Photon (4.8) M5

briandealwis:


briandealwis:
+1

elharo:
Haven't seen a lot of flakes lately so if you want to move ahead with this we can.

PR #2817, opened 2018-02-20
Switch projects execution environments to Java 8

briandealwis:
Enable use of Java 8 constructs.

Part of #2557


briandealwis:
Oops, I mistakenly based this off the `photon-m5` branch.

README.md
```diff
@@ -116,7 +116,7 @@ target platform whenever dependencies are updated.
 ### Steps to import into the Eclipse IDE
 
 
-1. Setup JDK 7 in Eclipse
+1. Setup JDK 8 in Eclipse
 
   1. Select `Window/Preferences` (on Mac `Eclipse/Preferences`).
```
chanseokoh: Down below is "Select the folder that contains the JDK 7 installation".

briandealwis:
- The `third_party` code shouldn't be touched
- The `plugins/com.google.cloud.tools.eclipse.ui.test/build.properties` is already fixed

Not sure why the jdt.core.prefs aren't being hammers in the .status bundles!

briandealwis:
Ah! Some of our bundles are parented against the root, not `plugins/`.

briandealwis:
Missed those errors as .status bundles weren't in my working set.

I'm going to remove these `javacSource` and `javacTarget` entries from `build.properties` to drop redundant warnings ("_The 'javacSource' build entry is not required because it matches the default value_").

briandealwis:
PTAL @chanseokoh

PR #2819, opened 2018-02-20
fix warning

elharo:
@chanseokoh


briandealwis:
Is this the only databinding warning?

chanseokoh:
Just one. For example, I see more databinding warnings in `AppEngineDeployPreferencesPanel`.

PR #2823, opened 2018-02-20
Drop our Consumer shim and use of Guava Optional and Function

briandealwis:


plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerDelegateTest.java
```diff
@@ -113,12 +111,7 @@ public void testCheckProjectFacets_deletedModule() throws CoreException {
   @Test
   public void testCheckConflictingId_defaultServiceIds() throws CoreException {
     delegate = getDelegateWithServer();
-    delegate.serviceIdFunction = new Function<IModule, String>() {
-      @Override
-      public String apply(IModule module) {
-        return "default";
-      }
-    };
+    delegate.serviceIdFunction = Functionsmodule -> "default";
```
chanseokoh: `Functionsmoudle` --> `module`?

briandealwis: Oops.  That was from an experiment to see if there was a Guava-like `Functions.constant()`

PR #2826, opened 2018-02-20
Migrate to Java 8 lambdas

chanseokoh:
Some more Java 8 adoptions.

Let me see if tests will pass. I have some problem locally inside Eclipse. Not sure why.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java
@@ -36,12 +36,12 @@ import com.google.cloud.tools.eclipse.ui.util.event.OpenUriSelectionListener.Que
 import com.google.cloud.tools.eclipse.ui.util.images.SharedImages;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Predicate;
 import com.google.common.base.Strings;
 import com.google.common.net.UrlEscapers;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.eclipse.core.databinding.Binding;
@@ -502,19 +502,13 @@ public abstract class AppEngineDeployPreferencesPanel extends DeployPreferencesP
   @VisibleForTesting
   public Job latestGcpProjectQueryJob;  // Must be updated/accessed in the UI context.
 
-  private Predicate<Job> isLatestQueryJob = new Predicate<Job>() {
-    @Override
-    public boolean apply(Job job) {
-      return job == latestGcpProjectQueryJob;
-    }
-  };
-
   private void refreshProjectsForSelectedCredential() {
     projectSelector.setProjects(Collections.<GcpProject>emptyList());
     latestGcpProjectQueryJob = null;
 
     Credential selectedCredential = accountSelector.getSelectedCredential();
     if (selectedCredential != null) {
+      Predicate<Job> isLatestQueryJob = job -> job == latestGcpProjectQueryJob;
       latestGcpProjectQueryJob = new GcpProjectQueryJob(selectedCredential,
           projectRepository, projectSelector, bindingContext, isLatestQueryJob);
       latestGcpProjectQueryJob.schedule();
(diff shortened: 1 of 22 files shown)


chanseokoh:
This never ends, so I think I'll stop here.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/jobs/PluggableJob.java
```diff
@@ -17,9 +17,8 @@
 package com.google.cloud.tools.eclipse.util.jobs;
 
 import com.google.common.base.Preconditions;
-import com.google.common.base.Predicate;
-import com.google.common.base.Predicates;
 import java.util.concurrent.Callable;
+import java.util.function.Predicate;
```
elharo: Weird. Why do we import this if we don't use it?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/AppEngineDeployPreferencesPanel.java
```diff
@@ -502,19 +502,13 @@ private Composite createBucketSection(Composite parent) {
...
-    }
-  };
-
   private void refreshProjectsForSelectedCredential() {
     projectSelector.setProjects(Collections.<GcpProject>emptyList());
     latestGcpProjectQueryJob = null;
 
     Credential selectedCredential = accountSelector.getSelectedCredential();
     if (selectedCredential != null) {
+      Predicate<Job> isLatestQueryJob = job -> job == latestGcpProjectQueryJob;
```
briandealwis: It might be easier to inline this and use `latestGcpProjectQueryJob::equals` — `Job` doesn't override equals anyways.  Or alternatively, `Predicate.isEqual(latestGcpProjectQueryJob)` returns a predicate, useful in case `latestGcpProjectQueryJob` may be `null`.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/ProjectSelectorSelectionChangedListener.java
```diff
@@ -82,6 +76,7 @@ public void selectionChanged(SelectionChangedEvent event) {
       }
     } else {  // The project has never been queried.
       Credential credential = accountSelector.getSelectedCredential();
+      Predicate<Job> isLatestQueryJob = job -> job == latestQueryJob;
```
briandealwis: ditto here with `latestQueryJob::equals`

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ConvertJobSuspenderTest.java
```diff
@@ -41,7 +40,7 @@
   @Before
   public void setUp() {
     // Assume all test jobs are "ConvertJob"s.
-    ConvertJobSuspender.isConvertJob = Predicates.alwaysTrue();
+    ConvertJobSuspender.isConvertJob = job -> true;
```
briandealwis: Aside: I personally find Guava's `alwaysTrue()` and `alwaysFalse()` to be clearer — I don't have to parse the actual lambda and see what it's doing.  That said, until we update to a later Guava, the `Predicate`s and `Function`s aren't interchangeable.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/jobs/PluggableJob.java
briandealwis: `stalenessCheck` on [line 32](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2826/files#diff-5de80434c98534b2471cb48f5a26d8e3R32)

briandealwis:
> I have some problem locally inside Eclipse.

I don't see issues?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ConvertJobSuspenderTest.java
elharo: We should update to a later Guava, but we should probably start further down the stack in appengine-plugins-core and ide-login.

chanseokoh:
> I don't see issues?

Looks like it's unrelated.

PR #2832, opened 2018-02-21
explorer not editor

elharo:
@briandealwis


elharo:
In fact I couldn't do it from the editor's context menu. That's why I sent this.

briandealwis:
That's odd, because it's there too:

screen shot 2018-02-22 at 8 50 08 am

I wonder if you had it open in a non-TPD editor.

PR #2845, opened 2018-02-23
close json reader

elharo:
@briandealwis


briandealwis:
I saw that warning too, but hadn't dealt with it.  The warning is spurious as the underlying reader is closed in the caller.  I think closing this `JsonReader` is wrong since it's is a streaming reader and theoretically there could be still be pending data.  I think `@SuppressWarnings("resource")` is the right fix.

elharo:
So here's my thoughts:

1. The JSONReader may itself contain resources that need to be cleaned up. Closing the underlying reader is not necessarily sufficient.

2. Closing is usually idempotent.

3. We don't read from the reader after we're done with the JSONReader. Given the possibility of buffering, doing this would normally be very error-prone.

PR #2847, opened 2018-02-23
Refactor and clean up code

chanseokoh:
WDYT?


plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -127,4 +137,15 @@ protected void canceling() {
       jobThread.interrupt();
     }
   }
+
+  /**
```
briandealwis: I didn't do this move as it wasn't clear to me whether all modify jobs would want this.  But it doesn't hurt.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
```diff
@@ -41,22 +41,25 @@
...
     }
     return instance;
   }
 
   @VisibleForTesting
-  CloudSdkManager() {}
-
-  // readers = using SDK, writers = modifying SDK
-  @VisibleForTesting final ReadWriteLock modifyLock = new ReentrantReadWriteLock();
+  CloudSdkManager(ReadWriteLock modifyLock) {
```
briandealwis: I wonder how this will work with mocking.  When mocking a normal class, it seems everything is pass-through except for the explicitly mocked methods.  So it's possible that we'll hit NPEs.  Maybe not a bad thing.

chanseokoh: I think you are talking about [partial mocking](https://stackoverflow.com/questions/14970516/use-mockito-to-mock-some-methods-but-not-others), and looks like a mock doesn't call actual methods unless explicitly told to do so or you spy instead of mock.

chanseokoh:
> It seems ugly to contort the API for no reason other than testing.

Yeah, so I'm trying to un-contort. For example, `modifyLock` was exposed (`@VisibleForTesting`), and I'm making it `private`. I am also removing the `forceManagedSdkFeature` field here and there that only exists for testing.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java
briandealwis: Well now I'm baffled — I'm sure I hit a situation with a partial mock where I had non-mocked methods executing. I must be mis-remembering something ¯\_(ツ)_/¯

PR #2850, opened 2018-02-23
Make GCP status polling job a system job and externalize

briandealwis:
Previously the job would appear in the progress view by default.


plugins/com.google.cloud.tools.eclipse.ui.status/src/com/google/cloud/tools/eclipse/ui/status/PollingStatusServiceImpl.java
```diff
@@ -58,22 +58,29 @@
  */
 @Component(name = "polling")
 public class PollingStatusServiceImpl implements GcpStatusMonitoringService {
+  /** Simple recurring job to refresh the status. */
+  private class PollJob extends Job {
+    public PollJob() {
+      super(Messages.getString("poll.job.name")); //$NON-NLS-1$
```
chanseokoh: Looks like both `Messages` and the property are not defined.

briandealwis: Where's the :hang-head-in-shame: emoji?

PR #2851, opened 2018-02-26
Minor fixes

chanseokoh:


plugins/com.google.cloud.tools.eclipse.ui.status/src/com/google/cloud/tools/eclipse/ui/status/Messages.java
```diff
@@ -16,23 +16,24 @@
...
-} catch (MissingResourceException e) {
-return '!' + key + '!';
-}
-}
+  private Messages() {
+  }
+
+  public static String getString(String key, Object... params) {
+    try {
+      return MessageFormat.format(RESOURCE_BUNDLE.getString(key), params);
```
elharo: This change might deserve a test.

Issue #2853, opened 2018-02-26
Update to appengine-plugins-core 0.4.4


briandealwis:
Update to appengine-plugins-core 0.4.4's changed APIs for managed Cloud SDK.

PR #2854, opened 2018-02-26
Update to appengine-plugins-core 0.4.4

briandealwis:
Towards #2853: 0.4.4 must wait it is on Maven Central.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/ProgressWrapper.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/ProgressWrapper.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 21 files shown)


plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/ProgressWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class ProgressWrapper implements ProgressListener {
+  private final IProgressMonitor wrapped;
+
+  private final SubMonitor progress;
+
+  public ProgressWrapper(IProgressMonitor progressMonitor) {
+    // Store wrapped monitor as SubMonitor#convert() says that caller is responsible for calling
+    // done() on argument
+    progress = SubMonitor.convert(progressMonitor);
```
briandealwis: Hmm. I wonder if it's safe to use a SubMonitor directly, since it may trigger an OperationCancelledException when appengine-plugins-core is in operation.  What's the worst case?

chanseokoh:
0.4.4 available: http://repo1.maven.org/maven2/com/google/cloud/tools/appengine-plugins-core/

pom.xml
```diff
@@ -95,7 +95,7 @@
       <!-- bundle: com.google.cloud.tools.appengine -->
       <groupId>com.google.cloud.tools</groupId>
       <artifactId>appengine-plugins-core</artifactId>
-      <version>0.4.2</version>
+      <version>0.4.4-SNAPSHOT</version>  <!-- for compilation -->
```
chanseokoh: Don't forget to drop `SNAPSHOT`!

chanseokoh:
Shouldn't we do similar progress monitor changes to `CloudSdkUpdateJob`?

briandealwis:
That won't be in until 0.4.5.

PR #2855, opened 2018-02-26
Update version to 1.6.0 for release

chanseokoh:
Are we good with appengine-plugins-core 0.4.4?


chanseokoh:
Um... actually, we need to first enable the managed Cloud SDK feature. That is, we need to make `CloudSdkManager.isManagedSdkFeatureEnabled()` return true. @briandealwis should the code modified or is there a way to make `debugOptions.getBooleanOption()` return true without modifying our code? If we have to modify the code, I guess it will require fixing tests too.

chanseokoh:
Maybe flipping the value of `plugins/com.google.cloud.tools.eclipse.sdk/.options` will do, since `build.properties` specifies it will be included it in the bundle binary?

briandealwis:
Given @loosebazooka release appengine-plugins-core 0.5.0, maybe we should wait and move to that?

chanseokoh:
Chatted with @elharo about that. Since 0.5.0 has upgraded Guava, we actually think it's safer to stick with 0.4.4 for the next release.

briandealwis:
Makes sense!

chanseokoh:
Blocked until #https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/551 is fixed and released.

briandealwis:
Do we want to ask to backport it for a 0.4.5?

chanseokoh:
Let me check the backporting with @loosebazooka.

PR #2856, opened 2018-02-26
Enable managed Cloud SDK feature

chanseokoh:


plugins/com.google.cloud.tools.eclipse.sdk/.options
```diff
@@ -1,4 +1,4 @@
 # Set true for development. To be loaded in a hosted debug Eclipse instance,
 # open the "gcloud-eclipse-tools" run configuration, go to the "Tracing" tab,
 # and check "Enable tracing" and "com.google.cloud.tools.eclipse.sdk".
-com.google.cloud.tools.eclipse.sdk/enable.managed.cloud.sdk=false
+com.google.cloud.tools.eclipse.sdk/enable.managed.cloud.sdk=true
```
briandealwis: This isn't enough: this value is only loaded if the `DebugOptions` service is available, and `DebugOptions` is only loaded if run with `-debug`.  We need to change the default in [`CloudSdkManager#isManagedSdkFeatureEnabled()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/CloudSdkManager.java#L71).

chanseokoh: Yeah, I did a try and verified this doesn't work. Closing.

PR #2861, opened 2018-02-27
Remove Eclipse Public License feature

briandealwis:
We pull in the `org.eclipse.license` feature, a vestige from GPE.  But since CT4E is under the ASL, it's unnecessary.

Note: this patch builds on #2860 since it requires regenerating the .target files.


briandealwis:
Change was rebased onto master.

PR #2866, opened 2018-02-27
Prevent NPE when validating XMLs in non-faceted projects

### LINKED ISSUE #2863: NPE when there are non-faceted projects with any XML files
In `XmlSourceValidator.validate()`, we validate (any) XML files only if the project has the App Engine facet.

```java
      IFacetedProject facetedProject = ProjectFacetsManager.create(project);
      if (AppEngineStandardFacet.hasFacet(facetedProject)) {
        ...
        validate(reporter, source, bytes);
      }
```

However,  `hasFacet()` requires that `project` is faceted.
```java
  public static boolean hasFacet(IFacetedProject project) {
    Preconditions.checkNotNull(project);
    return project.hasProjectFacet(FACET);
  }
```

Not sure if we should check null inside `hasFacet()` or before calling it.

chanseokoh:
Fixes #2863.


plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
```diff
@@ -118,7 +118,7 @@ public void setInitializationData(IConfigurationElement config, String propertyN
       String className = (String) data;
       Class<?> clazz = Class.forName(className);
       Constructor<?> constructor = clazz.getConstructor();
-      this.setHelper((XmlValidationHelper) constructor.newInstance(new Object[] {}));
+      setHelper((XmlValidationHelper) constructor.newInstance(new Object[] {}));
```
briandealwis: This code isn't correct: it's resolving the className from the calling bundle's context, rather than the bundle contributing the extension.  It should use `config.createExecutableElement(className)`.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidatorTest.java
```diff
@@ -173,4 +180,29 @@ public void testGetProject() throws CoreException {
     assertEquals(project, testProject);
   }
 
+  @Test
+  public void testNoNullPointerExceptionOnNonFacetedProject()
+      throws CoreException, ValidationException {
+    IProjectDescription projectDescription =
+        ResourcesPlugin.getWorkspace().newProjectDescription("non-faceted-project");
```
briandealwis: Would it would be better to move this into the TestProjectCreator, and change it to only add the `FacetedProjectNature.NATURE_ID` if there are facets to install?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
chanseokoh: File a new bug? I think we need to fix this in a separate PR. There may be more instances like this.

briandealwis: #2874

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidatorTest.java
chanseokoh: Done. Let's see if this cause any test failure in other classes.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/FacetUtil.java
```diff
@@ -68,7 +68,7 @@
   Set<IFacetedProject.Action> facetInstallSet = new HashSet<>();
 
   public FacetUtil(IFacetedProject facetedProject) {
-    Preconditions.checkNotNull(facetedProject);
+    Preconditions.checkNotNull(facetedProject, "project not faceted");
```
briandealwis: This isn't quite true: it was passed `null`.  It may be that the caller tried to get the faceted project which didn't exist.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidatorTest.java
```diff
@@ -173,4 +180,21 @@ public void testGetProject() throws CoreException {
     assertEquals(project, testProject);
   }
 
+  @Test
+  public void testNoNullPointerExceptionOnNonFacetedProject()
```
briandealwis: This method name is a bit too specific to the bug!  Maybe `testIgnoreNonFacetedProject()`?

chanseokoh: I'll name it as `testNoErrorOnNonFacetedProject()`, because the test doesn't actually check if it ignores and does not validate XMLs in a non-faceted project. (It just checks if it will crash.

chanseokoh:
Let's see if tests will pass.

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/NatureUtilsTest.java
```diff
@@ -63,10 +63,9 @@ public void testRemoveNature() throws CoreException {
 
   @Test
   public void testRemoveNature_nonExistingNature() throws CoreException {
-    // By default, project has faceted project nature.
-    assertArrayEquals(new String[] {FACETED_NATURE_ID}, project.getDescription().getNatureIds());
+    assertEquals(0, project.getDescription().getNatureIds().length);
 
     NatureUtils.removeNature(project, JavaCore.NATURE_ID, monitor);
-    assertArrayEquals(new String[] {FACETED_NATURE_ID}, project.getDescription().getNatureIds());
+    assertEquals(0, project.getDescription().getNatureIds().length);
```
briandealwis: This situation shouldn't happen, but it is kinda handy to have the natures shown when it fails?

Issue #2865, opened 2018-02-27
LibrarySelectorGroup appears cramped on Windows


briandealwis:
Our `LibrarySelectorGroup` generates a GridLayout using [`GridLayoutFactory.fillDefaults()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui/src/com/google/cloud/tools/eclipse/appengine/libraries/ui/LibrarySelectorGroup.java#L104) which looks very cramped on Windows:

![screen shot 2018-02-27 at 1 23 27 pm](https://user-images.githubusercontent.com/202851/36746950-99219890-1bc1-11e8-9ae7-c3bb6adba96b.PNG)

briandealwis:
A quick fix to this showed that the Maven project coordinates section seemed off:

<img width="631" alt="screen shot 2018-02-27 at 3 55 24 pm" src="https://user-images.githubusercontent.com/202851/36756877-411a0610-1bde-11e8-875e-f127446e9eff.PNG">

It might be worth using the new Nebula [Checkbox widget](https://wiki.eclipse.org/Nebula_CheckboxGroup).

PR #2871, opened 2018-02-27
Indent Maven Coordinates box to clarify control structure

briandealwis:
As noted in #2865, the Maven Coordinates box is slight indented compared to the Libraries API section on the New App Engine Standard Environment wizard.  This patch indents the Maven Coordinates box by 20 pixels to make it clear that it is controlled by the checkbox:

screen shot 2018-02-27 at 5 13 38 pm

It will affect the App Engine Flexible Environment wizard too.


briandealwis:
@elharo has reservations:
> Let's discuss this further. I'm not 100% convinced the original was incorrect. The Maven project coordinates group should be disabled when the checkbox isn't checked.

The contents are disabled.  This indentation still looks odd.  TBH, I'd prefer moving the controlling checkbox inside.

PR #2877, opened 2018-03-01
Use IConfigurationElement.createExecutableExtension() to create instances

### LINKED ISSUE #2874: Use extension registry for object creation
We have code which attempts to create an object using `java.lang.reflect` from a class-name provided via an extension point.  This code works as the only extensions are defined within that same bundle.  But `Class.forName()` attempts to resolve a class from the calling bundle's classloader, rather than the extension's bundle, and so will fail for extensions from a different bundle.

briandealwis:
Fixes #2874 

I stripped out the helper-creating exception catching & logging code as the implementations don't handle a case where there is a `null` validation helper.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
@@ -20,8 +20,6 @@ import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
 import com.google.cloud.tools.eclipse.util.status.StatusUtil;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Map;
 import java.util.logging.Level;
@@ -53,7 +51,8 @@ import org.w3c.dom.Document;
 import org.xml.sax.SAXException;
 
 /**
- * Contains the logic for source validation and message creation.
+ * Contains the logic for source validation and message creation. The actual validation logic is
+ * delegated to an {@link XmlValidationHelper}.
  */
 public class XmlSourceValidator implements ISourceValidator, IValidator, IExecutableExtension {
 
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
```diff
@@ -111,19 +109,11 @@ void validate(IReporter reporter, IFile source, byte[] bytes) throws IOException
...
-      }
-      String className = (String) data;
-      Class<?> clazz = Class.forName(className);
-      Constructor<?> constructor = clazz.getConstructor();
-      this.setHelper((XmlValidationHelper) constructor.newInstance(new Object[] {}));
-    } catch (ClassNotFoundException | NoSuchMethodException | SecurityException
-        | InstantiationException | IllegalAccessException | IllegalArgumentException
-        | InvocationTargetException ex) {
-      logger.log(Level.SEVERE, ex.getMessage());
+    if (data == null || !(data instanceof String)) {
```
chanseokoh: I'm tempted to remove `data == null`, as `!(data instanceof String)` subsumes it.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidator.java
```diff
@@ -95,19 +93,11 @@ void validate(IFile resource, byte[] bytes) throws CoreException, IOException {
...
-      }
-      String className = (String) data;
-      Class<?> clazz = Class.forName(className);
-      Constructor<?> constructor = clazz.getConstructor();
-      setHelper((XmlValidationHelper) constructor.newInstance());
-    } catch (ClassNotFoundException | NoSuchMethodException | SecurityException
-        | InstantiationException | IllegalAccessException | IllegalArgumentException
-        | InvocationTargetException ex) {
-      logger.log(Level.SEVERE, ex.getMessage());
+    if (data == null || !(data instanceof String)) {
```
chanseokoh: ditto

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
briandealwis: tests are failing so I'll add this.

briandealwis:
Oh dear — I remember this now.  We're dealing with two different types of validators: file validators (that operate on files), and source validators (that operate on `Document`s — essentially strings).  So Abbie had two different validators that pushed the actual validation logic into a common helper classes.
```xml
  	  <validator
		  scope="total"
		  class="com.google.cloud.tools.eclipse.appengine.validation.XmlSourceValidator:com.google.cloud.tools.eclipse.appengine.validation.AppEngineWebXmlValidator"
		  id="appEngineWebXmlSourceValidator">
```
But `createExecutableExtension()` is provided the _attribute_ name containing the class name to resolve and instantiate, and we want to instantiate that second part.  And so we can't use `createExecutableExtension()`.

Ok, I'll add comments to explain, but will get rid of the superfluous null test.

briandealwis:
PTAL: added comments as to why the `newInstance()` is necessary.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
```diff
@@ -112,17 +111,24 @@ void validate(IReporter reporter, IFile source, byte[] bytes) throws IOException
...
-        | InstantiationException | IllegalAccessException | IllegalArgumentException
-        | InvocationTargetException ex) {
+      this.setHelper((XmlValidationHelper) clazz.newInstance());
+    } catch (ClassNotFoundException
+        | SecurityException
+        | InstantiationException
+        | IllegalAccessException
+        | IllegalArgumentException ex) {
       logger.log(Level.SEVERE, ex.getMessage());
+      throw new CoreException(StatusUtil.error(this, "Unable to instantate helper", ex));
```
chanseokoh: instantiate

PR #2879, opened 2018-03-02
Make AESv8-related tests work in Eclipse

### LINKED ISSUE #2868: Examine some tests in .validation.test that fail if Java 8 facet is available
Examine if the following tests work as intended.

- `WebXmlValidatorTest.testValidateJavaServlet()`
- `XmlSourceValidatorTest.testValidate()`

`WebXmlValidatorTest.testValidateJavaServlet()` expects to see a validation error reported: "App Engine Standard does not support this servlet version" when `web.xml` declares Servlet version 3.1 (i.e., `<web-app version=3.1>`) in an App Engine Java7 project. (BTW, it'd be nice to update the error message to say "App Engine Standard Java 7 ...".)

However, these tests fail when running inside Eclipse. They don't report any validation error. We haven't noticed the failures with `mvn verify`, because the `.validation.test` isn't importing the `.java8` bundle. When we force including the `.java8` bundle (e.g., adding it to `MANIFEST.MF`), `mvn verify` fails as well.

Looking at the code, `AppEngineStandardFacet.checkServletApiSupport()` seems to show different behaviors depending on whether the Java 8 facet is present or not.

I believe `XmlSourceValidatorTest.testValidate()` fails for the same reason or at least for a similar reason.

briandealwis:
- adds a facet version property `appengine.runtime` to the AES JRE8 facet version to ensure the `<runtime>` value
- use the `appengine.runtime` property to set the `<runtime>` value as required
- rename `AppEngineStandardFacetVersionChangeTests` to ensure they are run by Tycho surefire tests
- fix up tests to reflect the current reality from #2591 

Fixes #2868

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacetTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacetTest.java
@@ -29,6 +29,7 @@ import org.eclipse.wst.common.project.facet.core.IProjectFacetVersion;
 import org.eclipse.wst.common.project.facet.core.ProjectFacetsManager;
 import org.eclipse.wst.server.core.IRuntimeType;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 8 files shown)


briandealwis:
```
Running com.google.cloud.tools.eclipse.appengine.standard.java8.AppEngineStandardFacetVersionChangeTest
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 12.638 sec - in com.google.cloud.tools.eclipse.appengine.standard.java8.AppEngineStandardFacetVersionChangeTest```

briandealwis:
Should mention that we don't need to default version for App Engine Standard facet.  When unspecified, WTP chooses the latest version of the facet as the default.  We configure our App Engine Standard Facet to use the StringVersionComparator to ensure `JRE7` sorts before `JRE8`, and so the JRE8 will appear as the default (if present).

In practice, this doesn't matter since our _New Project_ wizard prompts the user for the runtime, and the default is set in `AppEngineStandardWizardPage`.

PR #2885, opened 2018-03-05
Create and show SDK install/update console for background jobs

### LINKED ISSUE #2838: Stream managed Cloud SDK install output to console for local run
We stream the install output to the console when deploying but not for the local run. Currently, it is just a simple job progress, and even if we integrate the newer progress monitoring from appengine-plugins-core later, users won't see the detailed output.

selection_006

We do have a console for the dev appserver already, so we could simply stream to it.

When install fails, it'd be easier to see what went wrong with the install output. An example is https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/541.

chanseokoh:
Fixes #2838.

Currently, we see the detailed SDK install/update output only when deploying, for which we have a deploy console. For the background install/update jobs, only shown are the brief monitor task names.

This PR creates a new message console for those jobs. The console is created when the job is instantiated. There will be only one console which will be reused across multiple (sequential) install/update jobs.

When users open the job progress view, they will see task names as hyperlinks:

selection_005

When clicking on the link, it will bring the console to the front.

selection_006

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.sdk/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.sdk/META-INF/MANIFEST.MF
@@ -9,17 +9,21 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-ActivationPolicy: lazy
 Bundle-Localization: plugin
 Require-Bundle: com.google.cloud.tools.appengine;bundle-version="0.5.0"
-Import-Package: com.google.cloud.tools.eclipse.util.jobs,
+Import-Package: com.google.cloud.tools.eclipse.ui.util,
+ com.google.cloud.tools.eclipse.util.jobs,
  com.google.cloud.tools.eclipse.util.status,
  com.google.common.annotations;version="[21.0.0,22.0.0)",
  com.google.common.base;version="[21.0.0,22.0.0)",
  org.eclipse.core.runtime;version="3.5.0",
  org.eclipse.core.runtime.jobs,
  org.eclipse.core.runtime.preferences,
+ org.eclipse.jface.action,
  org.eclipse.jface.preference,
  org.eclipse.osgi.service.debug,
+ org.eclipse.swt.events,
  org.eclipse.ui.console,
  org.eclipse.ui.preferences,
+ org.eclipse.ui.progress,
  org.osgi.framework;version="[1.8.0,2.0.0)",
  org.osgi.service.prefs
 Eclipse-RegisterBuddy: com.google.cloud.tools.appengine
--- a/plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
@@ -31,9 +31,6 @@ import org.eclipse.core.runtime.IProgressMonitorWithBlocking;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.Job;
-import org.eclipse.jface.action.Action;
-import org.eclipse.ui.console.ConsolePlugin;
-import org.eclipse.ui.console.IConsoleManager;
 import org.eclipse.ui.console.MessageConsole;
 import org.eclipse.ui.console.MessageConsoleStream;
 import org.eclipse.ui.progress.IProgressConstants;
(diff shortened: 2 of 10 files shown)


plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -54,11 +57,22 @@
...
     super(jobName);
-    this.consoleStream = consoleStream;
+    this.consoleStream = consoleStream != null ? consoleStream : createNewMessageConsole();
     this.cloudSdkLock = cloudSdkLock;
     setRule(MUTEX_RULE);
   }
 
+  @VisibleForTesting
+  MessageConsoleStream createNewMessageConsole() {
+    final MessageConsole console = MessageConsoleUtilities.findOrCreateConsole(
```
elharo: why final?

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -116,11 +127,9 @@ static void clearBlocked(IProgressMonitor monitor) {
 
   protected void subTask(IProgressMonitor monitor, String description) {
     monitor.subTask(description);
-    if (consoleStream != null) {
-      // make output headers distinguishable on the console
-      String section = String.format("[%s]", description); // $NON-NLS-1$
-      consoleStream.println(section);
-    }
+    // make output headers distinguishable on the console
+    String section = String.format("[%s]", description); // $NON-NLS-1$
```
elharo: simple string concatenation might be clearer

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
chanseokoh: Oops, a remnant of an anonymous class creation.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -54,11 +57,22 @@
...
+    this.consoleStream = consoleStream != null ? consoleStream : createNewMessageConsole();
     this.cloudSdkLock = cloudSdkLock;
     setRule(MUTEX_RULE);
   }
 
+  @VisibleForTesting
+  MessageConsoleStream createNewMessageConsole() {
+    MessageConsole console = MessageConsoleUtilities.findOrCreateConsole(
+        Messages.getString("configuring.cloud.sdk"), // $NON-NLS-1$
+        name -> new MessageConsole(name, null /* imageDescriptor */));
```
elharo: Use of lambda here feels a little gratuitous. What do we gain by not simply passing in the name ?

chanseokoh: The API accepts a factory class. In this case, we don't need a custom console creator, but others like the deploy console, they need to create and return customized consoles. Maybe we could add a new API that just accepts a name. Let me look into it.

chanseokoh: Actually, there did exist another method that just accepts a name. It works well.

plugins/com.google.cloud.tools.eclipse.sdk/META-INF/MANIFEST.MF
```diff
@@ -9,17 +9,21 @@ Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-ActivationPolicy: lazy
 Bundle-Localization: plugin
 Require-Bundle: com.google.cloud.tools.appengine;bundle-version="0.5.0"
-Import-Package: com.google.cloud.tools.eclipse.util.jobs,
+Import-Package: com.google.cloud.tools.eclipse.ui.util,
```
briandealwis: Sorry I didn't get to reviewing this until now.  The other real comment is that we had the SDK stuff separated into a .sdk and .sdk.ui bundles.  This change has now pulled in UI stuff into both bundles.  I'm not sure this separation bought us much, so it might be worth merging the two bundles.

PR #2886, opened 2018-03-05
Upgrade to appengine-plugins-core 0.5.1

chanseokoh:


chanseokoh:
Ah, there are compilation problems.

PR #2888, opened 2018-03-06
Update CT4E feature label

chanseokoh:


chanseokoh:
If we change the feature ID, should we also change the directory name `featues/com.google.cloud.tools.eclipse.suite.e45.feature` to match it?

briandealwis:
+1

elharo:
Should the feature ID really specify an Eclipse version? In hindsight that seems...odd.

briandealwis:
I think we took the oaten from GPE where they had patches/workarounds bugs.  I don't see any reason that we need it.

chanseokoh:
I think it's best for @briandealwis to work on this, which involves modifying `p2.inf`.

PR #2893, opened 2018-03-06
Rename .e45 feature and support auto-update from old feature

briandealwis:
This patch strips out the `.e45` from our `com.google.cloud.tools.eclipse.suite.e45.feature` feature, and adds a p2 update directive to allow seamlessly updating from the old feature as well as from older versions of the now-renamed feature.


briandealwis:
Unfortunately there isn't an easy way to test the update except by hand.

briandealwis:
Hmm, this isn't producing the p2 metadata now.  Trying to figure out what's going on.

PR #2900, opened 2018-03-06
Do not clear existing console

### LINKED ISSUE #2895: Cloud SDK install/update message console may be cleared in the middle of output
If a message console is not explicitly given for Cloud SDK modification, we auto-create one. Once created, the auto-created console will be reused across multiple SDK modify jobs.

One problem is that, because of the way how `MessageConsoleUtilities.getMessageConsole()` works,  the auto-created console is cleared whenever an SDK modify job is created. So, for example, when SDK install/updating is in progress and streaming the detailed output, if the user tries to do a local run (which will create a new job), it will clear the console and the old output is forever lost.

I think we can call `getMessageConsole()` lazily to fix this.

chanseokoh:
Fixes #2895.

I made `getMessageConsole()` not clear an existing console. I think it actually doesn't make much sense to clear when reusing it. If a caller does want to clear it, the caller could simply do `.clearConsole()`. BTW, the only caller of `getMessageConsole()` is our SDK modify job.

Other changes are refactoring to reuse code, besides making the class non-static for testing.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
@@ -154,9 +154,9 @@ public abstract class DeployCommandHandler extends AbstractHandler {
     IPath workDirectory = createWorkDirectory();
     DeployPreferences deployPreferences = getDeployPreferences(project);
 
-    String consoleName = getConsoleName(deployPreferences.getProjectId());
     DeployConsole messageConsole =
-        new MessageConsoleUtilities().createConsole(consoleName, new DeployConsole.Factory());
+        MessageConsoleUtilities.createConsole(getConsoleName(deployPreferences.getProjectId()),
+                                              new DeployConsole.Factory());
     IConsoleManager consoleManager = ConsolePlugin.getDefault().getConsoleManager();
     consoleManager.showConsoleView(messageConsole);
     ConsoleColorProvider colorProvider = new ConsoleColorProvider();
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployCommandHandler.java
```diff
@@ -154,9 +154,9 @@ private void launchDeployJob(IProject project, Credential credential)
     IPath workDirectory = createWorkDirectory();
     DeployPreferences deployPreferences = getDeployPreferences(project);
 
+    String consoleName = getConsoleName(deployPreferences.getProjectId());
     DeployConsole messageConsole =
-        MessageConsoleUtilities.createConsole(getConsoleName(deployPreferences.getProjectId()),
-                                              new DeployConsole.Factory());
+        new MessageConsoleUtilities().createConsole(consoleName, new DeployConsole.Factory());
```
elharo: why non-static? This really looks like it should be static here.

plugins/com.google.cloud.tools.eclipse.sdk/src/com/google/cloud/tools/eclipse/sdk/internal/CloudSdkModifyJob.java
```diff
@@ -64,7 +64,7 @@ public CloudSdkModifyJob(String jobName, MessageConsoleStream consoleStream,
 
   @VisibleForTesting
   MessageConsoleStream createNewMessageConsole() {
-    MessageConsole console = MessageConsoleUtilities.getMessageConsole(
+    MessageConsole console = new MessageConsoleUtilities().getMessageConsole(
```
elharo: constantly creating an object, calling one method on it, and then letting the object be garbage collected strongly suggests you really want a static method

chanseokoh: That was for testing. I'll revert them to `static` and take a different approach to inject a `IConsoleManager`.

PR #2902, opened 2018-03-07
default locations

### LINKED ISSUE #2897: When SDK is found in customary location, preferences don't show this
image

- Cloud Tools for Eclipse version: 1.6.0.201803061547
- Google Cloud SDK version: 188.0.0
- Eclipse version: 4.7.2.v20171130-0510
- OS: Mac OS X 10.13.3
- Java version: 1.8.0_144

**What did you do?**

**What did you see instead?**

Screenshots and stacktraces are helpful.

elharo:
@chanseokoh @briandealwis fix #2897

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
@@ -158,16 +158,13 @@ public class CloudSdkPreferenceAreaTest {
   }
   
   @Test
-  public void testI2897() {
+  // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2897
+  public void testSearchSdkIfSdkLocationIsEmpty() {
     when(cloudSdkManager.isManagedSdkFeatureEnabled()).thenReturn(true);
     when(preferences.getString(CloudSdkPreferences.CLOUD_SDK_PATH)).thenReturn("");
+    when(preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT)).thenReturn("MANUAL");
     
     createPreferenceArea();
-    chooseSdk.setSelection(true);
-    sdkLocation.setText("");
-    
-    // now need to invoke update
-    area.updateSelectedVersion();
     
     assertTrue(sdkLocation.getText().endsWith("google-cloud-sdk"));
     assertTrue(sdkVersion.getText().length() > 3);
--- a/plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
@@ -156,6 +156,24 @@ public class CloudSdkPreferenceAreaTest {
     assertNotNull(sdkLocation);
     assertNotNull(sdkVersion);
   }
+  
+  @Test
+  public void testI2897() {
+    when(cloudSdkManager.isManagedSdkFeatureEnabled()).thenReturn(true);
+    when(preferences.getString(CloudSdkPreferences.CLOUD_SDK_PATH)).thenReturn("");
+    
+    createPreferenceArea();
+    chooseSdk.setSelection(true);
+    sdkLocation.setText("");
+    
+    // now need to invoke update
+    area.updateSelectedVersion();
+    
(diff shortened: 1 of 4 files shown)


elharo:
Yes, that happens. I'm not sure it shouldn't happen like that, and I definitely think that's an improvement on the current state. If the user tries to remove the cloud SDK by emptying out the field, we will try to pick one and if we happen to pick a managed SDK that's not a problem.

briandealwis:
Agreed — the managed SDK is still a valid SDK location, but if they are in MANUAL then it won't be managed.

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
```diff
@@ -171,13 +173,33 @@ private void updateManagedSdk() {
     // would be nice if we could updateSelectedVersion() when finished
   }
 
-  private void updateSelectedVersion() {
+  // todo anyway to force this through the UI widgets rather than direct invocation?
```
briandealwis: I don't understand this comment

chanseokoh:
@briandealwis @elharo that could make sense if the page would show the managed SDK location consistently. That is, if you close and open the page for the second time, our resolver will be disabled (since the preference is now to not return the managed location) and the page won't show the managed location.

chanseokoh:
Okay, if we can prevent saving an empty value, I agree this could work.

plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
```diff
@@ -155,6 +156,23 @@ public void testUi_sdkManagementFeature() {
     assertNotNull(sdkLocation);
     assertNotNull(sdkVersion);
   }
+  
+  @Test
+  public void testI2897() {
+    when(cloudSdkManager.isManagedSdkFeatureEnabled()).thenReturn(true);
```
chanseokoh: Add `when(preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT)).thenReturn("MANUAL");`, and you don't need to manually call `chooseSdk.setSelection(true)` below. `sdkLocation.setText("")` is unnecessary too. I prefer this way, as it also verifies that the page is correctly initialized with the given preferences. Also, because you don't manually call `sdkLocation.setText("")`, you don't need `area.updateSelectedVersion()` either.

plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
```diff
@@ -155,6 +156,23 @@ public void testUi_sdkManagementFeature() {
     assertNotNull(sdkLocation);
     assertNotNull(sdkVersion);
   }
+  
+  @Test
+  public void testI2897() {
```
chanseokoh: `testSearchSdkIfSdkLocationIsEmpty()`?

plugins/com.google.cloud.tools.eclipse.sdk.ui/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceArea.java
chanseokoh: I think @elharo is talking about calling `area.updateSelectedVersion()` directly in the test (`testI2897()`). But in my testing, I didn't need to call it directly with correct changes.

plugins/com.google.cloud.tools.eclipse.sdk.ui.test/src/com/google/cloud/tools/eclipse/sdk/ui/preferences/CloudSdkPreferenceAreaTest.java
```diff
@@ -155,6 +156,23 @@ public void testUi_sdkManagementFeature() {
...
+    when(cloudSdkManager.isManagedSdkFeatureEnabled()).thenReturn(true);
+    when(preferences.getString(CloudSdkPreferences.CLOUD_SDK_PATH)).thenReturn("");
+    
+    createPreferenceArea();
+    chooseSdk.setSelection(true);
+    sdkLocation.setText("");
+    
+    // now need to invoke update
+    area.updateSelectedVersion();
```
chanseokoh:
```java
    chooseSdk.setSelection(true);
    sdkLocation.setText("");
    area.updateSelectedVersion();
```

All of these three are not needed once you add `    when(preferences.getString(CloudSdkPreferences.CLOUD_SDK_MANAGEMENT)).thenReturn("MANUAL");` as I said earlier.

PR #2903, opened 2018-03-07
Fix .e45 feature p2-metadata generation and add tests

### LINKED ISSUE #2890: Change the feature ID to not specify "4.5"
That is `com.google.cloud.tools.eclipse.suite.e45.feature` should be `com.google.cloud.tools.eclipse.suite.feature`

Add an update section in an accompanying p2.inf (see [Update descriptor advice](https://help.eclipse.org/neon/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fp2_customizing_metadata.html))
We should also change the directory name featues/com.google.cloud.tools.eclipse.suite.e45.feature to match.

briandealwis:
I get inconsistent results with #2893 where the p2 update metadata wasn't always being picked up.  It turns out I had a missing `range()` in the p2ql expression, and fixing that now ensures the metadata is always generated.  I realized that I can can use ant to test that the metadata was generated.

Fixes #2890


briandealwis:
Committing since Travis push build failed on #1798 (sporadic) and all other builds passed.

Issue #2899, opened 2018-03-06
Remove Java 7 support


elharo:
Java 7 runtime support was deprecated on December 13, 2017, and will be shut down entirely on January 16, 2019. It is replaced by the Java 8 Runtime Environment.

At some point before then we should stop offering to create new Java 7 projects. We might still want to allow servlet 2.5 (web.xml instead of annotation projects) though.

briandealwis:
We could change the Java 7 label to indicate that it's deprecated?

patflynn:
+1. I'd be tempted to remove it entirely at this point.  Why send someone
down a path that ends within a year?

elharo:
I don't want t drop it too fast. IO know there are still customers using this, but let's mark it deprecated in thew wizard.

PR #2915, opened 2018-03-08
Mark Java 7 runtime as deprecated

briandealwis:
#2899


briandealwis:
I think 3.1 is backwards compatible.

Issue #2872, opened 2018-02-28
javax.servlet-api-3.1.0.jar and javax.servlet.jsp-api-2.3.0.jar not fully downloaded by Cloud Tools For Eclipse


jamesmortensen:
- Cloud Tools for Eclipse version: 1.5.1
- Google Cloud SDK version:  latest
- Eclipse version: Oxygen
- OS:  Windows (Not yet confirmed if it's also happening on other platforms)
- Java version:  1.8.0._152 

**What did you do?**

After installing the Cloud Tools for Eclipse, we create a Standard App Engine project.  When we try to run it using Run As -> App Engine, it fails because it can't read the following JAR files:

- ~/.m2/repository/javax/servlet/javax.servlet-api/3.1.0/javax.servlet-api-3.1.0.jar
- ~/.m2/repository/javax/servlet/jsp/javax.servlet.jsp-api/2.3.0/javax.servlet.jsp-api-2.3.0.jar

**What did you expect to see?**

We expect the newly created project to run without issues, just like in the old deprecated Google App Engine for Eclipse plugin.

I will send screenshots and more information once I get confirmation that this isn't a duplicate.  We've seen this on literally every computer in our organization for months, so I'm surprised no one has reported before and that it's been broken this long.

What we do to resolve the issue is this:

- Check the .m2 repository cache in the home folder
- Observe that the two JAR files listed above are present.
- Next, download fresh copies from Maven through Google Chrome
- Observe that the file sizes are different
- Copy the newly downloaded files into the Maven cache.

Once we overwrite the files CT4E downloads, newly created projects will then run.   But on a fresh system with a freshly installed Cloud Tools for Eclipse plugin installed, we have to fix this manually.  Also, it seems the installer isn't installing everything we need.  We don't want to use Maven, but the CT4E appears to rely on these files in the Maven cache in order to even create and run a non-maven project. So we create a Maven project just to get maven to do what it seems like the installer isn't doing, getting all the dependencies it needs in order to function properly....

elharo:
Thanks for the report. It sounds like you may have had corrupt jar files in your .m2 directory. When you say "it fails because it can't read the following JAR files:" exactly what message did you see? Was there a stack trace on the console?

jamesmortensen:
After taking some time to look at this again on a fresh machine, I made some screenshots of the process I follow to create a new project.

[...]

PR #2936, opened 2018-03-13
Cause java.util.logging to shunt to slf4j

briandealwis:
In debugging #2872, a key observation was that activities were happening on different threads.  `java.util.logging` doesn't log the thread name, [nor can it be configured to do so](https://stackoverflow.com/a/6889151/600339).  Whereas logback can output the thread, and [our logback configuration does so](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.test.dependencies/logback.xml#L4).  We already shunt log4j and commons-logging into SLF4j and thence into Logback in our tests.

This fix brings in the [java.util.logging → SLF4j bridge](https://www.slf4j.org/api/org/slf4j/bridge/SLF4JBridgeHandler.html) and causes it to be installed by a new bundle-activator for `com.google.cloud.tools.eclipse.test.dependencies`.

Another advantage is that this change will reduce our log size since `java.util.logging`'s default output takes two lines per log record.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.dependencies/META-INF/MANIFEST.MF
+++ b/plugins/com.google.cloud.tools.eclipse.test.dependencies/META-INF/MANIFEST.MF
@@ -4,6 +4,7 @@ Bundle-Name: %pluginName
 Bundle-SymbolicName: com.google.cloud.tools.eclipse.test.dependencies;singleton:=true
 Bundle-Version: 0.1.0.qualifier
 Bundle-Vendor: %providerName
+Bundle-Activator: com.google.cloud.tools.eclipse.test.dependencies.LogConfigurator
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-Localization: plugin
 Require-Bundle: org.junit;bundle-version="[4.12.0,4.12.1)";visibility:=reexport,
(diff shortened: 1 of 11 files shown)


plugins/com.google.cloud.tools.eclipse.test.dependencies/META-INF/MANIFEST.MF
```diff
@@ -107,4 +109,7 @@ Export-Package: org.mockito;provider=google;version="1.10.19";
  org.mockito.verification;provider=google;version="1.10.19";uses:="org.mockito.internal.verification.api,org.mockito.internal.util",
  org.objenesis;provider=google;version="2.2.0";uses:="org.objenesis.instantiator,org.objenesis.strategy"
 Bundle-Classpath: lib/mockito-core-1.10.19.jar,
- lib/objenesis-2.2.jar
+ lib/objenesis-2.2.jar,
+ .
+Import-Package: org.osgi.framework
+Bundle-ActivationPolicy: lazy
```
chanseokoh: Move this above?

plugins/com.google.cloud.tools.eclipse.test.dependencies/build.properties
```diff
@@ -1,7 +1,11 @@
...
 # NB: only include the lib/.jars actually required as lib/ will include referenced bundles
 # including broken hamcrest-1.1
 bin.includes = META-INF/,\
                lib/mockito-core-1.10.19.jar,\
                lib/objenesis-2.2.jar,\
                logback.xml,\
-               plugin.properties
+               plugin.properties,\
+               .
+source.. = src/
```
chanseokoh: Duplication. Already in the first line.

briandealwis: Weird.

briandealwis:
Hmm, something's not quite right as now there are SLF4j binding errors.

briandealwis:
Hmm, those aren't new.  But I had exorcised them so I'm not sure how they reappeared.
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

Anyways, `.test.dependencies` now imports `org.slf4j.impl`, which is the package where any SLF4j bindings are defined. So we'll pull in some binding, and we make the Logback binding available in the target.  No more binding errors reported.

chanseokoh:
OK , sounds like we are good.

PR #2938, opened 2018-03-14
Add Configure > Reconfigure for App Engine Java 8 runtime

briandealwis:
Add new _Configure > Reconfigure for App Engine Java 8 runtime_ for App Engine standard Java 7 projects.  Adds a suite of context-menu verification tests, as well as an integration test that the conversion works.

I initially used "Convert to" or "Configure for" but it didn't read well IMHO.

Fixes #2914

PR #2939, opened 2018-03-14
Refactor code

chanseokoh:
Inspired by #2938.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/RunAppEngineShortcutTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/RunAppEngineShortcutTest.java
@@ -28,20 +28,16 @@ import com.google.cloud.tools.eclipse.ui.util.WorkbenchUtil;
 import java.io.ByteArrayInputStream;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jst.common.project.facet.core.JavaFacet;
 import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
-import org.eclipse.swt.widgets.Menu;
-import org.eclipse.swt.widgets.MenuItem;
 import org.eclipse.swtbot.eclipse.finder.SWTWorkbenchBot;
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/RunAppEngineShortcutTest.java
```diff
@@ -85,10 +81,8 @@ public void testRunAppEngine_hiddenEvenIfAppEngineProjectFileIsOpen() throws Cor
...
     SWTBotTreeItem selected = SwtBotProjectActions.selectProject(
         new SWTWorkbenchBot(), project.getName());
-    Menu runAsMenu = selected.contextMenu("Run As").widget.getMenu();
-    Matcher<MenuItem> matcher = WidgetMatcherFactory.withRegex("App Engine");
-
-    List<MenuItem> menuItems = new MenuFinder().findMenus(runAsMenu, matcher, false);
-    return !menuItems.isEmpty();
+    List<String> menuItems = selected.contextMenu("Run As").menuItems();
+    Predicate<String> isAppEngineMenu = name -> name.contains("App Engine");
+    return menuItems.stream().anyMatch(isAppEngineMenu);
```
briandealwis: The SWTBot solution can be optimized too.  Might want to contrast the streams to:
```
try {
   selected.contextMenu("Run As").menu(WidgetMatcherFactory.withRegex("App Engine"), false /* recursive */, 0 /* index */);
   return true;
} catch (WidgetNotFoundException ex) {
   return false;
}
```
Which has a slight advantage in that it will retry if the menu generation is expensive?

chanseokoh: The problem with `.menu(...)` is that it will wait for a long time until it throws `WidgetNotFoundException`:

```java
	public SWTBotMenu menu(Matcher<MenuItem> matcher, final boolean recursive, final int index) throws WidgetNotFoundException {
		WaitForObjectCondition<MenuItem> waitForMenuItem = Conditions.waitForMenuItem(this, matcher, recursive, index);
		new SWTBot().waitUntilWidgetAppears(waitForMenuItem);
		return new SWTBotMenu(waitForMenuItem.get(0), matcher);
	}
```

So it works for finding an existing menu, but it takes too long to verify an item doesn't exist.

chanseokoh: I remember that is why I went for `selected.contextMenu("Run As").widget.getMenu()` previously.

Issue #2909, opened 2018-03-07
Tests in .validation.test take too long


chanseokoh:
I have been working on cutting the time in this bundle.

PR #2942, opened 2018-03-15
Do not add App Engine facet and Dynamic Web Module facet

chanseokoh:
Branched off of #2941, so please review #2941 first.

Part of #2909.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidatorTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidatorTest.java
@@ -23,28 +23,28 @@ import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
-import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
 import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Predicate;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.jface.text.IDocument;
-import org.eclipse.jst.common.project.facet.core.JavaFacet;
-import org.eclipse.jst.j2ee.web.project.facet.WebFacetUtils;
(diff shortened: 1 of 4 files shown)


chanseokoh:
Merged with `master` and ready for review. All the other similar PRs are also ready.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidatorTest.java
```diff
@@ -56,18 +56,24 @@
 
   private final IncrementalReporter reporter = new IncrementalReporter(null);
 
-  @Rule public TestProjectCreator nonFacetedProject = new TestProjectCreator();
+  @Rule public TestProjectCreator projectCreator = new TestProjectCreator();
```
briandealwis: Do we actually need real projects?  Can we use `IProject` mocks?

chanseokoh: I'm afraid not. One main reason is to create an `IFile`, which I believe we haven't found a way to mock it in these tests that actually reads contents.

Another complications is that we make use of a lot of actual components which accepts `IProject` and `IFile`, and they interact. They would work properly only when they are real projects and files:

```java
    IWorkbench workbench = PlatformUI.getWorkbench();
    WorkbenchUtil.openInEditor(workbench, file);
    ITextViewer viewer = ValidationTestUtils.getViewer(file);
```
```java
    IncrementalHelper helper = new IncrementalHelper(document, project);
```
```java
      IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
      IFile file = root.getFile(path);
```

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlSourceValidator.java
```diff
@@ -56,12 +57,22 @@
...
   private XmlValidationHelper helper;
 
+  private final Predicate<IFacetedProject> hasAppEngineStandardFacet;
+
+  public XmlSourceValidator() {
+    this(AppEngineStandardFacet::hasFacet);
+  }
+
+  @VisibleForTesting
+  XmlSourceValidator(Predicate<IFacetedProject> hasAppEngineStandardFacet) {
```
elharo: What's the advantage of adding this? It seems more complex without adding functionality. The old code did not need visible for testing.

chanseokoh: Yeah, it's a trade-off between less code vs. less time to run tests, by allowing mocking some underlying component. Since this change isn't the matter of whether we can write tests or not, I'm not pushing hard for this change. So, if you don't like this kind of change, I'll probably drop it.

PR #2943, opened 2018-03-15
Do not always create Dynamic Web project

chanseokoh:
For #2909.

- Does not always create a DWP. This is the main change. Other changes in the same class are basically refactoring.
- `testValidate_badXml()` is broken down to two: the later part separated out into `testValidate_badXml_dynamicWebProject()`.


briandealwis:
How much time does the DWP introduce?

chanseokoh:
Around 4 seconds on my machine.

PR #2944, opened 2018-03-15
Avoid Dynamic Web project

chanseokoh:
Part of #2909.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ToServlet25SourceQuickFixTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ToServlet25SourceQuickFixTest.java
@@ -16,23 +16,17 @@
 
 package com.google.cloud.tools.eclipse.appengine.validation;
 
-import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
-import com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet;
-import com.google.cloud.tools.eclipse.test.util.project.ProjectUtils;
 import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
 import com.google.cloud.tools.eclipse.ui.util.WorkbenchUtil;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.ITextViewer;
-import org.eclipse.jst.common.project.facet.core.JavaFacet;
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ToServlet25SourceQuickFixTest.java
```diff
@@ -68,12 +58,10 @@ public void testConvertServlet() throws CoreException {
     IDocument document = viewer.getDocument();
     String contents = document.get();
     assertFalse(contents.contains("version='3.1'"));
+    assertTrue(contents.contains("version=\"2.5\""));
 
     // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/1527
     editorPart.doSave(new NullProgressMonitor());
-
-    ProjectUtils.waitForProjects(project);
-    assertEquals(0, file.findMarkers(SERVLET_MARKER, true, IResource.DEPTH_ZERO).length);
```
briandealwis: The justification of removing this code being that it's an end-to-end test, and this outcome is already tested elsewhere.  Right?  Seems reasonable, just want to verify.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorPluginTest.java
```diff
@@ -48,10 +46,12 @@
   private IJavaProject javaProject;
   private IResource resource;
   @Rule public TestProjectCreator projectCreator = new TestProjectCreator().withFacets(
-      JavaFacet.VERSION_1_7, WebFacetUtils.WEB_25, AppEngineStandardFacet.JRE7);
+      JavaFacet.VERSION_1_7);
 
-  @Before
-  public void setUp() throws CoreException {
+  private void setUpProject(boolean dynamicWebProject) throws CoreException {
```
briandealwis: Perhaps we'd be better separating the tests into two classes, one with tests requiring only a plain project, the other for tests requiring the DWP.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/ToServlet25SourceQuickFixTest.java
chanseokoh: This is not tested elsewhere actually, but I did want to downgrade this end-to-end test to a smaller unit test. I think it's a reasonable change considering the new assertion.

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/WebXmlValidatorPluginTest.java
chanseokoh: Let me try that.

chanseokoh: Done!

PR #2951, opened 2018-03-16
Wait until classes are built: LocalAppEnginePublishOperationTest

chanseokoh:
Hopefully this is the last PR for #1798.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperationTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperationTest.java
@@ -66,10 +66,13 @@ public class LocalAppEnginePublishOperationTest {
     serverProject = projects.stream().filter(isServerProject).findFirst().get();
     sharedProject = projects.stream().filter(isSharedProject).findFirst().get();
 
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.localserver.test/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEnginePublishOperationTest.java
```diff
@@ -66,10 +66,13 @@ public void setUp() throws IOException, CoreException, InterruptedException {
...
     for (int i = 0; i < 100; i++) {
       serverModule = ServerUtil.getModule(serverProject);
       sharedModule = ServerUtil.getModule(sharedProject);
-      if (serverModule != null && sharedModule != null) {
+      if (serverModule != null && sharedModule != null
+          && serverProject.getFile("bin/sox/server/GreetingServiceImpl.class").exists()
+          && sharedProject.getFile("bin/sox/shared/GreetingService.class").exists()) {
         break;
       }
       Thread.sleep(100);
```
briandealwis: Could you put in a call to `ThreadDumpingWatchdog` here so that we can see the pending jobs?

chanseokoh: Added `ThreadDumpingWatchdog.report("Until modules are fully ready", stopwatch)`.

The only possible problem could be that we may never take our time to look for a report if this PR really fixes #1798 and tests always succeed.

PR #2955, opened 2018-03-20
Track Maven vs. native project creation for App Engine Standard and Flex

### LINKED ISSUE #2649: Track Maven vs. native project creation both for App Engine standard and flex
In the past we were tracking in terms of wizards we had: standard native wizard, standard Maven wizard, and the flex wizard.

Now the plugin evolved to have two wizards: std and flex, while each wizard can optionally configure Maven.

chanseokoh:
#2649


chanseokoh:
This kind of Analytics testing (testing whether we are sending correct actual values) is difficult to test, unless we change our ping API to enable further mocking. What we've been doing is to test the underlying Analytics framework instead, so that if correct values are given, the values will be sent correctly.

BTW, I did manually testing (creating std/flex projects with and without Maven).

PR #2956, opened 2018-03-20
add a BOM to our generated pom.xml files

### LINKED ISSUE #2923: BOM support
Make sure all library dependencies sync up. See https://docs.google.com/document/d/1dmhS0JSepXDizZScUf2tPBXqK1SumKRFOOI2WxMQLzk/edit?ts=5a9d63fd

elharo:
@chanseokoh fix #2923 for Maven projects


plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.flex.ftl
```diff
@@ -22,6 +22,18 @@
   <prerequisites>
     <maven>3.3.9</maven>
   </prerequisites>
+  
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud</artifactId>
+        <version>${googleCloudJavaVersion}</version>
```
briandealwis: `googleCloudJavaBomVersion`?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
```diff
@@ -237,6 +238,45 @@ private Element validatePom()
...
+                      coord.getTextContent().trim());
+                  Assert.assertTrue(
+                      bomVersion.compareTo(new DefaultArtifactVersion("0.40.0-alpha")) >= 0);
+                } else if (coord.getNodeName().equals("type")) {
+                  typePresent = true;
+                } else if (coord.getNodeName().equals("scope")) {
+                  scopePresent = true;
+                }
+              }
+              Assert.assertTrue(scopePresent);
```
briandealwis: This would  be simplified using XPath as in `PomXmlValidator`?

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -208,7 +214,7 @@ private static void createPomXml(IProject project, AppEngineProjectConfig config
     String sdkVersion = getCurrentVersion(
         "com.google.appengine", //$NON-NLS-1$
         "appengine-api-1.0-sdk", //$NON-NLS-1$
-        "1.9.62"); //$NON-NLS-1$
+        "1.9.63"); //$NON-NLS-1$
```
briandealwis: I think it's worth pulling these artifact constants (including the BOM artifact too) out into explicit constants rather than have them scattered in the code.

elharo: Makes sense. Can you file an issue for that?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/Maven4NamespaceContext.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.google.common.collect.ImmutableList;
+import java.util.Iterator;
+import javax.xml.namespace.NamespaceContext;
+
+/**
+ * Maps all prefixes to http://maven.apache.org/POM/4.0.0 and
+ * all namespace URIs to "m". Not suitable for general use.
+ */
+class Maven4NamespaceContext implements NamespaceContext {
```
briandealwis: You can reuse `com.google.cloud.tools.eclipse.appengine.validation.MavenContext`

chanseokoh:
Trying to understand the direction of this BOM stuff. So, with a BOM, I believe eventually we will get rid of the `<version>` element of Cloud libraries we add to `pom.xml` through the library page, right? Just checking, as this isn't happening in this PR.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
briandealwis: #2958

elharo:
With the BOM we can remove explicit versions from the google-cloud-java dependencies. However this may require splitting our library addition code into two: one for google-cloud-java dependencies and one for everything else. And we'll need to figure out how to tell whether any given dependency, especially a transitive dependency, is or is not covered by the BOM.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/Maven4NamespaceContext.java
elharo: I'd forgotten about that class. Looking at it now, it has the same issues this class does. That is, it works only for very limited code and is actively broken for all other uses. Given that, I don't feels safe exposing it as part of the public API. We could look for or build a more correct namespace context that fully satisfies the interface's contract.

elharo:
There's a bug where our code is expecting there to be only one `dependencies` element but gets confused when there's more than one in the document:

```
  <dependencyManagement>
    <dependencies>
      <dependency>
        <artifactId>google-cloud</artifactId>
        <version>0.40.0-alpha</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    <dependency>
<groupId>com.google.endpoints</groupId>
<artifactId>endpoints-framework</artifactId>
<version>2.0.13</version>
</dependency>
<dependency>
<groupId>com.googlecode.objectify</groupId>
<artifactId>objectify</artifactId>
<version>5.1.22</version>
</dependency>
</dependencies>
  </dependencyManagement>
```

I'm surprised tests didn't catch this.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2018 Google LLC.
```
chanseokoh: No period after "LLC".

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Google Inc.
+ * Copyright 2017 Google LLC. All Rights Reserved.
```
chanseokoh: Should end with "LLC" with no period.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
```diff
@@ -0,0 +1,89 @@
...
+    DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+    factory.setNamespaceAware(true);
+    DocumentBuilder builder = factory.newDocumentBuilder();
+    DOMImplementation impl = builder.getDOMImplementation();
+
+    Document doc = impl.createDocument("http://maven.apache.org/POM/4.0.0", "project", null);
+
+    Node rootElement = doc.getDocumentElement();
+
+    dependencyManager =
```
chanseokoh: Is this `dependencyManager` or `dependencyManagement`? (I'm confused.)

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/testdata/testpom.xml
```diff
@@ -13,6 +13,18 @@
   <artifactId>com.google.cloud.tools.eclipse.appengine.libraries.test</artifactId>
   <version>0.1.0-SNAPSHOT</version>
   <packaging>eclipse-test-plugin</packaging>
+  
+  <dependencyManagement>
```
chanseokoh: Is this `dependencyManager` or `dependencyManagement`? (I'm confused.)

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Set;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+import org.w3c.dom.Element;
+
+/**
+ * 
+ */
```
chanseokoh: Remove?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
```diff
@@ -0,0 +1,96 @@
...
+class Bom {
+  
+  private final static XPathFactory factory = XPathFactory.newInstance();
+
+  private static Set<String> artifacts;
```
chanseokoh: How about `artifactIds`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -135,12 +153,24 @@ void updateDependencies(Collection<Library> selectedLibraries,
...
+          document.getDocumentElement(),
+          XPathConstants.NODESET);
+      if (dependenciesNodes.getLength() > 0) {
+        dependencies = (Element) dependenciesNodes.item(0);
+      } else {
+        dependencies = document.createElementNS(
+            "http://maven.apache.org/POM/4.0.0", "dependencies");
+      }
+    } catch (XPathExpressionException ex) {
+      IStatus status = StatusUtil.error(Pom.class, ex.getMessage(), ex);
```
chanseokoh: This doesn't seem it will happen. `RuntimeException`?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
```diff
@@ -0,0 +1,96 @@
...
+
+    try {
+      String bomGroupId = (String) xpath.evaluate(
+          "string(./m:dependencies/m:dependency/m:groupId)",
+          dependencyManager,
+          XPathConstants.STRING);
+      String bomArtifactId = (String) xpath.evaluate(
+          "string(./m:dependencies/m:dependency/m:artifactId)",
+          dependencyManager,
+          XPathConstants.STRING);
```
chanseokoh: Let's say users include another 3rd-party BOM in addition to ours. I guess this code may not work, since it could return the group ID and artifact ID of the other 3rd-party POM?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
```diff
@@ -0,0 +1,96 @@
...
+class Bom {
+  
+  private final static XPathFactory factory = XPathFactory.newInstance();
+
+  private static Set<String> artifacts; 
+  
+  static {
+    try {
+      List<CloudLibrary> cloudLibraries =
+          com.google.cloud.tools.libraries.CloudLibraries.getCloudLibraries();
```
chanseokoh: To confirm: so we simply assume here our Cloud BOM includes all the libraries that appengine-plugins-core returns, and a future work is to read the actual content of the BOM?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+  
+  static boolean defines(Element dependencyManager, String groupId, String artifactId) {
+    
+    XPath xpath = factory.newXPath();
+    xpath.setNamespaceContext(new Maven4NamespaceContext());
+
+    try {
+      String bomGroupId = (String) xpath.evaluate(
+          "string(./m:dependencies/m:dependency/m:groupId)",
```
briandealwis: Wouldn't you want to get with `type=pom` and `scope=import`?

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
```diff
@@ -237,11 +240,50 @@ private Element validatePom()
...
+    
+    XPath xpath = XPathFactory.newInstance().newXPath();
+    xpath.setNamespaceContext(new Maven4NamespaceContext());
+    NodeList dependencyManagementNodes = (NodeList) xpath.evaluate(
+        "./m:dependencyManagement",
+        root,
+        XPathConstants.NODESET);
+    Assert.assertEquals(1, dependencyManagementNodes.getLength());
+    
+    String bomGroupId = (String) xpath.evaluate(
```
briandealwis: Worth checking that there is a single `<dependencies>` element?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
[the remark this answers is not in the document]
chanseokoh: We have a same-name class, so I think it's good to signify that this is from appengine-plugins-core.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
elharo: management, done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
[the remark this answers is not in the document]
elharo: right

elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -151,19 +164,19 @@ public void testAddDependencies_withDuplicates()
...
+
+      NodeList dependencyNodes = (NodeList) xpath.evaluate(
+            "./m:dependencies/m:dependency",
+            actual.getDocumentElement(),
+            XPathConstants.NODESET);
       
-      NodeList dependencies = actual.getElementsByTagName("dependencies");
-      NodeList children = ((Element) dependencies.item(0)).getElementsByTagName("dependency");
-      
-      Assert.assertEquals(2, children.getLength());
```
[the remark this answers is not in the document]
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -212,9 +225,9 @@ public void testRemoveUnusedDependencies_selectAll()
     try (InputStream contents = pomFile.getContents()) {
       Document actual = parse(contents);
       NodeList dependenciesList = actual.getElementsByTagName("dependencies");
-      Assert.assertEquals(1, dependenciesList.getLength());
-      Assert.assertTrue(dependenciesList.item(0) instanceof Element);
-      Element dependencies = (Element) dependenciesList.item(0);
+      Assert.assertEquals(2, dependenciesList.getLength());
+      
+      Element dependencies = (Element) dependenciesList.item(1);
```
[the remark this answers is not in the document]
elharo: added comment

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/PomTest.java
```diff
@@ -286,7 +297,7 @@ public void testResolveLibraries() throws CoreException {
     Library library2 = newLibrary("id2", file1, file2);
 
     pom.updateDependencies(Arrays.asList(library1), Arrays.asList(library1, library2));
-
+
```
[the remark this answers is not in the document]
elharo: Google style is officially not to worry about trailing whitespace

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/testdata/testpom.xml
elharo: dependencyManagement

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.Set;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+import org.w3c.dom.Element;
+
+/**
+ *
```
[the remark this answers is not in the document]
elharo: I expect it's going to become more generic. For now it's package private.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
elharo: yes

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
elharo: not necessary

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Bom.java
elharo: Good point. Let me think about that.
FYI, right now in prod the code is broken with any dependencyManagement section including ours.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
elharo: since we can return a status here, that's safer. But your'e right, it shouldn't happen.

plugins/com.google.cloud.tools.eclipse.appengine.newproject.test/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplatesTest.java
elharo: No, we're not. We're getting the first top-level dependencies element.

chanseokoh:
Things seem to work fine. Only one minor problem:

selection_008

"Duplicating managed version 4.12 for junit" in `pom.xml`. BOM seems to pull in the JUnit version too. Removing the `<version>` element clears the warning. Not sure why the BOM exposes a test dependency version. Will there be a way to exclude it?

elharo:
The ultimate fix to the JUnit warning is to expand the dependency checking code to actually query the BOM. For now, though, it's just a low priority warning with no actual effect on the user's code.

patflynn:
I don't see junit in the
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-bom/pom.xml
(and wouldn't expect it there)

elharo:
The BOM pulls in transitive dependencies, not just the immediate declarations. It's a lot broader and deeper than it looks.

chanseokoh:
Yeah, it pulls in `guava` too. The `<version>` element is not needed when importing the BOM.

patflynn:
Did you guys trace which dep is pulling in junit?  We should file a bug.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
```diff
@@ -0,0 +1,100 @@
...
+    Node rootElement = doc.getDocumentElement();
+
+    dependencyManagement =
+        doc.createElementNS("http://maven.apache.org/POM/4.0.0", "dependencyManagement");
+    Element dependencies = doc.createElementNS("http://maven.apache.org/POM/4.0.0", "dependencies");
+
+    Element springDependency =
+        configureBom(doc, "org.springframework.boot", "spring-boot-dependencies", "2.0.0.RELEASE");
+    dependencies.appendChild(springDependency);
+    Element cloudDependency = configureBom(doc, "com.google.cloud", "google-cloud", "0.40.0-alpha");
```
briandealwis: Aha. This should be `google-cloud-bom`.

chanseokoh: > Did you guys trace which dep is pulling in junit?  We should file a bug.

@patflynn turns out we were not pulling in a BOM.

PR #2961, opened 2018-03-21
Downgrade GCP Status logging severity

briandealwis:
This change dials down the logging severity level of the retrieved GCP status as INFO is too high.


chanseokoh:
Can you merge with master? I expect the TPD diff to disappear.

briandealwis:
I also wondered about having the GCP Status → logging severity (i.e., GCP down would be logged as an error).  But this log message is just something that's put into the UI, and it's the usual case.  So I'll remove it entirely.

PR #2964, opened 2018-03-22
Add "All Rights Reserved."

### LINKED ISSUE #2916: Google LLC
New Google standard is that this takes the form 

```
 * Copyright 2016 Google LLC. All Rights Reserved.
```

That is,

* Google LLC, not Google Inc.

https://g3doc.corp.google.com/company/teams/opensource/copyright.md?cl=head

briandealwis:
Fixes #2916 to add "All Rights Reserved."  There are three commits:
  1. the first amends the templates
  2. the second fixes existing usages
  3. the third replaces all existing Google Inc. with the new text

We can back out (3) but since the language has changed, the change can be deemed to be significant, even if it's of [questionable effect](https://en.wikipedia.org/wiki/All_rights_reserved).


elharo:
NVM. There's conflicting internal documentation about whether All Rights Reserved is required or not. I kicked this up the chain, and the official response is we shouldn't use "All Rights Reserved".

PR #2973, opened 2018-03-25
get rid of prerequisites in pom.xml templates

### LINKED ISSUE #2966: Remove prerequisites from generated pom.xml in new project wizards
Built a CT4E created Maven project at command line and got this:

```
[WARNING] The project c.g:fg:war:0.1.0-SNAPSHOT uses prerequisites 
which is only intended for maven-plugin projects but not for non maven-plugin projects. 
For such purposes you should use the maven-enforcer-plugin. 
See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
```

elharo:
@briandealwis fix #2966


chanseokoh:
I think it's good to have a comment about what this thing does, why it's there, and if it's optional. It's quite long and could look arcane to users new to Maven.

PR #2975, opened 2018-03-25
Query the Boms

### LINKED ISSUE #2971: Query the BOM
That is, rather than simply recognizing what we expect to be in the google-cloud-java BOM, actually look at the dependencies declared by the actual BOMs (can be more than one).

elharo:
fix #2971 
@chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
@@ -27,7 +27,7 @@ public class BomTest {
   
   @Before
   public void setUp() throws CoreException {
-    cloudBom = Bom.loadBom("com.google.cloud", "google-cloud", "0.41.0-alpha", null);
+    cloudBom = Bom.loadBom("com.google.cloud", "google-cloud-bom", "0.41.0-alpha", null);
   }
 
   @Test
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
@@ -16,85 +16,43 @@
 
 package com.google.cloud.tools.eclipse.appengine.libraries;
 
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
+import org.eclipse.core.runtime.CoreException;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
-import org.w3c.dom.DOMImplementation;
-import org.w3c.dom.Document;
-import org.w3c.dom.Element;
-import org.w3c.dom.Node;
 
 public class BomTest {
 
-  private Element dependencyManagement;
+  private Bom cloudBom;
   
   @Before
-  public void setUp() throws ParserConfigurationException {
-    DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
-    factory.setNamespaceAware(true);
-    DocumentBuilder builder = factory.newDocumentBuilder();
-    DOMImplementation impl = builder.getDOMImplementation();
-
-    Document doc = impl.createDocument("http://maven.apache.org/POM/4.0.0", "project", null);
-
-    Node rootElement = doc.getDocumentElement();
-
-    dependencyManagement =
-        doc.createElementNS("http://maven.apache.org/POM/4.0.0", "dependencyManagement");
-    Element dependencies = doc.createElementNS("http://maven.apache.org/POM/4.0.0", "dependencies");
-
-    Element springDependency =
-        configureBom(doc, "org.springframework.boot", "spring-boot-dependencies", "2.0.0.RELEASE");
-    dependencies.appendChild(springDependency);
-    Element cloudDependency = configureBom(doc, "com.google.cloud", "google-cloud", "0.40.0-alpha");
-    dependencies.appendChild(cloudDependency);
-    dependencyManagement.appendChild(dependencies);
-    rootElement.appendChild(dependencyManagement);
-  }
-
-  private Element configureBom(Document doc, String groupId, String artifactId, String version) {
-    Element dependency = doc.createElementNS("http://maven.apache.org/POM/4.0.0", "dependency");
-    Element groupIdElement = doc.createElementNS("http://maven.apache.org/POM/4.0.0", "groupId");
-    Element artifactIdElement =
-        doc.createElementNS("http://maven.apache.org/POM/4.0.0", "artifactId");
-    Element versionElement = doc.createElementNS("http://maven.apache.org/POM/4.0.0", "version");
-
-    groupIdElement.setTextContent(groupId);
-    artifactIdElement.setTextContent(artifactId);
-    versionElement.setTextContent(version);
-
-    dependency.appendChild(groupIdElement);
-    dependency.appendChild(artifactIdElement);
-    dependency.appendChild(versionElement);
-    return dependency;
+  public void setUp() throws CoreException {
+    cloudBom = Bom.loadBom("com.google.cloud", "google-cloud", "0.41.0-alpha", null);
   }
 
   @Test
   public void testCloudSpeech() {
-    Assert.assertTrue(Bom.defines(dependencyManagement, "com.google.cloud", "google-cloud-speech"));
+    Assert.assertTrue(cloudBom.defines("com.google.cloud", "google-cloud-speech"));
   }
 
   @Test
   public void testGroupdId() {
-    Assert.assertFalse(Bom.defines(dependencyManagement, "com.google", "google-cloud-speech"));
+    Assert.assertFalse(cloudBom.defines("com.google", "google-cloud-speech"));
   }
   
   @Test
   public void testCloudVision() {
-    Assert.assertTrue(Bom.defines(dependencyManagement, "com.google.cloud", "google-cloud-vision"));
+    Assert.assertTrue(cloudBom.defines("com.google.cloud", "google-cloud-vision"));
   }
   
   @Test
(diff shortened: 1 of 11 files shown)


elharo:
For some reason this doesn't work for google-cloud-dlp but works for others. I don't know why.

chanseokoh:
Apparently, the 0.41.0-alpha BOM doesn't have `google-cloud-dlp` defined in its `<dependencyManagement>` section. I don't know what that means though.

elharo:
I see google-cloud-dlp in the BOM at HEAD. There's also something about v2 though. And Eclipse does show the warning here: "Duplicating managed version 0.41.0-beta for google-cloud-dlp"

elharo:
This looks suspicious. There are two dlp dependencies in the BOM:

```
<dependency>
<artifactId>google-cloud-dlp</artifactId>
<version>${dlp.version}</version>
</dependency>
<dependency>
<artifactId>google-cloud-dlp</artifactId>
<version>${dlp.version}</version>
<type>test-jar</type>
</dependency>
```

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
```diff
@@ -16,85 +16,43 @@
...
-    groupIdElement.setTextContent(groupId);
-    artifactIdElement.setTextContent(artifactId);
-    versionElement.setTextContent(version);
-
-    dependency.appendChild(groupIdElement);
-    dependency.appendChild(artifactIdElement);
-    dependency.appendChild(versionElement);
-    return dependency;
+  public void setUp() throws CoreException {
+    cloudBom = Bom.loadBom("com.google.cloud", "google-cloud", "0.41.0-alpha", null);
```
briandealwis: Should this not be `com.google.cloud:google-cloud-bom`? (note trailing `-bom`)

elharo: good catch. There is a google-cloud too but I don't know what that is.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -235,10 +259,10 @@ void updateDependencies(Collection<Library> selectedLibraries,
    */
```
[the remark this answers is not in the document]
elharo: I think not yet. Worth exploring.

elharo:
https://maven.apache.org/resolver/apidocs/org/eclipse/aether/collection/DependencyCollectionContext.html#getManagedDependencies() looks promising

elharo:
Try `RepositorySystem.readArtifactDescriptor(RepositorySystemSession, ArtifactDescriptorRequest).getManagedDependencies()`

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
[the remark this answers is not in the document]
elharo: Haven't tested but that should work now since this is all handled by Maven resolver.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
```diff
@@ -16,85 +16,43 @@
...
   @Test
   public void testUnknown() {
-    Assert.assertFalse(Bom.defines(dependencyManagement, "com.google.cloud", "unknown"));
+    Assert.assertFalse(cloudBom.defines("com.google.cloud", "unknown"));
   }
 
   @Test
-  public void testGuava() {
-    Assert.assertFalse(Bom.defines(dependencyManagement, "com.google.guava", "guava"));
+  public void testGuava() { // transitive dependency
```
chanseokoh: Remove the comment.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -87,8 +88,31 @@ static Pom parse(IFile pomFile) throws SAXException, IOException, CoreException
       DocumentBuilder builder = builderFactory.newDocumentBuilder();
       Document document = builder.parse(pomFile.getContents());
       Pom pom = new Pom(document, pomFile);
+      
+      XPath xpath = xpathFactory.newXPath();
+      xpath.setNamespaceContext(new Maven4NamespaceContext());
+
+      NodeList bomNodes = (NodeList) xpath.evaluate(
+          "//m:dependencyManagement/m:dependencies/m:dependency",
```
chanseokoh: Actually, I just got this question. Is everything in this section is actually a BOM?

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
```diff
@@ -57,7 +60,8 @@
       String groupId, String artifactId, String version, IProgressMonitor monitor)
           throws CoreException {
 
-    final Artifact artifact = new DefaultArtifact(groupId + ":" + artifactId + ":" + version);
+    final Artifact artifact =
```
chanseokoh: Seems like `final` is unnecessary starting with Java 8.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
```diff
@@ -108,4 +112,44 @@
     repositories.add(repository);
     return repositories;
   }
+
+  public static Collection<Dependency> getManagedDependencies(String groupId, String artifactId,
+      String version, IProgressMonitor monitor) throws CoreException {
+    
+    // todo we'd prefer not to depend on m2e here
+    final Artifact artifact =
```
chanseokoh: Seems like `final` is unnecessary starting with Java 8.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BomTest.java
elharo: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
elharo: Every dependency element

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
elharo: done

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
elharo: done

chanseokoh:
My idea is to change
```java
    ICallable<List<Artifact>> callable = new ICallable<List<Artifact>>() {
      @Override
      public List<Artifact> call(IMavenExecutionContext context, IProgressMonitor monitor)
          throws CoreException {
        DependencyFilter filter = DependencyFilterUtils.classpathFilter(JavaScopes.RUNTIME);
```
to
```
    ICallable<List<Artifact>> callable = (executionContext, unusedMonitor) -> {
```

Also, `Artifact artifact = new DefaultArtifact(groupId + ":" + artifactId + ":" + version)` can be pushed inside to the `callable`.

Lastly, `IMavenExecutionContext context = MavenPlugin.getMaven().createExecutionContext()` can be moved down just before the `return` statement.

chanseokoh:
Another cleanup is to change
```java
        List<Dependency> managedDependencies;
        try {
          managedDependencies = system.readArtifactDescriptor(session, request).getManagedDependencies();
        } catch (ArtifactDescriptorException ex) {
          ...
        }
        
        return managedDependencies;
 ```
to
```java
        try {
          List<Dependency> managedDependencies =
              system.readArtifactDescriptor(session, request).getManagedDependencies();
          return managedDependencies;
        } catch (ArtifactDescriptorException ex) {
          ...
        }
```
These apply to all the three methods in `DependencyResolver`, actually.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
briandealwis: BOMs are [`<type>pom</type><scope>import</scope>`](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Importing_Dependencies).
> *import* (only available in Maven 2.0.9 or later)
> This scope is only supported on a dependency of type pom in the <dependencyManagement> section. It indicates the dependency to be replaced with the effective list of dependencies in the specified POM's <dependencyManagement> section. Since they are replaced, dependencies with a scope of import do not actually participate in limiting the transitivity of a dependency.

The non-import dependencies matter, but they aren't BOMs.  For example, [we define a dependencyManagement section](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/pom.xml#L159) in our pom.xml.

chanseokoh: Okay, sounds like we need to check the type and the scope before passing it to `Bom.loadBom()`.

PR #2977, opened 2018-03-26
google-cloud --> google-cloud-bom

elharo:
@briandealwis @chanseokoh fix BOM artifact ID until I figure out how to read dependencyManagement element with Aether


chanseokoh:
There are test failures. Some libraries seem not defined in the BOM?

chanseokoh:
I see. This won't expose JUnit, Guava, etc. Makes sense.

elharo:
kokoro appears hung but tests are passing

PR #2987, opened 2018-03-28
Add script to copy Kokoro-built repo

chanseokoh:
I tried Python this time. I'm completely new to Python.

This is a post-build script, to copy a Kokoro-built repo to a final, permanent GCS location (e.g., `gs://cloud-tools-for-eclipse/1.6.1`).

What it does is actually simple:
1. Asks the GCS URL of a Kokoro-built URL.
1. Asks the CT4E version (for the name of a final GCS location).
1. Halts if the destination directory already exists.
1. Copy the repo.
1. Make the repo publicly accessible.

### DIFF
--- a/release/copy_repo_to_final_location.py
+++ b/release/copy_repo_to_final_location.py
@@ -22,7 +22,7 @@ def _AskRepoOrigin():
   print
   print "#"
   print "# Enter the GCS URL of the Kokoro-built repo."
-  print '# ("Artficat location" in the "jar_signing" success email.)'
+  print '# ("Artifact location" in the "jar_signing" success email.)'
   url = raw_input("URL? ")
 
   gcs_url = _FormatGcsUrl(url)
--- a/release/copy_repo_to_final_location.py
+++ b/release/copy_repo_to_final_location.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python2
+
+"""Copies a Kokoro-built CT4E release to a permenant location and makes it
+publicly available.
+
+The GCS location of a Kokoro-built repo is not suitable for direct release or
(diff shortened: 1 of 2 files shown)


release/copy_repo_to_final_location.py
```diff
@@ -0,0 +1,98 @@
...
+
+import re
+import subprocess
+import sys
+
+
+def _AskRepoOrigin():
+  print
+  print "#"
+  print "# Enter the Cloud Console GCS URL of the Kokoro-built repo."
```
elharo: delete "Cloud Console"

release/copy_repo_to_final_location.py
```diff
@@ -0,0 +1,98 @@
...
+  else:
+    print "Wrong URL. Try again."
+    return _AskRepoOrigin()
+
+
+def _AskVersion():
+  print
+  print "#"
+  print "# Enter the CT4E version (e.g., 9.9.9)."
+  print '# (The name for a subdirectory of "gs://cloud-tools-for-eclipse/'
```
elharo: You don't need these three lines. 43-45

chanseokoh:
> Can we reorganize this as a library whose individual functions can be unit tested?

I'm new to Python, so what exactly should be done? I defined some methods, and looks like anyone can already import and call them.

```
$ python
Python 2.7.13 (default, Nov 24 2017, 17:33:09) 
[GCC 6.3.0 20170516] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import copy_repo_to_final_location
>>> copy_repo_to_final_location._GcsLocationExists("gs://cloud-tools-for-eclipse/1.6.0")
True
```

PR #2988, opened 2018-03-28
Remove obsolete flag

chanseokoh:
I removed `-DenableFlexPropertySettingsPage=false` from the launch config. Other changes are auto-regeneration.


chanseokoh:
Reverted other changes. It's always safe to keep them unchanged.

PR #2993, opened 2018-04-02
replace deprecated method

elharo:
@chanseokoh


chanseokoh:
I remember this. Unfortunately, compilation on neon fails.

elharo:
Ah, the replacement for the deprecated method probably only exists in Oxygen. :-(

PR #2991, opened 2018-03-30
appengine-plugins-core 0.5.2

elharo:

PR #2997, opened 2018-04-02
Track managed SDK vs. non-managed SDK usage

### LINKED ISSUE #2983: Track usage of self-CloudSDK vs managed-SDK on local run and deploy

chanseokoh:
(Branched off of #2991, so it will become cleaner when it is merged.)

Fixes #2983.

I decided to send this info when deploying and running locally, since they are the only two cases where the Cloud SDK is used.

I found it is not readily possible to write a test. `CloudSdkPreferences.isAutoManaging()` is a static method from another class that reads Eclipse preferences. If necessary, I could try tweaking the interface, but it may not look nice.

Did some actual testing (canary):

selection_001

selection_003

selection_002


plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsEvents.java
```diff
@@ -40,10 +40,11 @@
   // Metadata keys
   public static final String APP_ENGINE_LOCAL_SERVER_MODE = "mode";
   public static final String APP_ENGINE_NEW_PROJECT_WIZARD_TYPE = "type";
-  public static final String APP_ENGINE_NEW_PROJECT_WIZARD_BUILD_TOOL = "build_tool";
+  public static final String APP_ENGINE_NEW_PROJECT_WIZARD_BUILD_TOOL = "build.tool";
```
chanseokoh: I think it's OK to change this. We haven't set up a graph for this yet.

chanseokoh: Oh, actually, [this was merged 13 days ago](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2955/files#diff-90f045e3a2a727c3315674098de0cabcR43) and [1.6.0B was tagged 26 days ago](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/releases/tag/v1.6.0B), so we can freely change this.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsEvents.java
```diff
@@ -55,4 +56,7 @@
 
   public static final String NATIVE_PROJECT = "native.project";
   public static final String MAVEN_PROJECT = "maven.project";
+
+  public static final String MANAGED_CLOUD_SDK = "managed";
+  public static final String USER_CLOUD_SDK = "user.managed";
```
briandealwis: It would make more sense if these followed the `CloudSdkManagementOption` `MANUAL` vs `AUTOMATIC`.

PR #3011, opened 2018-04-06
Call getResolvedClasspath repeatedly

chanseokoh:
#2996. Hope this fixes it.


chanseokoh:
So we think https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2996#issuecomment-379330728 is fairly plausible. Makes sense so far, so closing this workaround.

PR #3013, opened 2018-04-06
Disable library list check listener in test

### LINKED ISSUE #2758: Travis Flake in BuildPathTest.testCheckLibraries
WARNING: Project no longer exists: test0.5965265269703218
Tests run: 4, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.98 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.libraries.BuildPathTest
testAddNativeLibrary(com.google.cloud.tools.eclipse.appengine.libraries.BuildPathTest)  Time elapsed: 1.556 sec
testCheckLibraries(com.google.cloud.tools.eclipse.appengine.libraries.BuildPathTest)  Time elapsed: 0.761 sec  <<< FAILURE!
	at org.junit.Assert.assertFalse(Assert.java:64)
	at org.junit.Assert.assertFalse(Assert.java:74)
	at com.google.cloud.tools.eclipse.appengine.libraries.BuildPathTest.testCheckLibraries(BuildPathTest.java:110)

chanseokoh:
Fixes #2758.

The listener spawns a workspace job that calls `BuildPath.checkLibraryList()`, which is the method of interest in this test. Removing the listener seems to fix it. Running `testCheckLibraries()` thousand times in a loop doesn't seem to fail after this change.

Also attempts to fix the flakyness in #2996 (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2996#issuecomment-379330728).

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPathTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/BuildPathTest.java
@@ -52,6 +52,9 @@ public class BuildPathTest {
 
   @Before
   public void setUp() throws JavaModelException {
+    // https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2758
+    Activator.removeLibraryListCheckListener();
+
     project = projectCreator.getJavaProject();
     initialClasspathSize = project.getRawClasspath().length;
     assertNull(BuildPath.findMasterContainer(project));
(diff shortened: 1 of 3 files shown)


briandealwis:
Hmm, maybe that Activator listener should be much more focused: if there's no master-container, but the _library.container file exists, then remove it.  That can be done in-place.

chanseokoh:
Okay, let me try that. I was actually thinking if I had to move `Activator.removeLibraryListCheckListener()` to `setUp()`.

chanseokoh:
We are already deleting the _library.container file only when there is no master container in the class path. Nonetheless, such checking is useless because there is no synchronization.

chanseokoh:
After giving some thoughts, I think the best way to handle this concurrency is to run the operations in the listener in the main thread (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2758#issuecomment-379786162). I opened #3014. It's therefore natural to close this PR, but I'm marking it as blocked in case we choose this path.

PR #3017, opened 2018-04-10
Add utility for NamespaceContext for arbitrary prefix-namespace mapping

### LINKED ISSUE #2965: General NamespaceContext utility class
Something that allows binding arbitrary prefixes to URLs. We have one offs in several places.

chanseokoh:
Fixes #2965.


chanseokoh:
Flaky test failure on Kokoro (#3020). Re-triggering with an empty commit.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowMavenModel.java
```diff
@@ -193,29 +194,9 @@ public void trackDataflowDependency(IProgressMonitor monitor) throws CoreExcepti
...
-      public String getPrefix(String namespaceURI) {
-        throw new UnsupportedOperationException();
-      }
-
-      @Override
-      public Iterator<?> getPrefixes(String namespaceURI) {
-        throw new UnsupportedOperationException();
-      }
-    };
+    static final NamespaceContext POM_NS_CONTEXT = new MappedNamespaceContext(ImmutableMap.of(
```
elharo: Why is this visible for testing?

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContextTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+  @Test
+  public void testConstructor_nullPrefix() {
+    try {
+      Map<String, String> map = new HashMap<>();
+      map.put(null, "scheme://host/path/");
+      new MappedNamespaceContext(map);
+      fail();
+    } catch (IllegalArgumentException ex) {
+      assertEquals("prefix and URI can't be null", ex.getMessage());
```
elharo: delete "and URI"

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContextTest.java
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  @Test
+  public void testConstructor_singleMappingNullUri() {
+    try {
+      new MappedNamespaceContext("p", null);
+      fail();
+    } catch (IllegalArgumentException ex) {
+      assertEquals("prefix and URI can't be null", ex.getMessage());
```
elharo: delete prefix and

plugins/com.google.cloud.tools.eclipse.util.test/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContextTest.java
```diff
@@ -0,0 +1,174 @@
...
+          "prefix", "scheme://multiple/prefixes/",
+          "maven", "http://maven.apache.org/POM/4.0.0"));
+
+  @Test
+  public void testConstructor_singleMappingNullPrefix() {
+    try {
+      new MappedNamespaceContext(null, "scheme://host/path/");
+      fail();
+    } catch (IllegalArgumentException ex) {
+      assertEquals("prefix and URI can't be null", ex.getMessage());
```
elharo: delete and URI; and adjust code accordingly

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.xml.XMLConstants;
+import javax.xml.namespace.NamespaceContext;
+
+public class MappedNamespaceContext implements NamespaceContext {
+
+  private final Map<String, String> prefixMapping;
+
+  public MappedNamespaceContext(String prefix, String namespaceUri) {
+    if (prefix == null || namespaceUri == null) {
+      throw new IllegalArgumentException("prefix and URI can't be null");
```
elharo: be specific about which is null

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.xml.XMLConstants;
+import javax.xml.namespace.NamespaceContext;
+
+public class MappedNamespaceContext implements NamespaceContext {
+
+  private final Map<String, String> prefixMapping;
```
elharo: The xml prefix needs to be unconditionally and irrevocably mapped to http://www.w3.org/XML/1998/namespace

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public MappedNamespaceContext(String prefix, String namespaceUri) {
+    if (prefix == null || namespaceUri == null) {
+      throw new IllegalArgumentException("prefix and URI can't be null");
+    }
+    prefixMapping = new HashMap<>();
+    prefixMapping.put(prefix, namespaceUri);
+  }
+
+  public MappedNamespaceContext(Map<String, String> prefixMapping) {
```
elharo: do we really need this constructor? We could instead add a method to add a prefix and a URI.

chanseokoh: There's a trade-off then. Not having a mutable method makes the class truly immutable (unless you go for the builder pattern). I'm find either way. Go for a simple method to add a prefix and a URI?

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowMavenModel.java
chanseokoh: `DataflowMavenModelTest` has been making use of it. I also added tests to check the `NamespaceContext` itself, as it was already `@VisibleForTesting`.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    this.prefixMapping.put("xml", XMLConstants.XML_NS_URI);
+    this.prefixMapping.putAll(prefixMapping);
+  }
+
+  private static void checkMapping(String prefix, String namespaceUri) {
+    if (prefix == null) {
+      throw new IllegalArgumentException("Prefix can't be null");
+    } else if (prefix.equals("xml")) {
+      throw new IllegalArgumentException("Cannot redefine the 'xml' prefix");
```
elharo: This is only true if the namespace URI is not http://www.w3.org/XML/1998/namespace

plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowMavenModelTest.java
```diff
@@ -171,7 +171,7 @@ public void testXpathExpression() throws Exception {
         new ByteArrayInputStream(simplifiedModel.getBytes()));
 
     XPath realXpath = XPathFactory.newInstance().newXPath();
-    realXpath.setNamespaceContext(DataflowMavenModelFactory.POM_NS_CONTEXT);
+    realXpath.setNamespaceContext(DataflowMavenModelFactory.pomNamespaceContext);
```
elharo: we should be able to use a separate namespace context object here; no real need to reach into the private parts

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -29,28 +28,30 @@
...
 
-  public MappedNamespaceContext(Map<String, String> prefixMapping) {
-    for (Entry<String, String> entry : prefixMapping.entrySet()) {
-      checkMapping(entry.getKey(), entry.getValue());
-    }
-    this.prefixMapping.put("xml", XMLConstants.XML_NS_URI);
-    this.prefixMapping.putAll(prefixMapping);
+  public MappedNamespaceContext(String prefix, String namespaceUri) {
+    this();
+    addMapping(prefix, namespaceUri);
```
elharo: It's a bit dangerous to call a public-non-final method from a constructor. Perhaps share a private method between them.

plugins/com.google.cloud.tools.eclipse.dataflow.core.test/src/com/google/cloud/tools/eclipse/dataflow/core/project/DataflowMavenModelTest.java
chanseokoh: I made the field private. I removed three tests too as a result, but those are the ones that I added in this PR.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,86 @@
...
+    prefixMapping.put("xml", XMLConstants.XML_NS_URI);
+  }
+
+  public MappedNamespaceContext(String prefix, String namespaceUri) {
+    this();
+    addMappingInternal(prefix, namespaceUri);
+  }
+
+  public MappedNamespaceContext addMapping(String prefix, String namespaceUri) {
+    return addMappingInternal(prefix, namespaceUri);
```
elharo: nit: I'd `return this` here rather than in addMappingInternal

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,86 @@
...
+  public MappedNamespaceContext() {
+    prefixMapping.put("xml", XMLConstants.XML_NS_URI);
+  }
+
+  public MappedNamespaceContext(String prefix, String namespaceUri) {
+    this();
+    addMappingInternal(prefix, namespaceUri);
+  }
+
+  public MappedNamespaceContext addMapping(String prefix, String namespaceUri) {
```
elharo: addMapping --> declareNamespace for consistency with terminology in the namespaces specification.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import javax.xml.XMLConstants;
+import javax.xml.namespace.NamespaceContext;
+
+public class MappedNamespaceContext implements NamespaceContext {
+
+  private final Map<String, String> prefixMapping = new HashMap<>();
```
elharo: prefixMap or perhaps namespaceMap

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/MappedNamespaceContext.java
```diff
@@ -0,0 +1,86 @@
...
+  public MappedNamespaceContext(String prefix, String namespaceUri) {
+    this();
+    addMappingInternal(prefix, namespaceUri);
+  }
+
+  public MappedNamespaceContext addMapping(String prefix, String namespaceUri) {
+    return addMappingInternal(prefix, namespaceUri);
+  }
+
+  private MappedNamespaceContext addMappingInternal(String prefix, String namespaceUri) {
```
elharo: One more special case to handle:

The prefix xmlns is used only to declare namespace bindings and is by definition bound to the namespace name http://www.w3.org/2000/xmlns/. It MUST NOT be declared . Other prefixes MUST NOT be bound to this namespace name, and it MUST NOT be declared as the default namespace. Element names MUST NOT have the prefix xmlns.

except

The prefix xmlns is used only for namespace bindings and is not itself bound to any namespace name.

That is, this rule changed between editions of the spec. I think we should throw an IllegalArgumentException if xmlns is passed to this method as a prefix.

PR #3018, opened 2018-04-10
Cleanup code

chanseokoh:


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/internal/AppYamlValidatorTest.java
```diff
@@ -42,7 +42,7 @@
 
   @Rule public TemporaryFolder tempFolder = new TemporaryFolder();
 
-  @Mock private IObservableValue appYamlPath;
+  @Mock private IObservableValue<String> appYamlPath;
```
elharo: I think this one breaks Mars compatibility, which is OK as long as it works in Neon.

chanseokoh: I believe we already broke Mars compatibility here and there. I never looked back once we dropped Mars.

Issue #3016, opened 2018-04-09
Investigate suitable version scheme for nightly builds


chanseokoh:
If we can somehow assign a working version to each nightly build in the way that the version is final and suitable for immediate and direct release, without any concern of confusing users or the Eclipse update system when released into the wild. Otherwise, we still need to trigger a manual cut only because we need to assign a working version.

Also, it is important to make versions between nightly builds work properly when it comes to automatic updates.

chanseokoh:
Fortunately, the current builds seem to already work well. The actual feature versions of CT4E that OGSi care about is already appended with a timestamp like "1.6.1.201804231645", so binaries built later are naturally recognized as updates. I've tried installing nightly versions on our official 1.6.1 through both the directly install URL and the Eclipse auto-update menu item, and they work well. So, we are almost there.

The only minor issue with the nightly builds is our static "branding" version we put into `gcp-repo/metadata.product`. For example, we currently put "1.6.1" after removing ".qualifier" . Perhaps we could just use, say, "nightly-after-1.6.1" for that "branding" version.

chanseokoh:
Oh, I made an interesting observation. The timestamp being appended to the version is the time of the last commit of the build, not the time of the build.

briandealwis:
Using git commit timestamp means that artifacts that haven't changed will have the same version and so p2 won't re-download them.  Tycho's timestamp provider is smart and also checks whether a referenced bundles/features have changed for features and products.

It might be worth configuring a different timestamp format for our non-feature/bundles.  For example, for Eclipse we use an `I` prefix for integration builds (e.g., `I20180404-2000`), or `N` for nightly builds (which we don't do any more).

chanseokoh:
Setting up nightly builds complete. Versions of nightlies will be like `1.6.1.201805010358-post-release-nightly`, which can be mixed naturally with regular versions. Updates from/to nightlies/regulars will work seamlessly.

PR #3034, opened 2018-04-24
Allow adding custom prefix to branding version

chanseokoh:
Regarding #3016.

My plan is to prefixing `nightly-after-` to the CT4E branding version for automated nightly builds, so that it looks like `nightly-after-1.6.1`. This prefixing is sort of optional, but I think it's good to avoid possible confusion.


chanseokoh:
Seems like the version should be in the proper format. Probably I should add a qualifier. Closing this until I get there.

PR #3035, opened 2018-04-25
Allow adding custom qualifier to branding version

### LINKED ISSUE #3038: Add build postfix only to the CT4E feature version
#3036 allows putting an arbitrary postfix to bundle versions, but it adds it to every bundle version. This works OK for now, but it seems better to add it only to the CT4E feature version.

Related: #3016

chanseokoh:
Regarding #3016. Fixes #3038.

My plan is to add `-post-release-nightly` as a suffix to the CT4E branding version and the feature version for automated nightly builds. The version will look like `1.6.1.201804026xxxx-post-release-nightly`. It is sort of optional to append such a suffix to nightlies, but I think it's good to avoid potential confusion in the future.


kokoro/ubuntu/new_release.sh
```diff
@@ -29,10 +30,14 @@ cd git/google-cloud-eclipse
 #
 # https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2363#issuecomment-327844378
 # https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2211
-readonly CT4E_DISPLAY_VERSION=$( \
-  xmlstarlet sel -t -v '/product/@version' gcp-repo/metadata.product )
+readonly CT4E_DISPLAY_VERSION="$( \
```
briandealwis: Might be worth a comment on what this is doing: retrieve the product version, which we assume either has a `.qualifier` or no version qualifier, and append the provided build qualifier (if any).

kokoro/ubuntu/new_release.sh
```diff
@@ -29,10 +30,14 @@ cd git/google-cloud-eclipse
...
+  xmlstarlet sel -t -v '/product/@version' gcp-repo/metadata.product )"
 readonly NO_QUALIFIER_VERSION="${CT4E_DISPLAY_VERSION%.qualifier}"
```
briandealwis: What should happen if we don't have a .qualifier?  This is presumably an error. Should we error?

chanseokoh: We always have ".qualifier" at the end in `gcp-repo/metadata.product`. But if it wasn't there, we would already be happy, and moreover, probably it makes sense to not touch the current value in that case (which is the behavior of the current code).

chanseokoh:
https://wiki.eclipse.org/Tycho/FAQ#How_do_I_embed_a_configurable_build_version.3F

chanseokoh:
Blocked for now. #3036 seems a better way.

chanseokoh:
My last commit removes the trickery of the automated removal of the `.qualifier` from `gcp-repo/metadata.product`, which became obsolete after #3036. Also renamed "postfix" to "suffix", which is shorter.

chanseokoh:
Unblocked. @briandealwis

kokoro/ubuntu/new_release.sh
```diff
@@ -23,24 +24,19 @@ test -n "${ANALYTICS_TRACKING_ID}"
...
-# https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2363#issuecomment-327844378
-# https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/2211
-readonly CT4E_DISPLAY_VERSION=$( \
-  xmlstarlet sel -t -v '/product/@version' gcp-repo/metadata.product )
-readonly NO_QUALIFIER_VERSION="${CT4E_DISPLAY_VERSION%.qualifier}"
-xmlstarlet ed --inplace -u '/product/@version' -v "${NO_QUALIFIER_VERSION}" \
-  gcp-repo/metadata.product
+VERSION_SUFFIX_PROPERTY=
+if [ "${PRODUCT_VERSION_SUFFIX}" ]; then
+  VERSION_SUFFIX_PROPERTY="'${PRODUCT_VERSION_SUFFIX}'"
```
briandealwis: Worth a comment here that the single-quotes are necessary so its interpretation as a date format in the `<format>` expression.

pom.xml
```diff
@@ -38,7 +38,7 @@
   <properties>
     <tycho.version>1.1.0</tycho.version>
     <tycho-extras.version>1.0.0</tycho-extras.version>
-    <product.version.qualifier.postfix/>  <!-- 0-length string by default -->
+    <product.version.qualifier.suffix/>  <!-- 0-length string by default -->
```
briandealwis: Drop `product.` since this is applying to all artifacts?

kokoro/ubuntu/new_release.sh
```diff
@@ -23,24 +24,19 @@ test -n "${ANALYTICS_TRACKING_ID}"
...
+fi
 
 # Need to unset `TMPDIR` for `xvfb-run` due to a bug:
 # https://bugs.launchpad.net/ubuntu/+source/xorg-server/+bug/972324
 TMPDIR= xvfb-run \
-  mvn -V -B -Doauth.client.id="${OAUTH_CLIENT_ID}" \
-            -Doauth.client.secret="${OAUTH_CLIENT_SECRET}" \
-            -Dga.tracking.id="${ANALYTICS_TRACKING_ID}" \
+  mvn -V -B \
+      -Dproduct.version.qualifier.suffix="${VERSION_SUFFIX_PROPERTY}" \
```
briandealwis: I think you could reduce this change to a one-liner with: `${PRODUCT_VERSION_SUFFIX:+-Dproduct.version.qualifier.suffix="'${PRODUCT_VERSION_SUFFIX}'"`.

chanseokoh: Ah, that's nice.

pom.xml
chanseokoh: Will fix #3038 soon, so `product` will be OK?

briandealwis: Sure

chanseokoh: Decided to fix #3038 here.
- Removed `<format>` from the top-level `pom.xml`.
- Added `<format>` to `gcp-repo/metadata.product`.
- Added `<format>` to `features/com.google.cloud.tools.eclipse.suite.feature/pom.xml`.

It works well. Only the feature version and the product branding version have the suffix when set.

features/com.google.cloud.tools.eclipse.suite.feature/pom.xml
```diff
@@ -13,6 +13,13 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-packaging-plugin</artifactId>
+        <configuration>
```
briandealwis: Is the feature version still based on the feature last-modification time? (I ask as I'm not sure the `<dependencies>` and `<timestampProvider>` are inherited?)

PR #3036, opened 2018-04-25
Configurable build version

chanseokoh:
Regarding #3016.

Initially I put the format configuration only in `gcp-repo/pom.xml`, which added the a postfix only to the display (product) version. I plan to add `-post-release-nightly` so that it would look like `1.6.1.201804251802-post-release-nightly`. However, I thought it is actually nice to put such a postfix to the feature version too, which will clearly show us if the installed version is a nightly build, e.g., in the bug reported from our menu. So I added it to the top-level `pom.xml`.

Things work naturally and properly when it comes to upgrading, as the postfix is the last item to compare. In terms of comparing versions, `1.6.1.201804251802-post-release-nightly` comes after `1.6.1.201804251802`, which matches my plan to build `1.6.1.201804251802-post-release-nightly` after releasing `1.6.1.201804251802`.

When not providing the postfix property, there would be no change to the current version scheme.

### DIFF
--- a/gcp-repo/pom.xml
+++ b/gcp-repo/pom.xml
@@ -13,8 +13,19 @@
   <version>0.1.0-SNAPSHOT</version>
   <packaging>eclipse-repository</packaging>
 
+  <properties>
+	<product.version.qualifier.prefix/>  <!-- 0-length string by default -->
+  </properties>
+
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-packaging-plugin</artifactId>
+        <configuration>
+          <format>${product.version.qualifier.prefix}yyyyMMddHHmm</format>
+        </configuration>
+      </plugin>
       <plugin>
         <groupId>org.eclipse.tycho</groupId>
         <artifactId>tycho-p2-repository-plugin</artifactId>
--- a/gcp-repo/pom.xml
+++ b/gcp-repo/pom.xml
@@ -13,19 +13,8 @@
   <version>0.1.0-SNAPSHOT</version>
   <packaging>eclipse-repository</packaging>
 
-  <properties>
-	<product.version.qualifier.prefix/>  <!-- 0-length string by default -->
-  </properties>
-
   <build>
     <plugins>
-      <plugin>
-        <groupId>org.eclipse.tycho</groupId>
-        <artifactId>tycho-packaging-plugin</artifactId>
-        <configuration>
-          <format>${product.version.qualifier.prefix}yyyyMMddHHmm</format>
-        </configuration>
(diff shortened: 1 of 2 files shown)


briandealwis:
This change should affect all build artifacts, including the bundles.  Since the timestamps are still from git commits, a user that installs a nightly and then updates to the next release may get warnings for bundles that haven't changed since the nightly as the qualifier will cause the nightly to seem _later_ than the release bundle.

briandealwis:
I guess if we used a suffix for both nightlies and releases, and the release suffix was lexicographically greater then we'd be ok.

chanseokoh:
I confirm that all our bundles also get the extra qualifier, as you said. However, interestingly enough, I did not get any warning when upgrading from a nightly build to the next non-postfix release build, and things seem to work properly. Below is what I did locally for testing:

1. Built `1.6.1.201804231645-post-release-nightly` locally (with `mvn -Dproduct.version.qualifier.postfix="'-post-release-nightly'" package`).
1. Installed it on a fresh Eclipse Oxygen (locally from `gcp-repo/target/repository`).
   - The version of the CT4E feature I can see from the _Installation Details_ page is `1.6.1.201804231645-post-release-nightly`, as expected.
   - All of our bundles are `0.1.0.2018xxxxxxxx-post-release-nightly`, whose timestamp is usually earlier than `201804231645`, as expected.
   - For example, the `.deploy` bundle is `0.1.0.201804022008-post-release-nightly`.
1. Bumped the version to `1.6.2` (modified `feature.xml`, `pom.xml`, and `metadata.product`) and committed the change.
1. Built `1.6.2` without a postfix (i.e., a normal release). The full version is `1.6.2.201804252343`, whose timestamp is later than the nightly build, as expected. 
1. Attempted installing it on the previous Eclipse instance. Eclipse correctly says the operation will be an update rather than an installation.
   - Upgrading completed without any warning.
   - The version of the CT4E feature installed is `1.6.2.201804252343` (no postfix).
   - There is no bundle with the postfix `-post-release-nightly`.
   - The `.deploy` bundle is `0.1.0.201804022008` (no postfix). Note that the timestamp didn't change, as expected. Only changed is the postfix that got removed.

briandealwis:
Ah right — the feature will ensure the rebuilt bundles are pulled in.  Thanks for checking.  So the only downside is that we'll download and re-download bundles.  I can live with that.  Let's give it a whirl and see how it works.

chanseokoh:
> So the only downside is that we'll download and re-download bundles. I can live with that.

Yeah, we should be good anyway, as we don't really expect users to install nightlies (at least for now). We are basically considering the nightlies for our internal usage. However, it'd be nice to eventually move to a release modal where can just promote a nightly to a public release easily.

PR #3039, opened 2018-04-27
more namespacing

elharo:
fix #2390 I think all other uses of these methods are for non-namespaced files like plugin.xml


chanseokoh:
The test failure seems related.

PR #3044, opened 2018-05-02
Allow setting service account key for Dataflow runs

### LINKED ISSUE #3027: Allow setting service account key or GOOGLE_APPLICATION_CREDENTIALS for Dataflow runs
In the distant past, the Dataflow plugin relied on the gcloud authentication, but we manged to wire up our CT4E authentication: #2002.

The actual implementation was to use `GOOGLE_APPLICATION_CREDENTIALS` and make it point to a CT4E OAuth2 login credential. Because we set the env var for our exclusive use for this purpose, we currently show a warning to users that "you cannot set the variable (inside Eclipse) (because we will have to overwrite it to pass a CT4E login credential)."

However, it's not infeasible to allow setting the variable, but this should be done with proper UI change. For example, the current Dataflow run config tab has a combo for selecting a login credential. Perhaps we could say "if you want to set `GOOGLE_APPLICATION_CREDENTIALS`, then do not select a login credential in the combo."

chanseokoh:
Fixes #3027 #3048.

selection_010

The key file is optional and when set, it will be used as the credential, instead of the login credential, for running the Dataflow application.


plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
```diff
@@ -139,6 +139,8 @@ private void setWidgetsEnabled(boolean enabled) {
     values.put(DataflowPreferences.STAGING_LOCATION_PROPERTY, defaultOptions.getStagingLocation());
     // TODO: Give this a separate input
```
elharo: todo still relevant?

chanseokoh:
The UI strings are in `messages.properties` included in this PR.

plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/launcher/DefaultedPipelineOptionsComponent.java
chanseokoh: It is relevant code-wise, although it is sort of a product decision whether we really want a separate input widget for this, or we will just continue to use staging directory as we do now.

chanseokoh:
Still not ready.

chanseokoh:
Still need to add tests.

chanseokoh:
Added tests, but let me see if I can reduce the GitHub diff for the tests.

chanseokoh:
Ready of review.

plugins/com.google.cloud.tools.eclipse.dataflow.core/src/com/google/cloud/tools/eclipse/dataflow/core/launcher/DataflowPipelineLaunchDelegate.java
```diff
@@ -167,7 +167,35 @@ private IProject getProject(ILaunchConfiguration configuration) throws CoreExcep
...
+  }
+
+  private void setServiceAccountCredential(ILaunchConfigurationWorkingCopy workingCopy,
+      String serviceAccountKey) throws CoreException {
+    Path jsonKey = Paths.get(serviceAccountKey);
+    if (Files.isDirectory(jsonKey)) {
+      String message = "Not a file but directory: " + jsonKey;
+      throw new CoreException(new Status(Status.ERROR, DataflowCorePlugin.PLUGIN_ID, message));
+    } else if (!Files.isReadable(jsonKey)) {
+      String message = "Cannot open a service account key file: " + jsonKey;
```
elharo: open a service account --> open service account

PR #3045, opened 2018-05-03
New JSP dependency

### LINKED ISSUE #3004: JSP 2.3.0 or 2.3.1
Creating a Java 8 Maven project with the wizard uses JSP 2.3.1.
Creating a Java 8 native project with the wizard uses JSP 2.3.0.

Which one is right? i.e. provided by the server? That's the one both types of project should use.

elharo:
fixes #3004 

@chanseokoh @briandealwis @ludoch This is the same JSP dependency used by the dev_appserver. I'm still trying to learn what version is used in prod. 

There are other fixes yet to be done, particularly around flex, but I would like some early feedback and figure out which tests this may break.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.flex.ftl
+++ b/plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.flex.ftl
@@ -40,9 +40,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>javax.servlet.jsp</groupId>
-      <artifactId>javax.servlet.jsp-api</artifactId>
-      <version>2.3.1</version>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>apache-jsp</artifactId>
+      <version>9.4.8.v20171121</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/features/com.google.cloud.tools.eclipse.suite.feature/pom.xml
+++ b/features/com.google.cloud.tools.eclipse.suite.feature/pom.xml
@@ -14,6 +14,13 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-packaging-plugin</artifactId>
+        <configuration>
+          <format>yyyyMMddHHmm${product.version.qualifier.suffix}</format>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-antrun-plugin</artifactId>
         <executions>
--- a/gcp-repo/pom.xml
+++ b/gcp-repo/pom.xml
@@ -17,6 +17,13 @@
     <plugins>
       <plugin>
         <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-packaging-plugin</artifactId>
+        <configuration>
+          <format>yyyyMMddHHmm${product.version.qualifier.suffix}</format>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
         <artifactId>tycho-p2-repository-plugin</artifactId>
         <version>${tycho.version}</version>
         <configuration>
--- a/kokoro/ubuntu/new_release.sh
+++ b/kokoro/ubuntu/new_release.sh
@@ -15,6 +15,7 @@ gcloud components install app-engine-java --quiet
 echo "OAUTH_CLIENT_ID: ${OAUTH_CLIENT_ID}"
 echo "OAUTH_CLIENT_SECRET: ${OAUTH_CLIENT_SECRET}"
 echo "ANALYTICS_TRACKING_ID: ${ANALYTICS_TRACKING_ID}"
+echo "PRODUCT_VERSION_SUFFIX: "${PRODUCT_VERSION_SUFFIX}
 
 # Exit if undefined (zero-length).
 test -n "${OAUTH_CLIENT_ID}"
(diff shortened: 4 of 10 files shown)


chanseokoh:
Now I think that, Jetty (and other container engines like Tomcat) would not expose classes that does not make sense to be unnecessarily exposed to user apps, if the engine has a sane design. I think it's a real possibility that user apps are deliberately run in an isolated context and the apps only see the standard Servlet and JSP interfaces. That is, no matter which implementation code the engine uses internally, such code is invisible and won't slip into user apps.

I did a small test of listing all available classes under the `org.apache` package in the dev appserver and on prod.

In the dev appsever:
```
class org.apache.taglibs.standard.util.XmlUtil$JstlEntityResolver
class org.apache.jasper.compiler.Node$JspOutput
class org.apache.tomcat.util.descriptor.web.ContextResourceEnvRef
class org.apache.taglibs.standard.tag.common.core.ParamSupport$ParamManager
class org.apache.jasper.compiler.ELInterpreterFactory$DefaultELInterpreter
class org.apache.jasper.runtime.PageContextImpl$4
class org.apache.jasper.servlet.JspServlet$1
class org.apache.taglibs.standard.lang.jstl.test.beans.PrivateBean2b
class org.apache.el.parser.AstLessThanEqual
interface org.apache.tomcat.PeriodicEventListener
interface org.apache.jasper.runtime.JspSourceDependent
class org.apache.tomcat.util.descriptor.tld.TldRuleSet
class org.apache.jasper.compiler.Node$ScriptingElement
class org.apache.tomcat.util.descriptor.web.SetOverrideRule
class org.apache.jasper.compiler.TagFileProcessor$TagFileLoaderVisitor
class org.apache.taglibs.standard.lang.jstl.BinaryOperator
class org.apache.taglibs.standard.lang.jstl.NullLiteral
class org.apache.jasper.compiler.Node$NamedAttribute$1AttributeVisitor
class org.apache.jasper.compiler.Node$ForwardAction
interface org.apache.tomcat.util.scan.Jar
...
```

On the other hand, on prod, only available to user apps is the `org.apache.taglibs.standard` package; there is no other package under `org.apache` than that:
```
class org.apache.taglibs.standard.extra.spath.SPathParser$JJCalls
class org.apache.taglibs.standard.tei.Util
class org.apache.taglibs.standard.lang.jstl.AndOperator
class org.apache.taglibs.standard.extra.spath.SPathParser
class org.apache.taglibs.standard.lang.jstl.BinaryOperatorExpression
class org.apache.taglibs.standard.lang.jstl.IntegerLiteral
class org.apache.taglibs.standard.lang.jstl.test.StaticFunctionTests
class org.apache.taglibs.standard.tag.common.core.ForEachSupport$SimpleForEachIterator
class org.apache.taglibs.standard.tlv.JstlXmlTLV$1
class org.apache.taglibs.standard.extra.spath.Predicate
class org.apache.taglibs.standard.lang.jstl.test.beans.PrivateBean2c
class org.apache.taglibs.standard.lang.jstl.Expression
...
```

BTW, the code I used to list is
```java
    Reflections reflections = new Reflections("org.apache", new SubTypesScanner(false));
    Set<Class<?>> subTypes = reflections.getSubTypesOf(Object.class);
    for (Class<?> subType : subTypes) response.getWriter().println(subType);
```
using the following library
```xml
    <dependency>
      <groupId>org.reflections</groupId>
      <artifactId>reflections</artifactId>
      <version>0.9.11</version>
    </dependency>
```

elharo:
If there is some such white listing happening, I want to push the App Engine team to document and commit to that. 

Note that we don't have a choice about exposing this in the IDE classpath. The only choice we can make is which jar we pull in; but we have to pull in something. Beware of status quo bias. The one we are pulling in today is by no means guaranteed to be more correct than this one.

chanseokoh:
Hmm... yeah, maybe it's whilelisting on prod, because those classes are still available on the dev appserver?

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/pom.xml.flex.ftl
```diff
@@ -40,9 +40,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>javax.servlet.jsp</groupId>
-      <artifactId>javax.servlet.jsp-api</artifactId>
-      <version>2.3.1</version>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>apache-jsp</artifactId>
+      <version>9.4.8.v20171121</version>
       <scope>provided</scope>
```
briandealwis: My concern here is that we're now requiring a particular *implementation* version, and thus foisting version maintenance onto the user.

elharo: We were doing that already. The only question is which implementation version we provide, not whether we provide one.

briandealwis: That's not quite true: we were pulling in an API spec jar that also has implementation classes.

With this replacement dependency, the user will be prompted by the `versions-maven-plugin` to update when a new version of Jetty is released.  Should they update?

elharo: Why do believe the previous version was an API spec jar? As best I can follow the various third party web sites, that is not the case. 

Regardless, CT4E is not requiring a particular implementation version. Both the dev_appserver and presumably the prod runtime are using a  particular implementation version. Becuase this jar is provided, CT4E can't change that. It can only match that as closely as possible.

elharo: Flex and standard don't seem to be using the same version of Jetty.  Standard uses 9.3. Flex uses 9.4. The version here for standard is 9.3, and for flex 9.4.

I don't believe there is any such BOM or canonical details from App Engine. I think there should be. I hope to convince the App Engine team they need to do a better job of documenting these details. Right now I'm reverse engineering all of this.

PR #3050, opened 2018-05-07
Do not always enable Dataflow options widgets but enable only when possible

### LINKED ISSUE #2430: Mini project selector and GCS bucket field enabled when using defaults in Dataflow launch config

chanseokoh:
Fixes #2430


plugins/com.google.cloud.tools.eclipse.dataflow.ui/src/com/google/cloud/tools/eclipse/dataflow/ui/preferences/RunOptionsDefaultsComponent.java
```diff
@@ -283,7 +291,7 @@ public void validate() {
       return;
     }
 
-    projectInput.setEnabled(true);
+    projectInput.setEnabled(canEnableChildren);
```
briandealwis: Could be done instead by expanding 234 to:
```
if (selectedCredential == null || !canEnableChildren) {
```

chanseokoh: That would make enablement correct at least, but we still want to validate all the global preference values to show proper validation errors even if the panel is locked from editing (disabled).

Issue #3055, opened 2018-05-11
Exception while loading Cloud Library cache LibraryClasspathContainerResolverService.libraryEntries


chanseokoh:
Can't reproduce it. Just noticed this is logged in my console. Might be just my hosted dev instance have gone borked, but at least I haven't touched code related to this.

The offending precondition violation:
```
                  ISchedulingRule currentRule = Job.getJobManager().currentRule();
                  Preconditions.checkState(
                      currentRule != null && currentRule.contains(MavenUtils.mavenResolvingRule()));
```

```
May 11, 2018 11:55:27 AM com.google.common.cache.LocalCache$Segment$1 run
WARNING: Exception thrown during refresh
java.util.concurrent.ExecutionException: java.lang.IllegalStateException
	at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:502)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:461)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:78)
	at com.google.common.util.concurrent.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:142)
	at com.google.common.cache.LocalCache$Segment.getAndRecordStats(LocalCache.java:2461)
	at com.google.common.cache.LocalCache$Segment$1.run(LocalCache.java:2439)
	at com.google.common.util.concurrent.MoreExecutors$DirectExecutor.execute(MoreExecutors.java:399)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:902)
	at com.google.common.util.concurrent.AbstractFuture.addListener(AbstractFuture.java:636)
	at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.addListener(AbstractFuture.java:100)
	at com.google.common.cache.LocalCache$Segment.loadAsync(LocalCache.java:2434)
	at com.google.common.cache.LocalCache$Segment.refresh(LocalCache.java:2508)
	at com.google.common.cache.LocalCache$Segment.scheduleRefresh(LocalCache.java:2486)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2891)
	at com.google.common.cache.LocalCache.getIfPresent(LocalCache.java:4135)
	at com.google.common.cache.LocalCache$LocalManualCache.getIfPresent(LocalCache.java:5054)
[...]

PR #3059, opened 2018-05-15
LibraryClasspathContainerResolverService cache should expire not refresh

### LINKED ISSUE #3005: CloudLibrariesPage should prevent creating multiple containers
We don't want to allow having multiple _Google Cloud Libraries_ classpath containers on a project.  Since JDT does not provide any mechanism to control creating new containers, then we should change `CloudLibrariesPage` to detect the situation and create an error page and prevent the creation/editing of the new container.

Basically our `#initialize()` is called with the project's current classpath entries.  `setSelection()` will be called with `null` to indicate a new container.  We should then check in `createControl()` if there's a current container.

briandealwis:
Fixes #3005

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
@@ -87,7 +87,7 @@ public class LibraryClasspathContainerResolverService
   /** Cached set of JDT classpath entries for Cloud Libraries. */
   private final LoadingCache<String, IClasspathEntry[]> libraryEntries =
       CacheBuilder.newBuilder()
-          .expireAfterWrite(10, TimeUnit.MINUTES)
+          .expireAfterAccess(10, TimeUnit.MINUTES)
           .build(
               new CacheLoader<String, IClasspathEntry[]>() {
                 @Override
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/repository/LibraryClasspathContainerResolverService.java
@@ -87,7 +87,7 @@ public class LibraryClasspathContainerResolverService
   /** Cached set of JDT classpath entries for Cloud Libraries. */
   private final LoadingCache<String, IClasspathEntry[]> libraryEntries =
       CacheBuilder.newBuilder()
-          .refreshAfterWrite(10, TimeUnit.MINUTES)
+          .expireAfterWrite(10, TimeUnit.MINUTES)
           .build(
               new CacheLoader<String, IClasspathEntry[]>() {
                 @Override


chanseokoh:
This doesn't seem right... The library container may disappear after the timeout.

1. Reduce the timeout to, say, 5 seconds for testing.
1. Start with a fresh, empty workspace.
1. Create an App Engine standard project.
1. Expand the project and note it has the "App Engine Standard Runtime" library container.
1. Wait for 10 seconds (sufficiently more than 5 seconds).
1. While selecting the project, hit F5 to refresh it.

Notice that the "App Engine Standard Runtime" container disappears. If you edit `HelloAppEngine.java` already open in the editor at all, you get build errors due to the Servlet library missing.

chanseokoh:
Perhaps the problem is unrelated to the cache expiration change but rather due to a legitimate pitfall of the current code.

My unconfirmed theory: `ServletClasspathProvider.resolveClasspathContainer()` always returns null while deferring the actual container update through the future callback mechanism. However, if the job fired from `LibraryClasspathContainerResolverService.resolveLibraryAttachSources()` completes really fast and the callback is executed immediately when calling `Futures.addCallback()`, you end up calling `requestClasspathContainerUpdate()` only to return null later (i.e., clear the container after all).

chanseokoh:
Well, I think the theory could still hold, but seems like there is actually a different reason than that for the problem.

chanseokoh:
Maybe there is a race condition too: I can see that the caller of `resolveClasspathContainer()` is the main thread, while the thread executing `requestClasspathContainerUpdate()` in the callback is a worker thread, which is the Job thread.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -100,6 +100,19 @@ public void onFailure(Throwable t) {
     return null;
   }
 
+  /** Request that a project's runtime server container be updated. */
+  private void requestClasspathContainerUpdate(
```
chanseokoh: nit: could be `static`?

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
```diff
@@ -100,6 +100,19 @@ public void onFailure(Throwable t) {
...
+      IProject project, IRuntime runtime, IClasspathEntry[] entries) {
+    /* Updating a container is done by either explicitly requesting an update from a
+     * {@code ClasspathContainerInitializer#requestClasspathContainerUpdate()} or by calling {@code
+     * JavaCore.setClasspathContainer()}. Both require knowledge of how the WTP Server Runtime
+     * containers are named. But {@code resolveClasspathContainerImpl()}'s implementation calls our
+     * {@code resolveClasspathContainer()} and calls {@code JavaCore.setClasspathContainer()} if the
+     * classpath entries change. */
```
chanseokoh: Honestly, I cannot get what the comment is trying to say, particularly "both require knowledge of how the WTP Server Runtime containers are named".

I think it's a good idea to add a link to your comment https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/3055#issuecomment-390242592 too.

briandealwis: Will rework the comment and link to this issue.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/ServletClasspathProvider.java
briandealwis: It calls instance methods.

PR #3078, opened 2018-05-22
Objectify 6.0

### LINKED ISSUE #3073: Objectify 6.0
Objectify 6.0 is out. Should we recommend this version to users in our wizards? Does it work on App Engine?

https://github.com/objectify/objectify/wiki/UpgradeVersion5ToVersion6

The release notes seem to reference Cloud SDK. I'm unclear why this matters to Objectify.

Objectify v6 does work with Flexible, and we seem to be picking up 6.0 automatically anyway. If we do want version 5, we have to pin the version.

elharo:
fix #3073 @briandealwis @chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -50,14 +50,13 @@
           tooltip="%objectify.tooltip"
           dependencies="include"
           siteUri="https://github.com/objectify/objectify/wiki" >
-      <libraryDependency id="appengine-api" />
       <libraryFile
           export="true"
-          javadocUri="https://www.javadoc.io/doc/com.googlecode.objectify/objectify/5.1.22">
+          javadocUri="https://www.javadoc.io/doc/com.googlecode.objectify/objectify/6.0">
         <mavenCoordinates
               artifactId="objectify"
               groupId="com.googlecode.objectify"
-              version="5.1.22" />
+              version="6.0" />
       </libraryFile>
     </library>
 
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -45,7 +45,7 @@
     
     <library
           id="objectify"
-          group="appengine"
+          group="appengine,flexible"
           name="Objectify"
           tooltip="%objectify.tooltip"
           dependencies="include"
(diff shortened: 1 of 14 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineWizardPage.java
```diff
@@ -96,6 +88,13 @@ public void createControl(Composite parent) {
     Dialog.applyDialogFont(container);
   }
 
+  private void configureLibrariesSelectorGroup(Composite container) {
+    appEngineLibrariesSelectorGroup = new LibrarySelectorGroup(container, getSupportedLibraries(),
+        Messages.getString("app.engine.libraries.group")); //$NON-NLS-1$
+  }
+
+  protected abstract String getSupportedLibraries();
```
chanseokoh: Given this returns a `String`, I think it's good to have a Javadoc or change the name to `getSupportedLibrariesGroup`.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -45,19 +45,18 @@
     
     <library
           id="objectify"
-          group="appengine"
+          group="appengine,flexible"
```
chanseokoh: Maybe we should just promote Datastore Cloud client library on flex?

https://cloud.google.com/appengine/docs/flexible/java/migrating#datastore

The doc also says (might be out of date),

> At this time ORM libraries that available in the standard environment such as ndb and Objectify are not supported outside of the standard environment.

elharo: That doc is out of date. 6.0 supports flex: "Objectify v5 and prior use the Google App Engine SDK and therefore can only be used within Google App Engine Standard. Objectify v6+ uses the new Google Cloud SDK and can be used from anywhere - GAE Standard, GAE Flex, GCE, or outside Google Cloud entirely."

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineWizardPage.java
elharo: changed the name

PR #3081, opened 2018-05-22
Fix recursing App Engine content blocks

### LINKED ISSUE #3077: App Engine content block circling indefintely
selection_001

I have an empty `datastore-indexes.xml`.

```xml
<?xml version="1.0" encoding="utf-8"?>
<datastore-indexes autoGenerate="true">
</datastore-indexes>
```

briandealwis:
Software can sometimes be too adaptable.

Unfortunately ditches the `MockWorkspace` as I need to rework the IFile/IFolder protocols to allow creating an `IFacetedProject`: the Resource API supports navigating to files and folders that don't exist (e.g., `if(project.getFile("non/existent/path").exists()) {…}`).  Tests are noticeably slower.

Fixes #3077


chanseokoh:
There is a way to run the tests fast: add the App Engine facet only when needed.

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/compare/i3077-simplify-fast-tests?expand=1

The `IFacetedProject`-to-`IProject` is not a hard requirement. Was trying to see if `IProject` could work here.

chanseokoh:
Oops... wrong link. Here is the right one:

https://github.com/GoogleCloudPlatform/google-cloud-eclipse/compare/i3077...i3077-simplify-fast-tests?expand=1

briandealwis:
Looks good! I merged in your branch and will commit once tests finish.

Issue #2825, opened 2018-02-20
Flaky DeployPropertyPageTest integration test


chanseokoh:
On Neon.

Note that there are two `DeployPropertyPageTest` classes. This one is about the `.integration.appengine` bundle.

```
testPropertyPageTitle_standardProject(com.google.cloud.tools.eclipse.integration.appengine.DeployPropertyPageTest)  Time elapsed: 8.496 sec  <<< ERROR!
org.eclipse.swtbot.swt.finder.exceptions.AssertionFailedException: assertion failed: 
	at org.eclipse.swtbot.swt.finder.utils.internal.Assert.isTrue(Assert.java:95)
	at org.eclipse.swtbot.swt.finder.utils.internal.Assert.isTrue(Assert.java:83)
	at org.eclipse.swtbot.swt.finder.widgets.SWTBotTreeItem.select(SWTBotTreeItem.java:516)
	at org.eclipse.swtbot.swt.finder.widgets.SWTBotTreeItem.select(SWTBotTreeItem.java:543)
	at com.google.cloud.tools.eclipse.integration.appengine.DeployPropertyPageTest.testPropertyPageTitle_standardProject(DeployPropertyPageTest.java:49)
```

The failed `select()` is in the following line:
````
bot.tree().expandNode("Google Cloud Platform").select("App Engine Deployment");
````

PR #3097, opened 2018-05-25
Tighten App Engine content block trigger to only support IProject

### LINKED ISSUE #3089: App Engine Content Block appears twice
image

Once under the project and once under the deployment descriptor

briandealwis:
Add an integration test to verify.

Fixes #3089

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
@@ -292,6 +292,8 @@
      </detector>
   </extension>
   
+  <!-- cannot use com.google.cloud.tools.eclipse.appengine.onlyForStandardProjects for the trigger
+    https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/3089 -->
   <extension
         point="org.eclipse.ui.navigator.navigatorContent">
      <navigatorContent
@@ -304,9 +306,15 @@
            name="%appengineNavigatorContentProvider"
            priority="high">
         <triggerPoints>
-          <reference
-                definitionId="com.google.cloud.tools.eclipse.appengine.onlyForStandardProjects">
-          </reference>
+           <and>
+              <instanceof
+                    value="org.eclipse.core.resources.IProject">
+              </instanceof>
+              <test
+                    property="org.eclipse.wst.common.project.facet.core.projectFacet"
+                    value="com.google.cloud.tools.eclipse.appengine.facets.standard">
+              </test>
+           </and>
         </triggerPoints>
         <possibleChildren>
            <instanceof
--- a/plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotWorkbenchActions.java
+++ b/plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotWorkbenchActions.java
@@ -16,8 +16,12 @@
 
 package com.google.cloud.tools.eclipse.swtbot;
 
+import static org.junit.Assert.assertNotNull;
+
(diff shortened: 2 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotWorkbenchActions.java
```diff
@@ -149,9 +155,23 @@ private static void openPreferencesDialogViaEvents(SWTBot bot) {
...
+  public static SWTBotView openViewById(SWTWorkbenchBot bot, String viewId) {
+    UIThreadRunnable.syncExec(
+        bot.getDisplay(),
+        () -> {
+          IWorkbenchWindow activeWindow = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+          assertNotNull(activeWindow);
+          try {
+            activeWindow.getActivePage().showView(viewId);
+          } catch (PartInitException ex) {
+            logger.log(Level.SEVERE, "Unable to open view " + viewId, ex);
```
elharo: Should the test fail in this case?

briandealwis: It will as the `bot.viewById(viewId)` will fail below since the view won't be found.

briandealwis:
Filed #3099

chanseokoh:
If the root of a project is the only place we will add the content block, I wonder if we can just simplify `AppEngineContentBlock.getProject()` (private method) to only return `IProject` only when `parentElement` itself is `IProject` (or `IFacetedProject` if still required).

briandealwis:
Of course the tests work fine for me locally. Sigh.

briandealwis:
The tests aren't lying: the subsides are "":

travis-12993-12993 1-linux-20180525-2123-plugins-com google cloud tools eclipse integration appengine-screenshots-testappenginestandardjava8 com google cloud tools eclipse integration appengine projectexplorerappengineblocktest

But `mvn verify` succeeds on my MacOS and Linux (docker) images.  I'll see if my Windows VM can reproduce the problem…

briandealwis:
The penny drops: this is #2825/#2829! 🤦‍♂️

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
```diff
@@ -292,6 +292,10 @@
      </detector>
   </extension>
   
+  <!-- 
+    cannot use com.google.cloud.tools.eclipse.appengine.onlyForStandardProjects for the trigger
```
elharo: what file is this in? I couldn't find it.

plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotTreeUtilities.java
```diff
@@ -51,6 +55,35 @@ public boolean test() throws Exception {
     return tree.getAllItems()[0];
   }
 
+  /**
+   * Wait until the given tree item has items, and return the first item.
```
elharo: and --> then

plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml
briandealwis: It's defined in this file [further above](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/105ea90a3b14a914222c00b4dcce26b7f6a35272/plugins/com.google.cloud.tools.eclipse.appengine.facets/plugin.xml#L143)

PR #3117, opened 2018-05-29
Fix bug to wait for all projects

chanseokoh:
Fix a bug in the test util class.


chanseokoh:
`failIfBuildErrors()` is correctly checking all projects (as it calls `getWorkspace().getRoot().getProjects()`).

However, without this new `failBuildErrors(String)`, `failBuildErrors("check any error globally")` ends up calling the varargs signature `failBuildErrors(String, IProject...)` without passing any project, which then does nothing.

briandealwis:
Good find.  `getAllBuildErrors(IProject…)` still suffers this issue.  Perhaps a better alternative would be to change `getAllBuildErrors()` to treat the empty array as all projects?

chanseokoh:
Makes sense. We can then cut off extra overrides that ultimately depends on `getAllBuildErrors()`.

PR #3119, opened 2018-05-30
Push down project retrieval logic

chanseokoh:
@briandealwis per https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3117#issuecomment-393013475

Nothing significant.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
@@ -152,10 +152,10 @@ public class ProjectUtils {
   }
 
   /**
-   * Waits until no error markers are found for the specified projects or any projects in the
-   * workspace if no project is specified. Returns true unless it times out after 300 seconds.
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/project/ProjectUtils.java
```diff
@@ -152,20 +152,9 @@
...
-  public static boolean waitUntilNoBuildErrors() throws CoreException {
-    IProject[] projects = getWorkspace().getRoot().getProjects();
-    return waitUntilNoBuildErrors(projects);
-  }
-
-  /**
-   * Waits until no error markers are found for the projects. Returns true unless it times out after
-   * 300 seconds. Returns false if it times out.
+   * Waits until no error markers are found for the specified projects or any projects in the
+   * workspace if no project is specified. Returns true unless it times out after 300 seconds.
```
briandealwis: I think it may sound better by separating out the if-no-projects-specified as a separate sentence?

> Waits until no error markers are found for the specified projects.  Use all projects in the workspace if {@code projects} is empty.  _Returns {@code true} **when no error markers are found and** {@code false} after time-out of 300 seconds.

PR #3123, opened 2018-05-31
Add Objectify filter when creating new projects

### LINKED ISSUE #3076: Objectify filter in web.xml
This one needs more investigation, especially around different versions of Objectify. However Objectify seems to need to add things to web.xml. Specifically:

  
```
     <filter>
        <filter-name>ObjectifyFilter</filter-name>
        <filter-class>com.googlecode.objectify.ObjectifyFilter</filter-class>
    </filter>
    <filter-mapping>
        <filter-name>ObjectifyFilter</filter-name>
        <url-pattern>/*</url-pattern>
    </filter-mapping>
```

If the user creates a new project and includes Objectify, we should add this automatically.  

See https://github.com/objectify/objectify/wiki/Setup

chanseokoh:
Fixes #3076

The filter is needed for both Objectify 5 and 6. The filter covers all incoming URLs (`/*`).

- Standard Java 7: `<filter> and <filter-mapping>` in `web.xml`
- Standard Java 8 and Flexible: new `ObjectifyWebFilter` class with the `@WebFilter` annotation (no change in `web.xml`)


chanseokoh:
Oops, this doesn't work for flex projects. I'll look into it.

PR #3124, opened 2018-05-31
lead align

### LINKED ISSUE #3121: Right align or left align in dialogs
Consider these two dialogs:

image

image

One of them aligns the labels incorrectly. I think it's the dataflow dialog, but verify with UI guidelines and fix as necessary.

elharo:
fix #3121 

If we want to do this:

image


elharo:
"Cloud Platform project ID" should probably be simply "Project ID" to sync with the dataflow documentation. Maybe we can make "Cloud Storage staging location" shorter too.

PR #3130, opened 2018-06-02
Grey icon and poll more frequently on connection error

### LINKED ISSUE #3110: GCP Status widget should retry more frequently on connection error
We currently refresh the GCP Status widget every 3 minutes or so.  But on startup I often see the widget showing an error status, which then takes a couple of minutes to resolve.

We should refresh more frequently on a connection exception.

briandealwis:
On connection error, show a greyed-out icon and poll every 30 seconds.

Status area:
screen shot 2018-06-01 at 10 25 00 pm

In the GCP menu:
screen shot 2018-06-01 at 10 24 52 pm

Fixes #3110 

Also added some more tests.


briandealwis:
- [ ] rename image to `gcp-disabled`
- [ ] set `gcp-disabled` as the default command image

Issue #3107, opened 2018-05-29
Flakey test: FlexWarStagingDelegateTest#testStage()


briandealwis:
```
Running com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexWarStagingDelegateTest
Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 5.973 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexWarStagingDelegateTest
testStage(com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexWarStagingDelegateTest)  Time elapsed: 2.916 sec  <<< FAILURE!
java.lang.AssertionError: null
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertTrue(Assert.java:52)
	at com.google.cloud.tools.eclipse.appengine.deploy.flex.FlexWarStagingDelegateTest.testStage(FlexWarStagingDelegateTest.java:59)
```

chanseokoh:
We need to see what went wrong by checking the resulting `status` first.

briandealwis:
Drat, the take-screenshot-on-error requires the `SWTBotJunit4ClassRunner`, which isn't enabled.

chanseokoh:
Looks like `WarPublisher.publishWar()` in FlexWarStagingDelegate` (which in turns calls `PublishUtil.publishZip()`) fails to leave the deployable WAR in the destination. I think most of the flakiness like this is due to a timing issue, e.g., publishing too fast while things have not been settled.

```java
  @Override
  protected IPath getDeployArtifact(IPath safeWorkDirectory, IProgressMonitor monitor)
      throws CoreException {
    IPath war = safeWorkDirectory.append("app-to-deploy.war");
    IPath tempDirectory = safeWorkDirectory.append("temp");
    WarPublisher.publishWar(project, war, tempDirectory, monitor);
    return war;
  }
```

briandealwis:
Do you think we have a threading issue?  The `TestProjectCreator`'s `waitForProjects()` should ensure the project has reached a fixpoint.

chanseokoh:
[...]

PR #3133, opened 2018-06-05
@RunWith(SWTBotJunit4ClassRunner) to capture screenshot on testStage() error

briandealwis:


chanseokoh:
Does this (just adding this annotation) magically make the test capture screenshots? Moreover, this is not an integration test, so what I actually see locally is just the "Welcome" screen.

briandealwis:
It captures screenshots on failure, which will then be uploaded to our travis_builds bucket.  I forgot about that dratted welcome screen though.

plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotWorkbenchCaretaker.java
```diff
@@ -0,0 +1,58 @@
...
+
+/** Take screenshot after test failure, and clean up the workbench. */
+public class SwtBotWorkbenchCaretaker extends TestWatcher {
+  private static final Logger logger = Logger.getLogger(SwtBotWorkbenchCaretaker.class.getName());
+
+  @Override
+  protected void starting(Description description) {
+    try {
+      SwtBotWorkbenchActions.closeWelcome(new SWTWorkbenchBot());
+    } catch (WidgetNotFoundException ex) {
```
elharo: In cases like this, I think we usually name the exception variable "ignored" which will turn off some static analysis checks.

plugins/com.google.cloud.tools.eclipse.swtbot/src/com/google/cloud/tools/eclipse/swtbot/SwtBotWorkbenchCaretaker.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.runner.Description;
+
+/** Take screenshot after test failure, and clean up the workbench. */
+public class SwtBotWorkbenchCaretaker extends TestWatcher {
+  private static final Logger logger = Logger.getLogger(SwtBotWorkbenchCaretaker.class.getName());
+
+  @Override
+  protected void starting(Description description) {
+    try {
+      SwtBotWorkbenchActions.closeWelcome(new SWTWorkbenchBot());
```
chanseokoh: Doesn't seem to work on Travis.

BTW, is there a way to wait until the Welcome view is closed? Seems like it is closed some time after the actual test method runs.

briandealwis: I blithely assumed from the log message ([_ Check the error message in the Console View_](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/3107#issuecomment-394708059)) that we had a workbench open here, and so wanted to take a screenshot of the _Package Explorer_ or _Project Explorer_ to see what it showed in the console view. But the workbench isn't open!

chanseokoh: Yeah, to be frank, I was thinking taking a screenshot won't help here.

PR #3134, opened 2018-06-05
Add support for refreshing 

### LINKED ISSUE #3090: App Engine Content Block refresh
1. Create a project with the wizard.
2. Add a cron.xml to the project in WEB-INF/lib
3. Observe no scheduled tasks in the content block.
4. Refresh
5. Scheduled Tasks shows up.
6. Delete cron.xml
7. Scheduled Tasks is still present.
8. Double click scheduled tasks:
9. The marketplace opens?!

image

briandealwis:
Fixes #3090 

This PR adds support for detecting and updating the App Engine content block on file changes.  Some of the changes that must be handled are:

  1. a project becoming or ceasing to be the default service via appengine-web.xml
  2. a configuration file being removed and added to `WEB-INF`
  3. the deployment assembly being rejigged so that different paths are mapped to `WEB-INF`

Summary of changes:
  - causes the AppEngineContentProvider to listen and react to resource changes to detect added, changed, or removed configuration files. This content provider now maintains a map of `IProject` &rarr; model representation, and the model representation is updated on resource changes.
      - also listens for changes to the special `.settings/org.eclipse.wst.common.component` file which describes the _ComponentCore_ virtual file system that maps different files into WEB-INF and beyond.  ComponentCore is an EMF-based model, and although I'm sure I could hook into the EMF notification mechanism, the changes are eventually written out to this file.
  - adds `ResourcesUtils#getAffectedFiles(IResourceChangeDelta)`, a utility method that walks the delta and pulls out the set of `IFile`s that are reported in the change delta.
  - renames the `AppEngineWebDescriptor` to `AppEngineStandardProjectElement`, since it pulls together everything that defines the project
  - moves the population of the model to `AppEngineStandardProjectElement`
  - adds a `#reload()` to the `AppEngineResourceElements` to allow an instance to reload itself — and return a replacement object.
  - moves some utility functions for creating configuration files from `AppEngineContentProviderTest` to a new `ConfigurationFileUtils` class


plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+
+/** Utility methods for creating App Engine configuration files for testing purposes. */
+public class ConfigurationFileUtils {
+
+  /** WebProjectUtil#createFileInWebInf() does not overwrite files. */
```
elharo: should it do so? and/or does this method belong there even with the different behavior?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
```diff
@@ -0,0 +1,94 @@
...
+  public static IFile createInWebInf(IProject project, IPath path, String contents) {
```
elharo: It would be nice if the method names here and in `WebProjectUtil#createFileInWebInf()` were either consistent or more clearly delineated.  E.g. this one could also be createFileInWebInf or createOrReplaceFileInWebInf

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
```diff
@@ -0,0 +1,94 @@
...
+      return WebProjectUtil.createFileInWebInf(
+          project, path, new ByteArrayInputStream(contents.getBytes(StandardCharsets.UTF_8)), null);
+    } catch (CoreException ex) {
+      fail(ex.toString());
+      /*NOTREACHED*/
+      return null;
+    }
+  }
+
+  public static IFile createEmptyCronXml(IProject project) {
```
elharo: If these aren't just for testing, it wouldn't hurt to have XML declarations and line breaks.

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/io/ResourceUtils.java
```diff
@@ -54,5 +56,17 @@ public static void createFolders(IContainer folder, IProgressMonitor monitor)
     }
   }
 
-
+  public static Collection<IFile> getAffectedFiles(IResourceDelta delta) throws CoreException {
+    Collection<IFile> files = new ArrayList<>();
+    delta.accept(
+        _delta -> {
```
elharo: This variable name is confusing, and I think it probably violates Google style.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.common.base.Strings;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+
+/** Utility methods for creating App Engine configuration files for testing purposes. */
```
chanseokoh: If this is for testing (and I see this is used only in testing too), move it to the test util bundle?

And if so, we can remove all the `// $NON-NLS-1$`.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/AppEngineContentProviderTest.java
```diff
@@ -56,65 +61,68 @@
...
-      descriptor = AppEngineDescriptor.parse(contents);
-    }
-    AppEngineWebDescriptor webDescriptor =
-        new AppEngineWebDescriptor(projectCreator.getProject(), appEngineWebXml, descriptor);
-
-    Object[] children = fixture.getChildren(webDescriptor);
+  public void testGetChildren_AppEngineStandardProjectElement_noDefault_cronIgnored()
+      throws AppEngineException {
+    ConfigurationFileUtils.createAppEngineWebXml(
+        projectCreator.getProject(), "service"); // $NON-NLS-1$
```
chanseokoh: Having `// $NON-NLS-1$` in tests seems too excessive.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
chanseokoh: Considering that this is only for testing, I've actually thought about making this dead simple, e.g., to just place the file in a specified location which will work whether the project has the WTP facet or not. [`CloudSdkStagingHelperTest`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/v1.7.0D/plugins/com.google.cloud.tools.eclipse.appengine.deploy.test/src/com/google/cloud/tools/eclipse/appengine/deploy/CloudSdkStagingHelperTest.java#L60) does similar things for testing, and maybe we could reuse code for both (later).

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
briandealwis: It's used primarily to create an `appengine-web.xml` if one doesn't already exist.  I'll change it to take a parameter to indicate whether to overwrite or not.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
briandealwis: They're used for the _Deployment Assembly_ case too.

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
```diff
@@ -69,9 +67,7 @@ private static IFile createFileInWebInf(IProject project, IPath path, String con
           new ByteArrayInputStream(contents.getBytes(StandardCharsets.UTF_8));
       return WebProjectUtil.createFileInWebInf(project, path, stream, true, null);
     } catch (CoreException ex) {
-      fail(ex.toString());
-      /*NOTREACHED*/
-      return null;
+      throw new AssertionError(ex.toString(), ex);
```
elharo: We probably shouldn't be throwing errors in non-test code.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/WebProjectUtil.java
```diff
@@ -89,11 +89,15 @@ public static IFile createFileInWebInf(
...
     IFile file = webInfDir.getFile(filePath);
-    if (overwrite || !file.exists()) {
-      SubMonitor progress = SubMonitor.convert(monitor, 2);
-      ResourceUtils.createFolders(file.getParent(), progress.newChild(1));
-      file.create(contents, true, progress.newChild(1));
+    if (file.exists() && !overwrite) {
+      return file;
     }
+    SubMonitor progress = SubMonitor.convert(monitor, 3);
+    if (overwrite) {
```
elharo: file.exists() && overwrite?

plugins/com.google.cloud.tools.eclipse.appengine.facets.test/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/ConfigurationFileUtils.java
briandealwis: This is a test utility

PR #3142, opened 2018-06-07
Prevent unzipping files outside destination

chanseokoh:
I've been notified of this problem. We've been fine since the zip util is used only in tests, but it's better to fix it correctly from the security point of view.

https://snyk.io/research/zip-slip-vulnerability#java

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ZipUtil.java
+++ b/plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ZipUtil.java
@@ -49,12 +49,20 @@ public class ZipUtil {
     } else if (!destination.isDirectory()) {
       return StatusUtil.error(ZipUtil.class, "Destination is not a directory: " + destination);
     }
+
     try (ZipFile zipFile = new ZipFile(zip)) {
+      String canonicalDestination = destination.getCanonicalPath();
+
       progress.setWorkRemaining(zipFile.size());
       for (Enumeration<? extends ZipEntry> entries = zipFile.entries(); entries.hasMoreElements()
           && !progress.isCanceled();) {
         ZipEntry entry = entries.nextElement();
         File entryLocation = new File(destination, entry.getName());
+        if (!entryLocation.getCanonicalPath().startsWith(canonicalDestination + File.separator)) {
+          return StatusUtil.error(
+              ZipUtil.class, "Blocked unzipping file outside the destination: " + entry.getName());
+        }
+
         if (entry.isDirectory()) {
           if (!entryLocation.exists()) {
             if (!entryLocation.mkdirs()) {


chanseokoh:
> The contents of this zip file have to be hand crafted. Archive creation tools don’t typically allow users to add files with these paths, despite the zip specification allowing it. However, with the right tools, it’s easy to create files with these paths.

Hand-crafting such a zip seems to require quite some effort and research. I think it's not that worth.

Also, writing a test for a test class seems a bit weird. However, I'm also fixing the zip classes in `appengine-plugins-core`, I'll have some tests there at least.

chanseokoh:
Aha: https://github.com/snyk/zip-slip-vulnerability

chanseokoh:
Turns out the samples are not ordinary archives. The usual `tar` and `zip` utilities as well as the Java libraries we use to extract them identifies them as invalid archive.

```
java.util.zip.ZipException: archive is not a ZIP archive
```

```
java.io.IOException: Error detected parsing the header
Caused by: java.lang.IllegalArgumentException: Invalid byte 34 at offset 0 in '" href="' len=8
```

```
$ tar ztvf zip-slip.tar.gz 
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
```

```
$ zip -Tv zip-slip.zip 
        zip warning: missing end signature--probably not a zip file (did you
        zip warning: remember to use binary mode when you transferred it?)
        zip warning: (if you are trying to read a damaged archive try -F)

zip error: Zip file structure invalid (zip-slip.zip)
```

chanseokoh:
Since this `ZipUtil` is under the `.test.util` bundle, I'm going to merge this.

briandealwis:
Hmm, the zip works for me:
```
$ wget -q https://github.com/snyk/zip-slip-vulnerability/raw/master/archives/zip-slip-win.zip
$ unzip -v zip-slip-win.zip 
Archive:  zip-slip-win.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
      19  Stored       19   0% 04-15-2018 15:04 f34f6f0f  good.txt
      20  Stored       20   0% 04-15-2018 22:04 397b4160  ..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\..\Temp\evil.txt
--------          -------  ---                            -------
      39               39   0%                            2 files
```

chanseokoh:
Ah, yeah, looks like downloading the files with Chrome must have saved them as text files: https://github.com/snyk/zip-slip-vulnerability/blob/master/archives/zip-slip.tar

chanseokoh:
Yeah, the vulnerability is real. Extracting the sample zip from our code will place a file at `/tmp/evail.txt`, outside a extraction destination.

chanseokoh:
@koto CC'ed per Bug 109861633

PR #3147, opened 2018-06-08
Update to Photon RC3

briandealwis:
- pulls out common dependencies into `eclipse/common-dependencies.tpd`
- uses latest SWTBot to ensure consistency across all supported platforms

### DIFF
--- a/eclipse/common-dependencies.tpd
+++ b/eclipse/common-dependencies.tpd
@@ -1,0 +1,23 @@
+/*
+ * Common Dependencies for the different platform Target Platform Dependencies.
+ */
+target "Cloud Tools for Eclipse: Common Dependencies"
+
+location "http://download.eclipse.org/tools/orbit/downloads/latest-R/" {
+  org.hamcrest.core [1.3.0,1.3.1)
+  org.hamcrest.integration [1.3.0,1.3.1)
+  org.hamcrest.library [1.3.0,1.3.1)
+  org.slf4j.api
+  org.slf4j.apis.jcl
+  org.slf4j.apis.log4j
+  org.slf4j.bridge.jul
+  ch.qos.logback.slf4j
+}
+
+location "http://dadacoalition.org/yedit/" {
+  org.dadacoalition.yedit.feature.feature.group
+}
+
+location "http://download.eclipse.org/technology/swtbot/milestones/photon-m7/" {
+  org.eclipse.swtbot.eclipse.feature.group
+}
--- a/eclipse/common-dependencies.tpd
+++ b/eclipse/common-dependencies.tpd
@@ -18,6 +18,8 @@ location "http://dadacoalition.org/yedit/" {
   org.dadacoalition.yedit.feature.feature.group
 }
 
-location "http://download.eclipse.org/technology/swtbot/milestones/photon-m7/" {
+// SWTBot 2.7.0 will be released with Photon, and supports Neon, Oxygen, and Photon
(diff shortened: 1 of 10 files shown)


eclipse/common-dependencies.tpd
```diff
@@ -0,0 +1,23 @@
...
+  org.slf4j.apis.log4j
+  org.slf4j.bridge.jul
+  ch.qos.logback.slf4j
+}
+
+location "http://dadacoalition.org/yedit/" {
+  org.dadacoalition.yedit.feature.feature.group
+}
+
+location "http://download.eclipse.org/technology/swtbot/milestones/photon-m7/" {
```
chanseokoh: Maybe a comment why we have a Photon-specific location in the common file.

eclipse/common-dependencies.tpd
```diff
@@ -0,0 +1,23 @@
+/*
+ * Common Dependencies for the different platform Target Platform Dependencies.
+ */
+target "Cloud Tools for Eclipse: Common Dependencies"
```
chanseokoh: Is the file will be included as-is? If so, is it OK to have multiple `target`s defined?

briandealwis: Yeah it's fine.  The .target files have all been generated from this.  I have another tweak coming to simplify the generation.

pom.xml
```diff
@@ -33,6 +33,7 @@
     <module>features</module>
     <module>plugins</module>
     <module>third_party</module>
+    <module>eclipse</module>
```
chanseokoh: So the difference will be that now we build `eclipse/neon`, `eclipse/oxygen`, and `eclipse/photon` by default at the top level (of course except `eclipse/ide-target-platform`)?

pom.xml
```diff
@@ -479,9 +474,6 @@
          <jettyMaxVersion>9.5</jettyMaxVersion>
       </properties>
       <!-- build against a known target platform -->
```
chanseokoh: Remove the comment together (also for the other two profiles).

That said, I wonder how these profiles will know and select which target platform to use.

pom.xml
briandealwis: We were already pulling in these three projects when importing with our previous instructions.  They will still be found and selected by default under the new instructions, though we will now say to unselect them.

pom.xml
briandealwis: That's done in the the next set of line for each of those, and why the comment is still relevant.

pom.xml
briandealwis: I should note that pulling in `eclipse/{neon,oxygen,photon}` by default adds a negligible overhead.  It's `eclipse/ide-target-platform` that takes a while:
```
[INFO] gcp-eclipse-neon 4.6.0-SNAPSHOT .................... SUCCESS [  0.003 s]
[INFO] gcp-eclipse-oxygen 4.7.0-SNAPSHOT .................. SUCCESS [  0.002 s]
[INFO] gcp-eclipse-photon 4.8.0-SNAPSHOT .................. SUCCESS [  0.002 s]
[INFO] ide-target-platform.repo ........................... SUCCESS [01:02 min]
[INFO] eclipse-setup 0.1.0-SNAPSHOT ....................... SUCCESS [  0.002 s]
```

briandealwis:
Ideally I'd like to have a way to tell M2Eclipse to not import the subprojects of `eclipse`, but I wasn't able to find any method with a quick search.

PR #3151, opened 2018-06-11
Prefer active window for opt-in dialog

chanseokoh:
@briandealwis I remember the original code was basically what you suggested. Now it looks to me it's better to flip the order to prefer an active window and use the workbench window as a last resort. At least this takes care of one case I just encountered where some dialog was open and the opt-in dialog is unresponsive. WDYT?


chanseokoh:
The situation you hit before was one case, and we fixed it by giving an explicit shell. If we flipped the order like in this PR, we could have avoided the problem.

The case I encountered is from #3150, when the SDK preference page was open. When clicking _Apply_ (leaving the dialog open), it triggers the SDK modify job, and pings form the job thread doesn't pass in any shell.

PR #3160, opened 2018-06-13
Update misleading variable names and string constants

chanseokoh:
Every variable name was blindly and confusingly `error`. Likewise, it was always `test error message`.

Besides that, no behavioral change.


briandealwis:
A definite improvement!

PR #3164, opened 2018-06-14
Propagate deploy artifact publish error

chanseokoh:
Related to #3107. We've been ignoring any possible error status from publishing failures. This will surface any failures on going.


plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -80,7 +81,10 @@ public static void publishWar(IProject project, IPath destination, IPath safeWor
 
     IModuleResource[] resources =
         flattenResources(project, safeWorkDirectory, subMonitor.newChild(10));
-    PublishUtil.publishZip(resources, destination, subMonitor.newChild(90));
+    if (resources.length == 0) {
+      logger.log(Level.WARNING, "no resource to publish");
```
elharo: If this is an issue should we return an error status immediately and not call publishZip?

chanseokoh: Makes sense. There must be something to publish all the time.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -58,14 +60,19 @@ public static void publishExploded(IProject project, IPath destination,
...
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
-    subMonitor.setTaskName(Messages.getString("task.name.publish.war"));
+    subMonitor.setTaskName(Messages.getString("task.name.publish.war")); //$NON-NLS-1$
 
     IModuleResource[] resources =
         flattenResources(project, safeWorkDirectory, subMonitor.newChild(10));
-    PublishUtil.publishFull(resources, destination, subMonitor.newChild(90));
+    if (resources.length == 0) {
+      IStatus error = StatusUtil.error(WarPublisher.class, project.getName()
+          + " has no resource to publish"); //$NON-NLS-1$
```
briandealwis: I think "resource" would be plural here?

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
```diff
@@ -76,11 +83,16 @@ public static void publishWar(IProject project, IPath destination, IPath safeWor
...
     SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
-    subMonitor.setTaskName(Messages.getString("task.name.publish.war"));
+    subMonitor.setTaskName(Messages.getString("task.name.publish.war")); //$NON-NLS-1$
 
     IModuleResource[] resources =
         flattenResources(project, safeWorkDirectory, subMonitor.newChild(10));
-    PublishUtil.publishZip(resources, destination, subMonitor.newChild(90));
+    if (resources.length == 0) {
+      IStatus error = StatusUtil.error(WarPublisher.class, project.getName()
+          + " has no resource to publish"); //$NON-NLS-1$
```
briandealwis: Here too (if you accept the premise)

plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/status/StatusUtil.java
```diff
@@ -74,6 +74,14 @@ public static MultiStatus multi(Object origin, String message, Throwable error)
     return new MultiStatus(getBundleId(origin), 0, message, error);
   }
 
+  public static MultiStatus multi(Object origin, String message, IStatus[] statuses) {
+    MultiStatus multiStatus = multi(origin, message);
+    for (IStatus status : statuses) {
```
briandealwis: Do you think it's worth filtering out OK statuses?  See `filter()` further below.  Otherwise they are reported in the UI as `OK`.

chanseokoh: Makes sense. I've added a check (`!isOK()`) in `FlexWarStagingDelegate` and `StandardStagingDelegate` to propagate return the multi-status only when it is not OK.

For this util method, I don't think we want to filter OK statuses. It's a general method to construct a multi-status from the given array of statues and not for error handing.

plugins/com.google.cloud.tools.eclipse.appengine.deploy/src/com/google/cloud/tools/eclipse/appengine/deploy/WarPublisher.java
chanseokoh: Thanks. I've used this as an opportunity to educate myself on this topic. I think I've conclusively understood the usage.

PR #3168, opened 2018-06-18
Update README.md

parkjam4:
Adding tags to the URL to track how many users come to the cloud.google.com pages from GitHub.


parkjam4:
Yes, I wanted to keep the URL as short as possible, so I just dropped Cloud.

PR #3169, opened 2018-06-18
Migrate bucket and remove unused m2-settings.xml

chanseokoh:
Migrating the m2 repo bucket for internal cleanup.


chanseokoh:
Verified Kokoro has no problem downloading and using the tarball in the Kokoro build log. We're good.

Issue #2867, opened 2018-02-27
Flaky DebugNativeAppEngineStandardProjectTest.testDebugLaunch()


chanseokoh:
On Neon.

We had seen flaky behaviors of this test method (e.g., #2741) and fixed them, but this time, it's a new one at

```java
    SwtBotTestingUtilities.waitUntilStyledTextContains(bot,
        "Module instance default is running at http://localhost:8080", consoleContents);
```

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 85.551 sec <<< FAILURE! - in com.google.cloud.tools.eclipse.integration.appengine.DebugNativeAppEngineStandardProjectTest

testDebugLaunch(com.google.cloud.tools.eclipse.integration.appengine.DebugNativeAppEngineStandardProjectTest)  Time elapsed: 84.926 sec  <<< ERROR!

org.eclipse.swtbot.swt.finder.widgets.TimeoutException: Timeout after: 30000 ms.: Text not found!
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:522)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:496)
	at org.eclipse.swtbot.swt.finder.SWTBotFactory.waitUntil(SWTBotFactory.java:484)
	at com.google.cloud.tools.eclipse.swtbot.SwtBotTestingUtilities.waitUntilStyledTextContains(SwtBotTestingUtilities.java:158)
	at com.google.cloud.tools.eclipse.integration.appengine.DebugNativeAppEngineStandardProjectTest.testDebugLaunch(DebugNativeAppEngineStandardProjectTest.java:127)
```

This may or may not be related to the error that logged just before the above stacktrace.
```
WARNING: Error terminating server: http://localhost:8080/_ah/admin/quit responded with Service Unavailable.

com.google.cloud.tools.appengine.api.AppEngineException: http://localhost:8080/_ah/admin/quit responded with Service Unavailable.
	at com.google.cloud.tools.appengine.cloudsdk.CloudSdkAppEngineDevServer1.stop(CloudSdkAppEngineDevServer1.java:176)
	at com.google.cloud.tools.eclipse.appengine.localserver.server.LocalAppEngineServerBehaviour.stop(LocalAppEngineServerBehaviour.java:150)
[...]

PR #3174, opened 2018-06-19
Loop on Run As menu

briandealwis:
An attempt at #2867?


plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/DebugNativeAppEngineStandardProjectTest.java
```diff
@@ -89,11 +91,25 @@ public void testDebugLaunch() throws Exception {
...
     assertNotNull(testProject);
-    SwtBotTestingUtilities.performAndWaitForWindowChange(bot, () -> {
-      System.out.println("==== menu items of Run > Debug As ====");
-      bot.menu("Run").menu("Debug As").menuItems().forEach(System.out::println);
-      bot.menu("Run").menu("Debug As").menu("1 Debug on Server").click();
-    });
+    SwtBotTestingUtilities.performAndWaitForWindowChange(
+        bot,
+        () -> {
+          bot.waitUntil(
```
chanseokoh: I think it's more readable to call this outside the runnable, before `performAndWaitForWindowChange()`. Also conceptually, the wait is an extra step that is independent of taking the click action.

briandealwis: The key here is that we perform some action (Run As > Run On Server) and then wait for the window change (the dialog to appear).  If we perform the action outside then the dialog may appear before we wait.

chanseokoh: Sorry, what I meant is

```java
    bot.waitUntil(new DefaultCondition() {
      @Override
      public boolean test() throws Exception {
        return Iterables.any(..., text -> "1 Debug on Server".equals(text));
      }
     ...
    });
    SwtBotTestingUtilities.performAndWaitForWindowChange(bot, () -> {
      bot.menu("Run").menu("Debug As").menu("1 Debug on Server").click();
    });
```

Looking into the code of `performAndWaitForWindowChange()`, there seems no real difference between this and the code in he PR.

briandealwis: That should work.

PR #3184, opened 2018-06-27
Update target platforms

briandealwis:
- Photon was released today
- Update Oxygen to SR3
- WTP seems to have removed their old releases from Neon and prior, so the SDK features aren't available

### DIFF
--- a/eclipse/neon/gcp-eclipse-neon.target
+++ b/eclipse/neon/gcp-eclipse-neon.target
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde?>
 <!-- generated with https://github.com/mbarbero/fr.obeo.releng.targetplatform -->
-<target name="GCP for Eclipse Neon" sequenceNumber="1528482238">
+<target name="GCP for Eclipse Neon" sequenceNumber="1530132654">
   <locations>
     <location includeMode="planner" includeAllPlatforms="false" includeSource="true" includeConfigurePhase="false" type="InstallableUnit">
       <unit id="org.eclipse.sdk.feature.group" version="4.6.3.v20170301-0400"/>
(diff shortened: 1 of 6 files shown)


briandealwis:
WTP SDKs are available from the [archives page](http://archive.eclipse.org/webtools/downloads/drops/?d) but since we're defaulting to Oxygen, I'm not going to bother.

elharo:
I had hoped we'd get one last Neon release before dropping it, but if it's too much trouble c'est la vie.

chanseokoh:
Neon builds still failing... the SDK seems required.
```
36812 [ERROR] Cannot resolve project dependencies:
36812 [ERROR]   Software being installed: com.google.cloud.tools.eclipse.dist 1.7.0.qualifier
36812 [ERROR]   Missing requirement: com.google.cloud.tools.eclipse.appengine.localserver 0.1.0.qualifier requires 'bundle org.eclipse.jst.server.ui 1.1.300' but it could not be found
```

elharo:
Not quite sure what's up here; but it seems our build depends on remote resources that do not have stable locations? If so, should we check those into our repo instead?

chanseokoh:
As I understand, the SDK for Neon is gone from the usual location but still available after moved to the archive location (https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3184#issuecomment-400826422).

We thought we wouldn't bother about pulling the SDK in for Neon, thinking it is only for retrieving source code and we don't target Neon in our dev environment. However, seems like we're wrong and need to pull in the SDK from the archive.

briandealwis:
I incorrectly thought we were pulling in the WTP dependencies from the release train locations (download.eclipse.org/releases/neon).  But we pull in all of the WTP dependencies using the SDK features.  I've updated the Neon TP to use the archive site for Neon.  I also updated SWTBot to the latest.

PR #3186, opened 2018-06-28
Add support for YAML files for App Engine content block

### LINKED ISSUE #3091: Content Block Descriptors should recognize YAML
E.g. cron.yaml as well as cron.xml; etc.

### LINKED ISSUE #3099: Support flex in App Engine content block
The App Engine content block should support App Engine flexible.

We'll need descriptor models for the various yaml files.

briandealwis:
Adds support for `app.yaml` and the various App Engine `.yaml` configuration files.  This change moves support for resolving App Engine configuration files into a utility class called `AppEngineConfigurationUtil`; we were previously lumping these files into `WEB-INF`.  The idea is that this class can use the project facets and language types to determine the right place for these configuration files.

But you'll notice that this `AppEngineConfigurationUtil` class doesn't currently do any of the special logic.  I started down the path of adding special logic for App Engine standard vs flexible for determining where these configuration files should be looked up, but giving some of [@ludoch's musings about new configuration files](https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/671) I thought it was better to support a `src/main/appengine` _if present_, and figure out the rest later.  It checks if there's an `src/main/appengine` and tries to resolve the file there, and otherwise looks in the `WEB-INF`.  This works for our existing projects as AES projects don't have `src/main/appengine` and all files are placed in `WEB-INF`, and AEF projects were using `src/main/appengine` anyways.  This has a few knock-on effects:

  - App Engine `.yaml` configuration files will be found and showed when they are located in `WEB-INF`: this is unlikely.
  - AES projects with `src/main/appengine` will break on run or deployment since our launch and deployment processes do not currently resolving files from there.  This is an unlikely to happen in practice.  And such projects can easily be made to work with by mapping `src/main/appengine` to `WEB-INF` the project's _Deployment Assembly_.

We may want to teach our projects to place these App Engine configuration files to an alternative well-known location via the deployment assembly (e.g., `WEB-INF/appengine`)?  

Fixes #3091 
Fixes #3099

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
@@ -17,7 +17,9 @@
 package com.google.cloud.tools.eclipse.appengine.facets;
 
 import com.google.cloud.tools.eclipse.util.io.ResourceUtils;
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
@@ -45,7 +47,26 @@ public class AppEngineConfigurationUtil {
    * Create an App Engine configuration file in the appropriate location for the project.
    *
    * @param project the hosting project
-   * @param relativePath the file name
+   * @param relativePath the path of the file relative to the configuration location
+   * @param contents the content for the file
+   * @param overwrite if {@code true} then overwrite the file if it exists
+   */
+  public static IFile createConfigurationFile(
+      IProject project,
+      IPath relativePath,
+      String contents,
+      boolean overwrite,
+      IProgressMonitor monitor)
+      throws CoreException {
+    InputStream bytes = new ByteArrayInputStream(contents.getBytes(StandardCharsets.UTF_8));
+    return createConfigurationFile(project, relativePath, bytes, overwrite, monitor);
+  }
+
+  /**
+   * Create an App Engine configuration file in the appropriate location for the project.
+   *
+   * @param project the hosting project
+   * @param relativePath the path of the file relative to the configuration location
    * @param contents the content for the file
    * @param overwrite if {@code true} then overwrite the file if it exists
    */
--- a/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.eclipse.appengine.facets;
+
+import com.google.cloud.tools.eclipse.util.io.ResourceUtils;
+import java.io.InputStream;
+import org.eclipse.core.resources.IContainer;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IFolder;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.Path;
+
+/**
+ * Utility class for resolving App Engine configuration files in a project. The goal of this class
+ * is to insulate callers from having to know how and when configuration files are placed within a
+ * project. For example, Java projects for the flexible environment typically place their YAML
+ * configuration files in {@code src/main/appengine} whereas Java projects for the standard
+ * environment place their XML configuration files within the {@code WEB-INF} directory.
+ *
+ * @see WebProjectUtil
+ */
+public class AppEngineConfigurationUtil {
+  /** A well-known location for App Engine configuration files. */
+  private static final IPath DEFAULT_CONFIGURATION_FILE_LOCATION = new Path("src/main/appengine");
+
+  /**
+   * Create an App Engine configuration file in the appropriate location for the project.
+   *
+   * @param project the hosting project
+   * @param baseName the file name
+   * @param contents the content for the file
+   * @param overwrite if {@code true} then overwrite the file if it exists
+   */
+  public static IFile createConfigurationFile(
+      IProject project,
+      String baseName,
+      InputStream contents,
+      boolean overwrite,
+      IProgressMonitor monitor)
+      throws CoreException {
+    IFolder appengineFolder = project.getFolder(DEFAULT_CONFIGURATION_FILE_LOCATION);
+    if (appengineFolder != null && appengineFolder.exists()) {
+      IFile destination = appengineFolder.getFile(baseName);
+      if (!destination.exists()) {
+        IContainer parent = destination.getParent();
+        ResourceUtils.createFolders(parent, monitor);
+        destination.create(contents, true, monitor);
+      } else if (overwrite) {
+        destination.setContents(contents, IResource.FORCE, monitor);
+      }
+      return destination;
+    }
+    return WebProjectUtil.createFileInWebInf(
+        project, new Path(baseName), contents, overwrite, monitor);
+  }
+
+  /**
+   * Attempt to resolve the given file within the project's {@code WEB-INF}. Note that this method
+   * may return a file that is in a build location (e.g., {@code
+   * target/m2e-wtp/web-resources/WEB-INF}) which may be frequently removed or regenerated.
+   *
+   * @return the file location or {@code null} if not found
+   */
+  public static IFile findConfigurationFile(IProject project, String baseName) {
+    IFolder appengineFolder = project.getFolder(DEFAULT_CONFIGURATION_FILE_LOCATION);
+    IFile destination = appengineFolder.getFile(baseName);
+    if (destination.exists()) {
+      return destination;
+    }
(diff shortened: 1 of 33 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class AppEngineConfigurationUtil {
+  /** A well-known location for App Engine configuration files. */
+  private static final IPath DEFAULT_CONFIGURATION_FILE_LOCATION = new Path("src/main/appengine");
+
+  /**
+   * Create an App Engine configuration file in the appropriate location for the project.
+   *
+   * @param project the hosting project
+   * @param relativePath the file name
```
chanseokoh: "relative path to the file"?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
```diff
@@ -0,0 +1,93 @@
...
+   * @param contents the content for the file
+   * @param overwrite if {@code true} then overwrite the file if it exists
+   */
+  public static IFile createConfigurationFile(
+      IProject project,
+      IPath relativePath,
+      InputStream contents,
```
chanseokoh: I was going to say this could just be `String`, but then `WebProjectUtil.createFileInWebInf()` takes `InputStream` too. Looking at the callers, they could both accept `String`, but I don't want to stretch this PR for that. I just think this can just be `String`, what about you?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
```diff
@@ -0,0 +1,93 @@
...
+      boolean overwrite,
+      IProgressMonitor monitor)
+      throws CoreException {
+    IFolder appengineFolder = project.getFolder(DEFAULT_CONFIGURATION_FILE_LOCATION);
+    if (appengineFolder != null && appengineFolder.exists()) {
```
chanseokoh: I don't think `getFolder()` ever returns null.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
```diff
@@ -0,0 +1,93 @@
...
+   * may return a file that is in a build location (e.g., {@code
+   * target/m2e-wtp/web-resources/WEB-INF}) which may be frequently removed or regenerated.
+   *
+   * @return the file location or {@code null} if not found
+   */
+  public static IFile findConfigurationFile(IProject project, IPath relativePath) {
+    IFolder appengineFolder = project.getFolder(DEFAULT_CONFIGURATION_FILE_LOCATION);
+    if (appengineFolder != null && appengineFolder.exists()) {
+      IFile destination = appengineFolder.getFile(relativePath);
+      if (destination.exists()) {
```
chanseokoh: No need to call `exists()` twice. This could basically be

```java
IFile destination = project.getFolder(DEFAULT_CONFIGURATION_FILE_LOCATION).getFile(relativePath);
```

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/AppEngineLabelProvider.java
```diff
@@ -61,70 +53,68 @@ public String getText(Object element) {
...
     }
-    return result;
+    return null; // carry onto the next label provider
   }
 
-  /** Returns a <em>project:service:version</em> tuple from the appengine-web descriptor. */
+  /** Returns a <em>project:service:version</em> tuple from the App Engine descriptor. */
   @VisibleForTesting
-  static String getVersionTuple(AppEngineDescriptor descriptor) throws AppEngineException {
+  static String getVersionTuple(String projectId, String projectVersion, String serviceId) {
```
chanseokoh: The order seems weird, given that the tuple will be "project:service:version".

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/AppEngineContentProvider.java
```diff
@@ -204,8 +209,7 @@ private void resourceChanged(IResourceChangeEvent event) {
           projectMapping.invalidate(project);
           toBeRefreshed.add(project);
         }
-      } else if (Iterables.any(
-          projectFiles, file -> file != null && "appengine-web.xml".equals(file.getName()))) {
+      } else if (anyMatch(projectFiles, Sets.newHashSet("appengine-web.xml", "app.yaml"))) {
```
chanseokoh: Can we reuse `AppEngineProjectElement.hasAppEngineDescriptor()`?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/model/AppEngineProjectElement.java
```diff
@@ -109,12 +145,20 @@ static boolean hasLayoutChanged(Collection<IFile> changedFiles) {
    */
   private IFile descriptorFile;
 
-  private AppEngineDescriptor descriptor;
+  private String projectId;
+  private String projectVersion;
+  private String serviceId;
+
+  /** Return the App Engine environment type (e.g., standard, flexible). */
```
chanseokoh: Looking at the code, this would be `flex` for the flexible and `null` for the standard, wouldn't it?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/ui/navigator/model/AppEngineProjectElement.java
```diff
@@ -109,12 +145,20 @@ static boolean hasLayoutChanged(Collection<IFile> changedFiles) {
...
+  private String environmentType;
+
+  /** Return the App Engine runtime (e.g., java7, java8). */
```
chanseokoh: The comment is not really correct. The values are not compatible between standard and flex, and I think we should clearly explain it. For example, for standard, the values are `null` and `java8`, and for flex, the values are `java`, `custom`, `python27`, etc. I haven't checked the actual `runtime:` value of `app.yaml` auto-generated from `appengine-web.xml` where `<runtime>` is not present (i.e., java7) or it is set to `java8`, but another point of confusion could be the possibly different values of `runtime:` and `<runtime>` after the auto-conversion.

briandealwis: Verified auto-generated `app.yaml` from an `appengine-web.xml` has `runtime: java7` or `runtime: java8`.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
chanseokoh: In that case, can't you mock `project` so that it does a mock `IFolder`, or is there some difficulty?

third_party/com.google.cloud.tools.eclipse.jdt.launching/.classpath
```diff
@@ -3,6 +3,5 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src/"/>
-    <classpathentry kind="src" path="tests"/>
```
chanseokoh: Just confirming (and for documentation), this is a cleanup?

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
briandealwis: The null check is defensive and possibly unnecessary, but  believe I have hit a situation where it returned null.  Unfortunately I don't recall how it happened.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
briandealwis: getFolder() returning null.

third_party/com.google.cloud.tools.eclipse.jdt.launching/.classpath
briandealwis: Oops sorry, this shouldn't have been included. Photon now supports scoped classpath entries, and some action occasionally causes the IDE to regenerate the .classpath files.

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineConfigurationUtil.java
chanseokoh: Okay, that's fair. I still don't get the inconsistency (checking null for `getFolder()` but not for `getFile()`, etc.), but this is not a big deal.

briandealwis: Point taken: I'll add a null check here too.

briandealwis: You're right the null check shouldn't be necessary and I will instead fix up any tests.  

But back to your original question about combining into one chain: we need to split it up as we only create or update the file *providing* the `src/main/appengine` exists.  If it doesn't exist, then we put it into `WEB-INF`.

briandealwis:
We have this `resource != null && resource.exists()` pattern throughout the code base.  It shouldn't be necessary, but I'll fix that separately.

PR #3207, opened 2018-07-03
Add library group for extra libraries to non-App Engine standard projects

chanseokoh:
Objectify 6 can be added to any non-App Engine standard projects, so I'm renaming the group name. (To be fair, I believe Objectify 6 can be used on App Engine standard if >= Java 8, but we are sticking to Objectify 5 for some technical reason.) In that light, it's really hard to give correct names to the field and the UI title. What do you think?

I'll add tests in the meantime.


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/model/CloudLibraries.java
```diff
@@ -41,17 +41,17 @@
   public static final String MASTER_CONTAINER_ID = "master-container";
 
   /**
-   * Library files for App Engine Standard environment applications; specifically
+   * Library files for App Engine standard environment applications; specifically
    * Objectify, App Engine API, and Google Cloud Endpoints.
```
elharo: Objectify 5
App Engine API 1.0 SDK

PR #3220, opened 2018-07-09
don't include optional dependencies

### LINKED ISSUE #3218: DependencyResolver pulls in optional dependencies
Notice the difference between the Maven project at the top and the non-Maven project at the bottom:

image

Both added Objectify, but only the standard project included joda-time and joda-money, both marked optional in objectify's pom.xml.

I'm not sure what the right answer is here, but I'm about 60% sure that we should not pull in the optional dependencies.

elharo:
fix #3218 @briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
+++ b/plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
@@ -83,7 +83,7 @@ public class DependencyResolver {
     CollectRequest collectRequest = new CollectRequest();
     collectRequest.setRoot(new Dependency(artifact, JavaScopes.RUNTIME));
     collectRequest.setRepositories(centralRepository(system));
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.util/src/com/google/cloud/tools/eclipse/util/DependencyResolver.java
```diff
@@ -71,16 +72,19 @@
...
     // todo we'd prefer not to depend on m2e here
 
     String coords = groupId + ":" + artifactId + ":" + version;
     Artifact artifact = new DefaultArtifact(coords);
     CollectRequest collectRequest = new CollectRequest();
     collectRequest.setRoot(new Dependency(artifact, JavaScopes.RUNTIME));
     collectRequest.setRepositories(centralRepository(system));
-    DependencyRequest request = new DependencyRequest(collectRequest, filter);
-
+    DependencyRequest request = new DependencyRequest(collectRequest, filter);
```
briandealwis: Remove added whitespace

PR #3222, opened 2018-07-11
Tweak ThreadDumpingWatchdog: sort by job name, adds more threads to ignore

briandealwis:
Found some additional useless threads to filter out of the reports.

Also adds a "Thread Dump" command accessible from the _Quick Access_.


plugins/com.google.cloud.tools.eclipse.test.util/src/com/google/cloud/tools/eclipse/test/util/ui/ThreadDumpHandler.java
```diff
@@ -0,0 +1,16 @@
+
```
elharo: Copyright

briandealwis:
Oops, yes, PTAL: these improvements for thread-dumps are useful.

PR #3244, opened 2018-09-05
explicit is now photon

elharo:
@briandealwis

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ By default, the build is targeted against Eclipse Oxygen / 4.7.
 You can explicitly set the `eclipse.target` property to
 `photon` (4.8).
 ```
-$ mvn -Declipse.target=oxygen package
+$ mvn -Declipse.target=photon package
 ```
 
 ### Adding a new bundle/fragment
--- a/README.md
+++ b/README.md
@@ -64,11 +64,11 @@ The tests need to find the Google Cloud SDK.  You can either:
 
 ### Changing the Eclipse Platform compilation and testing target
 
-By default, the build is targeted against Eclipse Oxygen / 4.7.
+By default, the build is targeted against Eclipse Photon / 4.8.
 You can explicitly set the `eclipse.target` property to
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -68,7 +68,7 @@ By default, the build is targeted against Eclipse Oxygen / 4.7.
 You can explicitly set the `eclipse.target` property to
```
chanseokoh: Above this line is "By default, the build is target against Eclipse Oxygen / 4.7."

That line should also be changed to "... against Eclipse Photon / 4.8."

README.md
```diff
@@ -68,7 +68,7 @@ By default, the build is targeted against Eclipse Oxygen / 4.7.
...
 `photon` (4.8).
 ```
-$ mvn -Declipse.target=oxygen package
+$ mvn -Declipse.target=photon package
```
chanseokoh: I noticed this still doesn't work. I think it should be

`$ (cd eclipse ; mvn -Pide-target-platform -Declipse.target=XXX package)`

Right, @briandealwis?

README.md
```diff
@@ -68,7 +68,7 @@ By default, the build is targeted against Eclipse Oxygen / 4.7.
 You can explicitly set the `eclipse.target` property to
 `photon` (4.8).
```
briandealwis: This line should be changed to:
> `oxygen` (4.7) or `2018-09` (4.9)

README.md
briandealwis: This should remain at `eclipse.target=oxygen`.  `photon` doesn't work as it's now the default; Maven doesn't allow an OR-style profile activation unfortunately.

chanseokoh: I also think you need `-Pide-target-platform` after moving into the `eclipse` sub-directory.

briandealwis: Yup, @chanseokoh's right.

README.md
elharo: done

PR #3284, opened 2018-11-19
Fix gcloud update issue in Kokoro

chanseokoh:
https://issuetracker.google.com/119096137 is making `gcloud components update` fail. Following the workaround suggested in the issue tracker.


elharo:
Why isn't the kokoro-ubuntu check firing on this one?

chanseokoh:
We don't have kokoro-ubuntu. There has been only one Kokoro build: kokoro-windows.

PR #3304, opened 2018-12-14
Remove Java 7 project creation

chanseokoh:
This works, but I will spend some more time on checking these locally. Basically putting up here to see if the tests run OK.


plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
```diff
@@ -3,26 +3,6 @@
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="${namespace} ${schemaUrl}"
          version="${servletVersion}">
-<#if servletVersion == "2.5">
```
elharo: Java 8 runtime still supports servlet 2.5 as well as 3.1. I wonder if we need a separate choice for that?

chanseokoh: I get that, but there is Servlet 4.0 (2017), and 2.5 is really ancient (2005). As a project creation wizard, I think we should discourage 2.5.

elharo: OK, we can take it out of the UI for now and put it back if there's demand, but let's leave in options to support it like this file.

chanseokoh: I restored `compilerVersion` (currently 1.8) and `servletVersion` (currently 3.1) in `pom.xml`.

For `web.xml`, we can git rid of this, because with >=3.0, we won't be setting `<servletMapping>` anymore.

Note that in my original PR, I did try to leave branching possibilities in other places for future changes (like Java 11), as we discussed. `pom.xml` was an exception because I just took `pom.flex.ftl`, which did not have template substitution at all.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -253,26 +232,19 @@ private static void createPomXml(IProject project, AppEngineProjectConfig config
...
-      }
-      createChildFile("pom.xml", Templates.POM_XML_STANDARD_TEMPLATE, //$NON-NLS-1$
-          project, properties, monitor);
-    } else {
-      createChildFile("pom.xml", Templates.POM_XML_FLEX_TEMPLATE, //$NON-NLS-1$
-          project, properties, monitor);
+      String sdkVersion = getCurrentVersion(
+          "com.google.appengine", //$NON-NLS-1$
+          "appengine-api-1.0-sdk", //$NON-NLS-1$
+          "1.9.64"); //$NON-NLS-1$
```
elharo: not for this PR but we do probably need to bump this to 1.9.70

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/messages.properties
```diff
@@ -1,2 +1 @@
-appengine.runtimes.java7=Java 7, Servlet 2.5 (deprecated)
```
elharo: How hard would it be to replace the Java 7, Servlet 2.5 runtime with Java 8, Servlet 2.5?

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
elharo: I'm not sure we can abandon servlet 2.5 so readily. There is a lot of legacy code using this, and a lot of tutorials and books and articles based on this.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/messages.properties
chanseokoh: I think this requires a bit of restructuring of our code. We use the `AppEngineRuntime` enum constants for this input, and their IDs are used set the `<runtime>` element in `appengine-web.xml`. We could have some ad-hoc conditionals on this, but ideally I think we should change the `AppEngineProjectConfig` interface to have `getServletVersion()`. For this reason and to be on the safe side, I think adding the Servlet 2.5 support should be another PR. Adding a new feature is also independent of deprecating something, and I think there is no reason to combine the two in a single step. However, we can leave the code to generate Servlet 2.5 templates.

plugins/com.google.cloud.tools.eclipse.util/templates/appengine/web.xml.ftl
chanseokoh: Alright, we can leave the code to generate Servlet 2.5 stuff.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
chanseokoh: Done in this PR.

plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/CodeTemplates.java
```diff
@@ -113,12 +111,7 @@ private static IFile createJavaSourceFiles(IProject project, AppEngineProjectCon
     Map<String, String> properties = new HashMap<>();
     properties.put("package", Strings.nullToEmpty(packageName)); //$NON-NLS-1$
 
-    boolean servlet25 = isStandardProject && isStandardJava7RuntimeSelected(config);
-    if (servlet25) {
-      properties.put("servletVersion", "2.5"); //$NON-NLS-1$ //$NON-NLS-2$
-    } else {
-      properties.put("servletVersion", "3.1"); //$NON-NLS-1$ //$NON-NLS-2$
-    }
+    properties.put("servletVersion", "3.1"); //$NON-NLS-1$ //$NON-NLS-2$
```
elharo: Could we rename isStandardJava7RuntimeSelected --> isServlet25Selected and have it return false for now? That will make it easier to add a servlet version pop-up in the future if we like.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/messages.properties
elharo: Make sense. Let's do that in another PR.

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui.test/META-INF/MANIFEST.MF
```diff
@@ -10,7 +10,8 @@ Bundle-Localization: plugin
 Comment: .appengine.localserver required for WTP server runtime definitions
 Require-Bundle: com.google.cloud.tools.eclipse.test.dependencies,
  com.google.cloud.tools.eclipse.appengine.localserver
-Import-Package: com.google.cloud.tools.eclipse.swtbot,
+Import-Package: com.google.cloud.tools.eclipse.appengine.standard.java8,
```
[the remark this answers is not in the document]
chanseokoh: I kind of agree. Filed #3317.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineRuntime.java
```diff
@@ -22,11 +22,10 @@
  * Description of the App Engine runtime environments.
  */
 public enum AppEngineRuntime {
-  STANDARD_JAVA_7(Messages.getString("appengine.runtimes.java7"), null), //$NON-NLS-1$
```
briandealwis: I don't think we should get rid of this yet.  But in hindsight this class is misguided.  IMHO we should instead be using the WTP IProjectFacet and decorating the App Engine facets with properties that can be queried.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/messages.properties
briandealwis: We can also use the WTP IProjectFacet APIs to query for dynamic web facet support.

plugins/com.google.cloud.tools.eclipse.appengine.ui/src/com/google/cloud/tools/eclipse/appengine/ui/AppEngineRuntime.java
chanseokoh: This is just an enum class whose values are used only in the project wizard UI and the `<runtime>` element in `appengine-web.xml` (this class is in the `.ui` bundle, and the Javadoc says "Description of the App Engine runtime environments"), so it should be fine and we can remove this. I guess you are thinking something else.

briandealwis: 👍  I had worried that it had spread beyond that.

PR #3314, opened 2018-12-18
Add quickfix to upgrade runtime

elharo:
@chanseokoh @briandealwis


plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidator.java
```diff
@@ -79,15 +79,15 @@
             "com.google.cloud.tools.eclipse.appengine.validation.runtimeMarker",
             IMarker.SEVERITY_WARNING,
             IMessage.NORMAL_SEVERITY,
-            userData, runtime.length(), null);
+            userData, runtime.length(), new UpgradeRuntimeQuickAssistProcessor());
         problems.add(problem);
       } else if ("java7".equals(runtime)) {
```
briandealwis: Do we need to check for a null runtime too?  Or does that implicitly become `java8`?

PR #3319, opened 2018-12-18
Warn when deploying obsolete runtime

### LINKED ISSUE #3296: Warn on deployment of Java 7

briandealwis:
Adds a new `appengine.runtime.obsolete` property for App Engine Standard facet versions, and tags the `JRE7` facet version as being obsolete.

screen shot 2018-12-18 at 12 07 09 pm

This warning can be controlled via a project property:

screen shot 2018-12-19 at 10 01 38 am

Fixes #3296

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.compat.test/.classpath
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.compat.test/.classpath
@@ -2,10 +2,6 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src/">
-		<attributes>
-			<attribute name="test" value="true"/>
-		</attributes>
-	</classpathentry>
+	<classpathentry kind="src" path="src"/>
 	<classpathentry kind="output" path="target/classes"/>
 </classpath>
--- a/plugins/com.google.cloud.tools.eclipse.appengine.compat.test/.classpath
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.compat.test/.classpath
@@ -2,6 +2,10 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="src/">
+		<attributes>
+			<attribute name="test" value="true"/>
+		</attributes>
+	</classpathentry>
 	<classpathentry kind="output" path="target/classes"/>
 </classpath>
--- a/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
@@ -32,7 +32,7 @@ vm.is.jre.proceed=\
  Continue with deployment?
 obsolete.runtime.title=Obsolete App Engine Runtime
 obsolete.runtime.proceed=\
- Project "{0}" is configured to use an obsolete App Engine runtime, and so deployment is likely to fail.\n\
+ Project "{0}" uses an obsolete App Engine runtime. Deployment is likely to fail.\n\
  \n\
  Continue with deployment?
 deploy.button=Deploy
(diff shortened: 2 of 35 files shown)


elharo:
@ludoch @patflynn What do you think of this option?

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/messages.properties
```diff
@@ -30,6 +30,11 @@ vm.is.jre.proceed=\
  Deploying JSPs requires a Java Development Kit (JDK), and so deployment is likely to fail.\n\
  \n\
  Continue with deployment?
+obsolete.runtime.title=Obsolete App Engine Runtime
+obsolete.runtime.proceed=\
+ Project "{0}" is configured to use an obsolete App Engine runtime, and so deployment is likely to fail.\n\
```
elharo: "is configured to use" --> "specifies" or perhaps just "uses"

"runtime, and so deployment" --> "runtime. Deployment"

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
```diff
@@ -38,15 +41,51 @@
...
   }
 
   @Override
   protected boolean checkProject(Shell shell, IProject project) throws CoreException {
+    return checkJspConfiguration(shell, project)
+        && checkAppEngineRuntimeCompatibility(shell, project);
+  }
+
+  private boolean checkAppEngineRuntimeCompatibility(Shell shell, IProject project) {
+    IFacetedProject facetedProject = null;
```
elharo: To avoid splitting declaration and initialization, include more in the try block

plugins/com.google.cloud.tools.eclipse.appengine.facets/src/com/google/cloud/tools/eclipse/appengine/facets/AppEngineStandardFacet.java
```diff
@@ -99,6 +99,28 @@ public static boolean hasFacet(IFacetedProject project) {
...
+  public static boolean usesObsoleteRuntime(IFacetedProject facetedProject) {
+    Preconditions.checkNotNull(facetedProject);
+    IProjectFacetVersion facetVersion = facetedProject.getProjectFacetVersion(FACET);
+    return facetVersion != null && usesObsoleteRuntime(facetVersion);
+  }
+
+  @VisibleForTesting
+  public static boolean usesObsoleteRuntime(IProjectFacetVersion facetVersion) {
+    Object obsoleteProperty = facetVersion.getProperty("appengine.runtime.obsolete");
+    return obsoleteProperty instanceof String
```
elharo: I had to read this a few times to decode it. Might be clearer as simple if-else

plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/standard/StandardDeployCommandHandler.java
briandealwis: `ProjectFacetsManager.create(project)` can return `null`, so it's a condition we need to handle.

elharo: Yes, you can handle it inside the try block.

briandealwis:
An alternative is to switch the obsolete-runtime problem markers to be errors, in which case our errors-prevent-deploy code kicks in.

screen shot 2018-12-18 at 3 59 53 pm

briandealwis:
Added a project deploy property to control whether the warning is shown or not.

I'm now wondering if we should make this a general project property that can be consulted by the validators.

patflynn:
👍 . As long as they've checked 'allow deploying to obsolete runtimes' the warning doesn't show up?

I'd also be perfectly fine if we just did nothing about the runtime at deploy time so that java7 targeting deployments would just fail in the console if they're not whitelisted.  The java7 population doing deployments should decline precipitously after java 7 is turned off.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/.classpath
```diff
@@ -2,6 +2,10 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="src" path="src/">
+		<attributes>
+			<attribute name="test" value="true"/>
```
chanseokoh: Probably we should revert these changes. Having the "test" attribute for "src/" doesn't seem right.

elharo:
I don't want to simply fail in the console. It's too inobvious for a problem we should be able to detect.

plugins/com.google.cloud.tools.eclipse.appengine.compat.test/.classpath
briandealwis: Oh!  I had marked those changes to be ignored — Eclipse 2018-09 or -12 added support for app vs test scoping of dependencies.

briandealwis:
Good point @chanseokoh — I'll add text to the warning dialog.

briandealwis:
screen shot 2018-12-19 at 12 58 55 pm

PR #3320, opened 2018-12-18
warn on missing runtime

### LINKED ISSUE #3298: Warn on Java 7 runtime in appengine-web.xml
red squiggly underline

This can be `<runtime>java</runtime>` or no runtime element at all.

elharo:
fix #3298

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
@@ -2,6 +2,7 @@
 <?eclipse version="3.4"?>
 <plugin>
   
+  <!-- todo this doesn't seem to be used any more -->
   <extension 
       id="blacklistedClass" 
       name="App Engine Whitelist Marker"
@@ -15,6 +16,7 @@
       name="%appengine.web"
       point="org.eclipse.wst.validation.validatorV2" 
       id="appEngineWebXmlValidator">
+    <!-- TODO should this specify more than one marker type? -->
     <validator
     	class="com.google.cloud.tools.eclipse.appengine.validation.XmlValidator:com.google.cloud.tools.eclipse.appengine.validation.AppEngineWebXmlValidator"
       	markerId="com.google.cloud.tools.eclipse.appengine.validation.appEngineBlacklistMarker">
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
@@ -188,7 +188,8 @@
       id="runtimeMarker"
       name="%appengine.runtime.marker"
       point="org.eclipse.core.resources.markers">
-    <super type="com.google.cloud.tools.eclipse.appengine.validation.runtimeMarker" />
+    <super type="org.eclipse.core.resources.problemmarker" />
+    <super type="org.eclipse.core.resources.textmarker" />
     <persistent value="true" />
   </extension>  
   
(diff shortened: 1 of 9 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/MissingRuntime.java
```diff
@@ -0,0 +1,29 @@
...
+package com.google.cloud.tools.eclipse.appengine.validation;
+
+/**
+ * No runtime element in appengine-web.xml.
+ */
+class MissingRuntime extends ElementProblem {
+
+  MissingRuntime() {
+    super("A runtime element is required.",
+        "com.google.cloud.tools.eclipse.appengine.validation.runtimeMarker");
```
briandealwis: This runtimeMarker needs to be defined in the plugin.xml

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ElementProblem.java
```diff
@@ -59,10 +60,20 @@
...
     this(message, "org.eclipse.core.resources.problemmarker",
         IMarker.SEVERITY_WARNING, IMessage.NORMAL_SEVERITY, new DocumentLocation(0, 0), 0, null);
   }
+  
+  /**
+   * A problem with no specific location in the document such as an element
+   * that should be present but isn't.
+   */
+  ElementProblem(String message, String markerId) {
+    this(message, markerId,
```
briandealwis: I wonder if we should be providing a (0,0) default rather than have the subclasses pass that down.

plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.xml
```diff
@@ -15,6 +16,7 @@
       name="%appengine.web"
       point="org.eclipse.wst.validation.validatorV2" 
       id="appEngineWebXmlValidator">
+    <!-- TODO should this specify more than one marker type? -->
```
briandealwis: I wonder if we could do away with the jdt marker.  When I experimented in setting the RuntimeProblem as being an ERROR, I'd see an overlapping warning and error sign in the editor's overlay ruler (the left-hand side).  I suspect we're not filling in a marker property used by the JDT marker, which is defaulting to WARNING.

elharo: Maybe. This is the first time I've looked at this code since @abbiefarr merged it last year. I'm sure there are some improvements to be made. I know some of the thing sI saw when I dug in surprised me.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/ElementProblem.java
elharo: 0, 0 is deliberate here to get it on the document instead of a line. Not sure what you want to change.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/MissingRuntime.java
elharo: done

PR #3326, opened 2018-12-20
Quick Fix for missing appengine runtime element

### LINKED ISSUE #3313: QuickFix for AppEngine Runtime
Upgrade to java8 runtime

elharo:
fix #3313 @chanseokoh 

This handles the obsolete and missing runtimes as variants of the same problem.


plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/.classpath
```diff
@@ -2,6 +2,10 @@
 <classpath>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
-	<classpathentry kind="src" path="src/"/>
+	<classpathentry kind="src" path="src/">
+		<attributes>
+			<attribute name="test" value="true"/>
```
chanseokoh: Need to revert this. The attribute "test=true" doesn't seem right against "src/".

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/UpgradeRuntimeQuickFixTest.java
```diff
@@ -61,10 +56,29 @@
...
+        + "<runtime>java7</runtime>"
+        + "</appengine-web-app>";
+    
+    checkUpgrade(appengineWebAppJava7);
+  }
+  
+  @Test
+  public void testAddMissingRuntime() throws CoreException {
+    String appengineWebAppJava7 =
+        "<appengine-web-app xmlns='http://appengine.google.com/ns/1.0'>"
```
chanseokoh: hit: I think you can close an element in itself, like `<element/>`?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
```diff
@@ -47,6 +45,12 @@
           .equals(marker.getType())) {
         IMarkerResolution fix = new VersionQuickFix();
         markerResolutions[0] = fix;
+      } else if ("com.google.cloud.tools.eclipse.appengine.validation.runtimeMarker"
+          .equals(marker.getType())) {
+        IMarkerResolution fix = new UpgradeRuntimeQuickFix();
```
chanseokoh: I saw a test failure. Should this be `UpgradeRuntimeSourceQuickFix` (unsure)?

plugins/com.google.cloud.tools.eclipse.appengine.validation/xslt/upgradeRuntime.xsl
```diff
@@ -20,6 +20,16 @@
...
+  <xsl:template match="appengine:appengine-web-app">
+    <xsl:copy>
+      <xsl:apply-templates select="node()|@*"/>
+      <xsl:if test="not(appengine:runtime)">  <runtime>java8</runtime><xsl:text>
+</xsl:text></xsl:if>
+    </xsl:copy>
+  </xsl:template>
+
+  <!-- upgrade existing runtime element -->
   <xsl:template match="appengine:runtime"><runtime>java8</runtime></xsl:template>
```
chanseokoh: Give two-space indentation before `<runtime>` like above?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/UpgradeRuntimeQuickFix.java
```diff
@@ -19,7 +19,7 @@
 /**
  * Changes runtime element to java8.
  */
-class UpgradeRuntimeQuickFix extends XsltSourceQuickFix {
+class UpgradeRuntimeQuickFix extends XsltQuickFix {
```
chanseokoh: This class is almost same as `UpgradeRuntimeSourceQuickFix`, except that this extends `XsltQuickFix` while the other `XsltSourceQuickFix`. Do we need both of them? What's the difference between a plain quick fix and the source quick fix?

elharo: Yes, we do need both. They implement different interfaces and are used in different places. We have similar dualism in all the quickfixes.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.ui.test/.classpath
elharo: I wonder why this keeps getting added? And whether .classpath files belong in the repo in the first place?

plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/UpgradeRuntimeQuickFixTest.java
elharo: possible, not required

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
elharo: fixed

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/UpgradeRuntimeQuickFix.java
elharo: yes, you need both for different interfaces. All the quick fixes have two.

plugins/com.google.cloud.tools.eclipse.appengine.validation/xslt/upgradeRuntime.xsl
elharo: no, the indentation here is off to try to get the correct indentation in the generated output. That is some though not all whitespace in XSLT is significant.

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
```diff
@@ -47,6 +45,12 @@
...
+        markerResolutions[0] = fix;
+      } else {
+        return new IMarkerResolution[0];
```
chanseokoh: I had a comment on this but noticed it got disappeared.

Probably we should do this when an exception occurs, so

```java
      IMarkerResolution fix = new UpgradeRuntimeQuickFix();
      markerResolutions[0] = fix;
    }
    return markerResolutions;
  } catch (CoreException ex) {
    ... // fall through
  }
  return new IMarkerResolution[0]; // <-- should return 0-length array
}
```
and better yet just
```java
     fix = new UpgradeRuntimeQuickFix();
      // markerResolutions[0] = fix;
    }
    return new IMarkerResolution[0] {fix};
  } catch (CoreException ex) {
    ... // fall through
  }
  return new IMarkerResolution[0]; // <-- should return 0-length array when exception
}
```
or even
```java
  try {
    if (...) {
      return new IMarkerResolution[0] {new ApplicationQuickFix()};
    } else if (...) {
      return new IMarkerResolution[0] {new VersionQuickFix()};
    } else if (...) {
      return new IMarkerResolution[0] {new UpgradeRuntimeQuickFix()};
    }
  } catch (CoreException ex) {
    ... // fall through
  }
  return new IMarkerResolution[0]; // <-- should return 0-length array
}
```

elharo: How about an ArrayList instead?

plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
```diff
@@ -33,25 +34,25 @@
   private static final Logger logger =
       Logger.getLogger(AppEngineWebMarkerResolutionGenerator.class.getName());
    
-  // TODO do we need to add runtime quick fixes here? If so we need to separate
-  // the obsolete runtime and missing runtime marker types
   @Override
   public IMarkerResolution[] getResolutions(IMarker marker) {
-    IMarkerResolution[] markerResolutions = new IMarkerResolution[1];
+    
+	ArrayList<IMarkerResolution> resolutions = new ArrayList<>();
```
chanseokoh: nit: indentation is off.

elharo: done

PR #3330, opened 2018-12-26
Fix doc formatting

chanseokoh:
Markdown requires 3-space indentation to work properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,33 +16,32 @@ Maven for building Eclipse bundles and features.
 
 ## Requirements
 
-1. The [Google Cloud SDK](https://cloud.google.com/sdk/); install
-  this somewhere on your file system.
+1. The [Google Cloud SDK](https://cloud.google.com/sdk/); install this somewhere on your file system.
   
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,19 +269,19 @@ The `.tpd` files are processed using the [TPD Editor](https://github.com/eclipse
...
-  2. Open the `.tpd` file in Eclipse.
-  3. Make any necessary changes and save.
-     - Note that the TPDs specify artifacts using their _p2 identifiers_.
-       Bundles are specified using their OSGi Bundle Symbolic Name (e.g.,
-       `org.eclipse.core.runtime`).
-       Features are specified using their Feature ID suffixed with `.feature.group`
-       (e.g., `org.eclipse.rcp.feature.group`).
-  4. Select the file in the Package Explorer, right-click, and choose _Create Target Definition File_
-     to update the corresponding .target file.
+   1. Install the TPD Editor, if necessary
```
elharo: I know markdown allows making every number in a list 1, but it is easier to read in source code if we don't do that.

chanseokoh: That's a good point. Updated with explicit numbers.

PR #3352, opened 2019-01-07
Update appengine-web.xsd from Cloud SDK 228.0.0

### LINKED ISSUE #3351: min-instances can be zero
Problem you have encountered:
There is inconsistency between documentation and validation of appengine-web.xml on whether `<min-instances>` can be set to 0 for `<automatic-scaling>`.

In appengine-web-xml reference (https://cloud.google.com/appengine/docs/standard/java/config/appref#scaling_elements_automatic_scaling), it says `<min-instances>` can be set to zero for `<automatic-scaling>`:
"Specify a value from 0 to 1000. You can set the parameter to the value 0 to allow scaling to 0 instances to lower costs when no requests are being served. Note that you are charged for the number of instances specified whether they are receiving traffic or not."

In eclipse, validation fails when trying to set <min-instances> to 0 with the following error message:
Multiple annotations found at this line:
        - cvc-type.3.1.3: The value '0' of element 'min-instances' is not valid.
        - cvc-minInclusive-valid: Value '0' is not facet-valid with respect to minInclusive '1' for type 
         'positiveInteger'.

What you expected to happen:
Either one should be able to set <min-instances> to 0, or the documentation should not imply this is possible.

Steps to reproduce:
In Eclipse xml editor, add the following to appengine-web.xml and save the file:
```
  <automatic-scaling>
    <min-instances>0</min-instances>
  </automatic-scaling>
```

chanseokoh:
Fixes #3351.

XSD from Cloud SDK 228.0.0. `min-instances` will now allow 0.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidationTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation.test/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebXmlValidationTest.java
@@ -40,7 +40,7 @@ public class AppEngineWebXmlValidationTest {
       JavaFacet.VERSION_1_7, WebFacetUtils.WEB_25, AppEngineStandardFacet.JRE7);
 
   @Test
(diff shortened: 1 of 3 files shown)


chanseokoh:
I've re-purposed an existing test (mainly because it is expensive to create a project installed with the App Engine facet).

PR #3355, opened 2019-01-08
no separate runtime category

### LINKED ISSUE #3347: Runtime element problem not properly categorized
This is at head.

image

elharo:
fix #3347 @chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.properties
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.properties
@@ -5,7 +5,6 @@ extension.name=App Engine Source Code Validation
 appengine.web=App Engine Web XML Validator
 pom.validator=Pom Validator
 appengine.blacklist.marker=App Engine Configuration Problem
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation/plugin.properties
```diff
@@ -5,7 +5,6 @@ extension.name=App Engine Source Code Validation
 appengine.web=App Engine Web XML Validator
 pom.validator=Pom Validator
 appengine.blacklist.marker=App Engine Configuration Problem
```
briandealwis: We should rename `appengine.blacklist.marker`?

elharo: Yes, we should. I'm splitting the big PR in #3344 that does this into smaller pieces to try to narrow down why some parts of that aren't working. This is just the first piece.

briandealwis: 👍

Issue #3340, opened 2019-01-04
Java 7 quickfix warning even if <runtime>java8 present


chanseokoh:
Not sure how I got into this. Was applying the quickfix and doing Ctrl+z / Ctrl+Shift+z. Note that the _Problems_ view does not have the corresponding warning entry.

![selection_002](https://user-images.githubusercontent.com/10523105/50704871-e1867d80-1026-11e9-8a37-45a10fb60cf8.png)

![selection_001](https://user-images.githubusercontent.com/10523105/50704874-e4816e00-1026-11e9-9664-22b188958688.png)

chanseokoh:
I found out.

- Remove `<runtime>java8</runtime>` to trigger the warning.
- Manually enter `<runtime>java8<runtime>`.

Then the warning in the _Problems_ view go away, but the warning marker in the left gutter doesn't disappear.

chanseokoh:
Also, by repeating the process, it adds more quick fixes.

![selection_004](https://user-images.githubusercontent.com/10523105/50705162-e697fc80-1027-11e9-9ece-ae012b652af3.png)

elharo:
Problem may be that we're not deleting markers in XmlSourceValidator.validate although we are in XmlValidator.validate.

elharo:
Nope, that didn't do it. :-(

elharo:
To repro place cursor at very first location in file and use Command-1

PR #3356, opened 2019-01-08
delete marker after fix

elharo:
@chanseokoh
deliberately keeping this change minimal to one line + tests


chanseokoh:
> fix #3344

You mean #3340, not #3344. But this doesn't seem to fix #3340, because #3340 is when I manually type and fix the problem and not using the quick fix.

briandealwis:
Is this really necessary?  When the file is saved, the validation is re-performed which [removes and recreates markers as required](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidator.java#L72).  (Though looking at [`XmlValidator#deleteMarkers()`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/XmlValidator.java#L125), perhaps we should be deleting just our own markers and not all `IMarker.PROBLEM`s?)

chanseokoh:
Close this if this isn't necessary? (But probably we should delete our own markers only as @briandealwis said. I will file an issue.)

PR #3367, opened 2019-01-09
reduce visibility

elharo:
@chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
@@ -29,7 +29,7 @@ import org.eclipse.ui.IMarkerResolutionGenerator;
  * Returns possible resolutions for given resource marker. Quick Fix resolutions
  * are available in the Problems view.
  */
-public class AppEngineWebMarkerResolutionGenerator implements IMarkerResolutionGenerator {
+class AppEngineWebMarkerResolutionGenerator implements IMarkerResolutionGenerator {
 
   private static final Logger logger =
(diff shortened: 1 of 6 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.validation/src/com/google/cloud/tools/eclipse/appengine/validation/AppEngineWebMarkerResolutionGenerator.java
```diff
@@ -29,7 +29,7 @@
  * Returns possible resolutions for given resource marker. Quick Fix resolutions
  * are available in the Problems view.
  */
-public class AppEngineWebMarkerResolutionGenerator implements IMarkerResolutionGenerator {
+class AppEngineWebMarkerResolutionGenerator implements IMarkerResolutionGenerator {
```
chanseokoh: This implements `org.eclipse.ui.IMarkerResolutionGenerator` and is referred to in `plugin.xml`. This might be read at runtime by external components? Is it safe to make it non-static?

briandealwis: The javadoc on [`IConfigurationElement#createExecutableExtension()`](https://git.eclipse.org/c/equinox/rt.equinox.bundles.git/tree/bundles/org.eclipse.equinox.registry/src/org/eclipse/core/runtime/IConfigurationElement.java?h=master#n52) says:
> The named attribute value must contain a fully qualified name
> of a Java class. The class can either refer to a class implementing
> the executable extension or to a factory capable of returning the
> executable extension.
> 
> The specified class is instantiated using its 0-argument public constructor.

elharo: It's still a tad ambiguous. For safety's sake, I'm going to restore public on these two classes for now.

PR #3384, opened 2019-01-14
Objectify 6 for App Engine Standard

elharo:
Just want to see what breaks

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -43,6 +43,7 @@
       </libraryFile>
     </library>
     
+    <!-- TODO unify objectify and objectify6 -->
     <library
           id="objectify"
           group="appengine"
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -43,22 +43,21 @@
...
       <libraryFile
-          pinned="true"
           export="true"
-          javadocUri="https://www.javadoc.io/doc/com.googlecode.objectify/objectify/5.1.22">
+          javadocUri="https://www.javadoc.io/doc/com.googlecode.objectify/objectify/6.0">
         <mavenCoordinates
               artifactId="objectify"
               groupId="com.googlecode.objectify"
-              version="5.1.22" />
+              version="6.0.2" />
```
chanseokoh: Actually, it may be better to keep Objectify 5, so that existing Java 7 projects can continue working correctly with the Add Library... functionality. Likewise, I think we won't be removing the JRE7 facet for some time.

PR #3395, opened 2019-02-12
Require Maven 3.5.0

### LINKED ISSUE #3393: maven new project wizard should define minimum maven version
App Engine docs specify 3.5.0 so that's probably what we should set. 

```
[WARNING] Project does not define minimum Maven version, default is: 2.0
[INFO] Plugins require minimum Maven version of: 3.0
[INFO] Note: the super-pom from Maven 3.5.0 defines some of the plugin
[INFO]       versions and may be influencing the plugins required minimum Maven
[INFO]       version.
[INFO] 
[ERROR] Project does not define required minimum version of Maven.
[ERROR] Update the pom.xml to contain
[ERROR]     <prerequisites>
[ERROR]       <maven>3.0</maven>
[ERROR]     </prerequisites>
[INFO] 
```

briandealwis:
Change the project `pom.xml` and the generated Maven-based App Engine project `pom.xml` to specify a requirement for Maven 3.5.0.

Fixes #3393


pom.xml
```diff
@@ -264,6 +264,25 @@
...
+        <executions>
+          <execution>
+            <id>enforce-maven</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireMavenVersion>
+                  <version>3.5.0</version>
```
elharo: This allows later versions too, right?

briandealwis: Since it's not a version range, the version specified is used as a lower bound.  (I run Maven 3.5.4 and there were no complaints.)

PR #3396, opened 2019-03-01
Clean up dev appserver 2 code

chanseokoh:
dev appserver2 uses a separate admin port, so that can be gone too.


chanseokoh:
Something got messed up. Will reopen after fixing it.

PR #3397, opened 2019-03-01
Clean up dev appserver2 code

chanseokoh:
dev appsever 2 had a separate admin port, so that can go away too.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
@@ -405,15 +405,17 @@ public class LocalAppEngineServerBehaviour extends ServerBehaviourDelegate
       } else if (line.contains("Error: A fatal exception has occurred. Program will exit")) { //$NON-NLS-1$
         // terminate the Python process
         stop(false);
-      } else if (shouldAutoDetectPort && (matcher = moduleRunningPattern.matcher(line)).matches()) {
+      } else if ((matcher = moduleRunningPattern.matcher(line)).matches()) {
         String serviceId = matcher.group("service");
         String url = matcher.group("url");
         moduleToUrlMap.put(serviceId, url);
 
-        String portString = matcher.group("port");
-        int port = parseInt(portString, 0);
-        if (port > 0 && (serverPort <= 0 || "default".equals(serviceId))) { // $NON-NLS-1$
-          serverPort = port;
+        if (shouldAutoDetectPort) {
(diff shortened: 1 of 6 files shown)


chanseokoh:
Looks like there is more code to dump.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -455,25 +405,15 @@ public void onOutputLine(String line) {
...
-        if (port > 0 && (serverPortCandidate == 0 || "default".equals(serviceId))) { // $NON-NLS-1$
-          serverPortCandidate = port;
-        }
-      } else if ((matcher = adminStartedPattern.matcher(line)).matches()
-          || (matcher = adminRunningPattern.matcher(line)).matches()) {
-        int port = parseInt(matcher.group("port"), 0);
-        if (port > 0 && adminPort <= 0) {
-          adminPort = port;
-        }
-        // Admin comes after other modules, so no more module URLs
```
chanseokoh: For reviewers: this isn't the case with Dev App Server 1. It prints out an admin URL for each module, so we shouldn't be using this to determine that no more URLs are coming. We should just ignore admin URL lines for Dev App Server 1. I've also reworked the "find-a-candidate-and-assign-it-later" approach and made it more straightforward.

plugins/com.google.cloud.tools.eclipse.appengine.localserver/src/com/google/cloud/tools/eclipse/appengine/localserver/server/LocalAppEngineServerBehaviour.java
```diff
@@ -130,11 +123,7 @@ public void stop(boolean force) {
     if (devServer != null && (!force || serverState != IServer.STATE_STOPPING)) {
       setServerState(IServer.STATE_STOPPING);
       StopConfiguration.Builder builder = StopConfiguration.builder();
-      if (isDevAppServer1()) {
-        builder.adminPort(serverPort);        
-      } else {
-        builder.adminPort(adminPort);
-      }
+      builder.adminPort(serverPort);
```
elharo: should appengine-plugins-core simply remove this method and use the server port?

chanseokoh: appengine-plugins-core removed admin port recently: https://github.com/GoogleCloudPlatform/appengine-plugins-core/pull/780/files#diff-cde49634f93d555a4ee296fe77c7294bL60

We will be able to remove this code too when we upgrade appengine-plugins-core once more.

PR #3402, opened 2019-04-08
Upgrade appengine-plugins-core to 0.7.5 and clean up more dev server2 code

chanseokoh:


plugins/com.google.cloud.tools.eclipse.sdk.ui/META-INF/MANIFEST.MF
```diff
@@ -14,8 +14,8 @@ Require-Bundle: org.eclipse.ui;bundle-version="3.107.0",
  org.eclipse.e4.core.contexts;bundle-version="1.4.0",
  com.google.cloud.tools.eclipse.sdk,
  com.google.cloud.tools.eclipse.preferences,
- com.google.cloud.tools.appengine;bundle-version="0.7.4"
-Import-Package: com.google.cloud.tools.appengine;version="0.7.4",
+ com.google.cloud.tools.appengine;bundle-version="0.7.5"
```
elharo: It's strange that we're requiring the bundle and importing the package here. Isn't it is usually one or the other but not both?

chanseokoh: That's true. Although we prefer `Import-Package`, I'll retain `Require-Bundle` to be on the safe side. I've seen weird cases where things don't work without `Require-Bundle`.

chanseokoh: Removed `Import-Package`.

PR #3410, opened 2019-05-09
Flag clearcut through environment variable

elharo:
Also log non-200 responses
@briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -260,11 +260,8 @@ public class AnalyticsPingManager {
   }
 
   private static boolean useClearCut() {
-    String s = System.getenv("USE_CLEARCUT");
-    if (s == null) {
-      return false;
-    }
-    return s.equalsIgnoreCase("true");
+    String flag = System.getenv("USE_CLEARCUT");
+    return flag.equalsIgnoreCase("true");
   }
 
   private static final Escaper METADATA_ESCAPER = new CharEscaperBuilder()
--- a/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
+++ b/plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
@@ -261,7 +261,7 @@ public class AnalyticsPingManager {
 
   private static boolean useClearCut() {
     String flag = System.getenv("USE_CLEARCUT");
-    return flag.equalsIgnoreCase("true");
+    return "true".equalsIgnoreCase(flag);
   }
 
   private static final Escaper METADATA_ESCAPER = new CharEscaperBuilder()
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -259,6 +259,14 @@ private void sendPing(PingEvent pingEvent) {
     }
   }
 
+  private static boolean useClearCut() {
+    String s = System.getenv("USE_CLEARCUT");
+    if (s == null) {
```
chanseokoh: nit: with `"true".equalsIgnoreCase(s)`, the null check is unnecessary.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -247,10 +244,13 @@ private void sendPing(PingEvent pingEvent) {
         }
       }
       
-      if (USE_CLEAR_CUT) {
+      if (useClearCut()) {
         try {
           String json = jsonEncode(pingEvent);
-          HttpUtil.sendPost(clearCutUrl, json, "application/json");
+          int resultCode = HttpUtil.sendPost(clearCutUrl, json, "application/json");
```
briandealwis: So I noticed that _google-http-client_'s `HttpRequest#execute()`, which is used under the hood here, may throw an `HttpResponseException` for an error response when `HttpRequest#getThrowExceptionsOnExecuteError()` is true, which defaults to true.

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
```diff
@@ -259,6 +259,14 @@ private void sendPing(PingEvent pingEvent) {
     }
   }
 
+  private static boolean useClearCut() {
+    String s = System.getenv("USE_CLEARCUT");
```
briandealwis: Are single letters allowed now?

plugins/com.google.cloud.tools.eclipse.usagetracker/src/com/google/cloud/tools/eclipse/usagetracker/AnalyticsPingManager.java
briandealwis: …which means this logging should be put below, I suspect.

elharo: I think we're using java.net.URLConnection, not HttpRequest, here.

briandealwis: No it's using google-http-client.

chanseokoh: I have to say it is not using Google HTTP Client. It is used only for the multi-part post that is not being used.

briandealwis: Sorry I was looking at `sendPostMultipart` :blush:

PR #3411, opened 2019-05-09
pick Objectify 6 on App Engine Standard

### LINKED ISSUE #3382: Objectify should select version 6
That is, stop pinning to Objectify 5 in the standard wizards. 

Related to #3111

We pinned to Objectify 5 because Java 7 couldn't handle Objectify 6. Now that we don't care about new projects for Java 7, we can stop doing this.

elharo:
fixes #3382

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -50,7 +50,6 @@
           tooltip="%objectify.tooltip"
           dependencies="include"
           siteUri="https://github.com/objectify/objectify/wiki" >
-      <libraryDependency id="appengine-api" />
       <libraryFile
           export="true"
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -52,13 +52,12 @@
           siteUri="https://github.com/objectify/objectify/wiki" >
       <libraryDependency id="appengine-api" />
```
chanseokoh: This dependency should go away.

But then, this `<library id="objectify"` becomes identical to the `<library id="objectify6"` below. So ideally, we may be able to git rid of this entirely.

chanseokoh: Oh, wait. That's not exactly true. The "group" attribute is different.

In any case, the `appengine-api` dependency should go away.

elharo: You're right, done.

PR #3423, opened 2019-05-14
Test case updated

lak-proddev:
I just updated the test case. within my knowledge Since we have changed the layout for composite. Someone needs to review it and may need to update it


lak-proddev:
accidentally created a new pull request. Not required

PR #3446, opened 2019-06-04
Update for Objectify 6.0.4

briandealwis:
[Objectify 6 requires Java 8](https://github.com/objectify/objectify/blob/master/pom.xml#L86) so also add `javaVersion="1.8"` to the library definition.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
@@ -48,6 +48,7 @@
           group="appengine"
           name="Objectify"
           tooltip="%objectify.tooltip"
+          javaVersion="1.8"
           dependencies="include"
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/plugin.xml
```diff
@@ -65,6 +66,7 @@
           group="non-appengine-standard"
           name="Objectify"
           tooltip="%objectify.tooltip"
+          javaVersion="1.8"
```
elharo: why was this needed?

briandealwis: While looking at why the test failed, I noticed the`Library` instance for Objectify had`javaVersion` set to `"1.7"`. Objectify 6 requires Java 8.

PR #3461, opened 2019-06-08
Pipeline Options area is not scrollable #3460

### LINKED ISSUE #3460: Pipeline Options area is not scrollable
As seen here when DataflowRunner is selected it contains many more options than fit in the box, even before they're expanded.

image

lak-proddev:
fixes #3460


lak-proddev:
I am not sure. You want me to add a test case?

elharo:
Yes, generally when we uncover a bug we try to write a failing test case before fixing it. That way we know we understand the bug, verify that the PR fixes it, and make sure future changes don't reintroduce it.

lak-proddev:
> Yes, generally when we uncover a bug we try to write a failing test case before fixing it. That way we know we understand the bug, verify that the PR fixes it, and make sure future changes don't reintroduce it.

Okay, will add a test case in a bit.

lak-proddev:
I didn't write much UI test cases earlier. Someone can correct or rewrite it.

PR #3476, opened 2019-06-17
Use plugin.properties to localize dataflow plugin.xml

### LINKED ISSUE #1744: Use plugin.properties to localize dataflow plugin.xml

lak-proddev:
fixes #1744


elharo:
Thanks @lak-proddev !

PR #3479, opened 2019-06-18
BOM in non-BOM Maven projects

### LINKED ISSUE #3380: BOM in non-BOM Maven projects
When adding a cloud client library to a Maven project (including non App Engine projects) that does not import the google-cloud-bom, add the BOM. There are two cases here:

1. The dependencyManagement section does not include google-cloud-bom.
2. There is no dependencyManagement section.

lak-proddev:
fixes #3380

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
@@ -104,15 +104,13 @@ public class Bug3380Test {
     Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-asset"));
     Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-automl"));
 
-    Element bomElement = null;
-    bomElement = (Element) xpath.evaluate(
+    Element bomElement = (Element) xpath.evaluate(
         "//m:dependencyManagement/m:dependencies/m:dependency[m:groupId='com.google.cloud'][m:artifactId='google-cloud-bom']",
         pom.document.getDocumentElement(), XPathConstants.NODE);
     Assert.assertNotNull(bomElement); // Make sure BOM element added
 
-    Element dependencies = null;
-    dependencies = (Element) xpath.evaluate("./m:dependencies", pom.document.getDocumentElement(),
-        XPathConstants.NODE);
+    Element dependencies = (Element) xpath.evaluate("./m:dependencies",
+        pom.document.getDocumentElement(), XPathConstants.NODE);
     Assert.assertNotNull(dependencies);
 
     Element dependency1 =
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.eclipse.appengine.libraries;
+
+import static org.junit.Assert.fail;
+
+import com.google.cloud.tools.eclipse.appengine.libraries.model.Library;
+import com.google.cloud.tools.eclipse.appengine.libraries.model.LibraryFile;
+import com.google.cloud.tools.eclipse.appengine.libraries.model.MavenCoordinates;
+import com.google.cloud.tools.eclipse.test.util.project.TestProjectCreator;
+import com.google.cloud.tools.eclipse.util.ArtifactRetriever;
+import com.google.cloud.tools.eclipse.util.MappedNamespaceContext;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpressionException;
+import javax.xml.xpath.XPathFactory;
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.CoreException;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.w3c.dom.Element;
+import org.w3c.dom.Node;
+import org.xml.sax.SAXException;
+
+public class Bug3380Test {
+
+  @Rule
+  public final TestProjectCreator projectCreator = new TestProjectCreator();
+  private static final DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+  private static final XPathFactory xpathFactory = XPathFactory.newInstance();
+
+  @BeforeClass
+  public static void configureParser() {
+    factory.setNamespaceAware(true);
+  }
+
+  private Pom pom;
+  private IFile pomFile;
+  private final XPath xpath = xpathFactory.newXPath();
+
+  @Before
+  public void setUp() throws SAXException, IOException, CoreException {
+    xpath.setNamespaceContext(new MappedNamespaceContext("m", "http://maven.apache.org/POM/4.0.0"));
+
+    IProject project = projectCreator.getProject();
+    pomFile = project.getFile("pom.xml");
+    try (InputStream in =
+        Files.newInputStream(Paths.get("testdata/testBug3380.xml").toAbsolutePath())) {
+      pomFile.create(in, IFile.FORCE, null);
+      pom = Pom.parse(pomFile);
+    }
+
+    Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+    logger.setLevel(Level.OFF);
+  }
+
+  @After
+  public void tearDown() {
+    Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+    logger.setLevel(null);
+  }
+
+  @Test
+  public void testBug3380() throws CoreException {
+    LibraryFile addFile = new LibraryFile(coordinates("com.google.cloud", "google-cloud-asset"));
+    Library addLib = newLibrary("asset", addFile);
+
+    LibraryFile existFile = new LibraryFile(coordinates("com.google.cloud", "google-cloud-automl"));
+    Library existLib = newLibrary("automl", existFile);
+
+    LibraryFile removeFile =
+        new LibraryFile(coordinates("com.google.cloud", "google-cloud-bigquery"));
+    Library removeLib = newLibrary("bigquery", removeFile);
+
+    pom.updateDependencies(Arrays.asList(addLib, existLib), Arrays.asList(removeLib));
+
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-asset"));
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-automl"));
+
+    Element bomElement = null;
+    try {
+      bomElement = (Element) xpath.evaluate(
+          "//m:dependencyManagement/m:dependencies/m:dependency[m:groupId='com.google.cloud'][m:artifactId='google-cloud-bom']",
+          pom.document.getDocumentElement(), XPathConstants.NODE);
+    } catch (XPathExpressionException e) {
+      fail(e.getMessage());
+    }
+    Assert.assertNotNull(bomElement); // Make sure BOM element added
+
+    Element dependencies = null;
+    try {
+      dependencies = (Element) xpath.evaluate("./m:dependencies", pom.document.getDocumentElement(),
+          XPathConstants.NODE);
+    } catch (XPathExpressionException e) {
+      fail(e.getMessage());
+    }
+    Assert.assertNotNull(dependencies);
+
+    Element dependency = pom.findDependency(dependencies, "com.google.cloud", "google-cloud-asset");
+    Assert.assertNotNull(dependency);
+
+    Node versionElement = Pom.findChildByName(dependency, "version");
+    Assert.assertNull(versionElement); // Version should be removed If BOM added
+
+    dependency = pom.findDependency(dependencies, "com.google.cloud", "google-cloud-automl");
+    Assert.assertNotNull(dependency);
+
+    versionElement = Pom.findChildByName(dependency, "version");
+    Assert.assertNull(versionElement); // Version should be removed If BOM added
+
+  }
+
+  private static Library newLibrary(String libraryId, LibraryFile... libraryFiles) {
+    Library library = new Library(libraryId);
+    library.setLibraryFiles(Arrays.asList(libraryFiles));
+    return library;
+  }
+
+  private static MavenCoordinates coordinates(String groupId, String artifactId) {
+    return new MavenCoordinates.Builder().setGroupId(groupId).setArtifactId(artifactId).build();
+  }
+}
--- a/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.tools.eclipse.appengine.libraries;
 
-import static org.junit.Assert.fail;
-
 import com.google.cloud.tools.eclipse.appengine.libraries.model.Library;
 import com.google.cloud.tools.eclipse.appengine.libraries.model.LibraryFile;
 import com.google.cloud.tools.eclipse.appengine.libraries.model.MavenCoordinates;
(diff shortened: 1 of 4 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -270,6 +280,66 @@ void updateDependencies(Collection<Library> selectedLibraries,
...
+        Element dependency = document.createElementNS(NAMESPACE_URI, DEPENDENCY);
+        dependency.appendChild(createChild(GROUP_ID, Bom.GOOGLE_BOM_GROUP_ID));
+        dependency.appendChild(createChild(ARTIFACT_ID, Bom.GOOGLE_BOM_ARTIFACT_ID));
+        dependency.appendChild(createChild(TYPE, Bom.GOOGLE_BOM_TYPE));
+        dependency.appendChild(createChild(SCOPE, Bom.GOOGLE_BOM_SCOPE));
+
+        String version = getBestVersion(Bom.GOOGLE_BOM_GROUP_ID, Bom.GOOGLE_BOM_ARTIFACT_ID);
+        if (version != null && !version.isEmpty()) {
+          dependency.appendChild(createChild(VERSION, version));
+          Bom bom = Bom.loadBom(Bom.GOOGLE_BOM_GROUP_ID, Bom.GOOGLE_BOM_ARTIFACT_ID, version, null);
```
lak-proddev: Adding Bom here is correct?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -270,6 +280,66 @@ void updateDependencies(Collection<Library> selectedLibraries,
...
+          return;
+
+        Element dependency = document.createElementNS(NAMESPACE_URI, DEPENDENCY);
+        dependency.appendChild(createChild(GROUP_ID, Bom.GOOGLE_BOM_GROUP_ID));
+        dependency.appendChild(createChild(ARTIFACT_ID, Bom.GOOGLE_BOM_ARTIFACT_ID));
+        dependency.appendChild(createChild(TYPE, Bom.GOOGLE_BOM_TYPE));
+        dependency.appendChild(createChild(SCOPE, Bom.GOOGLE_BOM_SCOPE));
+
+        String version = getBestVersion(Bom.GOOGLE_BOM_GROUP_ID, Bom.GOOGLE_BOM_ARTIFACT_ID);
+        if (version != null && !version.isEmpty()) {
```
lak-proddev: Retrieving version is the right way?

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -65,6 +65,15 @@
 import org.xml.sax.SAXException;
 
 class Pom {
+  private static final String NAMESPACE_URI = "http://maven.apache.org/POM/4.0.0";//$NON-NLS-1$
+  private static final String DEPENDENCY_MANAGEMENT = "dependencyManagement"; //$NON-NLS-1$
```
elharo: I've seen it done, but I wouldn't define constants for element and attribute names. The constants are no clearer than the string literals, arguably less so.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -270,6 +280,66 @@ void updateDependencies(Collection<Library> selectedLibraries,
     }   
   }
 
+  private void createBOMIfNeeded(XPath xpath) throws CoreException {
+    try {
+      Element bomElement = (Element) xpath.evaluate(
+          "//m:dependencyManagement/m:dependencies/m:dependency[m:groupId='"
+              + Bom.GOOGLE_BOM_GROUP_ID + "'][m:artifactId='" + Bom.GOOGLE_BOM_ARTIFACT_ID + "']",
```
elharo: Now that I think about it I wouldn't ad these constants to the Bom clkass either. They add nothing to the string literals.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
```diff
@@ -270,6 +280,66 @@ void updateDependencies(Collection<Library> selectedLibraries,
...
+        if (version != null && !version.isEmpty()) {
+          dependency.appendChild(createChild(VERSION, version));
+          Bom bom = Bom.loadBom(Bom.GOOGLE_BOM_GROUP_ID, Bom.GOOGLE_BOM_ARTIFACT_ID, version, null);
+          if (bom != null) {
+            boms.add(bom);
+          }
+        }
+        dependencies.appendChild(dependency);
+      }
+    } catch (Exception ex) {
```
elharo: do you have to catch java.lang.Exception ehre? Otherwise, be more specific.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
elharo: looks OK

chanseokoh:
This would be more complicated if we implement this complete.

The current library wizard must be specifying versions to all libraries (including the Google Client Libraries) in `pom.xml` when adding libraries to `<dependencies>`. When using a BOM, we should not specify the versions. Also note that libraries like App Engine API or Objectify is not included in the BOM.

And if the `pom.xml` already had Google Client Libraries added without BOM, we should remove versions from the existing dependencies as part of adding the BOM.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
chanseokoh: I'm not decisive yet, but I wonder if we should fail if we fail to retrieve `version`, because otherwise you end up adding a broken `<dependency>`.

chanseokoh:
> The current library wizard must be specifying versions to all libraries (including the Google Client Libraries) in `pom.xml` when adding libraries to `<dependencies>`. When using a BOM, we should not specify the versions.

Ah, looks like this is already being taken care of in the current code.

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
lak-proddev: done

plugins/com.google.cloud.tools.eclipse.appengine.libraries/src/com/google/cloud/tools/eclipse/appengine/libraries/Pom.java
lak-proddev: Modified method # getBestVersion()
new method returns "MavenCoordinates.LATEST_VERSION" if version is null. Someone please review it.
private String getBestVersion(String groupId, String artifactId) {
    ArtifactVersion latestVersion = ArtifactRetriever.DEFAULT.getBestVersion(groupId, artifactId);
    return latestVersion != null ? latestVersion.toString() : MavenCoordinates.LATEST_VERSION;
  }

lak-proddev:
Updated the test plan and code. Someone can review it now.
Always welcome for change suggestions :)

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+
+  @After
+  public void tearDown() {
+    Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+    logger.setLevel(null);
+  }
+
+  @Test
+  public void testBug3380() throws CoreException {
+    LibraryFile addFile = new LibraryFile(coordinates("com.google.cloud", "google-cloud-asset"));
```
elharo: addedFile

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+    Library addLib = newLibrary("asset", addFile);
+
+    LibraryFile existFile = new LibraryFile(coordinates("com.google.cloud", "google-cloud-automl"));
```
elharo: existingFile

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+  @After
+  public void tearDown() {
+    Logger logger = Logger.getLogger(ArtifactRetriever.class.getName());
+    logger.setLevel(null);
+  }
+
+  @Test
+  public void testBug3380() throws CoreException {
+    LibraryFile addFile = new LibraryFile(coordinates("com.google.cloud", "google-cloud-asset"));
+    Library addLib = newLibrary("asset", addFile);
```
elharo: addedLibrary

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+
+    LibraryFile existFile = new LibraryFile(coordinates("com.google.cloud", "google-cloud-automl"));
+    Library existLib = newLibrary("automl", existFile);
+
+    LibraryFile removeFile =
```
elharo: removedFile

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+        new LibraryFile(coordinates("com.google.cloud", "google-cloud-bigquery"));
+    Library removeLib = newLibrary("bigquery", removeFile);
```
elharo: removedLibrary

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-asset"));
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-automl"));
+
+    Element bomElement = null;
+    try {
+      bomElement = (Element) xpath.evaluate(
+          "//m:dependencyManagement/m:dependencies/m:dependency[m:groupId='com.google.cloud'][m:artifactId='google-cloud-bom']",
+          pom.document.getDocumentElement(), XPathConstants.NODE);
+    } catch (XPathExpressionException e) {
```
elharo: try-catch clause not needed; the exception will fail the test

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+      fail(e.getMessage());
+    }
+    Assert.assertNotNull(bomElement); // Make sure BOM element added
+
+    Element dependencies = null;
+    try {
```
elharo: ditto; no try-catch

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+    } catch (XPathExpressionException e) {
+      fail(e.getMessage());
+    }
+    Assert.assertNotNull(dependencies);
+
+    Element dependency = pom.findDependency(dependencies, "com.google.cloud", "google-cloud-asset");
+    Assert.assertNotNull(dependency);
+
+    Node versionElement = Pom.findChildByName(dependency, "version");
+    Assert.assertNull(versionElement); // Version should be removed If BOM added
```
elharo: nit: If --> if BOM was added

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+
+    dependency = pom.findDependency(dependencies, "com.google.cloud", "google-cloud-automl");
+    Assert.assertNotNull(dependency);
+
+    versionElement = Pom.findChildByName(dependency, "version");
```
elharo: I don't think you need to do this a second time.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+    Assert.assertNotNull(dependencies);
+
+    Element dependency = pom.findDependency(dependencies, "com.google.cloud", "google-cloud-asset");
+    Assert.assertNotNull(dependency);
+
+    Node versionElement = Pom.findChildByName(dependency, "version");
+    Assert.assertNull(versionElement); // Version should be removed If BOM added
+
+    dependency = pom.findDependency(dependencies, "com.google.cloud", "google-cloud-automl");
```
elharo: This should be a new variable, not a reused one

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/testdata/testBug3380.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+	xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+	<dependencies>
+		<dependency>
```
elharo: indent only two spaces per level, per google-style

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
lak-proddev: There are two different dependencies. One is a new one and another one already exists.

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
lak-proddev: Assert.assertNotNull(bomElement); // Make sure BOM element added
this check will make sure alreday added

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/testdata/testBug3380.xml
lak-proddev: Done

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,142 @@
...
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-asset"));
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-automl"));
+
+    Element bomElement = null;
+    bomElement = (Element) xpath.evaluate(
+        "//m:dependencyManagement/m:dependencies/m:dependency[m:groupId='com.google.cloud'][m:artifactId='google-cloud-bom']",
+        pom.document.getDocumentElement(), XPathConstants.NODE);
+    Assert.assertNotNull(bomElement); // Make sure BOM element added
+
+    Element dependencies = null;
```
elharo: combine with next line; no need to set this to null

plugins/com.google.cloud.tools.eclipse.appengine.libraries.test/src/com/google/cloud/tools/eclipse/appengine/libraries/Bug3380Test.java
```diff
@@ -0,0 +1,142 @@
...
+        new LibraryFile(coordinates("com.google.cloud", "google-cloud-bigquery"));
+    Library removedLibrary = newLibrary("bigquery", removedFile);
+
+    pom.updateDependencies(Arrays.asList(addedLibrary, existingLibrary),
+        Arrays.asList(removedLibrary));
+
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-asset"));
+    Assert.assertTrue(pom.dependencyManaged("com.google.cloud", "google-cloud-automl"));
+
+    Element bomElement = null;
```
elharo: combine with next line

lak-proddev: Sorry, forget to remove

PR #3481, opened 2019-06-20
remove old GA code

elharo:


chanseokoh:
Should be good, but I'd like to hold merging it until we complete dashboard migration. As I said the other day, not everything is working on the backend.

PR #3501, opened 2019-06-26
Changes from a fresh import into Eclipse

elharo:
@briandealwis I still wonder whether these files should simply be .gitignored instead.


briandealwis:
It really depends on our minimum version of Eclipse to be used for *development*.  I think your internal installs are based on 4.7 (Oxygen) and 4.9 (2018-09)?  JDT's support for _test sources_ in the `.classpath` came in 4.8 (Photon).  It looks like JDT tweaked some formatter settings.

Although the formatter settings shouldn't matter much (the only essential piece is [the configuration of Google formatter plugin](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/781341fb8973e016611427d9999d5f782b0112ff/eclipse/settings/files/org.eclipse.jdt.core.prefs#L456)), they're actually blasted into place from the files in [`eclipse/settings/`](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/tree/master/eclipse/settings).

PR #3506, opened 2019-06-27
Dataflow project not shown in New Menu

lak-proddev:
**JEE perspective**

image

**Java Perspective**

image

**Web Perspective**

image

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.dataflow.ui/plugin.xml
+++ b/plugins/com.google.cloud.tools.eclipse.dataflow.ui/plugin.xml
@@ -294,4 +294,16 @@
     <extension point="org.eclipse.help.contexts">
        <contexts file="helpContexts.xml" />
     </extension>
+	<extension point="org.eclipse.ui.perspectiveExtensions">
+		<perspectiveExtension
+			targetID="org.eclipse.jdt.ui.JavaPerspective">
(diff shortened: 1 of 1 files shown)


plugins/com.google.cloud.tools.eclipse.dataflow.ui/plugin.xml
```diff
@@ -294,4 +294,16 @@
     <extension point="org.eclipse.help.contexts">
        <contexts file="helpContexts.xml" />
     </extension>
+	<extension point="org.eclipse.ui.perspectiveExtensions">
```
elharo: Indent looks off. Are you using tabs? We try to stick to spaces.

lak-proddev: Yes,  done format with default settings. Updated to two spaces

lak-proddev:
> Thanks. I wonder if there's any way control the order in which menu items appear. Ideally we'd like App Engine to come before Dataflow.

I didn't find a  solution to order menus. that's why I sent SS to show. :)

lak-proddev:
> Thanks. I wonder if there's any way control the order in which menu items appear. Ideally we'd like App Engine to come before Dataflow.

Will push another change if i find any solution for ordering.

PR #3507, opened 2019-06-27
blacklist non 1.8 versions

elharo:
Experimenting...


elharo:
This doesn't seem to have any useful effect. I think we might have to simply bite the bullet and upgrade to beam.

PR #3523, opened 2019-10-16
Fix Travis build

elharo:
@briandealwis Make CLABot happy


briandealwis:
(I restarted two of the builds due to flakes: one on us, another around a connection-reset to Maven Central)

PR #3525, opened 2019-10-17
Add new target for latest Eclipse IDE release, remove Oxygen

briandealwis:
This PR adds a new target to track the latest Eclipse IDE release, and also removes Oxygen. 

The Eclipse IDE project now has a `download.eclipse.org/releases/latest` repository which tracks the latest release currently (2019-09), and other related projects offer similar `/latest` repos too.  We will still need to update the .tpd &rarr; .target files after a new release.  I did try to reference the repositories directly using a Maven `<repository>` to avoid this .tpd &rarr .target generation, but the artifact resolution times were terrible.

Jetty 9.4.20, which is used in Eclipse 2019-09, removed some `__`-prefixed variables like the `Request.__MULTIPART_CONFIG_ELEMENT` field that we used in our `TestHttpServer`, so I just replaced it with the actual field reference.


briandealwis:
Will restore Oxygen.

With this approach, on a new release, we just regenerate the `eclipse/eclipse-latest/gcp-eclipse-latest.target` file.  But it's imperfect as it's entirely possible that `/releases/latest` has new content but continues to host the old content too, so we're not actually testing on latest.

I'll close this and create a new target for 2019-09 (the current latest).  It looks like we can [generate a .target from a .tpd separately](https://github.com/eclipse-cbi/targetplatform-dsl#call-the-generator-from-the-command-line) which might make a nice weekly job.

chanseokoh:
FYI, about the stability of the build.

We are running nightly releases every day, and I haven't seen any single failure since Aug 4. And even the failure on Aug 4 was due to timeout from`download.eclipse.org/webtools`. If we exclude the Aug 4 failure, then the straight no failure stretches back to Jun 4. So I think the builds are very stable at least on Kokoro. However, the different between the nightly builds and a PR is that we have a large build matrix (Java 8 and 11, and 4.7, 4.8, 4.9, 4.10, latest) and that the Travis has its own constraints like "log too long" or the 50-minute limit.

elharo:
Yes. #3526 is one possibility for reducing the size of the build matrix. What do you think?

We should probably also consider if there's any way we can reduce the size of the logs. I'm sure we can quiet some parts we don't care about. I really wish Maven didn't feel compelled to produce output from every test that passes.

But the specific failures here on the new target do seem to be consistent.

briandealwis:
Will open new PR specifically for 2019-09

PR #3534, opened 2019-11-07
add entry point for Java 11

elharo:
@chanseokoh

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectConfig.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectConfig.java
@@ -40,6 +40,7 @@ public class AppEngineProjectConfig {
   private String mavenGroupId;
   private String mavenArtifactId;
   private String mavenVersion;
+  private String entryPoint;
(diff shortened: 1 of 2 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectConfig.java
```diff
@@ -114,4 +115,12 @@ String getMavenArtifactId() {
   String getMavenVersion() {
     return mavenVersion;
   }
+
+  void setEntryPoint(String entryPoint) {
```
chanseokoh: I didn't think you need to set a custom entrypoint. The [quickstart guide](https://cloud.google.com/appengine/docs/standard/java11/quickstart) does say anything about it, and the sample there doesn't have anything related to an entrypoint. Is this needed? Perhaps I don't understand the App Engine Java 11 runtime.

And will we expose this field in the wizard UI?

elharo: It's needed for thin jars. I think we'll eventually expose it in the UI. See https://cloud.google.com/appengine/docs/standard/java11/runtime

PR #3540, opened 2019-11-12
two page new project wizard

### LINKED ISSUE #3535: Split new app engine standard wizard into two pages
First page with most of existing content except libraries

Second page with libraries including three existing libraries--endpoints, app engine API, and Objectify--and the cloud client libraries.

First page has a finish button that skips the second page and does not add any libraries.

elharo:
fixes #3535

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
+++ b/plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
@@ -16,10 +16,15 @@
 
 package com.google.cloud.tools.eclipse.appengine.newproject;
 
+import com.google.cloud.tools.eclipse.appengine.libraries.model.CloudLibraries;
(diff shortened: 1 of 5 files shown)


plugins/com.google.cloud.tools.eclipse.appengine.newproject/src/com/google/cloud/tools/eclipse/appengine/newproject/AppEngineProjectWizard.java
```diff
@@ -84,7 +99,7 @@ private void retrieveConfigurationValues() {
       config.setEclipseProjectLocationUri(appEnginePage.getLocationURI());
     }
 
-    config.setAppEngineLibraries(appEnginePage.getSelectedLibraries());
+    config.setAppEngineLibraries(librariesPage.getSelectedLibraries());
```
chanseokoh: We should rename this to `setSelectedLibraries` (and all related changes in `AppEngineProjectConfig`). Doesn't have to be in this PR.

patflynn:
Could you link a bug or provide more context on this PR? Is this a UI change that could include screenshots?

elharo:
This is an idea @briandealwis came up with (and @chanseokoh and myself concurred) for making upcoming Java 11 new project wizards simpler. Brian had also noticed that the cloud libraries were hard to find and suspected that it would be useful to make them an optional part of new project creation.

See #3535

PR #3543, opened 2019-11-14
Convert tests to java8 runtime

elharo:
@briandealwis

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ProjectContextMenuTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ProjectContextMenuTest.java
@@ -59,25 +59,6 @@ public class ProjectContextMenuTest extends BaseProjectTest {
   }
 
   @Test
-  public void testDynamicWebProjectJava7() {
-    IProject project =
-        projectCreator.withFacets(JavaFacet.VERSION_1_7, WebFacetUtils.WEB_25).getProject();
-    SWTBotTreeItem selected = SwtBotProjectActions.selectProject(bot, project.getName());
-    assertThat(
-        selected.contextMenu("Debug As"), not(MenuMatcher.hasMenuItem(endsWith("App Engine"))));
-    assertThat(
-        selected.contextMenu("Run As"), not(MenuMatcher.hasMenuItem(endsWith("App Engine"))));
-    assertThat(
-        selected.contextMenu(), not(MenuMatcher.hasMenuItem("Deploy to App Engine Standard...")));
-    assertThat(
-        selected.contextMenu("Configure"),
-        MenuMatcher.hasMenuItem("Convert to App Engine Standard Project"));
-    assertThat(
-        selected.contextMenu("Configure"),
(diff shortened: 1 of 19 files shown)


plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ProjectContextMenuTest.java
```diff
@@ -58,25 +58,6 @@ public void testPlainJavaProject() {
         not(MenuMatcher.hasMenuItem("Reconfigure for App Engine Java 8 runtime")));
   }
 
-  @Test
-  public void testDynamicWebProjectJava7() {
```
briandealwis: This test should be kept — we should be able to convert a Java 7 web project to App Engine Java 8

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ProjectContextMenuTest.java
```diff
@@ -96,24 +77,6 @@ public void testDynamicWebProjectJava8() {
         not(MenuMatcher.hasMenuItem("Reconfigure for App Engine Java 8 runtime")));
   }
 
-  @Test
-  public void testAppEngineStandardJava7() {
```
briandealwis: This test should be kept until we remove App Engine Java 7

elharo: done

plugins/com.google.cloud.tools.eclipse.integration.appengine/src/com/google/cloud/tools/eclipse/integration/appengine/ProjectContextMenuTest.java
elharo: done

briandealwis:
Tests are failing as `JRE8` is defined in a separate bundle.
```
Caused by: java.lang.IllegalArgumentException: Version JRE8 of project facet com.google.cloud.tools.eclipse.appengine.facets.standard does not exist.
        at org.eclipse.wst.common.project.facet.core.util.internal.Versionable.getVersion(Versionable.java:80)
        at org.eclipse.wst.common.project.facet.core.internal.ProjectFacet.getVersion(ProjectFacet.java:1)
        at com.google.cloud.tools.eclipse.appengine.facets.AppEngineStandardFacet.<clinit>(AppEngineStandardFacet.java:59)
        ... 55 more
```

We can annotate the various `*.test/pom.xml` with requiring .standard.java8, but I suspect a slightly easier way is to use the `Require-Capability`/`Provide-Capability`.

briandealwis:
More specifically, I think we'd add to `.standard.java8`:
```
Provide-Capability: com.google.cloud.tools.facets.appengine;type:List="jre8"
```

and the `.test` fragments would then need:
```
Require-Capability: com.google.cloud.tools.facets.appengine;filter:="(type=jre8)"
```

PR #3546, opened 2019-11-19
Update to commons-compress' changed BSN

briandealwis:
Update our 3rd-party feature with commons-compress' changed OSGi symbolic name.  Fortunately the package names haven't changed and so `appengine-plugins-core`/`com.google.cloud.tools.appengine` is unaffected.

Filed [COMPRESS-498](https://issues.apache.org/jira/browse/COMPRESS-498) about the name change, which seems to have been a mistake.

### DIFF
--- a/eclipse/ide-target-platform/category.xml
+++ b/eclipse/ide-target-platform/category.xml
@@ -33,7 +33,7 @@
    <bundle id="com.google.guava" version="28.1.0.jre"/>
    <bundle id="com.google.gson" version="2.8.2"/>
    <bundle id="com.google.cloud.tools.appengine" version="0.0.0"/>
-   <bundle id="org.apache.commons.compress" version="0.0.0"/>
+   <bundle id="org.apache.commons.commons-compress" version="0.0.0"/>
    <bundle id="jackson-core-asl" version="1.9.13"/>
    <bundle id="com.fasterxml.jackson.core.jackson-core" version="0.0.0"/>
    <bundle id="org.yaml.snakeyaml" version="1.21"/>
--- a/features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml
+++ b/features/com.google.cloud.tools.eclipse.3rdparty.feature/feature.xml
@@ -40,7 +40,7 @@
          unpack="false"/>
 
    <plugin
-         id="org.apache.commons.compress"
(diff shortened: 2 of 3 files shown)


briandealwis:
The build was pulling `org.apache.commons.compress` 1.18 from Eclipse Orbit.  But I'm not quite sure why the `build/verify-feature-completeness` didn't fail since appengine-plugins-core does say it wants org.apache.commons.compress.* packages with version ≥ 1.19.

chanseokoh:
Could it be that it uses the old name on Photon but a new name on recent Eclipse?

briandealwis:
Ah, figured out the `verify-feature-completeness`:
  1. `verify-feature-completeness` attempts to install our `com.google.cloud.tools.eclipse.suite.feature` feature using our build artifacts in `gcp-repo/` and the `eclipse/ide-target-platform` as a proxy for the main Eclipse repositories.  
  2. (Using `eclipse/ide-target-platform` results in significant speedups.) 
  3. The `ide-target-platform` is assembled from the target platform **and** our specific Maven dependencies with OSGi metadata (from the root pom).  So it includes the `org.apache.commons.commons-compress` bundle.
  4. The only actual consumer of `commons-compress` is `appengine-plugins-core`, which imports its requisite `org.apache.commons.compress.*` as _packages_.
  5. The installation succeeds since the `org.apache.commons.commons-compress` bundle is present in the `ide-target-platform`.

If I change the test to instead reference `download.eclipse.org/releases/xxx` directly then it fails as the only bundle providing `org.apache.commons.compress.*` packages is Orbit's `org.apache.commons.compress` version 1.18.

briandealwis:
Our `com.google.cloud.tools.eclipse.3rdparty.feature` feature does include bundles that we require that aren't commonly available or that differ from Orbit, but we need to include them by bundle name rather than package imports.  So we specified `org.apache.commons.compress` which worked until commons-compress changed its bundle name in 1.18.  But Orbit's variant still has ` name `org.apache.commons.compress` and so that was considered satisfactory.

We could modify `com.google.cloud.tools.eclipse.3rdparty.feature` to specify required version numbers (1.19) which might have caught this situation.  It wouldn't differentiate if Orbit provided 1.19, but maybe that doesn't actually matter since we just care about 1.19 regardless of its source.

PR #3571, opened 2019-12-05
Fix user profile picture URL

### LINKED ISSUE #3570: Wrog image for logged in user
Not sure what's happening here but this is not my profile picture:

image

This is what it should be:

image

chanseokoh:
Fixes #3570.

For the image

https://lh3.googleusercontent.com/a-/AAuE7mBeBrCk1V_scMJv6_OOR68BaWs2OZ9OTE0byc94

we should request the size in the new URL format by appending `=s[size]`, e.g.,

https://lh3.googleusercontent.com/a-/AAuE7mBeBrCk1V_scMJv6_OOR68BaWs2OZ9OTE0byc94=s45

Clicking on the latter URL returns a small image.

### DIFF
--- a/plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/LabelImageLoader.java
+++ b/plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/LabelImageLoader.java
@@ -42,7 +42,8 @@ public class LabelImageLoader {
    *
    * Must be called in the UI context.
    */
-  void loadImage(String imageUrl, Label label) throws MalformedURLException {
+  @VisibleForTesting
+  public void loadImage(String imageUrl, Label label) throws MalformedURLException {
     Preconditions.checkNotNull(imageUrl);
 
     ImageData imageData = cache.get(imageUrl);
--- a/plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanelTest.java
+++ b/plugins/com.google.cloud.tools.eclipse.login.test/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanelTest.java
@@ -17,26 +17,34 @@
 package com.google.cloud.tools.eclipse.login.ui;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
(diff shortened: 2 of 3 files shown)


chanseokoh:
The test failures are very weird. Locally, `verify(imageLoader).loadImage(...)` works when I run a non-Plug-in test, but with Plug-in test, it seems to runs the code of the actual `LabelImageLoader.loadImage` method, even though `imageLoader` is a `@Mock`.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/LabelImageLoader.java
```diff
@@ -42,7 +42,8 @@
    *
    * Must be called in the UI context.
    */
-  void loadImage(String imageUrl, Label label) throws MalformedURLException {
+  @VisibleForTesting
+  public void loadImage(String imageUrl, Label label) throws MalformedURLException {
```
elharo: do you need to make this public? It looks like the test is in the same package.

chanseokoh: Mockito couldn't mock non-public method, and that was why I experienced the weird test failure that confused me.

elharo: Strange. Mockito should be able to mock a non-final package protected method.

chanseokoh: Yeah, I also think this is very weird. A non-plugin JUnit run worked, but it calls the actual method in the plugin JUnit run. Then I guessed maybe it's because the method is not public. I don't have any insights why it behaves differently between plugin and non-plugin runs.

elharo: so what's failing when this isn't public? Is that a Travis run or just a pluign test in Eclipse?

chanseokoh: Both a local plugin test in Eclipse and Travis fail. I guess `mvn verify` runs `.login.ui.test` in the plugin test mode.

PR #3581, opened 2019-12-10
Display "Other Libraries" label for non-App Engine libraries group

### LINKED ISSUE #3575: Objectify is no longer only an App Engine Standard Library
Noticed this while deploying a flex app:

image

I'm not sure what to do here. Perhaps simply rename the left hand column to "App Engine Libraries"?

chanseokoh:
Fixes #3575.


chanseokoh:
No need to change the message. It turns out that the project wizard was always giving the "App Engine Standard Libraries" label when creating a libraries page in the wizard. For non-App Engine libraries with a flex project, we have been showing "Other Libraries" in the standalone libraries page.

PR #3587, opened 2020-01-07
Source is not 1.7 any more

elharo:
@briandealwis Perhaps a coincidence, but fixing this seemed to let the project build in Eclipse without errors


briandealwis:
Hmm, odd.  I don't see errors?  And there are no Java 8'isms that I see in this code.  I suspect you don't have a Java 7 JRE configured?  If you modify _Preferences > Java > Installed JREs > Execution Environments_ and look at JavaSE-1.7, you should be able to select your Java 8 JVM.

I think I'd rather just remove these lines entirely?

elharo:
I did initially remove these lines entirely. However that led to other errors. 

I indeed do not have a Java 7 JVM installed, and going forward it's easier if we don't assume developers will have that, including Googlers with gLinux. Oracle  is making older JDKs increasingly hard to get hold of.

PR #3602, opened 2020-02-10
Redirect big logs

elharo:
@briandealwis Not complete yet but let's see if this at least helps the jobs pass


elharo:
Try 

-l,--log-file <arg>                    Log file where all build output
                                        will go (disables output color)

briandealwis:
Redirecting won't work as there will be no output for the 20-odd minutes taken by the build, and so Travis will abort the build.

I have a patch in progress to configure Logback to dedupe log entries.  Barring that, I'll just create a logback filter to skip logging these two problematic entries entirely.

briandealwis:
There's another alternative: don't log test output.
https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/2f2d4c8d6c83d884ebfe09d0cc1eb55b424d0f93/pom.xml#L53-L54

PR #3180, opened 2018-06-26
Log resolving errors from LibraryClasspathContainerResolverService

briandealwis:
I was hitting puzzling errors running the `ImportNativeAppEngineStandardProjectTest` integration tests locally where our classpath container had no entries.  It was really puzzling!  I discovered that our `LibraryClasspathContainerResolverService` wasn't logging a key exception, which clued me in that Maven had failed to download an artifact and left a little `.lastUpdated` dropping that prevents Maven from attempting to re-resolve for some amount of time.

This patch causes that exception to be logged.

```
java.lang.AssertionError: 
Expected: an array with size a value greater than <0>
     but: array size <0> was equal to <0>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at com.google.cloud.tools.eclipse.integration.appengine.ImportNativeAppEngineStandardProjectTest.updateOldContainers(ImportNativeAppEngineStandardProjectTest.java:131)
	at com.google.cloud.tools.eclipse.integration.appengine.ImportNativeAppEngineStandardProjectTest.importAppEngineStandardJava8_from1_3_1(ImportNativeAppEngineStandardProjectTest.java:102)
```

Issue #3095, opened 2018-05-24
Improve library resolution error reporting and handling


briandealwis:
I just hit the issue described in #2699, where the server runtime classpath container couldn't be resolved. It was difficult to figure out as there were no real indications of a problem, largely because our logging isn't actually visible (#3092).  I ended up having to launch a debug instance and put exception breakpoints on `CoreException` to figure out what was going wrong.

I hit this problem as my Windows VM's networking was non-functional and so attempting to resolve the server runtime artifacts failed (`servlet-api 3.1.0` and `jsp 2.3`).  Maven helpfully wrote out its `.lastUpdated` files which prevent subsequent attempts to re-resolve the artifacts because of [m2e's default update policy](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/issues/830#issuecomment-306843936) is set to never update.

Our reliance on Maven to resolve artifacts means that we're subject to m2e's policy decisions. But unlike m2e, we do not have a `pom.xml` to which to attach resolution problems.

We can actually ignore m2e's policies — and we actually do in our `DependencyResolver` (#2887).  But it turns out that we have two codepaths for resolving maven artifacts with different behaviours:

  - `MavenUtils`: uses m2e to resolve items, and is subject to the m2e preferences
  - `DependencyResolver`: uses aether directly, using m2e's configurations as a base that we can override

And our library resolving infrastructure uses `MavenUtils`, whereas our `BuildPath` support uses the `DependencyResolver`.  By using `DependencyResolver` everywhere, we can configure Aether to ignore the .lastUpdated files (`setUpdatePolicy(RepositoryPolicy.UPDATE_POLICY_*)`).

```
May 24, 2018 12:31:36 PM com.google.cloud.tools.eclipse.appengine.localserver.ServletClasspathProvider$1 onFailure
WARNING: Failed to resolve servlet APIs
java.util.concurrent.ExecutionException: org.eclipse.core.runtime.CoreException: Could not resolve artifact javax.servlet:servlet-api:jar:2.5
        at com.google.common.util.concurrent.AbstractFuture.getDoneValue(AbstractFuture.java:502)
        at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:461)
        at com.google.common.util.concurrent.AbstractFuture$TrustedFuture.get(AbstractFuture.java:78)
[...]

PR #3605, opened 2020-02-13
Check Android repository too

elharo:
@briandealwis


briandealwis:
These errors look suspiciously like #3180.  I suspect #3095 may play into it all too.

I looked briefly at how m2e's MavenPomEditor's _Dependency Hierarchy_ page resolves items.  It calls code somewhat like the following:
```java
String classpath = Artifact.SCOPE_RUNTIME_PLUS_SYSTEM;
IMavenProjectFacade facade = MavenPlugin.getMavenProjectRegistry().getProject(pomFile.getProject());

      DependencyNode root = MavenPlugin.getMavenModelManager().readDependencyTree(facade, mavenProject, classpath)
```

elharo:
I think #3180 and #3095 are symptoms, not causes. In all cases what we're seeing is a failure to find an artifact in Maven central. However there are different reasons that might be so. In this case the reason is simply that we're not checking the android repository.

I'm now inclined to go ahead with this PR since it does fix the problem. There might be a better way, but so far I don't see it.

elharo:
That didn't work (though it did locally). I May have to back it out.

PR #3256, opened 2018-09-21
Update target platform definitions

chanseokoh:
Fixes #3255.

PR #3614, opened 2020-03-10
remove photon support

elharo:
@briandealwis


briandealwis:
There are a few other Photon mentions.  We also marked two ITs as `@Ignored` in #3256 as they were flakey on Photon.

PR #3619, opened 2020-06-01
Add target for Eclipse 2020-03

briandealwis:


pom.xml
```diff
@@ -553,6 +553,40 @@
...
     </profile>
 
+    <profile>
+      <id>build-eclipse-2020-03</id>
+      <activation>
+        <property>
+          <name>eclipse.target</name> <value>2020-03</value>
+        </property>
+      </activation>
+      <properties>
```
chanseokoh, line 563: QQ: what are these, and until when should we carry them?

briandealwis: This is the selection mechanism for 2020-03 (vs the default, which is still 2018-09.

But maybe you mean the jetty properties?  I think we carried them over from GPE.  We could get rid of them since I think all Eclipse releases we support use Jetty 9.4.x.

chanseokoh: Ah, yeah, was talkinf about the jetty properties.

PR #3647, opened 2021-07-28
Set CLOUDSDK_PYTHON only after performing the Cloud SDK update

briandealwis:
The original installed Cloud SDK needs Python 2.  After being updated, it needs Python 3.

### DIFF
--- a/kokoro/ubuntu/release.sh
+++ b/kokoro/ubuntu/release.sh
@@ -8,14 +8,16 @@ set -o errexit
 # Display commands being run.
 set -o xtrace
 
-# More recent Cloud SDK requires Python 3.5 (b/194714889)
-export CLOUDSDK_PYTHON=python3.5
-
 gsutil -q cp "gs://ct4e-m2-repositories-for-kokoro/m2-cache.tar" - \
   | tar -C "${HOME}" -xf -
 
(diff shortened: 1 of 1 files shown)


kokoro/ubuntu/release.sh
```diff
@@ -8,14 +8,16 @@ set -o errexit
 # Display commands being run.
 set -o xtrace
 
-# More recent Cloud SDK requires Python 3.5 (b/194714889)
-export CLOUDSDK_PYTHON=python3.5
```
chanseokoh, line 12: I'm confused. You added this yesterday, and it did seem to fix Kokoro. Why did it work then?

briandealwis: I think that was Kokoro for Windows which likely has a more up-to-date Cloud SDK installed?

chanseokoh: Aha!

chanseokoh (approved):

kokoro/ubuntu/release.sh
chanseokoh: Well, I guess Windows uses `.bat`, but whatever. As long as this fixes all problems, we are good.

PR #3648, opened 2021-07-29
remove travis

elharo:
@briandealwis


briandealwis (changes requested):
Is there a compelling reason to remove Travis?  It at least lets us validate changes across a swathe of Eclipse versions.

elharo:
If we can keep it without charging my credit card, I'm OK with it.

PR #3663, opened 2021-12-15
chore: update to Tycho 2.5.0

briandealwis:
Update to Tycho 2.5.0. Note that this PR must be applied after #3668.

PR #3669, opened 2022-02-04
Update download link to point to archive

elefeint:
Version 3.11.0 was moved to archive.


chanseokoh:
@briandealwis has #3663 that I think subsumes this?

elefeint:
Even better. I missed that.

PR #3675, opened 2022-06-07
Try to find unit testing dependencies in an archive

elefeint:
Some of the old libraries seem to have moved away from the "latest" directories and are only available in dated archives. 

I've re-ran the last successful build, and it has these types of errors:
```
com.google.cloud.tools.eclipse.test.util: Failed to resolve target definition /home/runner/work/google-cloud-eclipse/google-cloud-eclipse/eclipse/eclipse-2018-09/gcp-eclipse-2018-09.target: Could not find "org.hamcrest.integration/1.3.0.v201305210900" in the repositories of the current location
```

Updating the target files to a dated folder fixes this issue, but updating the target files manually is not ideal.

To do it "properly", I installed an old version of Eclipse to match one of our integration test versions (4.15, from 2020-03), but I was not able to generate targets from TPD file, since it tries to load dependencies from `latest-R`.

### DIFF
--- a/eclipse/eclipse-2018-09/gcp-eclipse-2018-09.target
+++ b/eclipse/eclipse-2018-09/gcp-eclipse-2018-09.target
@@ -42,7 +42,7 @@
       <unit id="org.slf4j.apis.log4j" version="1.7.30.v20200204-2150"/>
       <unit id="org.slf4j.bridge.jul" version="1.7.30.v20200204-2150"/>
       <unit id="ch.qos.logback.slf4j" version="1.2.3.v20200428-2012"/>
-      <repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20200224183213/"/>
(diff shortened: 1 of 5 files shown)


elefeint:
@briandealwis Do you know how we might set TPD file repository location to go to someplace other than `latest-R`?

TPD can't generate target files for the same reason the build started failing:
```
Error occurred during resolution of 'https://download.eclipse.org/tools/orbit/downloads/latest-R/'. The IU 'org.hamcrest.integration' with range constraint '[1.3.0,1.3.1)' cannot be found.
```

This PR technically fixes the issues, but readme says it's a bad idea to edit target files directly.

chanseokoh:
Yeah, as the README says, `.target` files are auto-generated from .tpd`. So all I can say is that you should be editing the `.tpd` files. Not sure about the failure when generating them.

elefeint:
TPD tool is also trying to find these dependencies in R-latest.

elefeint:
Closing in favor of #3676.

PR #3676, opened 2022-06-09
Updated common TPD location and regenerated target files

elefeint:
Trying to do  #3675, but properly.

### DIFF
--- a/eclipse/eclipse-2020-03/gcp-eclipse-2020-03.target
+++ b/eclipse/eclipse-2020-03/gcp-eclipse-2020-03.target
@@ -1,37 +1,37 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde?>
 <!-- generated with https://github.com/eclipse-cbi/targetplatform-dsl -->
-<target name="GCP for Eclipse 2020-03" sequenceNumber="1591024101">
+<target name="GCP for Eclipse 2020-03" sequenceNumber="1654794613">
   <locations>
     <location includeMode="planner" includeAllPlatforms="false" includeSource="true" includeConfigurePhase="false" type="InstallableUnit">
-      <unit id="org.eclipse.sdk.feature.group" version="4.14.0.v20191210-0610"/>
-      <unit id="org.eclipse.jdt.feature.group" version="3.18.200.v20191210-0610"/>
-      <unit id="org.eclipse.m2e.feature.feature.group" version="1.14.0.20191209-1925"/>
-      <unit id="org.eclipse.m2e.sdk.feature.feature.group" version="1.14.0.20191209-1925"/>
-      <unit id="org.eclipse.m2e.wtp.feature.feature.group" version="1.4.3.20191209-1849"/>
-      <unit id="org.eclipse.m2e.wtp.sdk.feature.feature.group" version="1.4.3.20191209-1849"/>
+      <unit id="org.eclipse.sdk.feature.group" version="4.15.0.v20200305-0155"/>
+      <unit id="org.eclipse.jdt.feature.group" version="3.18.300.v20200305-0155"/>
+      <unit id="org.eclipse.m2e.feature.feature.group" version="1.15.0.20200310-1832"/>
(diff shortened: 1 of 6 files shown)


chanseokoh:
@briandealwis can you take a quick look?

briandealwis:
Oops, sorry @elefeint — ugh!  So it looks like we've been bitten by two changes:
  1. Orbit formerly had a repository of OSGi-ified jars and a new approach of building and bundling jars on Maven Central; it looks like the older repository had hamcrest 1.3.0 and they've been jettisoned
  2. Orbit moved to Hamcrest 2.2.0 which merged several jars into a single jar https://git.eclipse.org/c/orbit/orbit-recipes.git/commit/?id=493fdddeb1a9ba48f6321e428c5c88c5c22683a0

There are two approaches here:
  1. the newer versions of Tycho make it really easy to consume jars directly from Maven Central
  2. just update the tests to Hamcrest 2.2, which may be really easy

And we should drop old versions of Eclipse and just support the last 2 releases.

briandealwis:
But seeing as these are all test-time dependencies, this change works for now.

eclipse/eclipse-2020-03/gcp-eclipse-2020-03.target
```diff
@@ -1,37 +1,37 @@
...
+<target name="GCP for Eclipse 2020-03" sequenceNumber="1654794613">
   <locations>
     <location includeMode="planner" includeAllPlatforms="false" includeSource="true" includeConfigurePhase="false" type="InstallableUnit">
-      <unit id="org.eclipse.sdk.feature.group" version="4.14.0.v20191210-0610"/>
-      <unit id="org.eclipse.jdt.feature.group" version="3.18.200.v20191210-0610"/>
-      <unit id="org.eclipse.m2e.feature.feature.group" version="1.14.0.20191209-1925"/>
-      <unit id="org.eclipse.m2e.sdk.feature.feature.group" version="1.14.0.20191209-1925"/>
-      <unit id="org.eclipse.m2e.wtp.feature.feature.group" version="1.4.3.20191209-1849"/>
-      <unit id="org.eclipse.m2e.wtp.sdk.feature.feature.group" version="1.4.3.20191209-1849"/>
+      <unit id="org.eclipse.sdk.feature.group" version="4.15.0.v20200305-0155"/>
```
briandealwis, line 7: These changes are odd: the minor version shouldn't have changed here.

elefeint:
That makes sense -- I've thought of just upgrading Hamcrest, but the `hamcrest-integration` being unavailable in a newer version worried me. If they consolidated the functionality in core, then that should not be a concern. I'll upgrade the testing libraries.

elefeint:
I was not successful with upgrading Hamcrest in #3677. I think it's because JUnit4 brings in the old Hamcrest 1.3 -- I was able to use the [migration guide](http://hamcrest.org/JavaHamcrest/distributables#upgrading-from-hamcrest-1x) to make it not be a problem from maven side, but Eclipse classloading is a whole other thing:

```
java.lang.LinkageError: loader constraint violation: loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @eb2ce99 wants to load interface org.hamcrest.Matcher. A different interface with the same name was previously loaded by org.eclipse.osgi.internal.loader.EquinoxClassLoader @e280006. (org.hamcrest.Matcher is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @e280006, parent loader 'platform')
```

I would prefer to go forward with this current PR that repoints to a timestamped location where the old hamcrest library is actually present.

eclipse/eclipse-2020-03/gcp-eclipse-2020-03.target
briandealwis: The [2019-12 target](http://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/eclipse/eclipse-2019-12/gcp-eclipse-2019-12.target) show the Eclipse SDK feature is 4.14.x, so 2020-03 should have been 4.15.x. It looks like there was an error when generating the original 2020-03 target.

briandealwis (approved):
LGTM

PR #3715, opened 2023-02-24
feat: modify "Manage Google Accounts" button to rely on gcloud ADC

diegomarquezp:
This modifies the behavior of this button in the following way:
* If ADC are not set, it will show a clickable entry explaining the user is not logged in
image
* Clicking this will show [more information](https://cloud.google.com/docs/authentication/application-default-credentials)
image
* If set, it will offer to view the current account.
image
  * Clicking this button will display a singleton list of users (only user associated with ADC)
  * This used to have two buttons to sign in with a different user or log out. It will now show an avatar and email only. No action on click.

## NOTE 
(fixed - see below) The ADC credentials are cached in `google-auth-library-java` ([see](https://github.com/googleapis/google-auth-library-java/blob/e6c23d318dd98021cc8cc04ab2ed5eb1a72acfce/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L115-L124)). Running `gcloud auth application-default login|revoke` will not update this component properly:
1. plugin starts with ADC set
2. run `gcloud auth ... revoke`
3. menu button will still show as if ADC are set

this is fixed in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3716


plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/messages.properties
```diff
@@ -4,8 +4,8 @@ LOGIN_ERROR_CANNOT_OPEN_BROWSER=Error opening a system browser. Signing in requi
 LOGIN_ERROR_DIALOG_MESSAGE=An error occurred while trying to sign in.
 LOGIN_ERROR_DIALOG_TITLE=Sign in Failed
 LOGIN_ERROR_LOCAL_SERVER_RUN=Error listening for authorization code:\n\n{0}
-LOGIN_MENU_LOGGED_IN=Manage Google Accounts...
-LOGIN_MENU_LOGGED_OUT=Sign in to Google...
+LOGIN_MENU_LOGGED_IN=View Application Default Credentials account...
+LOGIN_MENU_LOGGED_OUT=Please log in via gcloud CLI...
```
burkedavison, line 8: If a new user see this, it seems reasonable they'd have one of two thoughts:
* I don't know how to do that.
* I should go execute `gcloud auth login`

But neither of these is the action we're looking to have completed.

Rather than fitting the request in the disabled menu item, do we need to open the dialog and display more information about specifically what we're asking the user to do?

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
```diff
@@ -105,14 +115,14 @@ void createAccountsPane(Composite accountArea) {
       GridLayoutFactory.fillDefaults().numColumns(2).applyTo(accountRow);
       GridLayoutFactory.fillDefaults().generateLayout(secondColumn);
 
-      if (account.getName() != null) {
-        name.setText(account.getName());
+      if (account.getName().isPresent()) {
+        name.setText(account.getName().get());
```
burkedavison, line 119: Since we're using `Optional`, consider replacing `isPresent()`+`get()` with:
```java
account.getName().ifPresent(name::setText);
```

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
```diff
@@ -82,13 +86,19 @@ protected Control createDialogArea(Composite parent) {
...
 
   @VisibleForTesting
   void createAccountsPane(Composite accountArea) {
-    for (Account account : loginService.getAccounts()) {
+    List<Account> accounts;
+    try {
+      accounts = Collections.singletonList(apiFactory.getAccount());
+    } catch (IOException ex) {
+      logger.log(Level.SEVERE, "Could not obtain ADC account", ex);
+      accounts = Collections.emptyList();
```
burkedavison, line 99: Syntax sugar (meaning: your choice, but consider for readability) -- a private `getAccounts()` method would prevent needing to have this relatively clunky "declare then initialize then use" order of events.

```java
private List<Account> getAccounts() {
  try {
    return Collections.singletonList(apiFactory.getAccount());
  } catch (IOException ex) {
    ...
    return Collections.emptyList();
  }
}
void createAccountsPane(...) {
  for (Account account : getAccounts()) {
    ...
  }
}
```

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/GoogleLoginCommandHandler.java
```diff
@@ -29,26 +34,57 @@
...
+  static final long LOGGED_IN_CHECK_DELAY = 2000l; // update logged in every 2 seconds
+  static boolean isLoggedIn;
+  Timer loggedInCheckTask = new Timer();
+  
+  public GoogleLoginCommandHandler() {
+    loggedInCheckTask.schedule(new TimerTask() {
+
+      @Override
+      public void run() {
+        checkIsLoggedIn();
```
burkedavison, line 47: Are we certain this is efficient enough to be running every two seconds?

If we were to change some assumptions (like, displaying the login status in the menu) could we turn this from a polling implementation to a on-demand check?

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/messages.properties
diegomarquezp: Makes sense - I will use a separate dialog

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/GoogleLoginCommandHandler.java
diegomarquezp: This polling approach was what I tried when noticing that updating the ADC did not reflect in the oauth library (fixed in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3716). I will move `checkLoggedIn` to `isEnabled`

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/IGoogleApiFactory.java
```diff
@@ -37,6 +37,8 @@ public interface IGoogleApiFactory {
    */
   public Account getAccount() throws IOException;
   
+  public boolean isLoggedIn();
```
diegomarquezp, line 40: Temporary as better approach is used in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3713

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountsPanel.java
```diff
@@ -81,45 +82,67 @@ protected Control createDialogArea(Composite parent) {
...
+    GridLayoutFactory.fillDefaults().numColumns(1).applyTo(messageRow);
+    GridLayoutFactory.fillDefaults().generateLayout(linkRow);
+    
+  }
+  
+  private Account getAccount() {
+    try {
+      return apiFactory.getAccount();
+    } catch (IOException ex) {
+      return null;
```
burkedavison, line 143: If this returns `null`, then the above `account.getName()` and `account.getEmail()` will throw a NullPointerException.

burkedavison (approved):

PR #3718, opened 2023-03-16
chore: enable or disable deploy button depending on ADC being found.

diegomarquezp:


plugins/com.google.cloud.tools.eclipse.appengine.deploy.ui/src/com/google/cloud/tools/eclipse/appengine/deploy/ui/DeployPreferencesDialog.java
```diff
@@ -102,7 +102,7 @@ protected Control createDialogArea(Composite parent) {
           @Override
           public int getMessageType(ValidationStatusProvider statusProvider) {
             int type = super.getMessageType(statusProvider);
-            setValid(type != IMessageProvider.ERROR);
+            setValid(type != IMessageProvider.ERROR && apiFactory.isLoggedIn());
```
burkedavison, line 105: `isLoggedIn()` is removed from the `IGoogleApiFactory` below, but then used here. I'm confused - how does this compile? Am I mistaken on which interface is being used here?

diegomarquezp: Oh, it seems like this slipped in when merging the `gcloud-creds-cli` branch. I changed it to `hasCredentialsSet()`.

PR #3731, opened 2023-05-11
feat: [wip] compute credentials from well-known ADC file path (apiary credentials)

diegomarquezp:


plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
```diff
@@ -63,43 +97,86 @@ public void initialize(HttpRequest httpRequest) throws IOException {
...
-    return getAccount().map(Account::getOAuth2Credential);
+    return currentCred;
   }
   
-  Account getAccount(Credential credential) throws IOException {
-    if (accountCache.containsKey(credential)) {
-      return accountCache.get(credential);
+  private Optional<Credential> computeCredential() {
+    Optional<File> credsFile = getCredentialFile();
+    if (!credsFile.get().exists()) {
```
burkedavison, line 146: If getCredentialFile returns Optional.empty(), this get() call would throw a NullPointerException.

If getCredentialFile will always return a File object (which, as implemented now, it will), then it doesn't need to be wrapped in an Optional.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
```diff
@@ -65,6 +65,9 @@ public AccountSelector(Composite parent, IGoogleApiFactory apiFactory) {
     GridDataFactory.fillDefaults().grab(true, false).applyTo(accountEmailComposite);
     GridLayoutFactory.fillDefaults().generateLayout(accountEmailComposite);
     GridLayoutFactory.fillDefaults().generateLayout(this);
+    apiFactory.addCredentialChangeListener(() -> {
```
burkedavison, line 68: Should this also have an associated `remove` call? What prevents this listener from living longer than the selector's usefulness?

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
```diff
@@ -16,16 +16,20 @@
...
 import org.eclipse.ui.commands.ICommandService;
 
 public class MenuContributionInitializer implements IStartup {
-
+  IGoogleApiFactory apiFactory = new GoogleApiFactory();
+  
   @Override
   public void earlyStartup() {
     updateLoginCommand();
+    apiFactory.addCredentialChangeListener(() -> updateLoginCommand());
```
burkedavison, line 32: Is there an on-menu-open event that can be used to query credential status only when opened, or does the menu have to be pre-calculated and proactively updated like this?

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
```diff
@@ -16,16 +16,20 @@
...
+import com.google.cloud.tools.eclipse.googleapis.IGoogleApiFactory;
+import com.google.cloud.tools.eclipse.googleapis.internal.GoogleApiFactory;
 import org.eclipse.ui.IStartup;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.commands.ICommandService;
 
 public class MenuContributionInitializer implements IStartup {
-
+  IGoogleApiFactory apiFactory = new GoogleApiFactory();
```
burkedavison, line 27: Is this instance different than the apiFactory being passed down to other parts of the application? Are there going to be any issue with have multiple instances?

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
```diff
@@ -25,30 +25,64 @@
...
+    adcPathPoller = new Timer();
+    TimerTask task = new TimerTask() {
+      @Override
+      public void run() {
+          checkIfAdcPathChanged();
+      }
+    };
+
+    // Schedule the task to run every second
+    adcPathPoller.scheduleAtFixedRate(task, 0, 1000);
```
burkedavison, line 73: Is file-polling the recommended approach by the auth team?

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
diegomarquezp: I removed the `Optional` wrapper.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
diegomarquezp: I think we can fall back to a singleton as a follow up. Initially this class was supposed to take different credentials depending on which part of the plugin needed to access the apis. Now we have the same apis returned everywhere because it depends solely on the `AccountProvider` used by this class

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/AccountSelector.java
diegomarquezp: I added a dispose listener

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
diegomarquezp: I reached out to auth to confirm. I noticed there is this [`WatchService`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/WatchService.html) class as well - it is _intended to map directly on to the native file event notification facility where available_

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
diegomarquezp: I'm not entirely sure if this is possible. We have this `plugin.xml` defining the extension  https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/cf10c1047efe35077565216274f909f938eb0466/plugins/com.google.cloud.tools.eclipse.login/plugin.xml#L23-L52

It tells it the `GoogleLoginCommandHandler` how to update the element and what it does when clicked in three points (File menu, Google Cloud menu and lower right status bar). These are three elements (of probably different types) - this is just a conjecture in favor of a hook of this kind not being available.

I'll keep looking at the docs 

[menu extensions](https://download.eclipse.org/rt/rap/doc/3.6/guide/reference/extension-points/org_eclipse_ui_menus.html) for reference

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
diegomarquezp: They got back saying that there is nothing built in in the library and a custom approach is fine as long as the credential is recreated.

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
diegomarquezp: Here is a [list of examples](https://wiki.eclipse.org/Menu_Contributions) of `menuContribution` - it seems like there is no way to subscribe to dropdown opening events. `updateElement` gets called even if the element is closed, and does not get called when the dropdown is opened/closed.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
burkedavison: +1000 to using WatchService over the custom timer.

diegomarquezp: Added Watch Service and tests

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
diegomarquezp: I added a follow up item in https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/3712

plugins/com.google.cloud.tools.eclipse.login/src/com/google/cloud/tools/eclipse/login/ui/MenuContributionInitializer.java
burkedavison: Thanks for looking into this.

plugins/com.google.cloud.tools.eclipse.googleapis/src/com/google/cloud/tools/eclipse/googleapis/internal/DefaultAccountProvider.java
```diff
@@ -25,30 +25,99 @@
...
+        try {
+          key = watchService.take();
+        } catch (InterruptedException ex) {
+          LOGGER.log(Level.SEVERE, "Error creating watch service", ex);
+          continue;
+        }
+        for (WatchEvent<?> event : key.pollEvents()) {
+          LOGGER.log(Level.FINE, "Detected change in ADC file");
+          Path affectedFile = Paths.get(adcFolderPath.toAbsolutePath().toString(), 
+              ((Path) event.context()).toString()).toAbsolutePath();
```
burkedavison, line 100: nit: Consider not going through strings and instead using `Path.resolve`
```java
adcFolderpath.toAbsolutePath().resolve((Path) event.context());
```

burkedavison (approved):

